PR #5, opened 2020-12-05
Init new action for the deletion of container(s)

fuweid:
Signed-off-by: Wei Fu <fuweid89@gmail.com>


AkihiroSuda (approved):
Thanks. I think we want to support short form of container ID, but can be a separate PR

Issue #11, opened 2020-12-06
Support `nerdctl run --seccomp=XXX`


soharaki:
Add the seccomp options to the `nerdctl run`.

AkihiroSuda:
Seccomp was implemented in #12,  AppArmor was implemented in 3e9a754ea082a5fc8fee8c91b265231348f6e03b .

We also need to support SELinux (`--security-opt label`)

TheDiveO:
@AkihiroSuda mentions that apparmor support has been implemented. Unfortunately, deploying my [lxkns discovery service](https://github.com/thediveo/lxkns) using the existing docker-compose.yaml fails, seemingly the security option from the composer file doesn't get applied?

Admittedly, this might have been composed in infosec theory hell 😁:
```yaml
version: '2.4'
services: 
    lxkns:
        image: lxkns:latest
        read_only: true
        build: 
            context: ../..
            dockerfile: deployments/lxkns/Dockerfile
        command:
            - "/lxkns"
            - "--initialcgroup"
            - "--docker=unix:///proc/1/root/run/docker.sock"
            - "--containerd=/proc/1/root/run/containerd/containerd.sock"
        ports:
            - "5010:5010"
        user: "65534"
        security_opt:
            # apparmor only accepts a profile name, which must have been loaded
            # by the system admin. You might want to use the lxkns-apparmor
            # profile to be found in this directory.
            - apparmor:unconfined #- apparmor:lxkns
            # Please note that Docker's default container seccomp profile is
            # adaptive: it allows all those syscalls for which the process has
            # matching capabilities. In consequence, this Docker default seccomp
            # profile thus does not interfere with lxkns. However, we still want
            # to slightly curb the unneeded parts of the powers bestowed upon us
            # by especially CRAP_SYS_ADMIN and CAP_SYS_PTRACE...
            - seccomp:deployments/lxkns/lxkns-seccomp.json
        cap_drop:
            - ALL
        cap_add:
            - SYS_ADMIN       # change namespaces
            - SYS_CHROOT      # change mount namespaces
            - SYS_PTRACE      # access nsfs namespace information
            - DAC_READ_SEARCH # for fully discovering bind-mounted namespaces
            - DAC_OVERRIDE    # for accessing the containerd API socket
        # We need the full PID view for correct scanning.
        pid: host
```

- The composer file specifies to **unconfine** the deployed service w.r.t. AppArmor.
[...]

PR #12, opened 2020-12-06
Add Seccomp Option

soharaki:
When running `nerdctl run`, we can use the seccomp flg.
Referred to "Docker's security option: seccomp", not the ctr commands.
#11

Signed-off-by: Sho Haraki <endurance.test.20380119@gmail.com>

### DIFF
--- a/run.go
+++ b/run.go
@@ -26,6 +26,7 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/containerd/console"
@@ -77,6 +78,10 @@ var runCommand = &cli.Command{
 			Usage: "Set custom DNS servers (only meaningful for \"bridge\" network)",
 			Value: cli.NewStringSlice("8.8.8.8", "1.1.1.1"),
 		},
+		&cli.StringSliceFlag{
+			Name:  "security-opt",
+			Usage: "Set security options (Currently enable Seccomp custom profile)",
+		},
 	},
 }
 
@@ -105,7 +110,6 @@ func runAction(clicontext *cli.Context) error {
 		oci.WithDefaultSpec(),
 		oci.WithDefaultUnixDevices,
 		oci.WithImageConfig(ensured.Image),
-		seccomp.WithDefaultProfile(),
 	)
 	cOpts = append(cOpts,
 		containerd.WithImage(ensured.Image),
@@ -151,6 +155,16 @@ func runAction(clicontext *cli.Context) error {
 		opts = append(opts, withCustomResolvConf(resolvConf.Name()))
 	default:
 		return errors.Errorf("unknown network %q", netstr)
+	}
+
+	securityOpts := clicontext.StringSlice("security-opt")
+	if securityOpts != nil {
+		securityOptsMaps := ConvertKVStringsToMap(securityOpts)
+		secopts := GenerateSecurityOps(securityOptsMaps)
+		opts = append(opts, secopts...)
+	} else {
+		// append the default profiles if no security-opt
+		opts = append(opts, seccomp.WithDefaultProfile())
 	}
 
 	var s specs.Spec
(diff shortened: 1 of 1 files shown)


run.go
```diff
@@ -77,6 +77,10 @@ var runCommand = &cli.Command{
 			Usage: "Set custom DNS servers (only meaningful for \"bridge\" network)",
 			Value: cli.NewStringSlice("8.8.8.8", "1.1.1.1"),
 		},
+		&cli.StringFlag{
```
soharaki, line 80: I wasn't sure whether to add this option to `run.go` or `main.go`.  And I selected `run.go` .
In the containerd project, it is `containerd/cmd/ctr/commands/commands.go`, 
but it is only used as an option for `ctr run`.

run.go
```diff
@@ -153,6 +156,17 @@ func runAction(clicontext *cli.Context) error {
 		return errors.Errorf("unknown network %q", netstr)
 	}
 
+	seccompProfile := clicontext.String("seccomp")
+	if seccompProfile != "" {
+		if seccompProfile != "unconfined" {
```
soharaki, line 161: The `seccomp` and `unconfined` come from Docker options.

run.go
```diff
@@ -153,6 +156,17 @@ func runAction(clicontext *cli.Context) error {
...
+			opts = append(opts, seccomp.WithProfile(seccompProfile))
+		} else {
+			// If passing "unconfined" to run a container without the default seccomp profile.
+		}
+	} else {
+		opts = append(opts, seccomp.WithDefaultProfile())
+	}
+
```
soharaki, line 169: This is how it works:

```console
# containerd's seccomp defaultProfile
$ sudo nerdctl run -it --seccomp="" alpine sh
/ # unshare
unshare: unshare(0x0): Operation not permitted

# no seccomp
$ sudo nerdctl run -it --seccomp="unconfined" alpine sh
/ # unshare --user
HOST:~$

# Enable custom seccomp
$ sudo nerdctl run -it --seccomp=_output/default.json alpine sh
/ # unshare
HOST:~$
```
refs: https://github.com/moby/moby/blob/f7bcb02f677257a5623dcb561f003e96d5f3ea3f/profiles/seccomp/default.json

AkihiroSuda (changes requested):
The CLI syntax needs to be compatible with Docker, please see https://docs.docker.com/engine/security/seccomp/

AkihiroSuda:
Also, could you use your real name for signing off the commit? Thanks

run.go
```diff
@@ -153,6 +156,17 @@ func runAction(clicontext *cli.Context) error {
 		return errors.Errorf("unknown network %q", netstr)
 	}
 
+	seccompProfile := clicontext.String("seccomp")
+	if seccompProfile != "" {
+		if seccompProfile != "unconfined" {
+			opts = append(opts, seccomp.WithProfile(seccompProfile))
+		} else {
```
AkihiroSuda, line 163: This empty else{} block is unneeded

AkihiroSuda (changes requested):

run.go
AkihiroSuda: This should be in run.go, but the flag syntax needs to be `--security-opt seccomp=<PROFILE>`

run.go
```diff
@@ -77,6 +79,10 @@ var runCommand = &cli.Command{
 			Usage: "Set custom DNS servers (only meaningful for \"bridge\" network)",
 			Value: cli.NewStringSlice("8.8.8.8", "1.1.1.1"),
 		},
+		&cli.StringFlag{
```
AkihiroSuda, line 82: cli.StringSliceFlag

run.go
```diff
@@ -153,6 +158,25 @@ func runAction(clicontext *cli.Context) error {
 		return errors.Errorf("unknown network %q", netstr)
 	}
 
+	securityOpt := clicontext.String("security-opt")
```
AkihiroSuda, line 161: `clicontext.StringSlice`

soharaki:
I tried to matche the behavior of `nerdctl` to docker-cli as closely as possible.
However, the behavior is different when not setting any arguments.

```console
$ docker container run -it --security-opt "" alpine sh
docker: Invalid --security-opt: "".
See 'docker run --help'.

$ sudo nerdctl run -it --security-opt "" alpine sh
/ #
$  sudo nerdctl run -it --security-opt "seccomp=" alpine sh
/ # #the same above
```

run.go
```diff
@@ -153,6 +158,25 @@ func runAction(clicontext *cli.Context) error {
...
+	if securityOpt != "" {
+		securityOpts := regexp.MustCompile("[\\,\\s]+").Split(securityOpt, -1)
```
AkihiroSuda, line 163: No need to use regexp

run.go
```diff
@@ -153,6 +158,25 @@ func runAction(clicontext *cli.Context) error {
...
+		securityOptsMaps := ConvertKVStringsToMap(securityOpts)
+		dictSecurityOpts := map[string]string{
+			"seccomp": "allow",
```
AkihiroSuda, line 166: unused value

run.go
soharaki: I've been thinking about that, but ` --security-opt "seccomp=hoge.json,apparmor=foo.profile"`

AkihiroSuda: Docker doesn't support such syntax either

soharaki: I supposed that docker are the docker seems to be able to handle the parser in the same way.

soharaki: >Docker doesn't support such syntax either
Oh, I see.

AkihiroSuda: `docker run --security-opt seccomp=hoge.json --security-opt apparmor=foo.profile" is the correct form

soharaki:
I fixed it.

AkihiroSuda (approved):
Thanks

PR #16, opened 2020-12-09
Add support for nerdctl images.

shishir-a412ed:
Signed-off-by: Shishir Mahajan <smahajan@roblox.com>

### DIFF
--- a/images.go
+++ b/images.go
@@ -1,0 +1,113 @@
+/*
+   Copyright (C) nerdctl authors.
+   Copyright (C) containerd authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/pkg/progress"
+	"github.com/containerd/containerd/platforms"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli/v2"
+)
+
+var imagesCommand = &cli.Command{
+	Name:   "images",
+	Usage:  "List images",
+	Action: imagesAction,
+	Flags: []cli.Flag{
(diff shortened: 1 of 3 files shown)


images.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+
+	return printImages(ctx, clicontext, imageList, cs)
+}
+
+func printImages(ctx context.Context, clicontext *cli.Context, imageList []images.Image, cs content.Store) error {
+	quiet := clicontext.Bool("quiet")
+
+	w := tabwriter.NewWriter(os.Stdout, 4, 8, 4, ' ', 0)
```
AkihiroSuda, line 73: Nit: stdout should be clicontext.App.Writer

shishir-a412ed:
```
root@vagrant:~/go/src/github.com/nerdctl# nerdctl images
REPOSITORY                  TAG       DIGEST ID       CREATED         SIZE
docker.io/library/alpine    latest    c0e9560cda11    25 hours ago    2.7 MiB
docker.io/library/redis     alpine    b0e84b6b9214    25 hours ago    10.0 MiB
root@vagrant:~/go/src/github.com/nerdctl# nerdctl images --quiet
c0e9560cda11
b0e84b6b9214
```

1. The formatting is slightly tilted in the above output, however, the formatting in the actual command ouput is correct.
2. IIRC in the `docker images` output, `IMAGE ID` is something specific to docker. Please let me know if that is not the case, and containerd also has some notion of `IMAGE ID`. For now, I have replaced imageID with DigestID.

/cc @AkihiroSuda

AkihiroSuda:
I think we can just show digest as "IMAGE ID"

images.go
```diff
@@ -0,0 +1,105 @@
...
+	if !quiet {
+		fmt.Fprintln(w, "REPOSITORY\tTAG\tDIGEST ID\tCREATED\tSIZE")
```
AkihiroSuda, line 75: DIGEST ID -> IMAGE ID

images.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	for _, img := range imageList {
+		size, err := img.Size(ctx, cs, platforms.Default())
+		if err != nil {
+			return errors.Wrap(err, "failed to compute image size")
+		}
+
+		repositoryTag := strings.Split(img.Name, ":")
```
AkihiroSuda, line 84: Can we trim "docker.io/library" and "docker.io/" prefix?

images.go
shishir-a412ed: Does [`ps tabwriter`](https://github.com/AkihiroSuda/nerdctl/blob/master/ps.go#L70) needs to be fixed too?

shishir-a412ed: Fixed.

images.go
shishir-a412ed: Fixed.

images.go
shishir-a412ed: Fixed.

AkihiroSuda (approved):
Thanks

PR #24, opened 2020-12-09
Add `--cni-bin-dir`/`CNI_BIN_DIR` to override default cni dir (`/opt/cni/bin`)

### LINKED ISSUE #23: allow for a different cni plugin directory?
Can you add an override environment variable so the cni plugin directory can be somewhere different to `/opt/cni/bin`? Or is that a bad idea?

https://github.com/AkihiroSuda/nerdctl/blob/master/run.go#L169
https://github.com/containerd/go-cni/blob/master/types.go#L22

06kellyjac:
resolves #23 

- [x] ran on my machine :+1: 
- [X] `gofmt`-ed
- [X] `git commit -s` signed off with real name

I can now successfully use `nerdctl run` with my cni plugins elsewhere.

The cli output is now the following, let me know if the wording or order needs changing:

```
NAME:
   nerdctl run - Run a command in a new container

USAGE:
   nerdctl run [command options] [arguments...]

OPTIONS:
   --tty, -t                     (Currently -t needs to correspond to -i) (default: false)
   --interactive, -i             (Currently -i needs to correspond to -t) (default: false)
   --detach, -d                  Run container in background and print container ID (default: false)
   --cni-bin-dir value           Set the cni-plugins binary directory (default: "/opt/cni/bin") [$CNI_BIN_DIR]
   --restart value               Restart policy to apply when a container exits (implemented values: "no"|"always") (default: "no")
   --rm                          Automatically remove the container when it exits (default: false)
   --pull value                  Pull image before running ("always"|"missing"|"never") (default: "missing")
   --network value, --net value  Connect a container to a network ("bridge"|"host"|"none") (default: "bridge")
   --dns value                   Set custom DNS servers (only meaningful for "bridge" network) (default: "8.8.8.8", "1.1.1.1")
   --security-opt value          Security options
   --privileged                  Give extended privileges to this container (default: false)
   --help, -h                    show help (default: false)
```

### DIFF
--- a/run.go
+++ b/run.go
@@ -67,6 +67,12 @@ var runCommand = &cli.Command{
 			Usage:   "Run container in background and print container ID",
 		},
 		&cli.StringFlag{
+			Name:    "cni-bin-dir",
+			Usage:   "Set the cni-plugins binary directory",
+			EnvVars: []string{"CNI_BIN_DIR"},
+			Value:   gocni.DefaultCNIDir,
+		},
+		&cli.StringFlag{
 			Name:  "restart",
 			Usage: "Restart policy to apply when a container exits (implemented values: \"no\"|\"always\")",
 			Value: "no",
@@ -158,6 +164,7 @@ func runAction(clicontext *cli.Context) error {
 	}
 	cOpts = append(cOpts, restartOpts...)
 
+	var cniBinDir = clicontext.String("cni-bin-dir")
 	var cniNetwork gocni.CNI
 	switch netstr := clicontext.String("network"); netstr {
 	case "none":
@@ -166,12 +173,12 @@ func runAction(clicontext *cli.Context) error {
 		opts = append(opts, oci.WithHostNamespace(specs.NetworkNamespace), oci.WithHostHostsFile, oci.WithHostResolvconf)
 	case "bridge":
 		for _, f := range requiredCNIPlugins {
-			p := filepath.Join(gocni.DefaultCNIDir, f)
+			p := filepath.Join(cniBinDir, f)
 			if _, err := exec.LookPath(p); err != nil {
 				return errors.Wrapf(err, "needs CNI plugin %q to be installed, see https://github.com/containernetworking/plugins/releases", p)
 			}
 		}
-		if cniNetwork, err = gocni.New(gocni.WithConfListBytes([]byte(defaultBridgeNetwork))); err != nil {
+		if cniNetwork, err = gocni.New(gocni.WithPluginDir([]string{cniBinDir}), gocni.WithConfListBytes([]byte(defaultBridgeNetwork))); err != nil {
 			return err
 		}
 		resolvConf, err := ioutil.TempFile("", "nerdctl-resolvconf")


run.go
```diff
@@ -46,6 +46,15 @@ import (
...
 )
 
+func getEnv(key, fallback string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return fallback
+}
+
+var cniDir = getEnv("CNI_DIR", gocni.DefaultCNIDir)
```
AkihiroSuda, line 56: I feel this should be CNI_BIN_DIR, and should be also defined as --cni-bin-dir flag in main.go

AkihiroSuda (changes requested):

06kellyjac:
I've got a first pass with #24 

~Let me know if I should put `getEnv` somewhere else and if the env var should be named something other than `CNI_DIR`~
~I guess other options could be `CNI_PLUGGINS_DIR` `CNI_BIN_DIR` `NERDCTL_CNI_PLUGINS_DIR` etc.~
I've seen your review

---

Also do you have a recommended distro for setting up containerd with cni plugins properly so I can isolate whether there are more changes to do to `nerdctl` rather than my current containerd setup being crap :sweat_smile:

run.go
AkihiroSuda: Also, could you sign off the commit with real name (git commit -a -s --amend), thanks

06kellyjac: I should have looked at the cli setup first :facepalm:

06kellyjac: Is there any particular order to the flags?

AkihiroSuda:
I use Ubuntu 20.10, but I don't use dpkg at all for Go and container stuff. I install almost everything from source.

06kellyjac:
I'm past my original /opt/cni/bin error of:

```
FATA[0000] needs CNI plugin "/opt/cni/bin/bridge" to be installed, see https://github.com/containernetworking/plugins/releases: exec: "/opt/cni/bin/bridge": stat /opt/cni/bin/bridge: no such file or directory
```

But I'm getting some new ones

for nerdctl I'm running:

```
sudo ./_output/nerdctl run --cni-bin-dir "/nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin" -it alpine
[sudo] password for jk:
ERRO[0000] network review                                error="failed to find plugin \"tuning\" in path [/opt/cni/bin]"
FATA[0000] failed to find plugin "bridge" in path [/opt/cni/bin]
```

For containerd I'm running

```
sudo containerd --config someconfig -l debug
```

with a config of:

```
[plugins]
  [plugins.cri]
    [plugins.cri.cni]
      bin_dir = "/nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin"
```
`runc` is in my path (specifically when `containerd`)

and it seems to be setting the right value:

```
INFO[2020-12-09T14:37:30.038588379Z] Start cri plugin with config {PluginConfig:{ContainerdConfig:{Snapshotter:overlayfs DefaultRuntimeName:runc DefaultRuntime:{Type: Engine: PodAnnotations:[] ContainerAnnotations:[] Root: Options:<nil> PrivilegedWithoutHostDevices:false BaseRuntimeSpec:} UntrustedWorkloadRuntime:{Type: Engine: PodAnnotations:[] ContainerAnnotations:[] Root: Options:<nil> PrivilegedWithoutHostDevices:false BaseRuntimeSpec:} Runtimes:map[runc:{Type:io.containerd.runc.v2 Engine: PodAnnotations:[] ContainerAnnotations:[] Root: Options:0xc0003eafc0 PrivilegedWithoutHostDevices:false BaseRuntimeSpec:}] NoPivot:false DisableSnapshotAnnotations:true DiscardUnpackedLayers:false} CniConfig:{NetworkPluginBinDir:/nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin NetworkPluginConfDir:/etc/cni/net.d NetworkPluginMaxConfNum:1 NetworkPluginConfTemplate:} Registry:{Mirrors:map[docker.io:{Endpoints:[https://registry-1.docker.io]}] Configs:map[] Auths:map[] Headers:map[]} ImageDecryption:{KeyModel:} DisableTCPService:true StreamServerAddress:127.0.0.1 StreamServerPort:0 StreamIdleTimeout:4h0m0s EnableSelinux:false SelinuxCategoryRange:1024 SandboxImage:k8s.gcr.io/pause:3.2 StatsCollectPeriod:10 SystemdCgroup:false EnableTLSStreaming:false X509KeyPairStreaming:{TLSCertFile: TLSKeyFile:} MaxContainerLogLineSize:16384 DisableCgroup:false DisableApparmor:false RestrictOOMScoreAdj:false MaxConcurrentDownloads:3 DisableProcMount:false UnsetSeccompProfile: TolerateMissingHugetlbController:true DisableHugetlbController:true IgnoreImageDefinedVolumes:false} ContainerdRootDir:/var/lib/containerd ContainerdEndpoint:/run/containerd/containerd.sock RootDir:/var/lib/containerd/io.containerd.grpc.v1.cri StateDir:/run/containerd/io.containerd.grpc.v1.cri}
```

specifically the cni part:

```
CniConfig:{NetworkPluginBinDir:/nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin NetworkPluginConfDir:/etc/cni/net.d NetworkPluginMaxConfNum:1 NetworkPluginConfTemplate:}
```

---

This looks like the area it's logging from

https://github.com/AkihiroSuda/nerdctl/blob/master/run.go#L237-L239

I'm guessing this needs to be told where the CNI plugins bin is too?
https://github.com/AkihiroSuda/nerdctl/blob/master/network_template.go#L25-L61

AkihiroSuda:
Nerdctl does not use CRI, so you don't need to modify config.toml. The dir can be specified as an option for gocni.New

https://godoc.org/github.com/containerd/go-cni#WithPluginDir

06kellyjac:
I've added `gocni.WithPluginDir([]string{cniBinDir})` but I'm still getting the same error :thinking: 

`if cniNetwork, err = gocni.New(gocni.WithConfListBytes([]byte(defaultBridgeNetwork)), gocni.WithPluginDir([]string{cniBinDir})); err != nil {`

shishir-a412ed:
@06kellyjac Do you have CNI plugins installed under `/opt/cni/bin`?
What is the output of `ls /opt/cni/bin -lt` on your system?

06kellyjac:
@shishir-a412ed the point of the PR is to not have the CNI binaries in `/opt/cni/bin` using `--cni-bin-dir`
`nerdctl run --cni-bin-dir "/some/dir/bin" -it alpine`

in this case my `/some/dir/bin` looks like so:

```
λ ls -al /nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin
dr-xr-xr-x root root   4 KB Thu Jan  1 01:00:01 1970  .
dr-xr-xr-x root root   4 KB Thu Jan  1 01:00:01 1970  ..
.r-xr-xr-x root root 2.9 MB Thu Jan  1 01:00:01 1970  bandwidth
.r-xr-xr-x root root 3.2 MB Thu Jan  1 01:00:01 1970  bridge
.r-xr-xr-x root root 7.6 MB Thu Jan  1 01:00:01 1970  dhcp
.r-xr-xr-x root root 3.4 MB Thu Jan  1 01:00:01 1970  firewall
.r-xr-xr-x root root 2.4 MB Thu Jan  1 01:00:01 1970  flannel
.r-xr-xr-x root root 2.9 MB Thu Jan  1 01:00:01 1970  host-device
.r-xr-xr-x root root 2.5 MB Thu Jan  1 01:00:01 1970  host-local
.r-xr-xr-x root root   3 MB Thu Jan  1 01:00:01 1970  ipvlan
.r-xr-xr-x root root 2.5 MB Thu Jan  1 01:00:01 1970  loopback
.r-xr-xr-x root root 3.1 MB Thu Jan  1 01:00:01 1970  macvlan
.r-xr-xr-x root root 2.8 MB Thu Jan  1 01:00:01 1970  portmap
.r-xr-xr-x root root 3.1 MB Thu Jan  1 01:00:01 1970  ptp
.r-xr-xr-x root root 2.6 MB Thu Jan  1 01:00:01 1970  sbr
.r-xr-xr-x root root 2.3 MB Thu Jan  1 01:00:01 1970  static
.r-xr-xr-x root root 2.6 MB Thu Jan  1 01:00:01 1970  tuning
.r-xr-xr-x root root   3 MB Thu Jan  1 01:00:01 1970  vlan
```

shishir-a412ed:
@06kellyjac I was just curious since your original `error` was still looking for CNI plugins under `/opt/cni/bin` even after the override.

06kellyjac:
Yeah I'm not sure what is causing it to still look there. Im guessing my change on line 181 isnt quite right.

I should probably double check the value is correctly being populated but it wouldnt get past the check on line 176

shishir-a412ed:
@06kellyjac The problem is your [`go-cni.New`](https://godoc.org/github.com/containerd/go-cni#New) call ignores your new CNI bin dir and always goes with the default directory.

https://github.com/containerd/go-cni/blob/8fbf3637eb5f67bb16eaacd4ec23a82e4d33b3ec/cni.go#L97

I changed (hardcoded) [`L83`](https://github.com/containerd/go-cni/blob/8fbf3637eb5f67bb16eaacd4ec23a82e4d33b3ec/cni.go#L83) and [`L89`](https://github.com/containerd/go-cni/blob/8fbf3637eb5f67bb16eaacd4ec23a82e4d33b3ec/cni.go#L89) to my new path `[]string{"/opt/cni/new-bin"}`, just to test it out and that works.

However, you might have to open a PR to https://github.com/containerd/go-cni first to make that path configurable, in order to set it.

```
root@vagrant:~/go/src/github.com/nerdctl# nerdctl run --cni-bin-dir /opt/cni/new-bin --rm -it alpine
/ # ls
bin    dev    etc    home   lib    media  mnt    opt    proc   root   run    sbin   srv    sys    tmp    usr    var
/ #
```

06kellyjac:
Thanks @shishir-a412ed !
Ill get on that when i can

06kellyjac:
hmm if I put `gocni.WithPluginDir([]string{"/nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin"})` manually instead of my value from the cli it works fine.

06kellyjac:
Ok it does work just fine with the var too.......

I've been having issues where my saving changes to files in my editor has a delay so I might have been running make & the command against an old version of the code :upside_down_face:


```
sudo ./_output/nerdctl --debug run --cni-bin-dir "/nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin" --network host -it alpine
/ # cat /etc/alpine-release
3.12.1
```

06kellyjac:
I've updated the commit to be signed off

06kellyjac:
`logrus.Debugf("----- '%+v'", cniNetwork)`

```
DEBU[0000] ----- '&{config:{pluginDirs:[/nix/store/2y22b9dzl0mcdqcs17kskmlpr1mgivhl-cni-plugins-0.8.7/bin] pluginConfDir:/etc/cni/net.d pluginMaxConfNum:1 prefix:eth} cniConfig:0xc0003ac400 networkCount:1 networks:[0xc000380f60] RWMutex:{w:{state:0 sema:0} writerSem:0 readerSem:0 readerCount:0 readerWait:0}}'
```

Looks like `config.pluginDirs` is set but `cniConfig.Path` isnt...


swapping the order to `gocni.New(gocni.WithPluginDir([]string{cniBinDir}), gocni.WithConfListBytes([]byte(defaultBridgeNetwork)));` works

Should I add a comment about this??

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Changed var name to CNI_PATH: https://github.com/AkihiroSuda/nerdctl/commit/6f67e6fcf16dd038f8a3b2728fa72dcbbaafb730

CNI_PATH seems the official convention: https://github.com/containernetworking/cni/blob/v0.8.0/pkg/invoke/args.go#L71

PR #32, opened 2020-12-11
Implement registry auth

AkihiroSuda:


pkg/imgutil/dockerconfigresolver/dockerconfigresolver.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/containerd/containerd/remotes"
+	"github.com/containerd/containerd/remotes/docker"
+	dockercliconfig "github.com/docker/cli/cli/config"
+	dockercliconfigtypes "github.com/docker/cli/cli/config/types"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// New instantiates a resolver using $DOCKER_CONFIG/config.json .
```
AkihiroSuda, line 31: @fuweid @ktock
Do you know whether there is an existing package for this?

fuweid: Unfortunately, don't see the package 😂

ktock: Me too.

Issue #33, opened 2020-12-11
Support `nerdctl save [CONTAINER IMAGE ID]`


soharaki:
Support nerdctl save [CONTAINER IMAGE ID].

soharaki:
@AkihiroSuda 
Hi, when I export a tarball directly using `client.Export` will output the OCI Image Spec tar instead of a Docker Image Spec v1.2 tar.

* Any `client.Export` options or alt-implementations which can export the Docker Image Spec v1.2?
* Should I explicitly explain that it is the OCI Image Spec tar in the help?

Any advices. Thank you.

* refs: 

```golang
func saveContainer(images []string, out io.Writer, saveOpts []archive.ExportOpt, clicontext *cli.Context) error {
	client, ctx, cancel, err := newClient(clicontext)
	if err != nil {
		return err
	}
	defer cancel()

	// Set the default platform
	saveOpts = append(saveOpts, archive.WithPlatform(platforms.Default()))

	is := client.ImageService()
	for _, img := range images {
		saveOpts = append(saveOpts, archive.WithImage(is, img))
	}

	return client.Export(ctx, out, saveOpts...)
}
```

AkihiroSuda:
OCI Archive with manifest.json is compatible with Docker Image Spec

PR #34, opened 2020-12-11
Add support for nerdctl save.

soharaki:
Refs: #33
Signed-off-by: Sho Haraki <endurance.test.20380119@gmail.com>


AkihiroSuda:
This seems an equivalent of `docker save`, not `docker export`.

`docker export` should produce a bare rootfs tar, while `docker save` should produce Docker Image Spec v1.2 tar.

AkihiroSuda:
Could you change the PR to `nerdctl save` ? Also, I think we are unlikely to support `skip-manifest-json` flag.

export.go
```diff
@@ -0,0 +1,107 @@
...
+package main
+
+import (
+	"io"
+	"os"
+
+	"github.com/containerd/containerd/images/archive"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli/v2"
+	terminal "golang.org/x/crypto/ssh/terminal"
```
AkihiroSuda, line 27: This library is smaller: https://github.com/mattn/go-isatty

export.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func exportContainer(id string, out io.Writer, exportOpts []archive.ExportOpt, clicontext *cli.Context) error {
+	client, ctx, cancel, err := newClient(clicontext)
+	if err != nil {
+		return err
+	}
+	defer cancel()
+
+	container, err := client.LoadContainer(ctx, id)
```
soharaki, line 93: I'll remove those src.

soharaki, line 93: I'll remove those lines.

export.go
```diff
@@ -0,0 +1,107 @@
...
+	Name:   "export",
+	Usage:  "Export a container's filesystem as a tar archive",
+	Action: exportAction,
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    "output",
+			Aliases: []string{"o"},
+			Usage:   "Write to a file, instead of STDOUT",
+		},
+		&cli.BoolFlag{
```
soharaki, line 40: Remove this option.

PR #37, opened 2020-12-12
Add Github action to run CI on pull requests

daxmc99:
I saw this project has a `.golangci.yml` file already. This should enable golangci-lint to run as part of pull requests. 

Signed-off-by: Dax McDonald <31839142+daxmc99@users.noreply.github.com>


.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+
+on:
+  push:
+    branches:
+    - '*'
+  pull_request:
+    branches:
+    - '*'
+
+jobs:
```
AkihiroSuda, line 11: Can we unify build.yml into this?

daxmc99: Unified, the `golangci-lint` should be in a separate step though since it utilizes caching to provide better performance.

AkihiroSuda (approved):
Thanks

PR #41, opened 2020-12-13
stop/kill: update actions with idutil

fuweid:
* support to stop/kill paused container(s)
* use task.Wait to check status of container

Signed-off-by: Wei Fu <fuweid89@gmail.com>


AkihiroSuda (approved):
Thanks

PR #44, opened 2020-12-14
Print message if container is not running.

shishir-a412ed:
Signed-off-by: Shishir Mahajan <smahajan@roblox.com>


shishir-a412ed:
Catching up a bit late on #41 changes, but IMO we should keep the behavior closer to `docker` behavior.

This is `docker` behavior when we try to `kill` a stopped container.

```
SM-smahajan628:~ smahajan$ docker kill 1ed12c6567b8
Error response from daemon: Cannot kill container: 1ed12c6567b8: Container 1ed12c6567b81a2ad417a464fcee7fcacd6e4c7dae956985233f1bfec5afb269 is not running
```

/cc @AkihiroSuda @fuweid

AkihiroSuda (approved):

PR #82, opened 2021-03-05
mv *.go ./cmd/nerdctl

AkihiroSuda:
- Users using `go install github.com/AkihiroSuda/nerdctl` will need to change their installation script to  `go install github.com/AkihiroSuda/nerdctl/cmd/nerdctl`
- Users using `make` or the binary distribution are unaffected


AkihiroSuda:
cc @06kellyjac [Your nix package](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/cluster/nerdctl/default.nix) will need to be updated, sorry

06kellyjac:
If you still want to move stuff into `./cmd/nerdctl` but don't have the time, I'm happy to give it a go.
If you prefer how it is now then no worries.

Also the nix package would probably build fine even with the move just because of how `buildGoModule` locates go files to build but I check it every version update :)

AkihiroSuda:
Thanks, but it seems better to keep the current file hierarchy, because moving the files to `./cmd/nerdctl` directory seemed to mess up registration of `go test` flags such as `-test.target=(nerdctl|docker)`.

AkihiroSuda:
Reopened as https://github.com/containerd/nerdctl/pull/269

Issue #121, opened 2021-03-18
Can't login to ghcr.io


AkihiroSuda:
```console
$ nerdctl pull ghcr.io/stargz-containers/python:3.9-org
FATA[0000] expected the hostname part of ac.ServerAddress ("ghcr.io") to be authConfigHostname="ghcr.io", got "" 
```

`~/.docker/config.json`:
```json
{
        "auths": {
                "ghcr.io": {
                        "auth":  "<base64>"
                }
        }
}
```

nerdctl version: v0.7.1

AkihiroSuda:
cc @ktock

PR #122, opened 2021-03-19
Allow to authenticate with non-URL ServerAddress

ktock:
Fixes: #121

Though `dockerconfigresolver` currently allows ServerAddress only to be URL, `docker/config.json` accepts non-URL ServerAddress that doesn't contain scheme (http://, https://). This results in nerdctl cannot authenticate to the registry using dockerconfig that contains non-URL ServerAddress.

This commit solves this issue by using `credentials.ConvertToHostname` of `github.com/docker/cli` instead of `net/url` pkg.

cc @AkihiroSuda


AkihiroSuda (approved):
Thanks!

PR #129, opened 2021-03-21
implement some runtime flags

fahedouch:

- [x]   nerdctl run --sysctl

- [x]   nerdctl run --cpuset-cpus

- [x]   nerdctl run --cpu-shares

### DIFF
--- a/run_runtime.go
+++ b/run_runtime.go
@@ -17,11 +17,15 @@
 package main
 
 import (
+	"context"
 	"strings"
 
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/containers"
+	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/plugin"
 	runcoptions "github.com/containerd/containerd/runtime/v2/runc/options"
+	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 )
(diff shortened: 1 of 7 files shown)


run_system.go
```diff
@@ -0,0 +1,41 @@
+/*
```
AkihiroSuda, line 1: I think we can have this in the existing `run_runtime.go`

run_runtime.go
```diff
@@ -54,3 +63,46 @@ func generateRuntimeCOpts(clicontext *cli.Context) ([]containerd.NewContainerOpt
...
+			s.Linux.Sysctl = make(map[string]string)
+		}
+		for k, v := range sysctls {
+			s.Linux.Sysctl[k] = v
+		}
+		return nil
+	}
+}
+
+func WithResources(resources *LinuxContainerResources) oci.SpecOpts {
```
AkihiroSuda, line 83: Should be around here: https://github.com/containerd/nerdctl/blob/master/run_cgroup.go#L55-L61

AkihiroSuda (approved):
Thanks,  could you also update README to add the new flags?

Would be great if we can had integration tests as well

AkihiroSuda:
Also please consider cleaning up the commit message (git commit -a -s —amend)

fahedouch:
> Thanks, could you also update README to add the new flags?
> 
> Would be great if we can had integration tests as well

Done. Can you review plz ?

AkihiroSuda (approved):
Thanks!

PR #132, opened 2021-03-22
suppress "install CNI isolation plugin" warning for `--net=bridge`

AkihiroSuda:
Previously, the warning was printed on every `nerdctl run` (`--net=bridge`).

The warning is now printed only on `nerdctl network create`.


fuweid (approved):
LGTM

fahedouch:
@AkihiroSuda i cannot figure out why you removed the isolation of the default network ?

AkihiroSuda:
I didn't remove anything, CNI upstream doesn't support isolation

Issue #146, opened 2021-04-01
CI: switch away from Docker Hub


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/ed8eec8dbb8a62e9f1634137be6a1bc6cdc4b34c/pkg/testutil/testutil.go#L276-L282

Docker Hub has rate limit, and OTOH `mirror.gcr.io` seems unstable:
```
docker: Error response from daemon: manifest for mirror.gcr.io/library/nginx:1.19-alpine not found: manifest unknown: Failed to fetch "1.19-alpine" from request "/v2/library/nginx/manifests/1.19-alpine".
```

We are currently using ~`mirror.gcr.io`~ (EDIT: switched back to plain old Docker Hub: https://github.com/containerd/nerdctl/pull/148)  but should switch to something else.

Requirements:
- No rate limit
- Stable
- Provides multi-arch images for `alpine:3.13`, `nginx:1.19-alpine`, and `registry:2`

AkihiroSuda:
@ktock 
Do you think we can use `org` images under `ghcr.io/stargz-snapshotter`? (Are they guaranteed to be multi-arch?)

ktock:
>  Do you think we can use org images under ghcr.io/stargz-snapshotter? (Are they guaranteed to be multi-arch?)

Makes sense.
They are copied using `crane` by ggcr so the entire image (including all platforms) is copied.

AkihiroSuda:
I switched back from mirror.gcr.io to plain old Docker Hub in https://github.com/containerd/nerdctl/pull/148.

I'll use `ghcr.io/stargz-snapshotter/<IMAGE>:<TAG>-org` if we hit rate limit again.

fahedouch:
we can specify credentials for container downloading: https://github.blog/changelog/2020-09-24-github-actions-private-registry-support-for-job-and-service-containers/  and increase the rate limit

AkihiroSuda:
Not seeing rate limit on macOS on GHA recently, closing

AkihiroSuda:
Seeing rate limit again

``        
521
[...]

PR #147, opened 2021-04-01
Dockerfile: verify fetched binaries

AkihiroSuda:
Verify the binaries to detect compromised ones.

This PR also switches back from `mirror.gcr.io` to Docker Hub to make CI green (#146)


ktock (approved):
LGTM

For making the contribution easier, we'll want helper scripts something like this ( https://github.com/kubernetes-sigs/kind/blob/8fe8b962521db017e3ffb22fc410f521dd8bc154/images/base/update-shasums.sh )

PR #152, opened 2021-04-05
stargz: add integration test

AkihiroSuda:


run_stargz_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestRunStargz(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	requiresStargz(base)
+	// if stargz snapshotter is functional, "/.stargz-snapshotter" appears
+	base.Cmd("--snapshotter=stargz", "run", "--rm", testutil.FedoraESGZImage, "ls", "/.stargz-snapshotter").AssertOK()
+}
```
ktock, line 31: Thanks for this. Can we also add a check if all snapshots are prepared as remote snapshots by Stargz Snapshotter? Stargz Snapshotter adds a label `containerd.io/snapshot/remote` for each remote snapshot so we can check it using this label.

e.g.

https://github.com/containerd/stargz-snapshotter/blob/17b648dd9cd4a4b859762c07b85d302e166c3387/script/pullsecrets/create-pod.sh#L84-L105

AkihiroSuda: We don't have CLI (`nerdctl image inspect --mode=native`) for that yet

AkihiroSuda: Can we merge this PR as-is?

ktock: Yes, we can.

ktock (approved):

PR #153, opened 2021-04-05
Support anonymous volumes (`nerdctl run -v /foo` and Dockerfile `VOLUME /foo`)

### LINKED ISSUE #67: Support anonymous volumes (`docker run -v /foo` and Dockerfile `VOLUME /foo`)
Support anonymous volumes:
- `docker run -v /foo`
- Dockerfile `VOLUME /foo`

Anonymous volumes are deleted on container deletion if `docker run` is invoked with `--rm`.

However, anonymous volumes are NOT deleted on invocation of `docker rm`.

AkihiroSuda:
Anonymous volumes are deleted on container deletion if `nerdctl run` is invoked with `--rm`.

However, anonymous volumes are NOT deleted on invocation of `nerdctl rm`, unless `--volumes` is specified.

Fix #67

### DIFF
--- a/pkg/mountutil/mountutil.go
+++ b/pkg/mountutil/mountutil.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/containerd/containerd/sys"
+	"github.com/containerd/nerdctl/pkg/idgen"
 	"github.com/containerd/nerdctl/pkg/mountutil/volumestore"
 	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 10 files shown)


pkg/mountutil/mountutil.go
```diff
@@ -29,63 +30,83 @@ import (
...
-	if len(split) < 2 || len(split) > 3 {
-		return nil, errors.Errorf("failed to parse %q", s)
-	}
-	src, dst := split[0], split[1]
-	if !strings.Contains(src, "/") {
-		// assume src is a volume name
-		vol, err := volStore.Get(src)
+	switch len(split) {
+	case 1:
+		dst = s
```
ktock, line 47: Don't we need a check if `dst` is an absolute path?

run.go
```diff
@@ -736,6 +731,13 @@ func withInternalLabels(ns, name, hostname, containerStateDir string, networks [
 	if logURI != "" {
 		m[labels.LogURI] = logURI
 	}
+	if len(anonVolumes) > 0 {
+		anonVolumeJSON, err := json.Marshal(anonVolumes)
+		if err != nil {
+			return nil, err
+		}
+		m[labels.AnonymousVolumes] = string(anonVolumeJSON)
```
ktock, line 739: Doesn't it hit 4096 char limit of label when many anonVolumes are specified?

https://github.com/containerd/containerd/blob/261c107ffc4ff681bc73988f64e3f60c32233b37/metadata/containers.go#L292-L296

pkg/mountutil/mountutil.go
AkihiroSuda: Surprisingly, we don't need: `docker run -v foo alpine` mounts an anon volume on `/foo`

run.go
AkihiroSuda: Who needs such number of anon volumes

ktock (approved):

Issue #28, opened 2020-12-10
Windows support


awakecoding:
I am looking for a way to use containerd on Windows Server 2019 with a Docker CLI replacement like this one. I see there are Linux builds, but has a Windows port been attempted?

Long story short, containerd for Windows should handle graceful shutdowns for containers, but the current version of Docker is still using HCS v1 + some really broken stuff under the hood, meaning Windows containers have never supported graceful shutdowns (https://github.com/microsoft/Windows-Containers/issues/16).

There is no clear migration path for Docker to adopt containerd on Windows in the short term, so I would be fine with installing containerd on Windows (https://kubernetes.io/docs/setup/production-environment/container-runtimes/#install-containerd) but then I just need a proper CLI on top of it, other than 'ctr' which is very primitive and marked as unsupported/unstable.

I am not familiar with Go, but how difficult would it be to attempt a Windows port of this tool?

AkihiroSuda:
Contribution is wanted 😃

jsturtevant:
I took a quick look and it would be doable but would require some work, there are assumption that this would run on linux like (there are probably more):

https://github.com/containerd/nerdctl/blob/06f3a7c8913abfae11095ee0ebbba0f013cc3f6f/exec.go#L229

https://github.com/containerd/nerdctl/blob/435f5331df3208b9e297290b4e8bec4fe2bd4a2e/run.go#L762-L766

https://github.com/containerd/nerdctl/blob/435f5331df3208b9e297290b4e8bec4fe2bd4a2e/pkg/lockutil/lockutil.go#L33

https://github.com/containerd/nerdctl/blob/435f5331df3208b9e297290b4e8bec4fe2bd4a2e/pkg/containerinspector/containerinspector.go#L73

jsturtevant:
I got started on this. Will likely want to break this down into many different PRs

AkihiroSuda:
Thanks!

jsturtevant:
Got it to build and connect 🚀 

```
PS C:\projects\nerdctl\_output> .\nerdctl.exe ps -a
CONTAINER ID    IMAGE                                        COMMAND                   CREATED          STATUS                   PORTS    NAMES
multi-arch      k8s.gcr.io/pause:3.4.1                       "/pause.exe"              3 days ago       Created
multi-arch2     k8s.gcr.io/pause:3.4.1                       "/pause.exe"              8 minutes ago    Created
[...]

PR #164, opened 2021-04-11
Windows support

jsturtevant:
This refactors to allow building a Windows binary as well as adds `nerdctl run`  support. This a draft to figure out/show the amount of work.   This should be broken down into smaller PRs that can be reviewed easier.  See #28 for discussion


Makefile
```diff
@@ -57,19 +57,19 @@ install:
 TAR_FLAGS=--transform 's/.*\///g' --owner=0 --group=0
 
 artifacts: clean
-	GOOS=linux GOARCH=amd64       make -C $(CURDIR)  binaries
+	GOOS=$(GOOS) GOARCH=amd64       make -C $(CURDIR)  binaries
```
AkihiroSuda, line 60: This produces `nerdctl-$(VERSION_TRIMMED)-linux-amd64.tar.gz`, so `GOOS` has to be hard-coded to `linux` here.

Instead, we should add `GOOS=windows` that produces  `nerdctl-$(VERSION_TRIMMED)-windows-amd64.zip`

client_windows.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, nil, nil, err
+	}
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
+	return client, ctx, cancel, nil
+}
+
+// getDataStore returns a string like "/var/lib/nerdctl/1935db59".
+// "1935db9" is from `$(echo -n "/run/containerd/containerd.sock" | sha256sum | cut -c1-8)``
+func getDataStore(clicontext *cli.Context) (string, error) {
```
AkihiroSuda, line 48: How is this different from Linux implementation?

client_windows.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/opencontainers/go-digest"
+	"github.com/urfave/cli/v2"
+)
+
+func newClient(clicontext *cli.Context) (*containerd.Client, context.Context, context.CancelFunc, error) {
```
AkihiroSuda, line 30: This does not necessarily need to be a separate file. 
The containerd address can be changed with `if runtime.GOOS == "windows"{}`.

pkg/defaults/defaults_windows.go
```diff
@@ -0,0 +1,55 @@
...
+package defaults
+
+import (
+	"fmt"
+)
+
+const AppArmorProfileName = ""
+
+func DataRoot() string {
+	return "C:\\ProgramData\\containerd\\root"
```
AkihiroSuda, line 26: Should read `%ALLUSERSPROFILE`?

pkg/defaults/defaults_windows.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func CNIPath() string {
+	return "C:\\Program Files\\containerd\\cni\\bin"
```
AkihiroSuda, line 30: Should read `%ProgramFiles%`?

pkg/infoutil/infoutil.go
```diff
@@ -17,106 +17,22 @@
...
-	for _, f := range utsname.Machine {
-		if f == 0 {
-			break
-		}
-		s += string(f)
-	}
-	return s
-}
-
 const UnameO = "GNU/Linux"
```
AkihiroSuda, line 34: Should be in `infoutil_linux.go`

pkg/lockutil/lockutil_windows.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"os"
+	"syscall"
+	"unsafe"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// LockFile from https://github.com/boltdb/bolt/blob/master/bolt_windows.go using MIT
```
AkihiroSuda, line 28: Please use permalink

pkg/netutil/netutil_windows.go
```diff
@@ -0,0 +1,58 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package netutil
+
+const (
+	DefaultNetworkName = "nat"
+	DefaultID          = 0
+	DefaultCIDR        = "10.1.0.0/24"
```
AkihiroSuda, line 22: Why is this different from Linux?

pkg/netutil/netutil_windows.go
```diff
@@ -0,0 +1,58 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package netutil
+
+const (
+	DefaultNetworkName = "nat"
```
AkihiroSuda, line 20: Why is this different from Linux?

pkg/ocihook/ocihook_linux.go
```diff
@@ -0,0 +1,30 @@
...
+
+package ocihook
+
+import (
+	"github.com/containerd/containerd/contrib/apparmor"
+	"github.com/containerd/nerdctl/pkg/defaults"
+	"github.com/sirupsen/logrus"
+)
+
+func loadapparmor() {
```
AkihiroSuda, line 25: nit: `loadAppArmor`

run.go
```diff
@@ -59,7 +51,7 @@ var runCommand = &cli.Command{
 	Name:         "run",
 	Usage:        "Run a command in a new container",
 	Action:       runAction,
-	BashComplete: runBashComplete,
+	BashComplete: runComplete,
```
AkihiroSuda, line 54: no

run.go
```diff
@@ -173,7 +165,7 @@ var runCommand = &cli.Command{
 		&cli.StringFlag{
 			Name:  "runtime",
 			Usage: "Runtime to use for this container, e.g. \"crun\", or \"io.containerd.runsc.v1\"",
-			Value: plugin.RuntimeRuncV2,
+			Value: "io.containerd.runhcs.v1",
```
AkihiroSuda, line 168: Linux runtime shouldn't change

run_windows.go
```diff
@@ -0,0 +1,135 @@
...
+
+	opts = append(opts, oci.WithImageConfig(image.Image))
+	cOpts = append(cOpts, containerd.WithImage(image.Image))
+	cOpts = append(cOpts, containerd.WithSnapshotter(image.Snapshotter))
+	cOpts = append(cOpts, containerd.WithNewSnapshot(id, image.Image))
+
+	//if len(args) > 0 {
+	//	opts = append(opts, oci.WithProcessArgs(args...))
+	//}
+	if cwd := clicontext.String("cwd"); cwd != "" {
```
AkihiroSuda, line 75: Flag name looks different from Linux

pkg/netutil/netutil_windows.go
TBBle: There isn't a "bridge" network type for Windows AFAIK, and the Windows CNI plugins (at least the ones from https://github.com/microsoft/windows-container-networking: `nat`, `sdnoverlay`, and `sdnbridge`) require that the network name equals the network type: https://github.com/microsoft/windows-container-networking/issues/57

`sdnoverlay` and `sdnbridge` require extra system setup beyond the CNI config file, so they seem like poor choices for default usage. Third-party CNI plugins are equally likely to require extra system configuration.

AkihiroSuda: The network name is expected to match Docker's default network, not the name of CNI plugin.

pkg/defaults/defaults_windows.go
TBBle: Specifically,
```
filepath.Join(os.Getenv("ProgramData"), "containerd", "root")
```
is how [`containerd.defaults.DefaultRootDir` is defined](https://github.com/containerd/containerd/blob/2486eb8ff8cf36b189108661d8e7f809fc288b60/defaults/defaults_windows.go#L27-L29).

pkg/defaults/defaults_windows.go
TBBle:
```
filepath.Join(os.Getenv("ProgramFiles"), "containerd", "cni", "bin")
```
is how [`containerd.pkg.cri.config.DefaultConfig().CniConfig.NetworkPluginBinDir` is defined](https://github.com/containerd/containerd/blob/2486eb8ff8cf36b189108661d8e7f809fc288b60/pkg/cri/config/config_windows.go#L33).

pkg/netutil/netutil_windows.go
TBBle: Ah, I misunderstood. I thought this was what ended up substituted as as `.Name` in `ConfigListTemplate`.

jsturtevant: When using Docker the default network on Windows is called `nat` so I called it that here. Are you saying this should stay `bridge`?

pkg/netutil/netutil_windows.go
jsturtevant: I don't think it needs to be, I was using the same values that I had manually configured to test

run.go
jsturtevant: yup! it was just to make it work.  This will need to be extracted so when built for Windows it is defaulted correctly.

pkg/netutil/netutil_windows.go
AkihiroSuda: nat SGTM then

pkg/defaults/defaults_windows.go
AkihiroSuda: The "data root" of nerdctl has nothing to do with containerd's data root.
So this should be like `filepath.Join(os.Getenv("ProgramData"), "nerdctl")`

jsturtevant:
closing in favor of a more iterative approach.  I did take much of the feedback here into consideration. Thanks for the early reviews!

PR #167, opened 2021-04-13
compose: implement config and secret objects

AkihiroSuda:
Only file-type objects are supported.
See also `./docs/compose.md` .


fuweid (commented):
LGTM

PR #184, opened 2021-04-23
Windows part 1

jsturtevant:
This is the first part step of adding Windows support. See https://github.com/containerd/nerdctl/issues/28#issuecomment-817348708 for the plan to do this in stages.   This is focused on make it build and mostly move Linux specific functionality to `_linux.go` files and stubs in `_windows.go` files.  The next PR will add ability to `nerdctl run` containers

Having all the changes in one PR (#164) made it difficult to see where differences were introduced and felt riskier than doing it in stages and should make reviewing faster.

This only enables Nerdctl to be built on Windows.  It doesn't add full support, which will come in later PR's.  Some functionality does work though, such as :

Pulling images
```
PS C:\projects\nerdctl\_output> .\nerdctl.exe pull jsturtevant/pause:3.6-windows-amd64-20H2
docker.io/jsturtevant/pause:3.6-windows-amd64-20H2:                               resolved       |++++++++++++++++++++++++++++++++++++++|
manifest-sha256:f0aad192a1b106ae51e3eef56d5f807830c97c2a086b038e3e04fcca34d2c0b2: done           |++++++++++++++++++++++++++++++++++++++|
config-sha256:23388efa1a0be06e8962f8fc8a756c860441b2402ededa8f098b055e7368a630:   done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:e4ebc87f3dce90c29691c1e41b3b18715d02b40889528750271bdf1ac116d8d9:    done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:9ef56e96c7a394f4f55704ab032a9092c38e27435e711a85ad3712de7e5b2b09:    done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:48ac2dabfb13cb3463b3938efc623ad9815a5d261b96b48eeeaf52eab055ba40:    done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:033263e01d18471a352ee800110d55b007ef31690c6fc40c1f97769ff84460d4:    done           |++++++++++++++++++++++++++++++++++++++|
elapsed: 59.8s                                                                    total:  1.3 Mi (23.1 KiB/s)      
```

listing images
```  
PS C:\projects\nerdctl\_output> .\nerdctl.exe image ls
REPOSITORY           TAG                       IMAGE ID        CREATED               SIZE
jsturtevant/pause    3.6-windows-amd64-20H2    f0aad192a1b1    About a minute ago    101.2 MiB
```

listing running containers
```
PS C:\projects\nerdctl\_output> .\nerdctl.exe ps
CONTAINER ID    IMAGE                                                 COMMAND         CREATED               STATUS    PORTS    NAMES
container1     docker.io/jsturtevant/pause:3.6-windows-amd64-20H2    "/pause.exe"    About a minute ago    Up          
```

listing namespaces:
```
PS C:\projects\nerdctl\_output> .\nerdctl.exe namespace ls
NAME       CONTAINERS    IMAGES    VOLUMES
default    1             1         0
```


Makefile
```diff
@@ -72,6 +75,9 @@ artifacts: clean
 	GOOS=linux GOARCH=s390x       make -C $(CURDIR) binaries
 	tar $(TAR_FLAGS) -czvf $(CURDIR)/_output/nerdctl-$(VERSION_TRIMMED)-linux-s390x.tar.gz   _output/nerdctl extras/rootless/*
 
+	GOOS=windows GOARCH=amd64       make -C $(CURDIR) binaries
+	tar $(TAR_FLAGS) -czvf $(CURDIR)/_output/nerdctl-$(VERSION_TRIMMED)-windows-amd64.tar.gz   _output/nerdctl extras/rootless/*
```
AkihiroSuda, line 79: Windows users may prefer zip over tar.gz? idk

AkihiroSuda: `extras/rootless` is useless for Windows

client_windows.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, nil, nil, err
+	}
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
+	return client, ctx, cancel, nil
+}
+
+// getDataStore returns a string like "/var/lib/nerdctl/1935db59".
+// "1935db9" is from `$(echo -n "/run/containerd/containerd.sock" | sha256sum | cut -c1-8)``
+func getDataStore(clicontext *cli.Context) (string, error) {
```
AkihiroSuda, line 48: Does not need to be in `_windows.go` perhaps

client_windows.go
```diff
@@ -0,0 +1,70 @@
...
+		return "", err
+	}
+	dataStore := filepath.Join(dataRoot, addrHash)
+	if err := os.MkdirAll(dataStore, 0700); err != nil {
+		return "", err
+	}
+	return dataStore, nil
+}
+
+func getAddrHash(addr string) (string, error) {
```
AkihiroSuda, line 64: Does not need to be in `_windows.go` perhaps

exec_linux.go
```diff
@@ -0,0 +1,41 @@
...
+*/
+
+package main
+
+import (
+	"github.com/containerd/containerd/pkg/cap"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func setCapabilities(pspec *specs.Process) error {
```
AkihiroSuda, line 24: `setExecCapabilities`, because this is different from the caps for `nerdctl run`

run.go
```diff
@@ -273,7 +272,7 @@ func runAction(clicontext *cli.Context) error {
 		oci.WithMounts([]specs.Mount{
 			{Type: "cgroup", Source: "cgroup", Destination: "/sys/fs/cgroup", Options: []string{"ro", "nosuid", "noexec", "nodev"}},
 		}),
-		oci.WithoutRunMount, // unmount default tmpfs on "/run": https://github.com/containerd/nerdctl/issues/157
+		WithoutRunMount(), // unmount default tmpfs on "/run": https://github.com/containerd/nerdctl/issues/157
```
AkihiroSuda, line 275: Just `if runtime.GOOS == "linux" { opts = append(opts, oci.WithoutRunMount) }`

AkihiroSuda:
Thanks, could you squash commits

Makefile
jsturtevant: We have `tar` on windows these days.  containerd uses tar.gz and i think it makes sense match that

run.go
jsturtevant: did it this way becuase `oci.WithoutRunMount` is in a Linux tagged file so it fails when building. :disappointed:

Makefile
jsturtevant: removed `extras/rootless`

client_windows.go
jsturtevant: I started to keep this all in one file but went back.  I take a look again

login_windows.go
```diff
@@ -0,0 +1,29 @@
...
+	"bufio"
+	"os"
+)
+
+func readPassword() (string, error) {
+	reader := bufio.NewReader(os.Stdin)
+	pwd, _ := reader.ReadString('\n')
+
+	return pwd, nil
+}
```
TBBle, line 29: Why not use the same logic as the Linux version, using `term.ReadPassword(syscall.Stdin)` if `term.IsTerminal`, but simply failing if it is not, rather than opening `/dev/tty` as a fallback?

main_windows.go
```diff
@@ -0,0 +1,37 @@
...
+	return
+}
+
+func bashCompleteNamespaceNames(clicontext *cli.Context) {
+	return
+}
+
+func bashCompleteSnapshotterNames(clicontext *cli.Context) {
+	return
+}
```
TBBle, line 37: Is there a reason to disable bash completion output on Windows builds? Or is this due to lack of Windows support in the nerdctl packages the bash-completion functions use?

jsturtevant: These functions wouldn't compile due to assumption for Linux. I didn't think that bash completion would function Windows so left them blank.

login_windows.go
jsturtevant: It was failing to build but looking closer I just needed to convert `Stdin` to `int` and it works.  Thanks!

client_windows.go
jsturtevant: updated

client_windows.go
jsturtevant: updated

login_windows.go
jsturtevant: updated

main_windows.go
TBBle: Fair 'nuff for now. We can come back to it later. Most of my team uses bash on Window as their default shell (which surprised me when I joined them, and brings a bunch of its own issues), but without this PR, we don't have bash completion output on Windows anyway, so it's not a regression.

Makefile
```diff
@@ -72,6 +75,9 @@ artifacts: clean
 	GOOS=linux GOARCH=s390x       make -C $(CURDIR) binaries
 	tar $(TAR_FLAGS) -czvf $(CURDIR)/_output/nerdctl-$(VERSION_TRIMMED)-linux-s390x.tar.gz   _output/nerdctl extras/rootless/*
 
+	GOOS=windows GOARCH=amd64       make -C $(CURDIR) binaries
+	tar $(TAR_FLAGS) -czvf $(CURDIR)/_output/nerdctl-$(VERSION_TRIMMED)-windows-amd64.tar.gz
```
AkihiroSuda, line 79: append `_output/nerdctl.exe`

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/containerd/nerdctl
 go 1.16
 
 require (
+	github.com/Microsoft/go-winio v0.4.17 // indirect
```
AkihiroSuda, line 6: `go mod tidy` seems needed to make CI happy

main_windows.go
AkihiroSuda: > Most of my team uses bash on Window as their default shell 

WSL bash? Cygwin bash? MinGW bash?

AkihiroSuda:
LGTM except CI failure, thanks!

main_windows.go
TBBle: Git Bash, I believe, so Cygwin-derived via MSYS2, but has a different env-var for disabling `/`-path conversion.

Makefile
jsturtevant: updated

Issue #50, opened 2021-01-01
`nerdctl images` should show unpacked size


AkihiroSuda:
`nerdctl images` is expected to show unpacked size as in `docker images`
```console
$ docker images
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
alpine       latest    389fef711851   2 weeks ago   5.58MB
```

The current implementation shows packed size:
```console
$ nerdctl images
REPOSITORY    TAG       IMAGE ID        CREATED          SIZE
alpine        latest    3c7497bf0c7a    5 seconds ago    2.7 MiB
```

https://github.com/AkihiroSuda/nerdctl/blob/5bb45da65b3e16ae69cde5ce8bf3a5b80c2d22ba/images.go#L83

fahedouch:
Hi @AkihiroSuda ,

I take a look to fix this bug. 
The idea is to calculate the sum of snapshots size corresponding to the each image.

I took the image  `ghcr.io/stargz-containers/golang:1.15.3-buster-esgz` as example.
```
snapshotter tree for ghcr.io/stargz-containers/golang:1.15.3-buster-esgz

sha256:807a937cec3d3bef7dab91841bc93cd43911c9e61d3a9b68be6afd6acd1b45f8
  \_ sha256:d08af225de79d044daf29b516dfb371ddb77e33b6a7fb07761816e456395491f
    \_ sha256:546ae4f9b99882e15dcccde39406b0893c9d906052d424a38ee1529858ad99c1
      \_ sha256:a9d4a1d0e99c11d05ddeca7b21ea8e96781cc91a4489dcc028c0c65e97310fcf
        \_ sha256:2806e6decb8e52eaadbd48e8e5176474b8f39ce9a92ef0b3ce78655c2b3b07af
          \_ sha256:7c4f4c436344ad8950f290525c9dffdde5dcc43f02cc8b87251f94aee138a389
            \_ sha256:0ff02ab503d2007f3d550bb1039be75ca2f2afa6f9db6ccfcf7b40257fe44c57
``` 

When I make the sum of each  above  layer size I got a number different than what docker show.
can you guide me in solving this bug please ?

thanks

AkihiroSuda:
implemented in https://github.com/containerd/nerdctl/pull/205 , thanks @fahedouch

PR #205, opened 2021-05-11
unpacked img size

fahedouch:
fixing https://github.com/containerd/nerdctl/issues/50
Considering that the image unpacked size is the  image ChainID snapshot size

### DIFF
--- a/images.go
+++ b/images.go
@@ -22,11 +22,12 @@ import (
 	"strings"
 	"text/tabwriter"
 
+	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/pkg/progress"
-	"github.com/containerd/containerd/platforms"
 	refdocker "github.com/containerd/containerd/reference/docker"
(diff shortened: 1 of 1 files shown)


images.go
```diff
@@ -94,7 +95,7 @@ func printImages(ctx context.Context, clicontext *cli.Context, imageList []image
 	}
 
 	for _, img := range imageList {
-		size, err := img.Size(ctx, cs, platforms.DefaultStrict())
+		size, err := unpackedImageSize(ctx, clicontext, client, img)
 		if err != nil {
 			return errors.Wrap(err, "failed to compute image size")
```
AkihiroSuda, line 100: This should be logrus.Warn, not return err

AkihiroSuda: (Multiple images may face this warning depending on snapshotter, so a warning should be printed only once)

fahedouch: @AkihiroSuda changes applied

AkihiroSuda (changes requested):
LGTM, but needs squashing

AkihiroSuda (approved):

PR #200, opened 2021-04-29
Support copying files in volume

fahedouch:
fixing  https://github.com/containerd/nerdctl/issues/165

PR #216, opened 2021-05-18
fix volume clean up & add test

fahedouch:
After/When Copying initial content into volume, the cleanup process should include this steps (which is not the case before this PR) :  

1-  unmount mounts
2-  clean snapshot
3-  remove tmp directory 

This fix display defer errors with logurs in order to avoid overriding  error returned by the function 

and I added `TestCopyingUpInitialContentsOnVolumeShouldRetainHardlink` test

@fuweid 
@AkihiroSuda 

related PR https://github.com/containerd/nerdctl/pull/200

### DIFF
--- a/run_mount.go
+++ b/run_mount.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"io/ioutil"
 	"os"
+	"os/signal"
 	"path/filepath"
 
 	"github.com/containerd/containerd"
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Please describe what is the original issue you are seeing, and how to confirm that the issue was fixed in this commit

fahedouch:
> Please describe what is the original issue you are seeing, and how to confirm that the issue was fixed in this commit

I updated the the description. I am facing a problem with  `test / test-docker-compatibility`

```
the test is TestCopyingUpInitialContentsOnVolumeShouldRetainHardlink  https://github.com/containerd/nerdctl/blob/ddd3f1bb8d769b2fb3b93ecf56457d3ced5b7289/run_mount_test.go#L161

As you see I am trying to test that mount volume keep hardlink.

My test should output  2  to stdout which is the result of CMD ["stat","-c","%s","../../../../../../../../../../../../../../../../../../etc/passwd"]  when executing the test command base.Cmd("run", "--rm", imageName).AssertOutContains(expected)

In my local every thing is good but in GA runners the result is always 1 it looks like not keeping hard link when running in docker GA or sthg is going wrong :confused:

I check my docker conf  avec the one in GA and they are the same. Any idea what should disturb my test
```

apostasie:
@fahedouch is this still something we need?

fahedouch:
Let's close it and open later if needed

PR #223, opened 2021-05-20
Setup nerdctl top cmd

fahedouch:
same as `docker top` 

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,8 @@ It does not necessarily mean that the corresponding features are missing in cont
     - [:whale: nerdctl events](#whale-nerdctl-events)
     - [:whale: nerdctl info](#whale-nerdctl-info)
     - [:whale: nerdctl version](#whale-nerdctl-version)
+  - [Stats management](#stats-management)
+    - [:whale: nerdctl top](#whale-nerdctl-top)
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
Ideally we should use github.com/containers/psgo rather than exec-ing `ps`, but this refactoring can be done later.

fahedouch:
running the ps command in a rootless container with cgroupv1 is not possible because rootless container does not rely on cgroupv1.
the problem does not arise in the case of rootlfull (v1 or v2) && rootless ( v2) 
The idea to get around this problem is to join the mount namespace of rootless container when we are dealing with rootless container on cgroupv1 and fetch process info inside the pid mount ns. 

FYI : for pid host is like doing a ps on the host and dealing with the host `/proc` so we should add this to the condition when implementing ` --pid=host`

I prefer making this refactoring later when using `psgo` rather than exec-ing ps. It handles better this use cases.

fahedouch:
Cc @AkihiroSuda

AkihiroSuda (approved):
LGTM if green, thanks

fahedouch:
Pid host is implemented here https://github.com/containerd/nerdctl/pull/273
The case of host pid should be handled when refacto with psgo

Issue #215, opened 2021-05-18
Low hanging fruits


AkihiroSuda:
PR is welcome.

####  Lint
- [x] https://github.com/containerd/nerdctl/issues/1133
- [x] #1134

Others: https://golangci-lint.run/usage/linters/
#### Docs
- Improve "FAQ & Troubleshooting" guide https://github.com/containerd/nerdctl/blob/master/docs/faq.md

#### Others

[`is:issue is:open label:"good first issue"`](https://github.com/containerd/nerdctl/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)

PR #226, opened 2021-05-24
add `nerdctl run --cidfile <FILE>`

smowafy:
First flag in the list in https://github.com/containerd/nerdctl/issues/215

One thing I'm not sure about is the category of this flag

### DIFF
--- a/run.go
+++ b/run.go
@@ -227,6 +227,10 @@ var runCommand = &cli.Command{
 			Name:  "label-file",
 			Usage: "Read in a line delimited file of labels",
 		},
+		&cli.StringFlag{
+			Name:  "cidfile",
+			Usage: "Write the container ID to the file",
+		},
 	},
 }
 
@@ -256,6 +260,12 @@ func runAction(clicontext *cli.Context) error {
 		cOpts []containerd.NewContainerOpts
 		id    = idgen.GenerateID()
 	)
+
+	if cidfile := clicontext.String("cidfile"); cidfile != "" {
+		if err := writeCIDFile(cidfile, id); err != nil {
+			return err
+		}
+	}
 
 	dataStore, err := getDataStore(clicontext)
 	if err != nil {
@@ -732,5 +742,23 @@ func withInternalLabels(ns, name, hostname, containerStateDir string, networks [
 func propagateContainerdLabelsToOCIAnnotations() oci.SpecOpts {
 	return func(ctx context.Context, oc oci.Client, c *containers.Container, s *oci.Spec) error {
 		return oci.WithAnnotations(c.Labels)(ctx, oc, c, s)
+	}
+}
+
+func writeCIDFile(path, id string) error {
+	if _, err := os.Stat(path); err == nil {
+		return errors.Errorf("container ID file found, make sure the other container isn't running or delete %s", path)
+	} else if errors.Is(err, os.ErrNotExist) {
+		f, err := os.Create(path)
+		defer f.Close()
+		if err != nil {
+			return errors.Errorf("failed to create the container ID file: %s", err)
+		}
+		if _, err := f.WriteString(id); err != nil {
+			return err
+		}
+		return nil
+	} else {
+		return err
 	}
 }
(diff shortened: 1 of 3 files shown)


run_test.go
```diff
@@ -137,3 +137,18 @@ func TestRunExitCode(t *testing.T) {
 	assert.Equal(base.T, "exited", inspect123.State.Status)
 	assert.Equal(base.T, 123, inspect123.State.ExitCode)
 }
+
+func TestRunCIDFile(t *testing.T) {
```
AkihiroSuda, line 141:
```
=== RUN   TestRunCIDFile
    run_test.go:145: assertion failed: expression is false: res.ExitCode == exitCode: time="2021-05-24T17:38:22Z" level=fatal msg="image name needs to be specified"
        
--- FAIL: TestRunCIDFile (0.02s)
```

run.go
```diff
@@ -734,3 +746,18 @@ func propagateContainerdLabelsToOCIAnnotations() oci.SpecOpts {
...
 }
+
+func writeCIDFile(path, id string) error {
+	if _, err := os.Stat(path); err == nil {
+		return errors.Errorf("Container ID file found, make sure the other container isn't running or delete %s", path)
+	}
+	f, err := os.Create(path)
+	defer f.Close()
+	if err != nil {
+		return errors.Errorf("Failed to create the container ID file: %s", err)
```
AkihiroSuda, line 757: nit: error string should start with a lower character in Go

run.go
```diff
@@ -257,6 +261,14 @@ func runAction(clicontext *cli.Context) error {
 		id    = idgen.GenerateID()
 	)
 
+	cidfile := clicontext.String("cidfile")
+
+	if cidfile != "" {
```
AkihiroSuda, line 266: `if cidfile := clicontext.String("cidfile"); cidfile != "" {`

run.go
```diff
@@ -227,6 +227,10 @@ var runCommand = &cli.Command{
 			Name:  "label-file",
 			Usage: "Read in a line delimited file of labels",
 		},
+		&cli.StringFlag{
+			Name:  "cidfile",
+			Usage: "Write the container ID to the file",
+		},
```
AkihiroSuda, line 233: Could you also update README.md to include `cidfile`

run.go
```diff
@@ -734,3 +744,18 @@ func propagateContainerdLabelsToOCIAnnotations() oci.SpecOpts {
 		return oci.WithAnnotations(c.Labels)(ctx, oc, c, s)
 	}
 }
+
+func writeCIDFile(path, id string) error {
+	if _, err := os.Stat(path); err == nil {
+		return errors.Errorf("container ID file found, make sure the other container isn't running or delete %s", path)
+	}
+	f, err := os.Create(path)
```
fahedouch, line 752: I think you should do smthing like this to handles err différent from `isNotExist`

```
if _, err := os.Stat("/path/to/whatever"); err == nil {
  // throw erro
} else if os.IsNotExist(err) {
  // your logic here

} else {
  // handle other erro
}
```

run.go
smowafy: The error message was inspired from https://github.com/docker/cli/blob/master/cli/command/container/create.go#L180, changed it to lowercase though

run.go
smowafy: To which category though?

AkihiroSuda: `Metadata flags:` would be fine, I think

run.go
smowafy: I added returning other errors, please check if this way makes sense. Also I used `errors.Is` according to https://golang.org/pkg/os/#IsExist

run_test.go
smowafy: Replaced by just asserting failure if the file exists. Is that enough or should it be more specific?

run.go
smowafy: Added

AkihiroSuda (approved):
Thanks

Issue #225, opened 2021-05-24
nerdctl load -i busybox.tar（digest not found


mpv945:
sudo nerdctl pull busybox:stable
sudo nerdctl save -o busybox.tar busybox:stable
sudo nerdctl rmi busybox:stable

sudo nerdctl load -i busybox.tar

FATA[0000] content digest sha256:9708dc541771170e42aadc4c3c4477ec15e91e595673493bbbb78b7c8b75b0b5: not found

AkihiroSuda:
Should be resolved in https://github.com/containerd/nerdctl/pull/227

Thanks @fahedouch

PR #227, opened 2021-05-25
set DefaultStrict platform for image loading

fahedouch:
fixing https://github.com/containerd/nerdctl/issues/225

### DIFF
--- a/load.go
+++ b/load.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/images/archive"
+	"github.com/containerd/containerd/platforms"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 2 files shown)


load.go
```diff
@@ -54,7 +55,7 @@ func loadAction(clicontext *cli.Context) error {
 }
 
 func loadImage(in io.Reader, clicontext *cli.Context) error {
-	client, ctx, cancel, err := newClient(clicontext)
+	client, ctx, cancel, err := newClient(clicontext, containerd.WithDefaultPlatform(platforms.DefaultStrict()))
```
AkihiroSuda, line 58: client.Import should receive the platform option, not client itself

AkihiroSuda (changes requested):

load.go
fahedouch: @AkihiroSuda  the `platformMatcher` refers to  client option
https://github.com/containerd/containerd/blob/96df20cf770479393342047a62182961f9205d62/import.go#L128

AkihiroSuda (approved):

Issue #228, opened 2021-05-29
task.Delete() should always be called in nerdctl run command without rm option


liubin:
When using another runtime than `runc`, for example:

```
$ nerdctl run --runtime='io.containerd.kata.v2' -it containerstack/alpine-stress:latest /bin/bash
```

After exiting from the shell, the kata shim process will remain, after comparing with `ctr`, I found [this commit](https://github.com/containerd/nerdctl/pull/112/files#diff-05d67ab64d2cc3aeb2a3d4b1b1de5b145c06a2371ddf94660d326f5aa265ab6d)

The must-calling  `task.Delete()` will only be called with `--rm` option, but the `--rm` is used to control if delete a container, but not a task.

Here is the [ctr's implementation](https://github.com/containerd/containerd/blob/v1.4.3/cmd/ctr/commands/run/run.go#L159-L161):

```golang
if context.Bool("rm") && !detach {
    defer container.Delete(ctx, containerd.WithSnapshotCleanup)
}
```

`ctr` is deleting container if `rm` option set.

liubin:
This issue has been fixed in the latest code.

PR #229, opened 2021-05-29
fix invalid action for `nerdctl run --rm`

liubin:
This commit includes two changes:

- always call task.Delete after task finished like ctr
- call container.Delete if `--rm` option set

Fixes: #228

Signed-off-by: bin liu <liubin0329@gmail.com>


AkihiroSuda:
Thanks, but CI is failing

liubin:
For the first failed case `TestRunExitCode`(`run_test.go`), the status has been set to [`Created`](https://github.com/containerd/nerdctl/blob/master/ps.go#L134-L137), the same behaver as `ctr`. So this case may need to adjust to as what `ctr` has done.

And the second failed case is because the task want to wait is deleted if not started with `-d` option. 

I updated the test cases, please check again.

AkihiroSuda:
> For the first failed case TestRunExitCode(run_test.go), the status has been set to Created, the same behaver as ctr. So this case may need to adjust to as what ctr has done.

The output of nerdctl should match docker output, regardless to the internal state of containerd.

liubin:
Docker uses a `config.json` file to store the container state(task state indeed) and display the state for `ps` command. Without this file, can't get the container state if the task is deleted.

@AkihiroSuda do you have any idea to solve this issue?

AkihiroSuda:
I thought OCI poststop hook could be used to store the state, but it might not support storing exit code...
https://github.com/opencontainers/runtime-spec/blob/1c3f411f041711bbeecf35ff7e93461ea6789220/runtime.md

AkihiroSuda:
I guess we need to start a small daemon (as an OCI hook) that watches exit status?

cc @fuweid

AkihiroSuda:
Can we just store the exit code as a new JSON file on `task.Delete`?

AkihiroSuda:
@liubin WDYT?

liubin:
@AkihiroSuda If save the task state in a JSON file, this tool will become stateful and have to take responsibility for manage the JSON file, but there has no daemon process like `dockerd`. Maybe it's hard to keep compatibility with Docker without saving any state.

liubin:
In the latest code the issue has been fixed, so close this.

PR #249, opened 2021-06-11
Add `nerdctl run --env-file <FILE>`

ilyee:
Signed-off-by: Zuti He <ilyeeelihe@gmail.com>

Related to #215

### DIFF
--- a/run.go
+++ b/run.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -215,6 +216,10 @@ var runCommand = &cli.Command{
 			Aliases: []string{"e"},
 			Usage:   "Set environment variables",
 		},
+		&cli.StringSliceFlag{
+			Name:  "env-file",
+			Usage: "Set environment variables from file",
+		},
 		// metadata flags
 		&cli.StringFlag{
 			Name:  "name",
@@ -305,6 +310,14 @@ func runAction(clicontext *cli.Context) error {
 
 	if wd := clicontext.String("workdir"); wd != "" {
 		opts = append(opts, oci.WithProcessCwd(wd))
+	}
+
+	if envFiles := strutil.DedupeStrSlice(clicontext.StringSlice("env-file")); len(envFiles) > 0 {
+		env, err := parseEnvVars(envFiles)
+		if err != nil {
+			return err
+		}
+		opts = append(opts, oci.WithEnv(env))
 	}
 
 	if env := strutil.DedupeStrSlice(clicontext.StringSlice("env")); len(env) > 0 {
@@ -782,4 +795,29 @@ func writeCIDFile(path, id string) error {
 	} else {
 		return err
 	}
+}
+
+func parseEnvVars(paths []string) ([]string, error) {
+	vars := make([]string, 0)
+	for _, path := range paths {
+		f, err := os.Open(path)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to open env file %s", path)
+		}
+		defer f.Close()
+
+		sc := bufio.NewScanner(f)
+		for sc.Scan() {
+			line := strings.TrimSpace(sc.Text())
+			// skip comment lines
+			if strings.HasPrefix(line, "#") {
+				continue
+			}
+			vars = append(vars, line)
+		}
+		if err = sc.Err(); err != nil {
+			return nil, err
+		}
+	}
+	return vars, nil
 }
--- a/run_test.go
+++ b/run_test.go
@@ -157,3 +157,27 @@ func TestRunShmSize(t *testing.T) {
 
 	base.Cmd("run", "--rm", "--shm-size", shmSize, testutil.AlpineImage, "/bin/grep", "shm", "/proc/self/mounts").AssertOutContains("size=32768k")
 }
+
+func TestRunEnvFile(t *testing.T) {
+	base := testutil.NewBase(t)
+
+	const pattern = "env-file"
+	file1, err := ioutil.TempFile("", pattern)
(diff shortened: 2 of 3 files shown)


run.go
```diff
@@ -783,3 +796,26 @@ func writeCIDFile(path, id string) error {
...
+	vars := make([]string, 0)
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, errors.Errorf("failed to open env file %s: %v", path, err)
+	}
+	defer f.Close()
+
+	sc := bufio.NewScanner(f)
+	for sc.Scan() {
+		line := sc.Text()
```
AkihiroSuda, line 810: strings.TrimSpace(line)

run.go
```diff
@@ -783,3 +796,26 @@ func writeCIDFile(path, id string) error {
 		return err
 	}
 }
+
+func getEnvVars(path string) ([]string, error) {
```
AkihiroSuda, line 800: Func name should be like "parseEnvFile"

run_test.go
```diff
@@ -157,3 +157,24 @@ func TestRunShmSize(t *testing.T) {
 
 	base.Cmd("run", "--rm", "--shm-size", shmSize, testutil.AlpineImage, "/bin/grep", "shm", "/proc/self/mounts").AssertOutContains("size=32768k")
 }
+
+func TestRunEnvFile(t *testing.T) {
+	base := testutil.NewBase(t)
+	const fileName = "env.file"
```
AkihiroSuda, line 163: https://golang.org/pkg/io/ioutil/#TempFile

run_test.go
```diff
@@ -157,3 +157,24 @@ func TestRunShmSize(t *testing.T) {
...
+
+	f, err := os.Create(fileName)
```
AkihiroSuda, line 165: https://golang.org/pkg/io/ioutil/#WriteFile

run.go
```diff
@@ -238,6 +239,10 @@ var runCommand = &cli.Command{
 			Name:  "shm-size",
 			Usage: "Size of /dev/shm",
 		},
+		&cli.StringFlag{
```
AkihiroSuda, line 242: Perhaps StringSliceFlag to support multiple files

AkihiroSuda: Also, should be defined right after "env" flag

run.go
```diff
@@ -783,3 +796,26 @@ func writeCIDFile(path, id string) error {
 		return err
 	}
 }
+
+func getEnvVars(path string) ([]string, error) {
+	vars := make([]string, 0)
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, errors.Errorf("failed to open env file %s: %v", path, err)
```
AkihiroSuda, line 804: errors.Wrapf

run.go
ilyee: > Perhaps StringSliceFlag to support multiple files

Agree with that

run_test.go
```diff
@@ -157,3 +157,29 @@ func TestRunShmSize(t *testing.T) {
...
+	file2, err := ioutil.TempFile("", pattern)
+	assert.NilError(base.T, err)
+	path2 := file2.Name()
+	defer file2.Close()
+	defer os.Remove(path2)
+	err = ioutil.WriteFile(path2, []byte("# this is a comment line\nTESTKEY2=TESTVAL2"), 0666)
+	assert.NilError(base.T, err)
+
+	base.Cmd("run", "--rm", "--env-file", path1, "--env-file", path2, testutil.AlpineImage,
+		"/usr/bin/env", "|", "grep", "TESTKEY1", "|", "awk", "-F=", "'{print $2}'").AssertOutContains("TESTVAL1")
```
AkihiroSuda, line 182: Why not just echo $TESTKEY1

run.go
```diff
@@ -215,6 +216,10 @@ var runCommand = &cli.Command{
 			Aliases: []string{"e"},
 			Usage:   "Set environment variables",
 		},
+		&cli.StringSliceFlag{
+			Name:  "env-file",
+			Usage: "Set environment variables from file",
```
fahedouch, line 221: files(s)

run_test.go
ilyee: Sry for the mistake I made, but here I found when I use echo as init command, the output env var is not correct, same thing happens using `env` flag like:
`nerdctl --namespace=nerdctl-test run --rm -e TESTKEY=TESTVALUE ghcr.io/stargz-containers/alpine:3.13-org echo $TESTKEY`
Output should be `TESTVALUE` but actually nothing. And when I attached to the container and run `echo $TESTKEY`, it seems working. (this also happens with `docker run`). And then I found host env will be outputed when echo is init command by `echo $PATH`.

Is it normal? Or something wrong with my host's env or config?

AkihiroSuda: Probably you forgot to quote `'$TESTKEY'`

ilyee: > Probably you forgot to quote `'$TESTKEY'`

quote doesn't work, instead `$TESTKEY` will be the output

AkihiroSuda: "sh" , "-c", "echo '$TESTKEY'"

fahedouch: you should escape  the $ with `\` in the runner shell .
Try : 
`nerdctl --namespace=nerdctl-test run --rm -e TESTKEY=TESTVALUE ghcr.io/stargz-containers/alpine:3.13-org sh -c " echo \$TESTKEY"`

ilyee: > you should escape the $ with `\` in the runner shell .

thx a lot, that works

fahedouch: it looks like test pass without escaping the $.  How did you manage to fix it ?

AkihiroSuda (approved):

AkihiroSuda:
Thanks

PR #264, opened 2021-06-24
nerdctl-full: compile runc with CGO_ENABLED=1

AkihiroSuda:
For opencontainers/runc#3037


fahedouch:
@AkihiroSuda  do we really need to activate the `GCO`  ? is it for cross-compiling needs?  because it has many drawbacks like making the compilation slower and then slower build.

AkihiroSuda:
Yes, we really need, for enabling seccomp.

ktock (approved):

Issue #266, opened 2021-06-24
nerdctl-full: compile containerd and runc as static binaries


AkihiroSuda:
As of 9fea1fe8bd17e4aac80f885b81cb2cb0ca5e5d6a , containerd and runc are dynamically linked with glibc and do not work on Alpine

```console
$ ldd bin/* libexec/cni/*
bin/buildctl:
        not a dynamic executable
bin/buildkitd:
        not a dynamic executable
bin/containerd:
        linux-vdso.so.1 (0x0000ffffb9419000)
        libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000ffffb6f5d000)
        libdl.so.2 => /lib/aarch64-linux-gnu/libdl.so.2 (0x0000ffffb6f49000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffffb6dcc000)
        /lib/ld-linux-aarch64.so.1 (0x0000ffffb93e6000)
bin/containerd-fuse-overlayfs-grpc:
        not a dynamic executable
bin/containerd-rootless-setuptool.sh:
        not a dynamic executable
bin/containerd-rootless.sh:
        not a dynamic executable
bin/containerd-shim-runc-v2:
        not a dynamic executable
bin/containerd-stargz-grpc:
        not a dynamic executable
bin/containerd-stress:
        linux-vdso.so.1 (0x0000ffff85bbd000)
        libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000ffff85b5a000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffff859dd000)
        /lib/ld-linux-aarch64.so.1 (0x0000ffff85b8a000)
bin/ctd-decoder:
        not a dynamic executable
bin/ctr:
        linux-vdso.so.1 (0x0000ffffbd6ae000)
        libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000ffffbd64b000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffffbd4ce000)
        /lib/ld-linux-aarch64.so.1 (0x0000ffffbd67b000)
bin/ctr-enc:
        not a dynamic executable
bin/ctr-remote:
        not a dynamic executable
bin/fuse-overlayfs:
        not a dynamic executable
bin/nerdctl:
        not a dynamic executable
[...]

PR #267, opened 2021-06-24
nerdctl-full: build runc as a static binary

AkihiroSuda:
#266


fahedouch:
@AkihiroSuda  `static` target on runc enable `CGO` by default 
https://github.com/opencontainers/runc/blob/51beb5c436b159ae2d483b219c37ecfde13b006a/Makefile#L29

no need to have `ENV CGO_ENABLED=1` in our dockerfile

AkihiroSuda:
Yes, but I want to specify that env explicitly

ktock (approved):

Issue #261, opened 2021-06-22
Support bind propagation options for `--volume`


ktock:
Currently, `rw` and `ro` are supported as options for `nerdctl run --volume`.  It would be great if we support bind propagation options like ones in docker (https://docs.docker.com/storage/bind-mounts/#configure-bind-propagation)

https://github.com/containerd/nerdctl/blob/39a46444754c581e020fffa63261f2a859c16bac/pkg/mountutil/mountutil.go#L87-L96

PR #268, opened 2021-06-25
Support bind propagation options for `--volume`

ktock:
Fixes https://github.com/containerd/nerdctl/issues/261

This PR adds the following [docker-compatible propagation option](https://docs.docker.com/storage/bind-mounts/#configure-bind-propagation) to bind mounts:

- `shared`: The host and the container share mount events in the bind mount each other.
- `slave`: The container receives mount events in the bind mount from the host. But the host doesn't receive events from the container.
- `private`: Mount events in the bind mount doesn't propagate each other.
- `rshared`: Similar as `shared` but the propagation type is applied to all mounts under the bind mount.
- `rslave`:  Similar as `slave` but the propagation type is applied to all mounts under the bind mount.
- `rprivate` (default): Similar as `private` but the propagation type is applied to all mounts under the bind mount.

For more details about propagation, please see also [shared subtree document of linux kernel](https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt).

```console
# nerdctl run --rm -it \
          -v /tmp/shared-dir:/shared-dir:shared \
          -v /tmp/slave-dir:/slave-dir:slave \
          -v /tmp/non-shared-dir:/non-shared-dir \
          ubuntu:20.04 /bin/bash
root@01302060790a:/# cat /proc/$$/mountinfo | grep -- "-dir"
2080 1905 0:113 /shared-dir /shared-dir rw,nosuid,nodev,relatime shared:865 - tmpfs tmpfs rw,mode=777
2081 1905 0:113 /slave-dir /slave-dir rw,nosuid,nodev,relatime master:865 - tmpfs tmpfs rw,mode=777
2082 1905 0:113 /non-shared-dir /non-shared-dir rw,nosuid,nodev,relatime - tmpfs tmpfs rw,mode=777
```

### DIFF
--- a/pkg/mountutil/mountutil.go
+++ b/pkg/mountutil/mountutil.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/sys"
 	"github.com/containerd/nerdctl/pkg/idgen"
 	"github.com/containerd/nerdctl/pkg/mountutil/volumestore"
@@ -38,13 +39,14 @@ type Processed struct {
 	Mount           specs.Mount
 	AnonymousVolume string // name
 	Type            string
+	Opts            []oci.SpecOpts
 }
 
 func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error) {
 	var (
 		res      Processed
 		src, dst string
-		ro       bool
+		options  []string
 	)
 	split := strings.Split(s, ":")
 	switch len(split) {
@@ -82,19 +84,19 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
 		if !filepath.IsAbs(dst) {
 			return nil, errors.Errorf("expected an absolute path, got %q", dst)
 		}
+		rawOpts := ""
 		if len(split) == 3 {
-			opts := strings.Split(split[2], ",")
-			for _, opt := range opts {
-				switch opt {
-				case "rw":
-					// NOP
-				case "ro":
-					ro = true
-				default:
-					logrus.Warnf("unsupported volume option %q", opt)
-				}
-			}
+			rawOpts = split[2]
 		}
+
+		// always call parseVolumeOptions for bind mount to allow the parser to add some default options
+		var err error
+		var specOpts []oci.SpecOpts
+		options, specOpts, err = parseVolumeOptions(res.Type, src, rawOpts)
+		if err != nil {
+			return nil, err
+		}
+		res.Opts = append(res.Opts, specOpts...)
 	default:
 		return nil, errors.Errorf("failed to parse %q", s)
 	}
@@ -102,10 +104,7 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
 		Type:        "none",
 		Source:      src,
 		Destination: dst,
-		Options:     []string{"rbind"},
-	}
-	if ro {
-		res.Mount.Options = append(res.Mount.Options, "ro")
+		Options:     append([]string{"rbind"}, options...),
 	}
 	if sys.RunningInUserNS() {
 		unpriv, err := getUnprivilegedMountFlags(src)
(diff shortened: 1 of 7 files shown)


pkg/mountutil/mountutil.go
```diff
@@ -82,30 +84,27 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
...
+		if err != nil {
+			return nil, err
 		}
+		res.Opts = append(res.Opts, specOpts...)
 	default:
 		return nil, errors.Errorf("failed to parse %q", s)
 	}
 	res.Mount = specs.Mount{
-		Type:        "none",
+		Type:        "bind",
```
AkihiroSuda, line 104: Why

pkg/mountutil/mountutil_linux.go
```diff
@@ -53,3 +62,154 @@ func getUnprivilegedMountFlags(path string) ([]string, error) {
...
+			}
+			if err := ensureMountOptionalValue(mi, "shared:", "master:"); err != nil {
+				return nil, nil, err
+			}
+
+			// See above comments about RootfsPropagation. Here we make sure that
+			// the mountpoint can be a slave of the host mount.
+			specOpts = append(specOpts, func(ctx context.Context, cli oci.Client, c *containers.Container, s *oci.Spec) error {
+				rootpg := s.Linux.RootfsPropagation
+				if rootpg != "shared" && rootpg != "rshared" && rootpg != "slave" && rootpg != "rslave" {
```
AkihiroSuda, line 168: Use switch{}

pkg/mountutil/mountutil.go
ktock: Because this is how docker does. But TIL runc ignores this for bind mounts.

https://github.com/opencontainers/runc/blob/v1.0.0/libcontainer/specconv/spec_linux.go#L344-L352

```golang
	if flags&unix.MS_BIND != 0 {
		// Any "type" the user specified is meaningless (and ignored) for
		// bind-mounts -- so we set it to "bind" because rootfs_linux.go
		// (incorrectly) relies on this for some checks.
		device = "bind"
		if !filepath.IsAbs(source) {
			source = filepath.Join(cwd, m.Source)
		}
	}
```

Reverted this to `"none"`.

pkg/mountutil/mountutil_linux.go
ktock: Fixed this.

pkg/mountutil/mountutil_linux.go
```diff
@@ -53,3 +62,158 @@ func getUnprivilegedMountFlags(path string) ([]string, error) {
...
+		switch opt {
+		case "rw":
+			writeModeRawOpts = append(writeModeRawOpts, opt)
+		case "ro":
+			writeModeRawOpts = append(writeModeRawOpts, opt)
+		case "private", "rprivate", "shared", "rshared", "slave", "rslave":
+			propagationRawOpts = append(propagationRawOpts, opt)
+		default:
+			logrus.Warnf("unsupported volume option %q", opt)
+		}
```
ktock, line 103: We should not print warn when `opt == ""`. I'll fix it.

ktock: done.

AkihiroSuda (approved):
thanks

pkg/mountutil/mountutil_windows.go
```diff
@@ -16,7 +16,43 @@
...
+	var writeModeRawOpts []string
+	for _, opt := range strings.Split(optsRaw, ",") {
+		switch opt {
+		case "rw":
+			writeModeRawOpts = append(writeModeRawOpts, opt)
+		case "ro":
+			writeModeRawOpts = append(writeModeRawOpts, opt)
+		default:
+			logrus.Warnf("unsupported volume option %q", opt)
+		}
```
ktock, line 49: Sorry. We should not print warn when `opt == ""` on windows as well. I'll fix it.

ktock: done

PR #269, opened 2021-06-25
mv *.go ./cmd/nerdctl

AkihiroSuda:
Reopen PR #82

- Users using `go install github.com/containerd/nerdctl` will need to change their installation script to  `go install github.com/containerd/nerdctl/cmd/nerdctl`
- Users using `make` or the binary distribution are unaffected


AkihiroSuda:
cc @06kellyjac  @fahedouch @ktock @fuweid PTAL

06kellyjac:
You can keep one main.go file in the root to keep the same go install path

something along the lines of this might work (after making Xmain exported):
```
package main

import (
	"github.com/containerd/nerdctl/cmd"
)

func main() {
	if err := cmd.Xmain(); err != nil {
		logrus.Fatal(err)
	}
}
```

AkihiroSuda:
Yes, but it complicates the package structure 😞

06kellyjac:
> Yes, but it complicates the package structure disappointed

True, you also say using `go install` is not recommended in the README so it's not a big deal causing changes there :+1:

06kellyjac (approved):
LGTM

ktock (approved):

fahedouch:
LGTM

Issue #278, opened 2021-07-06
How to skip certificate verification


jinnzy:
```
FATA[0000] failed to resolve reference "reg.xxx.cn/quay.io/coreos/etcd:v3.4.14": failed to do request: Head "https://reg.xxx.cn/v2/quay.io/coreos/etcd/manifests/v3.4.14": x509: certificate signed by unknown authority
```

containerd config.toml
```
    [plugins."io.containerd.grpc.v1.cri".registry]
      [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
      [plugins."io.containerd.grpc.v1.cri".registry.mirrors."reg.xxx.cn"]
        endpoint = ["https://reg.firstshare.cn"]
  [plugins."io.containerd.grpc.v1.cri".registry.configs]
      [plugins."io.containerd.grpc.v1.cri".registry.configs."reg.xxx.cn".tls]
        insecure_skip_verify = true
```

![image](https://user-images.githubusercontent.com/33449123/124562477-a6709980-de71-11eb-8e6f-784eaabf0a2e.png)

![image](https://user-images.githubusercontent.com/33449123/124562558-bee0b400-de71-11eb-9957-7e18292df593.png)

fahedouch:
Hi @jinnzy 

nerdctl doesn't rely on CRI, that's why the CRI configuration is ignored.

You can skips verifying HTTPS certs, and allows falling back to plain HTTP  in the client side with `nerdctl --insecure-registry`

jinnzy:
> Hi @jinnzy
> 
> nerdctl doesn't rely on CRI, that's why the CRI configuration is ignored.
> 
> You can skips verifying HTTPS certs, and allows falling back to plain HTTP in the client side with `nerdctl --insecure-registry`
> 
> I think that skipping tls it is not yet implemented for containerd API (not sure)

Hi @fahedouch 

Did I use it right? 

```
nerdctl  --insecure-registry=true  pull  reg.xxx.cn/quay.io/coreos/etcd:v3.4.14

WARN[0000] skipping verifying HTTPS certs for "reg.xxx.cn"
INFO[0000] trying next host                              error="failed to authorize: failed to fetch anonymous token: Get \"https://reg.xxx.cn/service/token?scope=repository%3Aquay.io%2Fcoreos%2Fetcd%3Apull&service=harbor-registry\": x509: certificate signed by unknown authority" host=reg.xxx.cn
[...]

PR #287, opened 2021-07-12
Apply skipping certification verification to authorization as well

ktock:
Fixes: #278

This commit fixes the issue that skipping certificate verification didn't applied to the access to the authorization server.
The behaviour that propagates skip-verification config to the authorizer is compatible to [the current behaviour of `ctr`](https://github.com/containerd/containerd/blob/main/cmd/ctr/commands/resolver.go#L95-L114).

### DIFF
--- a/pkg/imgutil/dockerconfigresolver/dockerconfigresolver.go
+++ b/pkg/imgutil/dockerconfigresolver/dockerconfigresolver.go
@@ -62,6 +62,11 @@ func New(refHostname string, optFuncs ...Opt) (remotes.Resolver, error) {
 		of(&o)
 	}
 	var authzOpts []docker.AuthorizerOpt
+	var insecureClient *http.Client
+	if o.skipVerifyCerts {
+		insecureClient = newInsecureClient()
+		authzOpts = append(authzOpts, docker.WithAuthClient(insecureClient))
+	}
 	if authCreds, err := NewAuthCreds(refHostname); err != nil {
 		return nil, err
 	} else {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Can we have a test like this? https://github.com/containerd/nerdctl/blob/f6f48cd29548226c2102f65c26bb7c807f0eca8d/cmd/nerdctl/push_test.go#L121

Can be a separate PR though

pkg/imgutil/dockerconfigresolver/dockerconfigresolver.go
```diff
@@ -77,15 +80,7 @@ func New(refHostname string, optFuncs ...Opt) (remotes.Resolver, error) {
...
-		tr := &http.Transport{
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-			},
-		}
-		client := &http.Client{
-			Transport: tr,
-		}
-		regOpts = append(regOpts, docker.WithClient(client))
+		regOpts = append(regOpts, docker.WithClient(newInsecureClient()))
```
AkihiroSuda, line 83: No need to call `newInsecureClient()` twice

ktock:
> Can we have a test like this?
> https://github.com/containerd/nerdctl/blob/f6f48cd29548226c2102f65c26bb7c807f0eca8d/cmd/nerdctl/push_test.go#L121
> Can be a separate PR though

I'll work on it. We'll need to host an authorization server in CI (maybe we can use https://github.com/cesanta/docker_auth).

AkihiroSuda:
@ktock Do you plan to work on tests before merging this PR or after?

ktock:
Sorry for the late. I add the test in this week but please merge it if this patch is needed urgently.

Issue #304, opened 2021-07-30
nerdctl can not use local load images to create containers


Huimintai:
I have the image tar package by `docker save`
I load the `registry.tar`by `nerdctl load -i  registry.tar`
```
# nerdctl load -i registry.tar
unpacking tkestack/registry-amd64:2.7.1 (sha256:3a8eef8d0a818b9bbb4bd17667253473e2d99935ccbbd37649af6bcaa064cf0d)...done
unpacking overlayfs@sha256:3a8eef8d0a818b9bbb4bd17667253473e2d99935ccbbd37649af6bcaa064cf0d (sha256:3a8eef8d0a818b9bbb4bd17667253473e2d99935ccbbd37649af6bcaa064cf0d)...done
```
Check the local images:
```
# nerdctl images
WARN[0000] unparsable image name "overlayfs@sha256:39b4bfdd8a183666591c8e2edeaaea7448539faddf4e8671c7abb04fdb520ed2"
WARN[0000] unparsable image name "overlayfs@sha256:3a8eef8d0a818b9bbb4bd17667253473e2d99935ccbbd37649af6bcaa064cf0d"
REPOSITORY                      TAG                           IMAGE ID        CREATED          SIZE
alpine                          latest                        234cb88d3020    9 minutes ago    5.9 MiB
tkestack/tke-installer-amd64    v1.6.0.113.g8e04f2e5.dirty    39b4bfdd8a18    3 minutes ago    8.0 KiB
                                                              39b4bfdd8a18    3 minutes ago    8.0 KiB
                                                              3a8eef8d0a81    4 seconds ago    8.0 KiB
tkestack/registry-amd64         2.7.1                         3a8eef8d0a81    4 seconds ago    8.0 KiB
```
Then I use nerdctl run the registry container:
```
# nerdctl run -it --rm tkestack/registry-amd64:2.7.1 sh
docker.io/tkestack/registry-amd64:2.7.1: resolving      |--------------------------------------|
elapsed: 29.9s                           total:   0.0 B (0.0 B/s)
INFO[0030] trying next host                              error="failed to do request: Head \"https://registry-1.docker.io/v2/tkestack/registry-amd64/manifests/2.7.1\": dial tcp 54.85.56.3:443: i/o timeout" host=registry-1.docker.io
[...]

PR #310, opened 2021-08-08
fix image size

fahedouch:
the actual image size is equal to the ChainID size.

this PR use the Walk API and Usage on each parent of  ChainID to calculate the total usage of unpacked image layers

related issue   https://github.com/containerd/nerdctl/issues/304 

### DIFF
--- a/cmd/nerdctl/images.go
+++ b/cmd/nerdctl/images.go
@@ -27,6 +27,7 @@ import (
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/pkg/progress"
 	refdocker "github.com/containerd/containerd/reference/docker"
+	"github.com/containerd/containerd/snapshots"
 	"github.com/containerd/nerdctl/pkg/imgutil"
 	"github.com/opencontainers/image-spec/identity"
 	"github.com/pkg/errors"
@@ -95,9 +96,11 @@ func printImages(ctx context.Context, clicontext *cli.Context, client *container
 		fmt.Fprintln(w, "REPOSITORY\tTAG\tIMAGE ID\tCREATED\tSIZE")
 	}
 
+	s := client.SnapshotService(clicontext.String("snapshotter"))
+
 	var errs []error
 	for _, img := range imageList {
-		size, err := unpackedImageSize(ctx, clicontext, client, img)
+		size, err := unpackedImageSize(ctx, clicontext, client, s, img)
 		if err != nil {
 			errs = append(errs, err)
 		}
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/images.go
```diff
@@ -143,19 +146,40 @@ func imagesBashComplete(clicontext *cli.Context) {
...
+	keys.add(ctx, s, &usage, keys[chainID])
 	return usage.Size, nil
 }
+
+// recursive function to calculate total usage of ChainID's parent
+func (keys snapshotKey) add(ctx context.Context, s snapshots.Snapshotter, usage *snapshots.Usage, key string) {
+	if key == "" {
+		return
+	}
+	u, _ := s.Usage(ctx, key)
```
ktock, line 182: Error shouldn't be ignored

cmd/nerdctl/images.go
```diff
@@ -143,19 +146,40 @@ func imagesBashComplete(clicontext *cli.Context) {
...
+	usage.Add(u)
+	keys.add(ctx, s, usage, keys[key])
+}
```
ktock, line 185: This (`snapshotKey.add`) should be put immediately following the type definition of `snapshotKey` (`type snapshotKey map[string]string`) .

ktock (commented):
overall LGTM but minor nits

fahedouch:
@ktock done

cmd/nerdctl/images.go
```diff
@@ -143,19 +146,47 @@ func imagesBashComplete(clicontext *cli.Context) {
...
 	}
 
+	keys := make(snapshotKey, 0)
+	walkFn := func(ctx context.Context, info snapshots.Info) error {
+		keys[info.Name] = info.Parent
+		return nil
+	}
+	if err := s.Walk(ctx, walkFn); err != nil {
+		return 0, err
+	}
```
ktock, line 185: Why does it scan all snapshots? Can we walk snapshots only in the chain? I think we can do this by following parents  from the topmost snapshot using `Stat` API.

fahedouch: @ktock thank you,  This is what I was looking for ==> `Stat` API . I made modification.

AkihiroSuda:
CI wasn't triggered?

cmd/nerdctl/images.go
```diff
@@ -143,19 +146,51 @@ func imagesBashComplete(clicontext *cli.Context) {
 	bashCompleteImageNames(clicontext)
 }
 
-func unpackedImageSize(ctx context.Context, clicontext *cli.Context, client *containerd.Client, i images.Image) (int64, error) {
+type snapshotKey string
```
AkihiroSuda, line 149: Why type?

AkihiroSuda (approved):
Thanks

PR #316, opened 2021-08-17
load: import all platforms (--all-platforms)

cpuguy83:
Currently if you try to import a non-native image the blobs will get
GC'd and unpack fails with with "not found" on the blobs.

Import doesn't really have a way to select which platforms to important,
so I don't think it makes sense to make this optional. After all the tar
could just be modified with the desired platforms.

### DIFF
--- a/cmd/nerdctl/load.go
+++ b/cmd/nerdctl/load.go
@@ -38,6 +38,10 @@ var loadCommand = &cli.Command{
 			Aliases: []string{"i"},
 			Usage:   "Read from tar archive file, instead of STDIN",
 		},
+		&cli.BoolFlag{
+			Name:  "all-platforms",
+			Usage: "Imports content for all platforms",
+		},
 	},
 }
 
@@ -62,7 +66,7 @@ func loadImage(in io.Reader, clicontext *cli.Context) error {
 	defer cancel()
 
 	sn := clicontext.String("snapshotter")
-	imgs, err := client.Import(ctx, in, containerd.WithDigestRef(archive.DigestTranslator(sn)))
+	imgs, err := client.Import(ctx, in, containerd.WithDigestRef(archive.DigestTranslator(sn)), containerd.WithAllPlatforms(clicontext.Bool("all-platforms")))
 	if err != nil {
 		return err
 	}


AkihiroSuda (changes requested):
This should be gated with `--all-platforms` as in `nerdctl image convert --all-platforms OLD NEW`

In addition to `nerdctl import`, `nerdctl pull` should follow the same semantics, though PR for `nerdctl pull` can be worked out separately
(Obviously nobody wants all the platforms to be pulled by default)

fahedouch:
Hello,


@cpuguy83  I do not agree with you. the import retrieves the platform to import from the client which is `platforms.DefaultStrict()`

the platform is applied on the client here : https://github.com/containerd/nerdctl/blob/47a4917e40f7923c213d12ec1eee1d9c58e0924a/cmd/nerdctl/load.go#L58

so then the `import` function will get it and initialize the `platformMatcher` with  `platforms.DefaultStrict()` : 
https://github.com/containerd/containerd/blob/fda782a7b941de5438e4fbbd38675bf23b69304b/import.go#L128


I agree with @AkihiroSuda making `all-platforms`  by default is not the it is not the best choice. I think we should keep `platforms.DefaultStrict()` and add `--all-platforms` flag

cpuguy83:
Either way is fine.
My main point is the user is supplying the tar to import. It seems strange to not import the whole tar, especially since it seems to be importing them anyway.

cpuguy83:
Added the `--all-platforms` flag.

fahedouch:
LGTM

cmd/nerdctl/load.go
```diff
@@ -38,6 +38,10 @@ var loadCommand = &cli.Command{
 			Aliases: []string{"i"},
 			Usage:   "Read from tar archive file, instead of STDIN",
 		},
+		&cli.BoolFlag{
+			Name:  "all-platforms",
+			Usage: "Imports content for all platforms, false by default",
```
AkihiroSuda, line 43: "false by default" can be removed, as `urfave/cli/v2` appends `(default: false)` automatically

AkihiroSuda (approved):
Thanks

PR #318, opened 2021-08-18
update Go (1.17), RootlessKit (0.14.5), fuse-overlayfs (1.7.1)

AkihiroSuda:


fahedouch:
I'm not sure if this needs to be updated or it's just an example : https://github.com/containerd/nerdctl/blob/47a4917e40f7923c213d12ec1eee1d9c58e0924a/Dockerfile#L199

AkihiroSuda:
it's just an example, no need to be updated

ktock (approved):

PR #315, opened 2021-08-17
pkg/defaults: Look at Debian's path for CNI plugins

vbatts:
Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

PR #321, opened 2021-08-20
pkg/defaults: CNI: search /usr/local/lib/cni, ~/.local/lib/cni

AkihiroSuda:
Follow-up to #315

### DIFF
--- a/pkg/defaults/defaults_linux.go
+++ b/pkg/defaults/defaults_linux.go
@@ -42,6 +42,7 @@ func DataRoot() string {
 func CNIPath() string {
 	candidates := []string{
 		"/usr/local/libexec/cni",
+		"/usr/local/lib/cni",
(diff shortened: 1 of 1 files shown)


vbatts:
what distro lands things in these paths?

AkihiroSuda:
Not for any distro, but /usr and /usr/local should be compatible

fahedouch:
FYI : podman is setting `/usr/local/lib/cni` path to look for cni plugin

https://github.com/containers/common/blob/8ca8dd9c0bdfc24e1e91211ec47dfb74d7c36b95/pkg/config/default.go#L82

vbatts (approved):
Right on. Any time there are a list of paths to search, it's good to know why.
A good next step would be to set order on which is searched. And then log it somewhere the path chosen.

PR #370, opened 2021-09-17
add `nerdctl run --ulimit <ulimit>`

piotrbelina:
Hello everybody,

I would like to propose my first change adding ulimits

Related to #215

This pull request adds `nerdctl run --ulimit <ulimit>`

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -263,6 +263,10 @@ var runCommand = &cli.Command{
 			Name:  "pidfile",
 			Usage: "file path to write the task's pid",
 		},
+		&cli.StringSliceFlag{
+			Name:  "ulimit",
+			Usage: "Ulimit options",
+		},
 	},
 }
 
@@ -510,6 +514,12 @@ func runAction(clicontext *cli.Context) error {
 			}
 		}
 	}
+
+	ulimitOpts, err := generateUlimitsOpts(clicontext)
+	if err != nil {
+		return err
+	}
+	opts = append(opts, ulimitOpts...)
 
 	rtCOpts, err := generateRuntimeCOpts(clicontext)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/nerdctl/run_ulimit.go
```diff
@@ -0,0 +1,56 @@
+/*
```
fahedouch, line 1: This may go under `run_runtime.go`

cmd/nerdctl/run.go
```diff
@@ -511,6 +515,12 @@ func runAction(clicontext *cli.Context) error {
 		}
 	}
 
+	ulimitOpts, err := generateUlimitsOpts(clicontext)
```
fahedouch, line 518: I think it is better to check `strutil.DedupeStrSlice(clicontext.StringSlice("ulimit"))` len here

cmd/nerdctl/run.go
```diff
@@ -511,6 +515,12 @@ func runAction(clicontext *cli.Context) error {
 		}
 	}
```
fahedouch, line 517: I think it is better to check the len of `strutil.DedupeStrSlice(clicontext.StringSlice("ulimit"))` here

cmd/nerdctl/run_ulimit.go
AkihiroSuda: run_ulimit.go is fine

cmd/nerdctl/run.go
AkihiroSuda: Why?

AkihiroSuda (approved):
Thanks

cmd/nerdctl/run.go
fahedouch: simply, no need to call `generateUlimitsOpts` if len == 0

fahedouch:
@AkihiroSuda if each time we will suffix the file run_x.go we will end up with several files run_x.go I think we must group by themes (runtime, network etc..) for a better reading

PR #391, opened 2021-09-27
Support local output in build

ehazlett:
This enables `--output` for build allowing different types of output (oci, local, etc).

### DIFF
--- a/cmd/nerdctl/build.go
+++ b/cmd/nerdctl/build.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"strings"
@@ -66,6 +67,11 @@ var buildCommand = &cli.Command{
 			Usage: "Do not use cache when building the image",
 		},
 		&cli.StringFlag{
+			Name:    "output",
+			Aliases: []string{"o"},
+			Value:   "type=docker",
+		},
+		&cli.StringFlag{
 			Name:  "progress",
 			Usage: "Set type of progress output (auto, plain, tty). Use plain to show container output",
 			Value: "auto",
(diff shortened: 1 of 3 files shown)


cmd/nerdctl/build.go
```diff
@@ -180,3 +188,7 @@ func generateBuildctlArgs(clicontext *cli.Context) (string, []string, error) {
 
 	return buildctlBinary, buildctlArgs, nil
 }
+
+func isLocalBuild(clicontext *cli.Context) bool {
+	return len(strutil.DedupeStrSlice(clicontext.StringSlice("tag"))) > 0
```
AkihiroSuda, line 193: This logic doesn't seem correct, why not parse --output

cmd/nerdctl/build_test.go
```diff
@@ -45,6 +45,32 @@ CMD ["echo", "nerdctl-build-test-string"]
...
+	defer os.RemoveAll(buildCtx)
+
+	if err := ioutil.WriteFile(filepath.Join(buildCtx, testFileName), []byte("nerdctl-test"), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	base.Cmd("build", "-o", fmt.Sprintf("type=local,dest=%s", outputDir), buildCtx).AssertOK()
+	if _, err := os.Stat(filepath.Join(outputDir, testFileName)); err != nil {
+		t.Fatal(err)
+	}
```
AkihiroSuda, line 71: Can we also assert the file content

cmd/nerdctl/build.go
ehazlett: That's how I had it but figured it was easier to just check if a tag is specified for the output.  I'll update it.  Thanks!

AkihiroSuda (approved):
Thanks, LGTM if CI gets green

cmd/nerdctl/build_test.go
```diff
@@ -45,6 +45,38 @@ CMD ["echo", "nerdctl-build-test-string"]
 	base.Cmd("run", "--rm", imageName).AssertOutContains("nerdctl-build-test-string")
 }
 
+func TestBuildLocal(t *testing.T) {
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 49: why?

ehazlett: Without it the test errored with not compatible.  I'm assuming that's because the regular `docker build` doesn't support "output" (it's in `buildx`).

PR #424, opened 2021-10-10
Add buildkitd-rootless startup script to extras

afbjorklund:
Similar to the containerd-rootless startup script

https://github.com/containerd/nerdctl/blob/master/docs/rootless.md

* `containerd-rootless.sh` (https://github.com/containerd/containerd/blob/main/docs/rootless.md)

* `buildkitd-rootless.sh` (https://github.com/moby/buildkit/blob/master/docs/rootless.md)


afbjorklund:
> I think this should be submitted to buildkit repo

But the containerd script is not in containerd repo ?

AkihiroSuda:
`containerd-rootless.sh` is maintained here because it had to be written with nerdctl in unison during early development phase of nerdctl.
We could move it to the main repo but migration is not straight forward.

Issue #443, opened 2021-10-19
`nerdctl run IMAGEID` does not work


AkihiroSuda:
`nerdctl run IMAGEID` is misinterpreted as `nerdctl run docker.io/library/IMAGEID`

```console
# nerdctl run -it --rm sha256:d61d5fd6f4418a52d4cdd28c66b83d11a402870b506347930827f6de7ee1f75c
docker.io/library/sha256:d61d5fd6f4418a52d4cdd28c66b83d11a402870b506347930827f6de7ee1f75c: resolving      |--------------------------------------| 
elapsed: 1.7 s                                                                             total:   0.0 B (0.0 B/s)                                         
INFO[0001] trying next host                              error="pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed" host=registry-1.docker.io
FATA[0001] failed to resolve reference "docker.io/library/sha256:d61d5fd6f4418a52d4cdd28c66b83d11a402870b506347930827f6de7ee1f75c": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed 
```

version: de55345b07b501b756385e50c770b0f9206ec8c7

PR #453, opened 2021-10-21
fix nerdctl run imageid does not work

junnplus:
fix https://github.com/containerd/nerdctl/issues/443


AkihiroSuda:
> Signed-off-by: junnplus

Please use real name, thanks

Issue #271, opened 2021-06-26
service option not work in compose command.


whitemay:
when using compose command such as below:
```
nerdctl compose up -d db
```
according to README，only a service named "db" should be started. 
but actually all services in compose file were started. so lacking the ability to 
control single task in compose file.

AkihiroSuda:
Fixed in https://github.com/containerd/nerdctl/pull/456

PR #456, opened 2021-10-21
compose up services in args

DamiaPoquet:
Use the `nerdctl compose up` arguments to choose which services start. 
 - #271


AkihiroSuda:
https://github.com/apps/dco

(run `git commit -a -s --amend`, and make sure “Signed-off-by” line with your real name is included in the commit message)

PR #491, opened 2021-11-02
fix `compose down` reference typo in README

asomethings:
This PR fixes typo where `compose down` command reference was misleading to `compose up` reference

Signed-off-by: Jae Hyeong Roh <asomethings@kakao.com>


AkihiroSuda (approved):
Thanks

PR #501, opened 2021-11-08
Add --cgroup-conf option to run subcommand

pippolo84:
The PR aims to add the `--cgroup-conf` option to `nerdctl run`.

The syntax for the option is the following:

```nerdctl run --cgroup-conf="KEY1=VALUE1,KEY2=VALUE2"```

just like the equivalent option in `podman run --cgroup-conf`.

Updates #215


AkihiroSuda (approved):
Thanks :+1:

PR #502, opened 2021-11-08
add runtime display  for nerdctl ps

ningmingxiao:


ningmingxiao:
[root@centos7 _output]# ctr c ls
CONTAINER                                                           IMAGE                               RUNTIME
d3fca4e7912c7006ec248a822b4c0b88eb2dea90c4cfc085bc67879d87c3f7c6    docker.io/library/busybox:latest    io.containerd.runc.v2
nerdctl should support  it

cmd/nerdctl/ps.go
```diff
@@ -70,14 +70,15 @@ func psAction(cmd *cobra.Command, args []string) error {
...
-	Status    string
+	Command     string
+	CreatedAt   string
+	ID          string
+	Image       string
+	Platform    string // nerdctl extension
+	Names       string
+	Ports       string
+	Status      string
+	RunTimeName string
```
AkihiroSuda, line 81: `RuntimeName string // nerdctl extension`

cmd/nerdctl/ps.go
```diff
@@ -106,7 +107,7 @@ func printContainers(ctx context.Context, cmd *cobra.Command, containers []conta
 	case "", "table":
 		w = tabwriter.NewWriter(os.Stdout, 4, 8, 4, ' ', 0)
 		if !quiet {
-			fmt.Fprintln(w, "CONTAINER ID\tIMAGE\tPLATFORM\tCOMMAND\tCREATED\tSTATUS\tPORTS\tNAMES")
+			fmt.Fprintln(w, "CONTAINER ID\tIMAGE\tPLATFORM\tCOMMAND\tCREATED\tSTATUS\tPORTS\tNAMES\tRUNTIME")
```
AkihiroSuda, line 110: I think we should add `RUNTIME` field only when we have sufficient TTY width (and when we don't have TTY)


https://pkg.go.dev/golang.org/x/term#GetSize

cmd/nerdctl/ps.go
```diff
@@ -78,6 +78,7 @@ type containerPrintable struct {
 	Names     string
 	Ports     string
 	Status    string
+	RunTime   string
```
AkihiroSuda, line 81:
```suggestion
	Runtime string // nerdctl extension`
```

AkihiroSuda (changes requested):

cmd/nerdctl/ps.go
```diff
@@ -78,6 +78,7 @@ type containerPrintable struct {
 	Names     string
 	Ports     string
 	Status    string
+	RunTime   string // nerdctl extension
```
AkihiroSuda, line 81:
```suggestion
	Runtime   string // nerdctl extension
```

cmd/nerdctl/ps.go
AkihiroSuda: Or maybe only when `--no-trunc` is specified

ningmingxiao: truncate means  "Don't truncate output" which isn't the right word to use here

AkihiroSuda: What about `--format wide`

ningmingxiao: format means make output  according to some standards,
maybe -o wide is ok, like kubectl get node -o wide?

ningmingxiao: but -o wide maybe conflict with  no-trunc

AkihiroSuda: I don't think we should have both `-o` and `--format`.

ningmingxiao: maybe  --format wide is right word

ningmingxiao:
I move platform  and runtime to --fortmat wide,bracause  platform's value often null

AkihiroSuda (approved):
Thanks, added a follow-up commit to approve this

PR #197, opened 2021-04-29
Windows Part 2 - run windows containers

jsturtevant:
Continuation of #184 to enable running Windows containers in  #28.  This add ability to run Windows Containers:

run a container (with image that is not pulled):
```
PS C:\projects\nerdctl> .\_output\nerdctl.exe run -d mcr.microsoft.com/oss/kubernetes/pause:3.4.1
mcr.microsoft.com/oss/kubernetes/pause:3.4.1:                                     resolved       |++++++++++++++++++++++++++++++++++++++|
index-sha256:e3b8c20681593c21b344ad801fbb8abaf564427ee3a57a9fcfa3b455f917ce46:    done           |++++++++++++++++++++++++++++++++++++++|
manifest-sha256:60b0987a9a89932a59a30f4d896ba061a612b23444cadb34827802298cb9db31: done           |++++++++++++++++++++++++++++++++++++++|
config-sha256:3914de10243c46fdec6e1f4583301185a0b1ead4d6c8964f46984b130674bf14:   done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:a17285143420aa470f0b8c1addb73751bd647cf21979c011bf668130c513c4d0:    done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:2021f649dd34b8a57b0ea5620d3d32b08fe9c96e7dd2e4d7c357e2e1a30f92d2:    done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:1a142ed7c6984745b2f5e82416e73231992fa374e43d3b546d93ff44499e156c:    done           |++++++++++++++++++++++++++++++++++++++|
elapsed: 3.1 s                                                                    total:  1.4 Mi (461.7 KiB/s)
6f38e163e20d83a5c2a7f704a78711ec0f33bc826c4ff7489c7353cf90eabf04
```

interact with it:
```
PS C:\projects\nerdctl> .\_output\nerdctl.exe ps
CONTAINER ID    IMAGE                                           COMMAND         CREATED           STATUS    PORTS    NAMES
6f38e163e20d    mcr.microsoft.com/oss/kubernetes/pause:3.4.1    "/pause.exe"    15 seconds ago    Up
PS C:\projects\nerdctl> .\_output\nerdctl.exe exec -it 6f3 cmd /c ver

Microsoft Windows [Version 10.0.19042.746]
              
PS C:\projects\nerdctl> .\_output\nerdctl.exe exec -it 6f3 cmd /c dir
[...]

PR #508, opened 2021-11-10
rootless: allow loading an existing AppArmor profile

### LINKED ISSUE #507: Support rootless AppArmor with `sudo nerdctl apparmor load`
While we still can't load an AppArmor profile without root, running a rootless container with a pre-loaded AppArmor profile should be possible.

We will need an additional command like `sudo nerdctl apparmor load`.

---

~~We will have to let RootlessKit bind-mount `/sys/kernel/security` from the parent mount namespace.~~ (PR #508 implements the proposal without relying on `/sys/kernel/security`)

AkihiroSuda:
### nerdctl run

Rootless nerdctl now applies the `nerdctl-default` profile to containers by default, if it is already loaded with `sudo nerdctl apparmor load`.

Still defaults to "unconfined" when the profile is not loaded.

### nerdctl info

Now `nerdctl info` shows "apparmor" when the AppArmor kernel module is enabled, regardless to whether the `nerdctl-default` profile is loaded or not.
When the profile is not available, `nerdctl info` shows a warning message.

Fix #507

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -189,6 +189,7 @@ FROM ubuntu:${UBUNTU_VERSION} AS base
 # fuse3 is required by stargz snapshotter
 RUN apt-get update && \
   apt-get install -qq -y --no-install-recommends \
+  apparmor \
   ca-certificates curl \
   iproute2 iptables \
   dbus systemd systemd-sysv \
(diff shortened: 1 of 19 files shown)


AkihiroSuda:
This PR is ready to review, but merging this PR will cause conflict with the Windows PR https://github.com/containerd/nerdctl/pull/197.

I'm planning to merge the Windows PR #197 before this one if possible.

ktock (approved):

PR #509, opened 2021-11-10
Add --blkio-weight option to run subcommand

pippolo84:
The PR aims to add the `--blkio-weight` option to `nerdctl run`, exposing the same behaviour as `docker run --blkio-weight`

Both cgroup v1 and cgroup v2 are supported. Besides, since the bfq IO scheduler exposes a different weight knob, a specific check is added for that.
Just like in docker, if the underlying kernel does not support blkio weight, a warning is generated and the weight is reset to 0.

Updates #215


AkihiroSuda (approved):
Thanks.

Pushed commit `compose: support blkio_config` as a follow-up

PR #515, opened 2021-11-15
build: Add VERBOSE option in Makefile

seemethere:
Adds a VERBOSE option in the Makefile to build the nerdctl binary with
the '-v' build flag to check on output of the go build

Usage:
```bash
nerdctl run --rm -it -v "$(pwd)":/v -w /v -e VERBOSE=1 golang make
```
Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,10 @@ REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet 
 
 export GO_BUILD=GO111MODULE=on CGO_ENABLED=0 GOOS=$(GOOS) $(GO) build -ldflags "-s -w -X $(PACKAGE)/pkg/version.Version=$(VERSION) -X $(PACKAGE)/pkg/version.Revision=$(REVISION)"
 
+ifdef VERBOSE
+	VERBOSE_FLAG := -v
+endif
+
 all: binaries
 
 help:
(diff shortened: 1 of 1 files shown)


fahedouch:
based  on your example `VERBOSE=1 golang make`, the target should be like this : 
```
# turn on more verbose build when VERBOSE=1
ifeq ($(VERBOSE),1)
VERBOSE_FLAG += -x
endif
```

may be changing `VERBOSE_FLAG` to `GO_BUILDFLAGS` , we can make sthg like this  :  
```
# Optional build flags passed to go
GO_BUILDFLAGS ?=

# turn on more verbose build when VERBOSE=1
ifeq ($(VERBOSE),1)
GO_BUILDFLAGS += -x
endif

seemethere:
> based on your example `VERBOSE=1 golang make`, the target should be like this :
> 
> ```
> # turn on more verbose build when VERBOSE=1
> ifeq ($(VERBOSE),1)
> VERBOSE_FLAG += -v
> endif
> ```
> 
> may be wraping `VERBOSE_FLAG` into `GO_BUILDFLAGS` too , we can have sthg like this :
> 
> ```
> # Optional build flags passed to go
> GO_BUILDFLAGS ?=
> 
> # turn on more verbose build when VERBOSE=1
> ifeq ($(VERBOSE),1)
> GO_BUILDFLAGS += -v
> endif
> ```

Not necessarily, I'm just checking for the existence of a `VERBOSE` variable, the actual value could really be anything

AkihiroSuda (approved):
Thanks

PR #536, opened 2021-11-18
Support output flag for nerdctl compose build

afbjorklund:
It was only supported for regular `nerdctl build` before.

### DIFF
--- a/cmd/nerdctl/compose_build.go
+++ b/cmd/nerdctl/compose_build.go
@@ -33,6 +33,7 @@ func newComposeBuildCommand() *cobra.Command {
 	}
 	composeBuildCommand.Flags().StringArray("build-arg", nil, "Set build-time variables for services.")
 	composeBuildCommand.Flags().Bool("no-cache", false, "Do not use cache when building the image.")
+	composeBuildCommand.Flags().StringP("output", "o", "", "Output destination (format: type=local,dest=path)")
 	composeBuildCommand.Flags().String("progress", "", "Set type of progress output")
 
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
How is this expected to work for multi-services?

Also, please update the Command Reference in README.md

afbjorklund:
> How is this expected to work for multi-services?

It probably does not. :-)

The image might not even "exist" i.e. when just saving it to a file
(not even sure how you would "template" the resulting file names)

Intended use was with a **remote** buildkitd, like `kube-pod://`

There is a obvious workaround, even for the single service, i.e.
to first download the image locally, and push it up to the registry.

afbjorklund:
That is, `name` is auto-generated but `dest` is not.

afbjorklund:
>  first download the image locally, and push it up to the registry

Spoke to soon, `compose push` and `compose pull` are also missing.

afbjorklund:
Added naive implementations of the missing `compose` commands:

* PR #546 
* PR #547

afbjorklund:
Would it make sense to have a more high-level parameter instead ? Like `push bool`.

Now we have `type=docker` (with loading from stdout pipe)

And also `type=image,push=true` (for pushing to registry)

djdongjin:
Do we still need this feature? Seems [`docker compose build`](https://docs.docker.com/engine/reference/commandline/compose_build/#options) doesn't have this option either and from the discussion the behavior is not defined on multi-service compose. cc @afbjorklund @AkihiroSuda

PR #547, opened 2021-11-21
Add support for --no-build to compose up

afbjorklund:
Try to pull the image instead of building it.

Otherwise it would _always_ rebuild images.

### DIFF
--- a/cmd/nerdctl/compose_up.go
+++ b/cmd/nerdctl/compose_up.go
@@ -30,6 +30,7 @@ func newComposeUpCommand() *cobra.Command {
 		SilenceErrors: true,
 	}
 	composeUpCommand.Flags().BoolP("detach", "d", false, "Detached mode: Run containers in the background")
+	composeUpCommand.Flags().Bool("no-build", false, "Don't build an image, even if it's missing.")
 	composeUpCommand.Flags().Bool("no-color", false, "Produce monochrome output")
 	composeUpCommand.Flags().Bool("no-log-prefix", false, "Don't print prefix in logs")
 	composeUpCommand.Flags().Bool("build", false, "Build images before starting containers.")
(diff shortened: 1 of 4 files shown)


pkg/composer/up_service.go
```diff
@@ -99,8 +99,8 @@ func (c *Composer) upServices(ctx context.Context, parsedServices []*servicepars
 	return nil
 }
 
-func (c *Composer) ensureServiceImage(ctx context.Context, ps *serviceparser.Service, force bool, bo BuildOptions) error {
-	if ps.Build != nil {
+func (c *Composer) ensureServiceImage(ctx context.Context, ps *serviceparser.Service, build bool, force bool, bo BuildOptions) error {
```
AkihiroSuda, line 102: `build` -> `allowBuild`; `foce` -> `forceBuild`

cmd/nerdctl/compose_up.go
```diff
@@ -42,6 +43,10 @@ func composeUpAction(cmd *cobra.Command, services []string) error {
 	if err != nil {
 		return err
 	}
+	noBuild, err := cmd.Flags().GetBool("no-build")
+	if err != nil {
+		return err
+	}
```
AkihiroSuda, line 49: Error out when noBuild && forceBuild

pkg/composer/up_service.go
AkihiroSuda: Added a follow-up commit by myself

cmd/nerdctl/compose_up.go
AkihiroSuda: Added a follow-up commit by myself

AkihiroSuda (approved):
Thanks

Issue #542, opened 2021-11-19
compose feature: shell env substitution support is missing


pythonwood:
docker-compose.yml
```
    ports:                  
      - ${ADDRESS:-127.0.0.1}:1909:1909
```

in docker:
```
ADDRESS=0.0.0.0 docker-compose up -d # work good
docker-compose -e ADDRESS=0.0.0.0 up -d # work good
```

in nerdctl:
```
nerdctl  -e ADDRESS=0.0.0.0 compose up -d # not work
ADDRESS=0.0.0.0  nerdctl  compose up -d # not work
```

a little like  #252

pythonwood:
another example use: 
```
    volumes:
      - $HOME/.data/:/var/spool/postfix
```
hope nerdctl support it so much.

fahedouch:
fixed by https://github.com/containerd/nerdctl/pull/570.

PR #570, opened 2021-11-25
Support os env var compose

fahedouch:
Hello,

fixing https://github.com/containerd/nerdctl/issues/542

`nerdctl  -e ADDRESS=0.0.0.0 compose up -d` is not supported

### DIFF
--- a/cmd/nerdctl/compose_up_test.go
+++ b/cmd/nerdctl/compose_up_test.go
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

benjaminpreiss:
@AkihiroSuda could you maybe give an example of the correct syntax to use this? I am having trouble using a .env file with the command line option `--env-file ./.env`. I am using Rancher Desktop v0.7.1 on Windows 10 WSL. The error I get is "FATA[0000] Couldn't find env file: ./.env"

PR #533, opened 2021-11-18
Windows: run CI with containerd 1.5.8

AkihiroSuda:

Issue #528, opened 2021-11-18
[Windows] flaky test: TestRunUserName: `msg="error unpacking image: lease \"427436400-Uv38\": already exists"`


AkihiroSuda:
```
=== CONT  TestRunUserName
    run_user_windows_test.go:42: assertion failed: 
        Command:  C:\Windows\system32\config\systemprofile\go\bin\nerdctl.exe --namespace=nerdctl-test run --rm gcr.io/k8s-staging-e2e-test-images/busybox:1.29-2 whoami
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   time="2021-11-18T07:09:32Z" level=fatal msg="error unpacking image: lease \"427436400-Uv38\": already exists"
        
        
        Failures:
        ExitCode was 1 expected 0
        Expected stdout to contain "ContainerAdministrator"
        Expected no error
```

https://cirrus-ci.com/task/5676643743170560

jsturtevant:
another instance: https://cirrus-ci.com/task/6157490336301056?logs=build#L353

apostasie:
Have not seen this in the past six months.
Unless it is reproducible, shall we close?

PR #580, opened 2021-12-01
Add test matrix for windows ctrd versions

jsturtevant:
Adds test matrix for the current windows ctrd versions

follow up from https://github.com/containerd/nerdctl/pull/533#issuecomment-982772530

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -18,6 +18,13 @@ docker_builder:
   name: windows
   platform: windows
   os_version: 2019
+  matrix:
+    - name: Windows 1.5 tests
+      env:
+        ctrdVersion: 1.5.8
+    - name: Windows 1.6 tests
+      env:
+        ctrdVersion: 1.6.0-beta.3
   env:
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -18,6 +18,16 @@ docker_builder:
   name: windows
   platform: windows
   os_version: 2019
+  matrix:
+    - name: Windows 1.4 tests
+      env:
+        ctrdVersion: 1.4.12
```
AkihiroSuda, line 24: I think we can drop v1.4 support for Windows

jsturtevant: sounds good, was thinking the same thing.  With out looking to deeply I think it might be related to https://github.com/containerd/containerd/commit/791e175c79930a34cfbb2048fbcaa8493fd2c86b

jsturtevant:
the 1.5 failure is the flake #528: `msg="error unpacking image: lease \"241796900-Uv38\": already exists"`

Issue #581, opened 2021-12-01
`nerdctl load` should support loading tgz


AkihiroSuda:
Currently `nerdctl load` only accepts uncompressed tar balls

PR #587, opened 2021-12-02
load cmd support gzip

Junnplus:
Fixes https://github.com/containerd/nerdctl/issues/581

### DIFF
--- a/cmd/nerdctl/load.go
+++ b/cmd/nerdctl/load.go
@@ -23,6 +23,7 @@ import (
 	"os"
 
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/archive/compression"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/images/archive"
 	"github.com/containerd/containerd/platforms"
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/load.go
```diff
@@ -68,6 +72,16 @@ func loadAction(cmd *cobra.Command, args []string) error {
...
 
+	inbuf := bufio.NewReader(in)
+	buff, _ := inbuf.Peek(3)
+	if bytes.HasPrefix(buff, []byte("\x1f\x8b")) {
+		in, _ = gzip.NewReader(inbuf)
+	} else if bytes.HasPrefix(buff, []byte("\x42\x5a\x68")) {
+		in = bzip2.NewReader(inbuf)
+	} else {
+		in = inbuf
+	}
```
AkihiroSuda, line 83: Thanks for working on this, but I'd suggest using https://pkg.go.dev/github.com/containerd/containerd/archive/compression#DecompressStream

Junnplus: `DecompressStream` only supports gzip and zstd.
https://github.com/containerd/containerd/blob/main/archive/compression/compression.go#L169

but `docker load` support gzip and bzip2.
https://docs.docker.com/engine/reference/commandline/load/#extended-description

AkihiroSuda: Then we should fix containerd/containerd

AkihiroSuda: Anyway, supporting bzip2 is probably not important. Is anyone really using bzip2 for Docker v1 archives?

Junnplus: @AkihiroSuda I first change to  using`DecompressStream`,

AkihiroSuda (approved):
Thanks

Issue #586, opened 2021-12-02
Setting ulimit prompts an error


dmanbu:
containerd version: v1.5.5
nerdctl: version v0.14.0

Start a container:
```
nerdctl run -d --name haproxy --ulimit nofile=65536:131072 haproxy:2.5.0-alpine3.15
```
Prompted an error:  
FATA[0000] OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: process_linux.go:446: setting rlimits for ready process caused: error setting rlimit type 7: invalid argument: unknown

But it works like this, no problem:
```
nerdctl run -d --name haproxy --ulimit nofile=131072:131072 haproxy:2.5.0-alpine3.15
```

‘soft’ must be equal to ‘hard’

AkihiroSuda:
Fixed in #588

PR #588, opened 2021-12-02
fix rlimits

fahedouch:
fix `POSIXRlimit`  . issue https://github.com/containerd/nerdctl/issues/586


AkihiroSuda (approved):
Thanks, good catch

Maybe we should have a test too?

AkihiroSuda (approved):
Thanks

Issue #344, opened 2021-09-01
nerdctl compose issues with ulimit


amitkgupta:
I see [here](https://github.com/containerd/nerdctl/blob/master/docs/compose.md) no mention that `nerdctl compose` does _not_ support `ulimits`. This is something Docker Compose supports, see [here](https://github.com/containerd/nerdctl/blob/master/docs/compose.md).

In my `compose.yml` I have:

```
version: '1'
services:
  foo:
    ulimits:
      nofile:
        soft: 1048576
        hard: 1048576
...
```

However, when I `nerdctl compose up -d` and then exec into that container, the ulimit setting hasn't applied:

```
$ nerdctl exec -it 810c4ccc66a6 /bin/bash
[appuser@foo ~]$ cat /proc/1/limits
Limit                     Soft Limit           Hard Limit           Units
Max cpu time              unlimited            unlimited            seconds
Max file size             unlimited            unlimited            bytes
Max data size             unlimited            unlimited            bytes
Max stack size            8388608              unlimited            bytes
Max core file size        unlimited            unlimited            bytes
Max resident set          unlimited            unlimited            bytes
Max processes             15536                15536                processes
Max open files            1024                 1024                 files
Max locked memory         513875456            513875456            bytes
Max address space         unlimited            unlimited            bytes
Max file locks            unlimited            unlimited            locks
Max pending signals       15536                15536                signals
Max msgqueue size         819200               819200               bytes
Max nice priority         0                    0
Max realtime priority     0                    0
Max realtime timeout      unlimited            unlimited            us
```

I'm new to nerdctl, but I wonder if this has something to do with rootless?

How can I increase the open file limits for one of my compose containers when using `nerdctl`?

AkihiroSuda:
Unimplemented yet, both for rootful and rootless

> I see here no mention that nerdctl compose does not support ulimits. 

Covered by the following sentences. We don't have `docker run --ulimit` yet.

> ### Unimplemented YAML fields
> - Fields that correspond to unimplemented `docker run` flags, e.g., `services.<SERVICE>.links` (corresponds to `docker run --link`)

oscar60310:
Seems we have implemented `nerdctl run --ulimit` at https://github.com/containerd/nerdctl/pull/370. Do we have any plan to bring it to compose too?

AkihiroSuda:
PR is welcome

PR #593, opened 2021-12-03
compose: support ulimits

Junnplus:
fixed: https://github.com/containerd/nerdctl/issues/344

### DIFF
--- a/pkg/composer/serviceparser/serviceparser.go
+++ b/pkg/composer/serviceparser/serviceparser.go
@@ -78,6 +78,7 @@ func warnUnknownFields(svc compose.ServiceConfig) {
 		"User",
 		"WorkingDir",
 		"Volumes",
+		"Ulimits",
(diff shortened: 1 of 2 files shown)


pkg/composer/serviceparser/serviceparser.go
```diff
@@ -78,6 +78,7 @@ func warnUnknownFields(svc compose.ServiceConfig) {
 		"User",
 		"WorkingDir",
 		"Volumes",
+		"Ulimits",
```
AkihiroSuda, line 81: Please keep this list sorted

AkihiroSuda (approved):
Thanks, looks good but a nit

pkg/composer/serviceparser/serviceparser.go
AkihiroSuda: (The list seems to have been already unsorted, so merging)

PR #595, opened 2021-12-05
add no-start flag for compose up

Junnplus:


pkg/composer/up_service.go
```diff
@@ -43,6 +43,10 @@ func (c *Composer) upServices(ctx context.Context, parsedServices []*servicepars
 		}
 	}
 
+	if uo.NoStart {
+		return nil
```
AkihiroSuda, line 47: This should call `nerdctl create` (which isn't implemented yet)

AkihiroSuda (changes requested):
This should call `nerdctl create` (which isn't implemented yet)

pkg/composer/up_service.go
Junnplus: `nerdctl compose create` ? `docker-compose create` has been deprecated. 
https://docs.docker.com/compose/reference/create/

AkihiroSuda: No, `nerdctl create` (`docker create`)  to create non-running containers

PR #617, opened 2021-12-14
Add compose kill command

Junnplus:

Add `nerdctl compose kill` command.

### DIFF
--- a/pkg/composer/kill.go
+++ b/pkg/composer/kill.go
@@ -1,0 +1,52 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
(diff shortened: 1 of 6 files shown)


pkg/composer/kill.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return err
+	}
+	for _, container := range containers {
+		args := []string{"kill", "-s", opts.Signal, container.ID()}
+		if err := c.runNerdctlCmd(ctx, args...); err != nil {
+			logrus.Warn(err)
+		}
+	}
+	return nil
```
AkihiroSuda, line 44: Shouldn’t we return an error if failed to kill at least one of the containers?

Junnplus: Or return the first error?

AkihiroSuda (approved):
Thanks

PR #619, opened 2021-12-14
Update imgcrypt dependency to v1.1.3

stefanberger:
This PR updates the imgcrypt dependency to v1.1.3.

There was a re-taggin of v1.1.2, which changed the checksum. This PR fixes this as well but then also updates the dependency to v1.1.3.


AkihiroSuda (approved):
Thanks

PR #621, opened 2021-12-15
support specifying service name for compose pull/push

Junnplus:
Support specifying service name for compose pull/push


AkihiroSuda (approved):
Thanks

PR #629, opened 2021-12-16
fix nerdctl inspect image error

ningmingxiao:
```
nerdctl  inspect busybox
FATA[0001] flag accessed but not defined: platform
```


cmd/nerdctl/inspect.go
```diff
@@ -51,6 +51,8 @@ func addInspectFlags(cmd *cobra.Command) {
 	cmd.RegisterFlagCompletionFunc("mode", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"dockercompat", "native"}, cobra.ShellCompDirectiveNoFileComp
 	})
+	cmd.Flags().String("platform", "", "Inspect a specific platform") // not a slice, and there is no --all-platforms
+	cmd.RegisterFlagCompletionFunc("platform", shellCompletePlatforms)
 }
```
AkihiroSuda, line 56: Thanks, but this flag is specific to `nerdctl image inspect`, and I don't think the top-level `nerdctl inspect` should have this.

Opened https://github.com/containerd/nerdctl/pull/630 as an alternative.

PR #470, opened 2021-10-26
add configfile support

ningmingxiao:
Signed-off-by: ning.mingxiao <ning.mingxiao@zte.com.cn>

PR #635, opened 2021-12-17
Support loading `nerdctl.toml` (`$NERDCTL_TOML`)

AkihiroSuda:
See `docs/config.md`

Replace PR #470

## Example

```toml
# This is an example of /etc/nerdctl/nerdctl.toml .
# Unrelated to the daemon's /etc/containerd/config.toml .

debug          = false
debug_full     = false
address        = "unix:///run/k3s/containerd/containerd.sock"
namespace      = "k8s.io"
snapshotter    = "stargz"
cgroup_manager = "cgroupfs"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1230,6 +1230,9 @@ Flags:
   - Default: "systemd" on cgroup v2 (rootful & rootless), "cgroupfs" on v1 rootful, "none" on v1 rootless
 - :nerd_face: `--insecure-registry`: skips verifying HTTPS certs, and allows falling back to plain HTTP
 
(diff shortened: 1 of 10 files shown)


ningmingxiao:
why  use   “github.com/pelletier/go-toml”   instead of  "github.com/spf13/viper“ ?
viper support more format like JSON, TOML, YAML , if user like JSON,YAML config file  it is easier  to change it.
and go-toml maybe is not as strong as viper.

AkihiroSuda:
> if user like JSON,YAML config file it is easier to change it.

I don't think we need such extensibility.

ktock (approved):

PR #636, opened 2021-12-17
validate output path

fahedouch:
`ValidateOutputPath `  validate path before save/export. this will also be used in the `cp` command to validate the remote and local path

### DIFF
--- a/pkg/imgutil/imgutil.go
+++ b/pkg/imgutil/imgutil.go
@@ -19,8 +19,10 @@ package imgutil
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"os"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/imgutil/imgutil.go
```diff
@@ -413,3 +415,32 @@ func ParseRepoTag(imgName string) (string, string) {
...
+	if fileInfo, err := os.Stat(path); !os.IsNotExist(err) {
+		if err != nil {
+			return err
+		}
+
+		if fileInfo.Mode().IsRegular() {
+			return nil
+		}
+		if err := ValidateOutputPathFileMode(fileInfo.Mode()); err != nil {
+			return fmt.Errorf("invalid output path: %q must be a regular file: %w", path, err)
```
AkihiroSuda, line 431: Why?

AkihiroSuda: Historically, tar balls were designed to be directly writable to tape devices in 19XX.

And it is still 100% legit to write tar balls to device files in 2021. (Rarely useful, though)

fahedouch:
Right @AkihiroSuda, i was disturbed with some tests thinking that devices are overwritten with output, but it is not the case because we are dealing with stream byte. Sorry for this

Issue #634, opened 2021-12-17
Unspam `WARN[0000] kernel support for cgroup blkio weight missing, weight discarded `


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/0f3e1e6d722459c92416b495011e8174ed07ee07/cmd/nerdctl/run_cgroup_linux.go#L149

This warning doesn't need to be printed unless the user explicitly specifies `--blkio-weight`

PR #637, opened 2021-12-19
Unspam warning: kernel support for cgroup blkio weight missing

JuozasVainauskas:
Do not show warning `kernel support for cgroup blkio weight missing, weight discarded` if --blkio-weight was not explicitly specified.

Issue: https://github.com/containerd/nerdctl/issues/634


AkihiroSuda (approved):
Thanks

PR #642, opened 2021-12-20
Support `/etc/containerd/certs.d/<HOST:PORT>/hosts.toml`

### LINKED ISSUE #639: Support `/etc/containerd/certs.d/<HOST>/hosts.toml` (with tests)
https://github.com/containerd/containerd/blob/main/docs/hosts.md
https://pkg.go.dev/github.com/containerd/containerd@v1.5.8/remotes/docker/config#HostDirFromRoot

~~Implementation is easy, but~~ we need to have good integration tests.

AkihiroSuda:
Fix #639

Depends on:
- https://github.com/containerd/containerd/pull/6396

### DIFF
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -21,16 +21,20 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"runtime"
+	"net/http"
+	"net/url"
 	"strings"
 
-	"github.com/containerd/nerdctl/pkg/version"
+	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/remotes/docker"
+	dockerconfig "github.com/containerd/containerd/remotes/docker/config"
+	"github.com/containerd/nerdctl/pkg/imgutil/dockerconfigresolver"
 	dockercliconfig "github.com/docker/cli/cli/config"
 	clitypes "github.com/docker/cli/cli/config/types"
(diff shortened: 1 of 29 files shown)


cmd/nerdctl/login.go
```diff
@@ -169,33 +174,119 @@ func GetDefaultAuthConfig(checkCredStore bool, serverAddress string, isDefaultRe
...
 	}
-	if insecureRegistry {
-		insecureRegistries = append(insecureRegistries, auth.ServerAddress)
+	dOpts = append(dOpts, dockerconfigresolver.WithHostsDirs(hostsDirs))
+
+	authCreds := func(acArg string) (string, string, error) {
+		if acArg == host {
+			if auth.RegistryToken != "" {
+				// Even containerd/CRI does not support RegistryToken as of v1.4.3,
+				// so, nobody is actually using RegistryToken?
```
dmcgowan, line 206: It's not really necessary and the only use of it I know of was for Docker swarm. It solves the problem of sending unscoped credentials to nodes, however, the tokens end up being long lived and passed directly to registries. The better solution is using scoped refresh tokens. It would be good to support scoping somewhere here, although that does not work well with the `docker login` approach.

AkihiroSuda: Thanks, the current goal is to just emulate Swarm-less `docker login`

ktock (approved):
LGTM

PR #647, opened 2021-12-23
Use containerd.errdefs instead of docker.errdefs

Junnplus:
https://github.com/containerd/containerd/blob/main/container.go#L393

Use containerd.errdefs instead of docker.errdefs


AkihiroSuda (approved):
Thanks

Issue #504, opened 2021-11-09
Support for `nerdctl build --rm` flag


gakio12:
There are some makefile systems that specify the `--rm` flag on a `build` command. Because there is no support for this in nerdctl, an `alias docker=nerdctl` won't work.

From the Docker documentation: 
> default: true
> Remove intermediate containers after a successful build

gakio12:
It looks like buildkit doesn't actually use `--rm`, and the Docker CLI doesn't look to be using it anymore (https://github.com/moby/buildkit/issues/1470).

I can add it as a flag, then mark it as hidden, so it will still function. If someone does use it, should it print a warning such as `ignoring depricated flag: '--rm=false'`, or should it just be quiet about it?

AkihiroSuda:
Maybe we should run `buildctl prune` when `--rm` is specified

PR #649, opened 2021-12-23
Add support for nerdctl build --rm flag

JuozasVainauskas:
Add support for `nerdctl build --rm` flag. 

User has the possibility to specify `--rm` flag with `build` command. If user specifies `--rm=false` flag, he gets informed that `--rm` flag with `false` value is deprecated and thus is ignored.

Issue: https://github.com/containerd/nerdctl/issues/504


AkihiroSuda (approved):
Thanks

PR #653, opened 2021-12-27
add label flag for build

Junnplus:


AkihiroSuda (approved):
Thanks

Can we support labels in `compose up --build` too ?
https://pkg.go.dev/github.com/compose-spec/compose-go@v1.0.8/types#BuildConfig

Junnplus:
> Can we support labels in compose up --build too ?

Yes, I will support it in the next PR.

Issue #657, opened 2021-12-31
network bridge name conflict


Junnplus:
```
--- PASS: TestComposeUpWithScale (0.53s)
=== RUN   TestComposeIPAMConfig
    compose_up_linux_test.go:324: projectName="nerdctl-compose-test2749460603"
    compose_up_linux_test.go:326: assertion failed: expression is false: res.ExitCode == exitCode: time="2021-12-27T10:48:36Z" level=info msg="Creating network nerdctl-compose-test2749460603_default"
        time="2021-12-27T10:48:37Z" level=info msg="Ensuring image ghcr.io/stargz-containers/alpine:3.13-org"
        time="2021-12-27T10:48:37Z" level=info msg="Creating container nerdctl-compose-test2749460603_foo_1"
        time="2021-12-27T10:48:37Z" level=fatal msg="failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: time=\"2021-12-27T10:48:37Z\" level=fatal msg=\"failed to call cni.Setup: plugin type=\\\"bridge\\\" failed (add): failed to set bridge addr: \\\"nerdctl1\\\" already has an IP address different from 10.1.100.1/24\"\nFailed to write to log, write /var/lib/nerdctl/1935db59/containers/nerdctl-test/0c6b3bc6622c1ec0cead5d768ad577f8d0347c0b0c74501c98aef03d85290411/oci-hook.createRuntime.log: file already closed: unknown"
        time="2021-12-27T10:48:37Z" level=fatal msg="error while creating container nerdctl-compose-test2749460603_foo_1: exit status 1"
--- FAIL: TestComposeIPAMConfig (0.24s)
```

https://cirrus-ci.com/task/4506924338118656?logs=rootful_test#L214

_Originally posted by @AkihiroSuda in https://github.com/containerd/nerdctl/issues/652#issuecomment-1001530513_

I found bridge name always like nerdctlN, however, bridgt does not seem to be recycled after network remove, it will cause bridge conflicts.

Junnplus:
I want to fix it simply: use a random string instead nerdctlN, like `nerdctl-b1625a`.

Why is the bridge not deleted? This needs to be investigated.

Junnplus:
> Why is the bridge not deleted? This needs to be investigated.

[...]

PR #658, opened 2021-12-31
make bridge name random

Junnplus:
related: https://github.com/containerd/nerdctl/issues/657


AkihiroSuda:
Thanks, but I think we should rather investigate the real cause of #657

Junnplus:
> Thanks, but I think we should rather investigate the real cause of #657

This is necessary, but the bridge of nerdctlN named is problematic and can easily cause conflicts.

PR #660, opened 2022-01-03
remove bridge network interface after network rm

Junnplus:

### DIFF
--- a/cmd/nerdctl/network_rm.go
+++ b/cmd/nerdctl/network_rm.go
@@ -82,6 +82,11 @@ func networkRmAction(cmd *cobra.Command, args []string) error {
 			if err := os.RemoveAll(l.File); err != nil {
 				return err
 			}
+			// Remove the bridge network interface on the host.
+			if l.Plugins[0].Network.Type == "bridge" {
+				netIf := fmt.Sprintf("nerdctl%d", *l.NerdctlID)
+				removeBridgeNetworkInterface(netIf)
+			}
 			fmt.Fprintln(cmd.OutOrStdout(), name)
 		}
 		return nil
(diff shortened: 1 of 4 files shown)


cmd/nerdctl/network_rm_linux.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/vishvananda/netlink"
+)
+
+func removeBridgeNetworkInterface(netIf string) {
+	link, err := netlink.LinkByName(netIf)
+	if err == nil {
+		if err = netlink.LinkDel(link); err != nil {
+			logrus.Infof("Failed to remove network interface %s: %v", netIf, err)
```
AkihiroSuda, line 28: logrus.Warnf

cmd/nerdctl/network_rm.go
```diff
@@ -82,6 +82,11 @@ func networkRmAction(cmd *cobra.Command, args []string) error {
 			if err := os.RemoveAll(l.File); err != nil {
 				return err
 			}
+			// Remove the bridge network interface on the host.
+			if l.Plugins[0].Network.Type == "bridge" {
+				netIf := fmt.Sprintf("nerdctl%d", *l.NerdctlID)
```
AkihiroSuda, line 87: This should be fetched from here https://github.com/containerd/nerdctl/blob/ed3d553a9bd6818b0917f2e040e43df5973ff8c2/pkg/netutil/netutil_unix.go#L40

Junnplus: It seems that there is no easy way to get the name of the bridge from plugins.

`l.Plugins[0].Network.Name` always is emtry.

AkihiroSuda (approved):

PR #676, opened 2022-01-07
ipfs registry: allow timeout and retry

ktock:
IPFS occasionally hangs during reading chunks and requires retrying the request.
To fix this isssue, this commit adds the following options.

- `--read-retry-num`: Times to retry query on IPFS (default 0 (no retry))
- `--read-timeout`: Timeout duration of a read request to IPFS (defualt 0 (no timeout))


AkihiroSuda (approved):
Thanks

PR #685, opened 2022-01-09
Enhancements for `compose up` command.

### LINKED ISSUE #342: nerdctl compose up support for remove-orphans
nerdctl currently seems not to support the `remove-orphans` option from docker-compose up. It's a handy way to remove unneeded created container

fhke:
This PR contains the following updates to the `compose up` command:

* Idempotency: rather than failing when a container already exists and has been created by compose, recreate the container. This allows users to iterate rapidly with `compose up`, rather than deleting & recreating their compose stack to effect changes.
* Add support for `--remove-orphans` flag: remove containers for services not defined in the Compose file. This fixes #342 

Signed-off-by: fergal kearns <fhke@protonmail.com>

### DIFF
--- a/pkg/composer/up.go
+++ b/pkg/composer/up.go
@@ -29,14 +29,15 @@ import (
 )
 
 type UpOptions struct {
-	Detach      bool
-	NoBuild     bool
-	NoColor     bool
-	NoLogPrefix bool
-	ForceBuild  bool
-	IPFS        bool
-	QuietPull   bool
-	Scale       map[string]uint64 // map of service name to replicas
+	Detach        bool
+	NoBuild       bool
+	NoColor       bool
+	NoLogPrefix   bool
+	ForceBuild    bool
+	IPFS          bool
+	QuietPull     bool
+	RemoveOrphans bool
+	Scale         map[string]uint64 // map of service name to replicas
 }
 
 func (c *Composer) Up(ctx context.Context, uo UpOptions, services []string) error {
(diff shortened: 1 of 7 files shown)


pkg/composer/up.go
```diff
@@ -86,7 +87,26 @@ func (c *Composer) Up(ctx context.Context, uo UpOptions, services []string) erro
 		return err
 	}
 
-	return c.upServices(ctx, parsedServices, uo)
+	if err := c.upServices(ctx, parsedServices, uo); err != nil {
+		return err
+	}
+
+	orphans, err := c.getOrphanContainers(ctx, parsedServices)
```
AkihiroSuda, line 94: This operation might be slow and should be called only when `RemoveOrphans` is true

fhke: Hi @AkihiroSuda, I appreciate this operation could introduce additional delays into the `compose up` workflow, but without performing it we cannot warn users that orphaned containers exist:

```go
	if numOrphans > 0 {
		if uo.RemoveOrphans {
			if err := c.downContainers(ctx, orphans, true); err != nil {
				return fmt.Errorf("error removing orphaned containers: %s", err)
			}
		} else {
			logrus.Warnf("%d orphaned containers exist. Run with --remove-orphans to remove them.", numOrphans)
		}
	}
```

This would be a deviation from the behaviour of `docker-compose`, which prints a similar warning. 

We can probably reduce the time taken without losing this error message by comparing the total number of containers in the `parsedServices` slice with the total number of running containers in the compose project; this would remove the requirement to call the `.Labels()` method for each container:

```go
	for _, container := range containers {
		containerLabels, err := container.Labels(ctx)
		if err != nil {
			return false, err
		}
		if name == containerLabels[labels.Name] {
			// container exists
			return true, nil
		}
	}
```

Please let me know if you're happy with that approach, and I'll make the changes accordingly.

AkihiroSuda: > This would be a deviation from the behaviour of docker-compose, which prints a similar warning.

I think this difference is acceptable.

AkihiroSuda: ping @fhke

fhke: Hey @AkihiroSuda, apologies for the delay. I've updated this PR with the requested changes.

AkihiroSuda (approved):
Thanks

PR #704, opened 2022-01-15
Add nerdctl version to User-Agent header

ningziwen:
Right now, nerdctl reports its user-agent string as "containerd/${containerd_version}`, so container registries can't distinguish nerdctl pulls from containerd pulls. It would be useful to re-set the user-agent as "nerdctl/${nerdctl_version} containerd/${containerd_version}".

Signed-off-by: Ziwen Ning <ningziwe@amazon.com>


AkihiroSuda:
> container registries can't distinguish nerdctl pulls from containerd pulls

Why do you need to distinguish?

samuelkarp:
Hey @AkihiroSuda,

Our apologies for not explaining this super clearly; we’ll be sure to open issues for changes like this first next time.

We’d like to see nerdctl have a distinct User-Agent for the same reasons we’d like to see other clients have distinct User-Agent strings: (a) if our customers use nerdctl and run into problems, a distinct User-Agent string helps us isolate relevant logs so we can more effectively troubleshoot and reproduce any behavior, and (b) understanding usage of different clients (like nerdctl) helps us understand if there are additional features we can develop for the registry or if there are opportunities for us to help out in the clients themselves.

Thanks!
Sam

samuelkarp (approved):

AkihiroSuda (changes requested):
Sorry, NOTLGTM.

I don't think we should leak unnecessary information that can be used for identifying users (and also for finding vulnerable versions of clients, potentially) without the user's consent.

AkihiroSuda:
Maybe we can extend `hosts.toml` to support specifying custom user agents (or any HTTP header), and Amazon can extend  `docker-credential-ecr-login` to automatically configure `hosts.toml` (with notice to users, perhaps).

samuelkarp:
Further discussion is on https://github.com/containerd/containerd/issues/6474.  I'd still like to see this change made as I think that having nerdctl version information is valuable to the ecosystem at large.

tompittson:
Another issue I have encountered that this will solve is that the User-Agent header is the default "Go-http-client" when executing a nerdctl login command.

apostasie:
Curious what is our thinking overall on this now, and if it has evolved over the last 3 years.

Personal 2 cents:
- agreed with @samuelkarp that a distinct UA string is not a threat to anonymity and that it has definite benefits for registry operators
- I also do not think there is a realistic scenario where an attacker would derive actual value from a UA string (can elaborate on that if needed)

Now:
- I do not see a reason to customize containerd UA string - nerdctl pulls *are* containerd pulls, right? - I do not see what kind of issues could arise with pull that would be nerdctl specific...
- I am more interested in the login command, which IS done client-side, and which IS fully-identifying the user already (making the anonymity argument moot), and in that case, I do believe we should have a default `nerdctl/vx.y.z` UA (that users can still override from hosts.toml if they wish) - and that one would certainly help with point b from https://github.com/containerd/nerdctl/pull/704#issuecomment-1020593794

Where do you folks stand?

AkihiroSuda:
At least we should hide the git commit hash

Not just for anonymity...
https://github.com/kubernetes/registry.k8s.io/issues/286#issuecomment-2216874638

apostasie:
> At least we should hide the git commit hash
> 
> Not just for anonymity... [kubernetes/registry.k8s.io#286 (comment)](https://github.com/kubernetes/registry.k8s.io/issues/286#issuecomment-2216874638)

Yep. `nerdctl/vx.y.z` is absolutely enough (and again, *only* for `login`).

I must say I am ambivalent on this. I do definitely support using a specific UA for point b. mentioned above as I said - but I do think server-side behavioral adaptation based on UA sniffing is just abhorrent (there is a reason the frontend world moved to feature testing rather than UA sniffing, and that any operator worth their salt are instead looking at request urls and overall behaviors rather than UA).

Someone in the kube discussion you provided is saying just as much - waf rules on UA are idiotic and very easy to workaround.

Issue #624, opened 2021-12-15
proposal: support namespace switch command


junnplus:
Switching between multiple namespaces is cumbersome.
Support switching namespace commands to make it easier for users to use.
- `nerdctl namespace switch <ns>` switch to the specified namespace
- `nerdctl namespace switch -` switch to the previous context

AkihiroSuda:
Perhaps `nerdctl namespace use`, or `nerdctl context use`, to mimic `docker context use`
https://docs.docker.com/engine/reference/commandline/context_use/

Junnplus:
@AkihiroSuda `namespace use` is OK, `nerdctl context use` is best to be consistent with `docker context use`.

Junnplus:
maybe need https://github.com/containerd/nerdctl/pull/470

fahedouch:
`namespace` should be related to `context` , `nerdctl context use` makes more sense to switch namespace ; 

this can be sthg like this:
`nerdctl context create context1 --namespace=ns1`
and then to switch to `ns1` : 
`nerdctl context use context1`

PR #707, opened 2022-01-17
Add namespace use command

Junnplus:

Fixes: https://github.com/containerd/nerdctl/issues/624

Why not use `nerdctl context use`?
1. `namespace use` is more direct
2. If the `context create context --namespace ns1` command is added later, also can use `namespace use ns2` to switch the namespace of the current context

### DIFF
--- a/cmd/nerdctl/namespace_use.go
+++ b/cmd/nerdctl/namespace_use.go
@@ -1,0 +1,84 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+
+	ncdefaults "github.com/containerd/nerdctl/pkg/defaults"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const (
+	LastNamespace = "last-namespace"
+)
+
+func newNamespaceUseCommand() *cobra.Command {
+	namespaceUseCommand := &cobra.Command{
+		Use:           "use",
+		Short:         "use a namespace",
+		RunE:          namespaceUseAction,
+		Args:          cobra.ExactArgs(1),
+		SilenceUsage:  true,
+		SilenceErrors: true,
+	}
+	return namespaceUseCommand
+}
+
+func namespaceUseAction(cmd *cobra.Command, args []string) error {
+	tomlPath := ncdefaults.NerdctlTOML()
+	if v, ok := os.LookupEnv("NERDCTL_TOML"); ok {
+		tomlPath = v
+	}
+	cfg, err := NewConfigFromTOML(tomlPath)
+	if err != nil {
+		return err
+	}
+	namespace := args[0]
+	if namespace == "-" {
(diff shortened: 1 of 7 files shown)


pkg/defaults/defaults_linux.go
```diff
@@ -94,15 +94,15 @@ func BuildKitHost() string {
 	return fmt.Sprintf("unix://%s/buildkit/buildkitd.sock", xdr)
 }
 
-func NerdctlTOML() string {
+func NerdctlPath() string {
```
AkihiroSuda, line 97:
```suggestion
func NerdctlConfigDir() string {
```

cmd/nerdctl/namespace_use.go
```diff
@@ -0,0 +1,84 @@
...
+	"os"
+	"path/filepath"
+
+	ncdefaults "github.com/containerd/nerdctl/pkg/defaults"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const (
+	LastNamespace = "last-namespace"
```
AkihiroSuda, line 29: nit: maybe `current-namespace`

AkihiroSuda: Also please explain this in https://github.com/containerd/nerdctl/blob/master/docs/dir.md#config

cmd/nerdctl/namespace_use.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func newNamespaceUseCommand() *cobra.Command {
+	namespaceUseCommand := &cobra.Command{
+		Use:           "use",
```
AkihiroSuda, line 34: Please explain this in Command Reference
https://github.com/containerd/nerdctl#command-reference

cmd/nerdctl/namespace_use.go
```diff
@@ -0,0 +1,84 @@
...
+		} else {
+			logrus.Debugf("missing last namespace: %v", err)
+			namespace = "default"
+		}
+	}
+	if err := writeLastNamespace(cfg.Namespace); err != nil {
+		return err
+	}
+	cfg.Namespace = namespace
+	if err := cfg.MarshalToTOML(tomlPath); err != nil {
```
AkihiroSuda, line 67: Why do we need to marshal TOML too?

AkihiroSuda: This results in losing comment lines in the toml

cmd/nerdctl/namespace_use.go
```diff
@@ -0,0 +1,84 @@
...
+	tomlPath := ncdefaults.NerdctlTOML()
+	if v, ok := os.LookupEnv("NERDCTL_TOML"); ok {
+		tomlPath = v
+	}
+	cfg, err := NewConfigFromTOML(tomlPath)
+	if err != nil {
+		return err
+	}
+	namespace := args[0]
+	if namespace == "-" {
```
AkihiroSuda, line 54: Needs explanation

cmd/nerdctl/namespace_use.go
Junnplus: `last-namespace` is used to record the last used namespace, used by `namespace use -`

cmd/nerdctl/namespace_use.go
Junnplus: > Why do we need to marshal TOML too?

My idea is to directly modify the namespace field of toml.

> This results in losing comment lines in the toml

Yes, I didn't think of a better way :(

cmd/nerdctl/namespace_use.go
```diff
@@ -0,0 +1,84 @@
...
+
+func namespaceUseAction(cmd *cobra.Command, args []string) error {
+	tomlPath := ncdefaults.NerdctlTOML()
+	if v, ok := os.LookupEnv("NERDCTL_TOML"); ok {
+		tomlPath = v
+	}
+	cfg, err := NewConfigFromTOML(tomlPath)
+	if err != nil {
+		return err
+	}
```
fahedouch, line 52: `NerdctlTOML`  should be emerged by cobra. I made a PR for this https://github.com/containerd/nerdctl/pull/716

PR #716, opened 2022-01-19
config path handled by cmd

fahedouch:
The config path should be handled by cobra cmd. This can be overridden by the user for subcommands

Signed-off-by: fahedouch <fahed.dorgaa@gmail.com>

### DIFF
--- a/cmd/nerdctl/main.go
+++ b/cmd/nerdctl/main.go
@@ -134,6 +134,9 @@ func initRootCmdFlags(rootCmd *cobra.Command, tomlPath string) error {
 			return err
 		}
 	}
+
+    // Overriding config path for subcommands
+	rootCmd.PersistentFlags().String("config", tomlPath, "location of config file")
 	rootCmd.PersistentFlags().Bool("debug", cfg.Debug, "debug mode")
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/main.go
```diff
@@ -134,6 +134,9 @@ func initRootCmdFlags(rootCmd *cobra.Command, tomlPath string) error {
 			return err
 		}
 	}
+
+    // Overriding config path for subcommands
+	rootCmd.PersistentFlags().String("config", tomlPath, "location of config file")
```
AkihiroSuda, line 139: Where is this referred?

fahedouch: looks like this PR need this https://github.com/containerd/nerdctl/pull/707#discussion_r789811259

AkihiroSuda: Not sure. But this PR seems wrong anyway, because tomlPath has to be determined before parsing the toml.

AkihiroSuda (changes requested):
Doesn't seem correct
https://github.com/containerd/nerdctl/pull/716/files#r790359282

cmd/nerdctl/main.go
fahedouch: Yes I know. The idea is to emerge the tomlPath to the application via cobra.

PR #724, opened 2022-01-23
Fixes 404 error when authenticating with v2 reg

### LINKED ISSUE #715: Valid login to ghcr.io returns 404 and pulling images still returning 401 due to failed flow
Hi,

When trying to login to Github Container Registry with correct credentials nerdctl returns 404 and I am unable to pull images due to being unauthorized. This is presumable due to the failed login flow. See below for output with --debug full

```
nerdctl login --debug-full -u dylanrhysscott ghcr.io
Enter Password:
DEBU[0000] Ignoring hosts dir "/etc/containerd/certs.d"  error="stat /etc/containerd/certs.d: no such file or directory"
DEBU[0000] len(regHosts)=1
ERRO[0001] failed to call tryLoginWithRegHost            error="unexpected status code 404" i=0
FATA[0001] unexpected status code 404
```

* I'm using [Rancher Desktop 1.0.0-beta1](https://github.com/rancher-sandbox/rancher-desktop/releases/tag/v1.0.0-beta.1)
* Nerdctl and containerd versions
* **Note: Swapping Rancher Desktop away from containerd to dockerd resolves the issue - This is probably related to containerd**

```
nerdctl version
Client:
 Version:	v0.16.0
 Git commit:	0ddaffd2ce304dd917267470883898feff9463dc

Server:
 containerd:
  Version:	v1.5.8-k3s2
  GitCommit:
```
* MacOS Big Sur 11.3.1
* Kernel version

```
uname -rv
20.4.0 Darwin Kernel Version 20.4.0: Thu Apr 22 21:46:47 PDT 2021; root:xnu-7195.101.2~1/RELEASE_X86_64
```

I see there was a related issue raised about ghcr.io specifically here but doesn't look like the same symptoms https://github.com/containerd/nerdctl/issues/121. I would like to adopt containerd / nerdctl as a replacement to Docker Desktop. I suspect having done some reading the container registry provided by Github doesn't support the /v2 endpoint or lacks the /v1 version. Can't be sure which nerdctl uses. Any ideas on how to get more debug out of the CLI to confirm?

Thanks

dylanrhysscott:
After applying this fix I can successfully authenticate with github registry

```
[vagrant@localhost nerdctl]$ nerdctl login -u <USERNAME> ghcr.io
Enter Password:
Login Succeeded
[vagrant@localhost nerdctl]$

[vagrant@localhost ~]$ nerdctl pull ghcr.io/<IMAGE>
ghcr.io/<IMAGE>:                            resolved       |++++++++++++++++++++++++++++++++++++++|
manifest-sha256:9d7715dd1e7a20a03c1fda6b625ab73aa8e1a9cacddad885b79876bd2a8ebce8: done           |++++++++++++++++++++++++++++++++++++++|
config-sha256:6f7f6bfd79125762caacefdb0e83ece7e101c85279d31fbc36f61d643f878a42:   done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:abd092694f3b3635a1642c6fed15a81f02bbcd44045fc5dfb138eda3531899bf:    done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:b53a42fddd847a0d1d550565601c9299fecae84e94ffcb8c9a9e5e98231e0ee0:    done           |++++++++++++++++++++++++++++++++++++++|
elapsed: 3.8 s                                                                    total:  7.8 Mi (2.1 MiB/s)
[vagrant@localhost ~]$

```

This maybe a slightly naive approach so welcome feedback on a better place for this - Not sure who to ping from containerd 

Closes #715

### DIFF
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -253,6 +253,12 @@ func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost) error {
 		Host:   rh.Host,
 		Path:   rh.Path,
 	}
+	if rh.Path == "/v2" {
+		// If the path is using /v2 endpoint but lacks trailing slash add it
+		// https://docs.docker.com/registry/spec/api/#detail. Acts as a workaround
+		// for containerd issue https://github.com/containerd/containerd/blob/2986d5b077feb8252d5d2060277a9c98ff8e009b/remotes/docker/config/hosts.go#L110
+		rh.Path = "/v2/"
+	}
 	ctx = docker.WithScope(ctx, "")
 	var ress []*http.Response
 	for i := 0; i < 10; i++ {


cmd/nerdctl/login.go
```diff
@@ -253,6 +253,12 @@ func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost) error {
 		Host:   rh.Host,
 		Path:   rh.Path,
 	}
+	if rh.Path == "/v2" {
+		// If the path is using /v2 endpoint but lacks trailing slash add it
+		// https://docs.docker.com/registry/spec/api/#detail. Acts as a workaround
+		// for containerd issue https://github.com/containerd/containerd/blob/main/remotes/docker/config/hosts.go#L110
```
AkihiroSuda, line 259: Thanks, could you open the PR in the containerd repo

AkihiroSuda: Also, please sign the commit for DCO
https://github.com/apps/dco

cmd/nerdctl/login.go
```diff
@@ -253,6 +253,12 @@ func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost) error {
 		Host:   rh.Host,
 		Path:   rh.Path,
 	}
+	if rh.Path == "/v2" {
```
alisson276, line 256:
```
if !strings.HasSuffix(rh.Path, "/") {
  rh.Path += "/"
```

This way you avoid hard-coded `v2` in the code. Don't you think?

dylanrhysscott: Thanks I'd forgotten about HasSuffix from that package - looks like this PR will be closed off and merged elsewhere but that would be a better approach

cmd/nerdctl/login.go
dylanrhysscott: I have raised https://github.com/containerd/containerd/pull/6470 with the needed DCO commit would you like me to close this PR?

AkihiroSuda: Yes, let me close this PR, assuming we can merge https://github.com/containerd/containerd/pull/6470

AkihiroSuda:
Thanks, I'm closing this, as we now have https://github.com/containerd/containerd/pull/6470

https://github.com/containerd/nerdctl/pull/724#discussion_r790505829

cmd/nerdctl/login.go
dylanrhysscott: Brilliant thanks will keep an eye out for any further changes in new PR

AkihiroSuda:
Reopened per discussion in https://github.com/containerd/containerd/pull/6470 .
Sorry for going back ans force.

cmd/nerdctl/login.go
dylanrhysscott: @AkihiroSuda @alisson276  I'm looking at this suggestion and I'm not sure to implement seems a good change but as far as I can tell the login issues are only affecting v2 and adding this applies it to every version of registry - Would it be best to just apply the leading slash to /v2 only or globally?

AkihiroSuda: I think just checking `if rh.Path == "/v2"` is fine.

cmd/nerdctl/login.go
AkihiroSuda, line 259: > https://github.com/containerd/containerd/blob/main/remotes/docker/config/hosts.go#L110

Please use a permanent link with the commit hash (or just the tag), e.g., 
https://github.com/containerd/containerd/blob/2986d5b077feb8252d5d2060277a9c98ff8e009b/remotes/docker/config/hosts.go#L110

cmd/nerdctl/login.go
alisson276: I just think that keep in sync with the containerd library is not a good idea. If, for any reason, they change that path, it must be synced here. If you are worried about always adding a slash, the safest approach is to try twice: first using the rh.Path (no matter what it is, if v2, v3, whatever) and if fails, do a second try appending the slach at the end (IMO).

alisson276: If I correctly understood, in the PR they are asking to add this explicit in the GET call... in other words, do a:
```
req, err := http.NewRequest(http.MethodGet, u.String() + "/", nil)
```
So, this `if` is ok, I just don't know if fixing the `v2` is a good idea, but this is just my opinion. I'm just trying to contribute because this problem afftects my environment too.

dylanrhysscott: @alisson276 v2 is part of the docker registry specification which I think is pretty static and as far as I know unlikely to change but it maybe a valid concern in future - with regards to doesn't happen with Docker question docker uses the dockerd daemon and not containerd directly as well as using /v1 and not /v2. This might explain the difference

cmd/nerdctl/login.go
dylanrhysscott: Fixed

AkihiroSuda (approved):
Thanks

Issue #594, opened 2021-12-04
/dev/shm support is broken (Was: `What options does the --tmpfs flag in the run statement allow`)


doberkofler:
I'm trying to a run an Oracle  Database in a Docker image in a lima environment on macOS and get a lot of errors related to the shm volume.

```bash
lima nerdctl run --detach --name="oracledb" --network="oraclenet" --memory="4g" --publish="1521:1521" --volume="/dev/shm" --tmpfs="/dev/shm:rw,exec,size=1g" qualiant/database
```

As there is no specific documentation, I wanted to ask a few specific questions:

1) Does `nerdctl` all the options `--tmpfs="/dev/shm:rw,exec,size=1g"` in the `--tmpfs` flag ?
2) `nerdctl` support the `--memory` option but how does this relate to lima? Is it guaranteed that the container gets 4g when using the `--memory` flag independently from lima?

Thank you!

doberkofler:
After some more testing , it seems as if at least the `exec` would option not be observed when using
`--tmpfs="/dev/shm:rw,exec,size=1g"`

AkihiroSuda:
1. All the tmpfs options are supported, however, there seems to be a bug specific to `/dev/shm` path, as nerdctl mounts its own `/dev/shm` 
```
$ nerdctl run --rm --tmpfs="/dev/shm:rw,exec,size=1g" alpine grep /dev/shm /proc/mounts
shm /dev/shm tmpfs rw,nosuid,nodev,noexec,relatime,size=65536k,uid=1001,gid=1001,inode64 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev,relatime,size=1048576k,uid=1001,gid=1001,inode64 0 0

$ docker run --rm --tmpfs="/dev/shm:rw,exec,size=1g" alpine grep /dev/shm /proc/mounts
tmpfs /dev/shm tmpfs rw,nosuid,nodev,relatime,size=1048576k,inode64 0 0
```
(nerdctl v0.14.0)

2. Unrelated. Lima memory is an amount of QEMU memory (which looks like the physical memory to the Linux guest), and nerdctl memory is just cgroup memory limits.

doberkofler:
@AkihiroSuda Thank you for your very helpful feedback. Do you expect this problem to be fixed in the near future?

doberkofler:
@AkihiroSuda I would be most appreciated, if you could give some update on this bug.

Junnplus:
[...]

PR #728, opened 2022-01-25
Fix duplicate mount /dev/shm

Junnplus:
Fixes: https://github.com/containerd/nerdctl/issues/594


cmd/nerdctl/run_mount_linux_test.go
```diff
@@ -226,6 +226,10 @@ func TestRunTmpfs(t *testing.T) {
 	base := testutil.NewBase(t)
 	f := func(allow, deny []string) func(stdout string) error {
 		return func(stdout string) error {
+			lines := strings.Split(strings.TrimRight(stdout, "\n"), "\n")
```
fahedouch, line 229: `strings.TrimSpace(stdout, "\n")`  in case of leading space too

AkihiroSuda (approved):
Thanks

PR #645, opened 2021-12-21
build for macOS

robberphex:
* Now can build nerdctl at macOS.
* `nerdctl version` works at macOS. Other command is WIP.
* Can build debug version via `make nerdctl GOGCFLAGS="all=-N -l" GOLDFLAGS=""`

PR #732, opened 2022-01-25
darwin build

thehajime:
This is yet another mac build PR.  darwin support of containerd is still underway, but would like to ask your opnions.

There are several notes on the current status:

- On amd64 darwin, this patch can do `nerdctl run` with `--net=none` option; with --net options, it won't work.  On arm64/darwin, nerdctl itself works, but there is no available runtime.
- go.mod tentatively put a commit hash which is not the upstreamed contaienrd; I will update this PR to non-draft one once https://github.com/containerd/containerd/pull/5935 is finalized.
- cross build for darwin requires osxcross (https://github.com/tpoechtrager/osxcross) because dependent package (i.e., moby/sys) needs CGO_ENABLED.

This might be a duplicate of https://github.com/containerd/nerdctl/pull/645


.github/workflows/test.yml
```diff
@@ -108,7 +114,24 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 1
+    - name: OSXCross for CGO Support
```
AkihiroSuda, line 117: Not sure about legal status of this.
Better not to use it, probably.

AkihiroSuda: We can just compile it on macOS instances of GHA

AkihiroSuda:
This has to be marked as an experimental feature in `docs/experimental.md`.
We also need a well written document, perhaps in `docs/darwin.md`, or `docs/lkl.md`.
Differences from `lima nerdctl` should be explained in the docs, too.

.github/workflows/test.yml
thehajime: > Not sure about legal status of this.
> Better not to use it, probably.

indeed..

> We can just compile it on macOS instances of GHA

I see.
if we don't have to publish the binaries via `make artifacts` over cross compiling on a Linux host, it should be okay.

I will seek a way to handle this.

thehajime:
> This has to be marked as an experimental feature in `docs/experimental.md`.
> We also need a well written document, perhaps in `docs/darwin.md`, or `docs/lkl.md`.
> Differences from `lima nerdctl` should be explained in the docs, too.

okay.  I'll also prepare documents to address the above concerns.

apostasie:
@thehajime curious if this is still something people want to pursue.

AkihiroSuda:
It is still nice to support Darwin hosts, especially for Wasm and VM runtimes

slonopotamus:
See #3763 for a new attempt.

Issue #736, opened 2022-01-26
`nerdctl inspect` faild to  get {{.Id}} but `docker` did


tanguofu:
[root@VM-141-67-centos ~]# nerdctl inspect registry:latest  --format '{{.Id}}'
FATA[0000] template: :1:2: executing "" at <.Id>: can't evaluate field Id in type *dockercompat.Image 
[root@VM-141-67-centos ~]# docker inspect registry:latest  --format '{{.Id}}'
sha256:b8604a3fe8543c9e6afc29550de05b36cd162a97aa9b2833864ea8a5be11f3e2

AkihiroSuda:
Workaround: `nerdctl image inspect docker.io/library/alpine:latest | jq -r .[0].Id` 
(This works for `docker` too)

tanguofu:
> 

ok ， thanks very much！

AkihiroSuda:
Fixed in https://github.com/containerd/nerdctl/pull/743

PR #743, opened 2022-01-27
enhance inspect

fahedouch:
- enhance https://github.com/containerd/nerdctl/issues/736
-  fix duplicate inspect entries

### DIFF
--- a/cmd/nerdctl/container_inspect.go
+++ b/cmd/nerdctl/container_inspect.go
@@ -81,6 +81,7 @@ func containerInspectAction(cmd *cobra.Command, args []string) error {
 	if len(errs) > 0 {
 		return fmt.Errorf("%d errors: %v", len(errs), errs)
 	}
+
 	return formatSlice(cmd, f.entries)
 }
 
--- a/cmd/nerdctl/fmtutil.go
+++ b/cmd/nerdctl/fmtutil.go
@@ -75,13 +75,41 @@ func formatSlice(cmd *cobra.Command, x []interface{}) error {
 		for _, f := range x {
 			var b bytes.Buffer
 			if err := tmpl.Execute(&b, f); err != nil {
-				return err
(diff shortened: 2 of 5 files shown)


cmd/nerdctl/container_inspect.go
```diff
@@ -69,18 +69,13 @@ func containerInspectAction(cmd *cobra.Command, args []string) error {
...
-		n, err := walker.Walk(ctx, req)
-		if err != nil {
-			errs = append(errs, err)
-		} else if n == 0 {
-			errs = append(errs, fmt.Errorf("no such object: %s", req))
-		}
-	}
-	if len(errs) > 0 {
-		return fmt.Errorf("%d errors: %v", len(errs), errs)
+	n, err := walker.Walk(ctx, args[0])
```
AkihiroSuda, line 72: Why only args[0]

cmd/nerdctl/fmtutil.go
```diff
@@ -75,7 +75,12 @@ func formatSlice(cmd *cobra.Command, x []interface{}) error {
 		for _, f := range x {
 			var b bytes.Buffer
 			if err := tmpl.Execute(&b, f); err != nil {
-				return err
+				if _, ok := err.(template.ExecError); ok {
+					// FallBack to Raw Format
```
AkihiroSuda, line 79: Why do we need this

fahedouch: We need this to encode struct with `tags` (RawFormat) and execute tmpl on the  RawFormat in the case where `--format` point to tag like  `Id` and not  `ID`

cmd/nerdctl/container_inspect.go
fahedouch: I forgot that we can inspect an image or a container directly 😅

AkihiroSuda (approved):
Thanks

PR #641, opened 2021-12-20
prompt for username when --username not used

simorgh1:
Signed-off-by: Bahram Maravandi <bahi@hey.com>

PR #752, opened 2022-01-28
enhance login prompt

fahedouch:
replace https://github.com/containerd/nerdctl/pull/641 ( thanks @simorgh1 ) 

### DIFF
--- a/cmd/nerdctl/login_unix.go
+++ b/cmd/nerdctl/login_unix.go
@@ -20,8 +20,10 @@
 package main
 
 import (
+	"bufio"
 	"fmt"
 	"os"
+	"strings"
 	"syscall"
 
 	"golang.org/x/term"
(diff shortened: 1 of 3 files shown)


simorgh1:
Not entering the username raises the following error
FATA[0002] error reading username: unexpected newline
instead of the validation error: username is required

I had the same issue when assigning the username,
options.username = username

By not entering username, the value will be a newline and option.username raises the above error.

AkihiroSuda:
Please squash

cmd/nerdctl/login_unix.go
```diff
@@ -46,3 +48,26 @@ func readPassword() (string, error) {
 
 	return string(bytePassword), nil
 }
+
+func readUsername() (string, error) {
+	var fd *os.File
+	if term.IsTerminal(syscall.Stdin) {
+		fd = os.Stdin
+	} else {
+		tty, err := os.Open("/dev/tty")
```
AkihiroSuda, line 57: Shouldn’t we just read stdin even when it is not a terminal?

simorgh1: I would like to know the use case that one would pass username **and** password by stdin and why, then password alone has such case, because of secret that is passed but username as stdin? docker login has only password stdin option.
cat secret-password | docker login --username foo --password-stdin

fahedouch: adding username to stdin is not intuitive without `username-stdin` which is rarely used.

AkihiroSuda: So, `readUsername()` should return an error when stdin is not a terminal?
I don't think we have to open `/dev/tty` here.

AkihiroSuda: Added a follow-up commit to avoid /dev/tty

PR #757, opened 2022-01-30
Fetch full PR branch

### LINKED ISSUE #756: Project Checks fail when PR branch is long
I have a medium size (by my standard) PR and the `Project Checks` DCO check fails because it doesn't have enough commits to see the end.

```
  fatal: bad revision 'c101a5ec56fef7bf18021e895d8a861edd117c64^..c101a5ec56fef7bf18021e895d8a861edd117c64'
   * c101a5e "..." ... FAIL
    - FAIL - errored with: exit status 128
```

There are other ways to write the CI so that it would work.

jsoref:
Closes #756

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,14 @@ jobs:
       with:
         path: src/github.com/containerd/nerdctl
         fetch-depth: 25
+        ref: ${{ github.event.pull_request.base.sha }}
+    - name: Retrieve PR Head
+      shell: bash
+      if: ${{ github.event.pull_request.head.sha }}
+      run: |
+        cd src/github.com/containerd/nerdctl;
+        git fetch origin refs/pull/${{ github.event.pull_request.number }}/merge;
+        git checkout FETCH_HEAD
     - uses: containerd/project-checks@v1.0.2
       with:
         working-directory: src/github.com/containerd/nerdctl


.github/workflows/test.yml
```diff
@@ -20,6 +20,14 @@ jobs:
       with:
         path: src/github.com/containerd/nerdctl
         fetch-depth: 25
```
AkihiroSuda, line 22: Can we just increase fetch-depth here?

jsoref: You could pick any arbitrary number, it'll always eventually be wrong.

I personally use a different approach for my own workflows, but it's more complicated than what you have (which is broken) or what I'm proposing (which will solve the PR case, but wouldn't technically solve a Push event with a similarly long set of commits.

AkihiroSuda: > You could pick any arbitrary number, it'll always eventually be wrong.

100 should be enough for everyone.

If you really need to fetch all commits you can just remove `fetch-depth` line, but I don't see necessity of doing that.

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/757#discussion_r795384162

jsoref:
I accidentally broke the link to this branch, closing -- will try to reopen.

PR #774, opened 2022-02-01
purge useless login func

fahedouch:
I don't see any need in redoing a loginClientSide here since it's already been done before

### DIFF
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -123,9 +123,6 @@ func loginAction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("error saving credentials: %w", err)
 	}
 
-	if _, err = loginClientSide(ctx, cmd, *authConfig); err != nil {
-		return err
-	}
 	fmt.Fprintln(cmd.OutOrStdout(), "Login Succeeded")
 
 	return nil
@@ -217,9 +214,6 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
 				// Even containerd/CRI does not support RegistryToken as of v1.4.3,
 				// so, nobody is actually using RegistryToken?
 				logrus.Warnf("RegistryToken (for %q) is not supported yet (FIXME)", host)
-			}
-			if auth.IdentityToken != "" {
-				return "", auth.IdentityToken, nil
 			}
 			return auth.Username, auth.Password, nil
 		}


cmd/nerdctl/login.go
```diff
@@ -123,9 +123,6 @@ func loginAction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("error saving credentials: %w", err)
 	}
 
-	if _, err = loginClientSide(ctx, cmd, *authConfig); err != nil {
-		return err
-	}
```
AkihiroSuda, line 128: The second call has a different `*authConfig`, so I'm not sure we can safely remove this.

fahedouch: this second login can be based on : 

1.  empty / Password (Identity token) --> useless because we previously did a login with username/password
2. Username / Password  (not Identity token) --> this is done previously 

I wondering what is the reason behind making double login operation especially when filled the `CredentialsStore` with creds

fahedouch:
Log extract the same action x 2 
```
WARN[0000] WARNING! Using --password via the CLI is insecure. Use --password-stdin. 
DEBU[0000] Ignoring hosts dir "/etc/containerd/certs.d"  error="stat /etc/containerd/certs.d: no such file or directory"
DEBU[0000] len(regHosts)=1                              
DEBU[0000] no scope specified for token auth challenge   host=vpodapiregistry.azurecr.io
DEBU[0000] Ignoring hosts dir "/etc/containerd/certs.d"  error="stat /etc/containerd/certs.d: no such file or directory"
DEBU[0000] len(regHosts)=1                              
DEBU[0000] no scope specified for token auth challenge   host=vpodapiregistry.azurecr.io
```

cmd/nerdctl/login.go
```diff
@@ -218,9 +215,6 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
 				// so, nobody is actually using RegistryToken?
 				logrus.Warnf("RegistryToken (for %q) is not supported yet (FIXME)", host)
 			}
-			if auth.IdentityToken != "" {
-				return "", auth.IdentityToken, nil
```
fahedouch, line 222: this line is never visited

Junnplus (approved):
LGTM on green.

I tested multiple image repositories locally and they all work.

fahedouch:
test with 

- [x]  GCR : username/password  `AND` oauth2accesstoken/accessToken
- [x]  ACR : username/password `AND`  00000000-0000-0000-0000-000000000000/accessToken
- [x]  Jfrog: username/password  `AND`    username/accessToken
- [x]  dockerHub
- [x]  AWS: username/password `AND`  AWS/accessToken

@AkihiroSuda  OK for most registries

AkihiroSuda (approved):
Thanks

Issue #787, opened 2022-02-07
`nerdctl run --runtime=(io.containerd.kata.v2|io.containerd.runsc.v1)` silently fails to create network devices (probably skipping OCI hooks)


AkihiroSuda:
`ip a` shows no network devices in Kata containers and gVisor containers.

- [ ] https://github.com/kata-containers/kata-containers/issues/1935
- [ ] https://github.com/google/gvisor/issues/7149

Seems they lack support for OCI `createRuntime` hooks

PR #788, opened 2022-02-07
Workarounds for gVisor (and Kata)

AkihiroSuda:
## Commit 1: `ocihook: use prestart (deprecated, though) instead of createRuntime`
`prestart` is deprecated in favor of `createRuntime` which was added in the OCI Runtime Spec v1.0.2,  however, gVisor and Kata still do not support createRuntime.
    
Workaround for #787
- https://github.com/kata-containers/kata-containers/issues/1935
- https://github.com/google/gvisor/issues/7149

## Commit 2: `ocihook: workaround for gVisor/Kata state annotations bug`
 gVisor and Kata do not propagate annotations automatically, so we have to copy the annotations from the spec to the state.

- https://github.com/kata-containers/kata-containers/issues/3629

## Commit 3: `CNI: enable "loopback" plugin by default`

gVisor requires "loopback" plugin to be enabled explicitly to bring up the loopback interface.

- - -

# gVisor

```console
# runsc --version
runsc version release-20220131.0
spec: 1.0.2-dev

# nerdctl run -it --rm --runtime=io.containerd.runsc.v1 alpine
/ # ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65522 
    link/loopback 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff
    inet 127.0.0.1/8 scope global dynamic 
    inet6 ::1/128 scope global dynamic 
2: eth0: <UP,LOWER_UP> mtu 1500 
    link/ether 1e:a7:b5:7a:4b:d4 brd ff:ff:ff:ff:ff:ff
    inet 10.4.0.19/24 scope global dynamic 
    inet6 fe80::1ca7:b5ff:fe7a:4bd4/64 scope global dynamic 
/ # apk update
fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/community/x86_64/APKINDEX.tar.gz
v3.15.0-258-gd5d32e35cf [https://dl-cdn.alpinelinux.org/alpine/v3.15/main]
v3.15.0-259-g8b98485bc8 [https://dl-cdn.alpinelinux.org/alpine/v3.15/community]
OK: 15857 distinct packages available
```

(still no support for rootless)

# Kata

```console
# /opt/kata/bin/kata-runtime --version
kata-runtime  : 2.3.2
   commit   : 1af292c9e693e9bc8e8324a9eb860dad45306fb5
   OCI specs: 1.0.2-dev

# nerdctl run -it --rm --runtime=io.containerd.kata.v2 alpine
FATA[0000] failed to create shim task: Unsupported network interface: bridge: unknown 
```

<details>
<summary>CNI config</summary>

<p>

```console
# nerdctl network inspect bridge --mode=native
[
    {
        "CNI": {
            "cniVersion": "0.4.0",
            "name": "bridge",
            "nerdctlID": 0,
            "nerdctlLabels": {},
            "plugins": [
                {
                    "type": "bridge",
                    "bridge": "nerdctl0",
                    "isGateway": true,
                    "ipMasq": true,
                    "hairpinMode": true,
                    "ipam": {
                        "ranges": [
                            [
                                {
                                    "gateway": "10.4.0.1",
                                    "subnet": "10.4.0.0/24"
                                }
                            ]
                        ],
                        "routes": [
                            {
                                "dst": "0.0.0.0/0"
                            }
                        ],
                        "type": "host-local"
                    }
                },
                {
                    "type": "portmap",
                    "capabilities": {
                        "portMappings": true
                    }
                },
                {
                    "type": "firewall"
                },
                {
                    "type": "tuning"
                },
                {
                    "type": "loopback",
                    "name": "lo"
                },
                {
                    "type": "isolation"
                }
            ]
        },
        "NerdctlID": 0,
        "NerdctlLabels": {}
    }
]
```

</p>

</details>

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -788,8 +788,11 @@ func withNerdctlOCIHook(cmd *cobra.Command, id, stateDir string) (oci.SpecOpts, 
 		if s.Hooks == nil {
 			s.Hooks = &specs.Hooks{}
 		}
-		crArgs := append(args, "createRuntime")
-		s.Hooks.CreateRuntime = append(s.Hooks.CreateRuntime, specs.Hook{
+		// prestart is deprecated in favor of createRuntime added in OCI Runtime Spec v1.0.2,
+		// however, as of Feb 2022, gVisor and Kata still do not support createRuntime.
+		// https://github.com/containerd/nerdctl/issues/787
+		crArgs := append(args, "prestart")
+		s.Hooks.Prestart = append(s.Hooks.CreateRuntime, specs.Hook{
 			Path: selfExe,
 			Args: crArgs,
 			Env:  os.Environ(),
--- a/docs/cni.md
+++ b/docs/cni.md
@@ -7,7 +7,7 @@ either `--network` or `--net` option.
 
 nerdctl support some basic types of CNI plugins without any configuration
 needed(you should have CNI plugin be installed), for Linux systems the basic
-CNI plugin types are `bridge`, `portmap`, `firewall`, `tuning`, for Windows
+CNI plugin types are `bridge`, `portmap`, `firewall`, `tuning`, `loopback`, for Windows
 system, the supported CNI plugin types are `nat` only.
 
 The default network `bridge` for Linux and `nat` for Windows if you
@@ -50,6 +50,10 @@ Configuration of the default network `bridge` of Linux:
     },
     {
       "type": "tuning"
+    },
+    {
+      "type": "loopback",
+      "name": "lo"
     },
     {
       "type": "isolation"
--- a/pkg/netutil/cni_plugin_unix.go
+++ b/pkg/netutil/cni_plugin_unix.go
@@ -95,6 +95,23 @@ func (*tuningConfig) GetPluginType() string {
 	return "tuning"
 }
 
+// loopbackConfig describes the loopback plugin (required for gVisor)
+type loopbackConfig struct {
+	PluginType string `json:"type"`
+	Name       string `json:"name"`
+}
+
+func newLoopbackPlugin() *loopbackConfig {
+	return &loopbackConfig{
+		PluginType: "loopback",
+		Name:       "lo",
+	}
+}
+
+func (*loopbackConfig) GetPluginType() string {
+	return "loopback"
+}
+
 // https://github.com/containernetworking/plugins/blob/v1.0.1/plugins/ipam/host-local/backend/allocator/config.go#L47-L56
 type hostLocalIPAMConfig struct {
 	Type        string        `json:"type"`
--- a/pkg/netutil/netutil_unix.go
+++ b/pkg/netutil/netutil_unix.go
@@ -55,7 +55,8 @@ func GenerateCNIPlugins(driver string, id int, ipam map[string]interface{}, opts
 		bridge.IsGW = true
 		bridge.IPMasq = true
 		bridge.HairpinMode = true
-		plugins = []CNIPlugin{bridge, newPortMapPlugin(), newFirewallPlugin(), newTuningPlugin()}
+		// gVisor needs "loopback" plugin to be added explicitly
+		plugins = []CNIPlugin{bridge, newPortMapPlugin(), newFirewallPlugin(), newTuningPlugin(), newLoopbackPlugin()}
 	default:
 		return nil, fmt.Errorf("unsupported cni driver %q", driver)
 	}
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -73,8 +73,13 @@ func Run(stdin io.Reader, stderr io.Writer, event, dataStore, cniPath, cniNetcon
 	}
 
 	switch event {
-	case "createRuntime":
-		return onCreateRuntime(opts)
+	// prestart is deprecated in favor of createRuntime added in OCI Runtime Spec v1.0.2,
+	// however, as of Feb 2022, gVisor and Kata still do not support createRuntime.
+	// https://github.com/containerd/nerdctl/issues/787
+	//
+	// "prestart" is NOT a typo of "preStart".
+	case "prestart":
+		return onPrestart(opts)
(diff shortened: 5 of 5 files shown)


AkihiroSuda:
Kata v2 no longer supports "bridged" CNI?
https://github.com/kata-containers/kata-containers/blob/3fdc97e11091e0c404c4db49ab1c354075d2c3a2/src/runtime/virtcontainers/network.go#L1117-L1153

cc @liubin

liubin:
@AkihiroSuda this workaround may not work with Kata Containers. Kata Containers runtime(shim) will create a new network namespace and run hooks in this created namespace. In fact, CNI plugins should run in the host network namespace.

Basically said, in Kubernetes or containerd, kubelet/ctr will create the network namespace and pass it to Kata Containers, nerdctl uses a hook to create the network interfaces, and doesn't take the charge of creating network namespace.

So Kata Containers must do some changes as long as nerdctl uses this pattern to handle CNI plugins.

AkihiroSuda:
Linking kata-side fix: https://github.com/kata-containers/kata-containers/pull/3630

Thanks to @liubin

sameo:
> Kata v2 no longer supports "bridge" CNI? https://github.com/kata-containers/kata-containers/blob/3fdc97e11091e0c404c4db49ab1c354075d2c3a2/src/runtime/virtcontainers/network.go#L1117-L1153
> 
> cc @liubin

@AkihiroSuda Kata is not bound to any particular CNI plugin, but tries to plumb whatever the plugin created in the pod's netns with an interface (e.g. a TAP interface) passed to the VMM. CNI bridge is supported.

So the bridge plugin will create a veth pair, and leave one end on the pod's netns. Kata scans this netns and [finds the veth end](https://github.com/kata-containers/kata-containers/blob/3fdc97e11091e0c404c4db49ab1c354075d2c3a2/src/runtime/virtcontainers/network.go#L1144). It will then create a TAP interfaces, set TC mirroring between the veth  and TAP and boot the VM with the TAP.

So Kata does not have CNI specific implementations but behaves differently depending on the set of interfaces found on the pod's netns.

sameo:
> @AkihiroSuda this workaround may not work with Kata Containers. Kata Containers runtime(shim) will create a new network namespace and run hooks in this created namespace. In fact, CNI plugins should run in the host network namespace.
> 
> Basically said, in Kubernetes or containerd, kubelet/ctr will create the network namespace and pass it to Kata Containers, nerdctl uses a hook to create the network interfaces, and doesn't take the charge of creating network namespace.
> 

@liubin @AkihiroSuda Excuse my ignorance, but what's the reason for nerdctl to not create a netns for the container/pod?

liubin:
> > @AkihiroSuda this workaround may not work with Kata Containers. Kata Containers runtime(shim) will create a new network namespace and run hooks in this created namespace. In fact, CNI plugins should run in the host network namespace.
> > Basically said, in Kubernetes or containerd, kubelet/ctr will create the network namespace and pass it to Kata Containers, nerdctl uses a hook to create the network interfaces, and doesn't take the charge of creating network namespace.
> 
> @liubin @AkihiroSuda Excuse my ignorance, but what's the reason for nerdctl to _not_ create a netns for the container/pod?

@sameo I think nerdctl uses a Docker-style network process mechanism, that uses an OCI hook to create the network.

Now Kata Containers [run hooks in the network namespace](https://github.com/kata-containers/kata-containers/blob/07b9d93f5f94673921dc833167710178a42c0d71/src/runtime/pkg/katautils/create.go#L158-L161
) that Kata Containers created:

```go
// Run pre-start OCI hooks.
err = EnterNetNS(sandboxConfig.NetworkConfig.NetworkID, func() error {
	return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
})
```

For nerdctl, the hook runs in the created namespace, so the CNI plugins also create `cni0`, `eth0` interfaces in that namespace, which leads Kata Containers to report an error.

sameo:
> > > @AkihiroSuda this workaround may not work with Kata Containers. Kata Containers runtime(shim) will create a new network namespace and run hooks in this created namespace. In fact, CNI plugins should run in the host network namespace.
> > > Basically said, in Kubernetes or containerd, kubelet/ctr will create the network namespace and pass it to Kata Containers, nerdctl uses a hook to create the network interfaces, and doesn't take the charge of creating network namespace.
> > 
> > 
> > @liubin @AkihiroSuda Excuse my ignorance, but what's the reason for nerdctl to _not_ create a netns for the container/pod?
> 
> @sameo I think nerdctl uses a Docker-style network process mechanism, that uses an OCI hook to create the network.

Right. And those hooks happen to call into CNI plugins iiuc.

> Now Kata Containers [run hooks in the network namespace](https://github.com/kata-containers/kata-containers/blob/07b9d93f5f94673921dc833167710178a42c0d71/src/runtime/pkg/katautils/create.go#L158-L161)(I think this should be reviewed) that Kata Containers created:
> 
> ```go
> // Run pre-start OCI hooks.
> err = EnterNetNS(sandboxConfig.NetworkConfig.NetworkID, func() error {
> 	return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
> })
> ```

According to the runtime spec, the `createRuntime` hook must be run in the *runtime* namespace, i.e. the host namespace. So I think we should run the nerdctl as a `createRuntime` hook as it's currently implemented, and from Kata we should not run it in the sandbox netns. Our `preStart` support is also wrong, it should not run in the sandbox netns. 

> For nerdctl, the hook runs in the created namespace, so the CNI plugins also create `cni0`, `eth0` interfaces in that namespace, which leads Kata Containers to report an error.

If we run the `createRuntime` in the host ns, and pass the created netns through the `linux.namespace` entry, wouldn't that work?

AkihiroSuda:
> I think nerdctl uses a Docker-style network process mechanism, that uses an OCI hook to create the network.

Yes. We rely on OCI hook so that we can support `nerdctl run --restart=always`.

Issue #777, opened 2022-02-02
Image size for multiple platform images are not shown correctly


amioranza:
nerdctl version:
`nerdctl version 0.15.0`

I'm using Rancher Desktop 1.0.0 on macOS Monterrey (12.1), M1 machine.

When building multiple platform images only the first image shows the correct size of it:
```
nerdctl build --platform=arm64,amd64 -t fluentbit . --progress plain
nerdctl image ls
REPOSITORY    TAG       IMAGE ID        CREATED          PLATFORM       SIZE
fluentbit     latest    724c7e5d5096    5 seconds ago    linux/arm64    397.2 MiB
fluentbit     latest    724c7e5d5096    5 seconds ago    linux/amd64    0.0 B
```

The build works and the image too, just the size report is not working.

```
#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 249B done
#1 DONE 0.0s

#2 [internal] load .dockerignore
#2 transferring context: 2B done
#2 DONE 0.0s

#4 [linux/arm64 internal] load metadata for public.ecr.aws/aws-observability/aws-for-fluent-bit:latest
#4 DONE 0.2s

#3 [linux/amd64 internal] load metadata for public.ecr.aws/aws-observability/aws-for-fluent-bit:latest
#3 DONE 0.2s

#6 [internal] load build context
#6 transferring context: 38B done
#6 DONE 0.0s

#5 [linux/arm64 1/2] FROM public.ecr.aws/aws-observability/aws-for-fluent-bit:latest@sha256:7e4b9ae7d3154d70d20d67c3b75251920d296b7ba8481f029684b44dd14e7c49
#5 resolve public.ecr.aws/aws-observability/aws-for-fluent-bit:latest@sha256:7e4b9ae7d3154d70d20d67c3b75251920d296b7ba8481f029684b44dd14e7c49 0.0s done
#5 CACHED

#9 [linux/amd64 1/2] FROM public.ecr.aws/aws-observability/aws-for-fluent-bit:latest@sha256:7e4b9ae7d3154d70d20d67c3b75251920d296b7ba8481f029684b44dd14e7c49
#9 resolve public.ecr.aws/aws-observability/aws-for-fluent-bit:latest@sha256:7e4b9ae7d3154d70d20d67c3b75251920d296b7ba8481f029684b44dd14e7c49 0.0s done
#9 CACHED

#8 [linux/amd64 2/2] ADD custom-logs.conf /custom-logs.conf
[...]

PR #789, opened 2022-02-07
fixe image size

fahedouch:
fixing https://github.com/containerd/nerdctl/issues/777
`ImageSize` is the size of the unpacked snapshots. for multi-platform image, snapshot host single platform layer(s). So the computing of image size shouldn't be based on image platform

### DIFF
--- a/cmd/nerdctl/images.go
+++ b/cmd/nerdctl/images.go
@@ -109,7 +109,8 @@ type imagePrintable struct {
 	ID           string // image target digest (not config digest, unlike Docker), or its short form
 	Repository   string
 	Tag          string // "<none>" or tag
-	Size         string
+	Size         string // the size of the unpacked snapshots.
+	BlobSize     string // the size of the blobs in the content store
 	// TODO: "SharedSize", "UniqueSize", "VirtualSize"
 	Platform string // nerdctl extension
 }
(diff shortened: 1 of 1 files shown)


AkihiroSuda (changes requested):
This only shows the size of the default platform

cmd/nerdctl/images.go
```diff
@@ -314,8 +314,8 @@ func (key snapshotKey) add(ctx context.Context, s snapshots.Snapshotter, usage *
 
 // unpackedImageSize is the size of the unpacked snapshots.
 // Does not contain the size of the blobs in the content store. (Corresponds to Docker).
```
AkihiroSuda, line 316: Maybe we should show both the blob size and the unpacked size to avoid confusion for multi-platform images

fahedouch:
related issue https://github.com/containerd/nerdctl/issues/796

cmd/nerdctl/images.go
fahedouch: fixed

cmd/nerdctl/images.go
```diff
@@ -109,7 +109,8 @@ type imagePrintable struct {
 	ID           string // image target digest (not config digest, unlike Docker), or its short form
 	Repository   string
 	Tag          string // "<none>" or tag
-	Size         string
+	UnpackedSize string // the size of the unpacked snapshots.
```
AkihiroSuda, line 112: the `Size` field needs to be retained for Docker compatibility

cmd/nerdctl/images.go
```diff
@@ -142,9 +143,9 @@ func printImages(ctx context.Context, cmd *cobra.Command, client *containerd.Cli
 		w = tabwriter.NewWriter(w, 4, 8, 4, ' ', 0)
 		if !quiet {
 			if digestsFlag {
-				fmt.Fprintln(w, "REPOSITORY\tTAG\tDIGEST\tIMAGE ID\tCREATED\tPLATFORM\tSIZE")
+				fmt.Fprintln(w, "REPOSITORY\tTAG\tDIGEST\tIMAGE ID\tCREATED\tPLATFORM\tSIZE\tBLOB_SIZE")
 			} else {
-				fmt.Fprintln(w, "REPOSITORY\tTAG\tIMAGE ID\tCREATED\tPLATFORM\tSIZE")
+				fmt.Fprintln(w, "REPOSITORY\tTAG\tIMAGE ID\tCREATED\tPLATFORM\tSIZE\tBLOB_SIZE")
```
AkihiroSuda, line 148: Do we use underscore in other places?

AkihiroSuda (approved):
Thanks

AkihiroSuda:
`nerdctl images` might be slower than before (which has been already slow).

We should do some benchmark and probably have `--quick` (`--fast`?, idk) flag to skip slow size calculation

fahedouch:
@AkihiroSuda I think we can merge this, I will open a new PR!

Issue #775, opened 2022-02-02
[Linux] --insecure-registry option not works for some circumstances 


Zheaoli:
For example, I have a docker registry with the hostname `registry.manjusaka.me`. I just open 80 port to support normal HTTP, not HTTPS. And now if I use `nerdctl --insecure-registry=true push registry.manjusaka.me/demo:v0.1` to push the image. There will be an error here bellow

```text
failed to do request: Head "xxxx": dial tcp x.x.x:443: connect: connection refused
```

I think the root cause is here

https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/push.go#L186

The `IsErrHTTPResponseToHTTPSClient` will check the error message is `server gave HTTP response to HTTPS client` or not.

In the real environment, people may close the 443 port and only use the 80 port. I think we need compact with this situation.

Ghostwritten:
me,too,
```
[root@bastion01 ~]# nerdctl --insecure-registry=true  push registry.demo:35000/envoyproxy/envoy:v1.22.5 
INFO[0000] pushing as a reduced-platform image (application/vnd.docker.distribution.manifest.list.v2+json, sha256:8fba6a3d40f33a7ee436bf02e1c5635211821efaed6644b3542b8b0cf82a2789) 
WARN[0000] skipping verifying HTTPS certs for "registry.demo:35000" 
index-sha256:8fba6a3d40f33a7ee436bf02e1c5635211821efaed6644b3542b8b0cf82a2789:    waiting        |--------------------------------------| 
manifest-sha256:727d10228de2f9895f9dddb7c11883672f462db3744c56fd5c8d83e4afa81f55: waiting        |--------------------------------------| 
config-sha256:e9c4ee2ce7207ce0f446892dda8f1bcc16cd6aec0c7c55d04bddca52f8af280d:   waiting        |--------------------------------------| 
elapsed: 0.3 s                                                                    total:   0.0 B (0.0 B/s)                                         
FATA[0000] failed to do request: Head "https://registry.demo:35000/v2/envoyproxy/envoy/blobs/sha256:7dcdb29d804aa1a6a09607656fc277b92ac7ba07df814cb2e340e5e45141a406": EOF 
[root@bastion01 ~]# nerdctl -v
nerdctl version 1.5.0

```

fahedouch:
> me,too,
> 
> ```
[...]

PR #790, opened 2022-02-07
Fix --insecure-registry not working in some circumstances for #775

Zheaoli:
Signed off by: Zheao Li <me@manjusaka.me>

#775

### DIFF
--- a/pkg/imgutil/imgutil.go
+++ b/pkg/imgutil/imgutil.go
@@ -144,7 +144,8 @@ func EnsureImage(ctx context.Context, client *containerd.Client, stdout, stderr 
 
 	img, err := PullImage(ctx, client, stdout, stderr, snapshotter, resolver, ref, ocispecPlatforms, unpack, quiet)
 	if err != nil {
-		if !IsErrHTTPResponseToHTTPSClient(err) {
+		// In some circumstance (e.g. people just use 80 port to support pure http), the error will contain message like "dial tcp <port>: connection refused".
+		if !IsErrHTTPResponseToHTTPSClient(err) && !IsErrConnectionRefused(err) {
 			return nil, err
 		}
 		if insecure {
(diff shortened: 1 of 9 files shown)


cmd/nerdctl/login.go
```diff
@@ -287,7 +289,14 @@ func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost) error {
 		}
 		res, err := ctxhttp.Do(ctx, rh.Client, req)
 		if err != nil {
-			return fmt.Errorf("failed to call rh.Client.Do: %w", err)
+			// In some circumstance (e.g. people just use 80 port to support pure http), the error will contain message like "dial tcp <port>: connection refused"
+			if (imgutil.IsErrHTTPResponseToHTTPSClient(err) || imgutil.IsErrConnectionRefused(err)) && insecure {
```
Junnplus, line 293: If the registry service does not support HTTPS and the user does not specify `--insecure-registry`, we should alert them, like https://github.com/containerd/nerdctl/blob/v0.16.1/cmd/nerdctl/push.go#L197-L201

cmd/nerdctl/login.go
```diff
@@ -256,7 +256,7 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
 	return "", err
 }
 
-func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost) error {
+func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost, insecure bool) error {
```
Junnplus, line 259: `insecure` doesn't seem to be used?

Junnplus (approved):
I don't know why the logic related to login was removed, other LGTM.

Zheaoli:
> I don't know why the logic related to login was removed, other LGTM.

Because I find the TestPushInsecureWithLogin will fail sometimes, not every time if I add plain HTTP logic for login action. Such as https://github.com/containerd/nerdctl/runs/5122259173?check_suite_focus=true. So I choose to just support plain HTTP for pull and push action, and I will make another PR for login action if I have figure out what's happen

pkg/imgutil/imgutil.go
```diff
@@ -200,6 +201,13 @@ func IsErrHTTPResponseToHTTPSClient(err error) bool {
 	return strings.Contains(err.Error(), unexposed)
 }
 
+// IsErrConnectionRefused return whether err is
+// "connect: connection refused"
+func IsErrConnectionRefused(err error) bool {
+	const errMessage = "connect: connection refused"
+	return strings.Contains(err.Error(), errMessage)
```
AkihiroSuda, line 208: Can't we use `errors.Is`?

AkihiroSuda (approved):
Thanks

pkg/imgutil/imgutil.go
Zheaoli: I think I can make a single PR to make all error message checks like `errors.Is`

PR #791, opened 2022-02-07
Limit the CI concurrency to allow one running CI per PR

Zheaoli:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,10 @@ on:
       - 'release/**'
   pull_request:
 
+concurrency:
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Could you explain more?

Zheaoli:
CleanShot 2022-02-07 at 22 55 13@2x

Please take a look at it here. If people get N'times committed in a single PR. The whole CI process will be triggered N'times. So in my PR, the config limit a single running process for one PR(That means the GitHub workflow will cancel the previous CI process in the same PR automatically)

ktock:
> limit a single running process for one PR

Do we really need this? Are there any cases where this behaviour is problematic as of now? (e.g. are there something like rate limitations?)

Issue #518, opened 2021-11-16
Add Windows hostprocess support


jsturtevant:
Add the correct specs fields to https://github.com/containerd/nerdctl/blob/f316402913c4239a2388b7467bcffc9735dd8327/cmd/nerdctl/run_windows.go#L42

Something like:

```
hpc, err := cmd.Flags().GetBool("host-process")
if err != nil {
  return nil, err
}
if hpc {
  //TODO append the right hpc opts
}
```

See https://github.com/containerd/containerd/pull/5131 for details on how to set the right spec fields.

dcantah:
Planning to look at this next week

dcantah:
@AkihiroSuda What's the story for how certain things should be handled if they require a specific Containerd version? hostprocess (Windows' version of a privileged container) will be a feature for Containerd 1.6 and currently only works in the beta releases until a full release.

AkihiroSuda:
> What's the story for how certain things should be handled if they require a specific Containerd version?

Just mentioning that in `docs/windows.md` would be fine

dcantah:
Ok awesome, thanks!

PR #795, opened 2022-02-08
Add Windows Hostprocess flag

jsturtevant:
Fixes: #518

Still needs:

- [x]  docs 
- [x]  a bit more thought on how to specific user
- [x] test

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,10 @@ Init process flags:
   Please make sure the binary exists in your `PATH`.
   - Default: `tini`
 
+Isolation flags:
+- :whale: :blue_square: :nerd_face: `--isolation=(default|process|host)`: Used on Windows to change process isolation level. `default` will use the runtime options configured in `default_runtime` in the [containerd configuration](https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cri-plugin-config-guide) which is `process` in containerd by default. `process` runs process isolated containers.  `host` runs [Host Process containers](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/).  Host process containers inherit permissions from containerd process unless `--user` is specified then will start with user specified and the user specified must be present on the host.  `host` requires Containerd 1.7+. Not implemented for Linux.
+
+
 Network flags:
 - :whale: `--net, --network=(bridge|host|none|container:<container>|<CNI>)`: Connect a container to a network.
   - Default: "bridge"
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -276,6 +276,15 @@ func setCreateFlags(cmd *cobra.Command) {
 	// #endregion
 
 	cmd.Flags().String("ipfs-address", "", "multiaddr of IPFS API (default uses $IPFS_PATH env variable if defined or local directory ~/.ipfs)")
+
+	cmd.Flags().String("isolation", "default", "Specify isolation technology for container. On Linux the only valid value is default. Windows options are host or process with process isolation as the default")
+	cmd.RegisterFlagCompletionFunc("isolation", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if runtime.GOOS == "windows" {
+			return []string{"default", "host", "process"}, cobra.ShellCompDirectiveNoFileComp
+		}
+		return []string{"default"}, cobra.ShellCompDirectiveNoFileComp
+	})
+
 }
 
 // runAction is heavily based on ctr implementation:
--- a/cmd/nerdctl/run_windows.go
+++ b/cmd/nerdctl/run_windows.go
@@ -19,19 +19,23 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/containerd/containerd/containers"
+	"github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/nerdctl/pkg/api/types"
 	"github.com/docker/go-units"
 	"github.com/spf13/cobra"
 )
 
-func WithoutRunMount() func(ctx context.Context, client oci.Client, c *containers.Container, s *oci.Spec) error {
-	// not valid on windows
-	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *oci.Spec) error { return nil }
-}
+const (
+	// HostProcessInheritUser inherits permissions of containerd process
+	hostProcessInheritUser = "microsoft.com/hostprocess-inherit-user"
+
+	// HostProcessContainer will launch a host process container
+	hostProcessContainer = "microsoft.com/hostprocess-container"
+)
 
 func capShellComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 	candidates := []string{}
@@ -71,9 +75,54 @@ func setPlatformOptions(
 		opts = append(opts, oci.WithMemoryLimit(uint64(mem64)))
 	}
 
+	// TODO implement hyper-v isolation
+	isolation, err := cmd.Flags().GetString("isolation")
+	if err != nil {
+		return nil, err
+	}
+	switch isolation {
+	case "host":
+		hpAnnotations := map[string]string{
+			hostProcessContainer: "true",
+		}
+
+		// If user is set we will attempt to start container with that user (must be present on the host)
+		// Otherwise we will inherit permissions from the user that the containerd process is running as
+		user, err := cmd.Flags().GetString("user")
+		if err != nil {
+			return nil, err
+		}
+		if user == "" {
+			hpAnnotations[hostProcessInheritUser] = "true"
+		}
+
+		opts = append(opts, oci.WithAnnotations(hpAnnotations))
+	case "process":
+		// override the default isolation mode in the case where
+		// the containerd default_runtime is set to hyper-v
+		opts = append(opts, WithWindowsProcessIsolated())
+	case "default":
+		// no op
+		// use containerd's default runtime option `default_runtime` set in the config.toml
+	default:
+		return nil, fmt.Errorf("unknown isolation value %q. valid values are 'host', 'process' or 'default'", isolation)
+	}
+
 	opts = append(opts,
 		oci.WithWindowNetworksAllowUnqualifiedDNSQuery(),
 		oci.WithWindowsIgnoreFlushesDuringBoot())
 
 	return opts, nil
+}
+
+func WithWindowsProcessIsolated() oci.SpecOpts {
+	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *specs.Spec) error {
+		if s.Windows == nil {
+			s.Windows = &specs.Windows{}
+		}
+		if s.Windows.HyperV != nil {
+			s.Windows.HyperV = nil
+		}
+		return nil
+	}
 }
--- a/cmd/nerdctl/run_windows_test.go
+++ b/cmd/nerdctl/run_windows_test.go
@@ -1,0 +1,67 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestRunHostProcessContainer(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	hostname, err := exec.Command("hostname").Output()
+	if err != nil {
+		t.Fatalf("unable to get hostname: %s", err)
+	}
+
+	base.Cmd("run", "--rm", "--isolation=host", testutil.WindowsNano, "hostname").AssertOutContains(string(hostname))
+	output := base.Cmd("run", "--rm", "--isolation=host", testutil.WindowsNano, "whoami").Out()
+	t.Logf("whoami %s", output)
+}
+
+func TestRunHostProcessContainerAsUser(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+
+	hostuser := "nt authority\\system"
+	base.Cmd("run", "--rm", "--isolation=host", "-u", "NT AUTHORITY\\SYSTEM", testutil.WindowsNano, "whoami").AssertOutContains(hostuser)
+}
+
+func TestRunHostProcessContainerAsService(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	hostuser := "nt authority\\local service"
+	base.Cmd("run", "--rm", "--isolation=host", "-u", "NT AUTHORITY\\Local Service", testutil.WindowsNano, "whoami").AssertOutContains(hostuser)
+}
+
+func TestRunHostProcessContainerAslocalService(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	hostuser := "nt authority\\network service"
+	base.Cmd("run", "--rm", "--isolation=host", "-u", "NT AUTHORITY\\Network Service", testutil.WindowsNano, "whoami").AssertOutContains(hostuser)
+}
+
+func TestRunProcessIsolated(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+
+	containerUser := "ContainerUser"
+	base.Cmd("run", "--rm", "--isolation=process", "-u", containerUser, testutil.WindowsNano, "whoami").AssertOutContains(containerUser)
+}
--- a/hack/configure-windows-ci.ps1
+++ b/hack/configure-windows-ci.ps1
@@ -5,8 +5,9 @@ choco install --limitoutput --no-progress -y git golang
 
 #install containerd
 $version=$env:ctrdVersion
+echo "Installing containerd $version"
 curl.exe -L https://github.com/containerd/containerd/releases/download/v$version/containerd-$version-windows-amd64.tar.gz -o containerd-windows-amd64.tar.gz
-tar xvf containerd-windows-amd64.tar.gz
+tar.exe xvf containerd-windows-amd64.tar.gz
 mkdir -force "$Env:ProgramFiles\containerd"
 mv ./bin/* "$Env:ProgramFiles\containerd"
 


cmd/nerdctl/run_windows.go
```diff
@@ -61,6 +64,17 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
...
 	}
 
+	hostProcess, err:= cmd.Flags().GetBool("host-process")
+	if err != nil {
+		return nil, err
+	}
+	if hostProcess {
+		opts = append(opts,
+				customopts.WithAnnotation(annotations.WindowsHostProcess, strconv.FormatBool(true)),
+				customopts.WithAnnotation("microsoft.com/hostprocess-inherit-user", strconv.FormatBool(true)),
```
dcantah, line 74: Does nerdctl not have a user flag?

cmd/nerdctl/run_windows.go
```diff
@@ -61,6 +64,17 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
 		opts = append(opts, oci.WithMemoryLimit(uint64(mem64)))
 	}
 
+	hostProcess, err:= cmd.Flags().GetBool("host-process")
+	if err != nil {
+		return nil, err
+	}
+	if hostProcess {
+		opts = append(opts,
+				customopts.WithAnnotation(annotations.WindowsHostProcess, strconv.FormatBool(true)),
```
dcantah, line 73: Oh duh, I was thinking we may need to get out a hcsshim release with this annotation exported (we recently made a package that exports all of these https://github.com/microsoft/hcsshim/tree/master/pkg/annotations but no tag available with this). Forgot it exists in cri

cmd/nerdctl/run_windows.go
dcantah: Well, the preferred behavior would probably be to inherit, and if a user was passed we skip supplying this annotation.

README.md
```diff
@@ -370,6 +370,7 @@ Security flags:
 - :whale: `--cap-add=<CAP>`: Add Linux capabilities
 - :whale: `--cap-drop=<CAP>`: Drop Linux capabilities
 - :whale: `--privileged`: Give extended privileges to this container
+- :blue_square: :nerd_face: `--host-process`: Allow running Host Process containers on Windows. Inherits permissions from containerd process unless `--user` is specified then will start with user specified.  The user specified must be present on the host.  Requires Containerd 1.6+.
```
AkihiroSuda, line 373: What about calling this `--isolation=host`

cmd/nerdctl/run.go
```diff
@@ -171,6 +171,7 @@ func setCreateFlags(cmd *cobra.Command) {
 	cmd.Flags().StringSlice("cap-drop", []string{}, "Drop Linux capabilities")
 	cmd.RegisterFlagCompletionFunc("cap-drop", capShellComplete)
 	cmd.Flags().Bool("privileged", false, "Give extended privileges to this container")
+	cmd.Flags().Bool("host-process", false, "Allow host access for Windows Containers")
```
AkihiroSuda, line 174: Should be added only when GOOS==windows

cmd/nerdctl/run_windows.go
```diff
@@ -19,9 +19,12 @@ package main
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/oci"
+	"github.com/containerd/containerd/pkg/cri/annotations"
```
AkihiroSuda, line 26: We can just redefine the constants without adding extra dependency on cri

go.mod
```diff
@@ -160,6 +161,7 @@ require (
 	google.golang.org/grpc v1.44.0 // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/square/go-jose.v2 v2.5.1 // indirect
+	k8s.io/cri-api v0.24.0-alpha.1 // indirect
```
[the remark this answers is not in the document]
dcantah: It's being pulled in because we're using 'github.com/containerd/containerd/pkg/cri/' for the annotations I take it... I'd prefer to just redefine the constants if anything like you said. @AkihiroSuda Would you be opposed to adding in hcsshim as a dependency here (not in this PR but in the future)? There's some other annotations that would be useful for hyper-v, but if not we can redefine all the same.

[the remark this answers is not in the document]
AkihiroSuda: Hcsshim dependency is probably fine

cmd/nerdctl/run_windows.go
jsturtevant: Yes, definitely makes sense to do that along with maybe updating to using hcsshim once a released package has these constants exported since we already rely on it

README.md
jsturtevant: this has grown on me overnight. I do have a concern on naming...  Right now we call this `hostprocess` containers in kubernetes and `job` containers in hcsshim and we would be adding another slightly different way of referencing this concept.  I think it is a minor difference but thought to bring it up. thoughts?

cmd/nerdctl/run.go
jsturtevant: since docker does have `--isolation` on Linux should we keep it? I left it for now.

README.md
dcantah: The 'job containers' name was supposed to be just for the devs on hcsshim as it was what we got accustomed to when referring to it internally, although it does kind of suck that the name leaks out in some error strings. That should maybe get addressed/cleaned up. If that was touched up would that alleviate some of the concern on the three way split @jsturtevant?

dcantah: Feel like the UX is a bit improved if all the Windows folks need to remember is --isolation and its three options vs. --isolation and then an option off to the side that also dramatically reduces the amount of isolation you have. Also might simplify the case where both are specified (--isolation and --host-process) as ideally we should error out if they are as they're conflicting options.

README.md
```diff
@@ -333,6 +333,10 @@ Basic flags:
 Platform flags:
 - :whale: `--platform=(amd64|arm64|...)`: Set platform
 
+Isolation flags:
+- :whale: :blue_square: :nerd_face: `--isolation=(default|host)`: Used on Windows to change process isolation level. Valid values are `default` and `host`. `default` runs process isolated containers.  `host` runs Host Process containers.  Host process containers inherit permissions from containerd process unless `--user` is specified then will start with user specified.  The user specified must be present on the host.  `host` requires Containerd 1.6+. Not implemented for Linux.
+
```
dcantah, line 338: Wasn't the Docker option for process isolated "process"? I think for Docker Desktop it defaulted to hyper-v if you didn't supply anything, and on server it defaulted to process. Regardless of that we should probably stick to the same naming as docker (unless I missed the discussion for naming this default 😆)

README.md
jsturtevant: It's a minor concern. Just wanted to call it out. Anytime we refer to the same thing multiple ways it adds complexity to understanding. 

That said thinking about hostprocess as an `--isolation` resonates with me. and keeps it simple from the cli UX perspective.  

This brings up one more thought, As of right now hostprocess doesn't not support hyper-v containers. Is it possible in the future it could? Does that even make sense?

README.md
jsturtevant: yes, I should probably have `process` as an explicit support flag but since hyper-v isn't really support fully in containerd we can default that.

The `default` flag uses whatever was set on the docker daemon via `--exec-opt isolation=hyperv` option: https://docs.docker.com/engine/reference/commandline/run/#specify-isolation-technology-for-container---isolation

I don't think containerd has this setting but the equivalent would be `default_runtime_name`?

README.md
dcantah: That's a good question.. I could imagine a world where we could get it to work, but I'm not sure what use it would bring off the top of my head. I don't know if its worth putting too much brain power into thinking of at the moment though

AkihiroSuda:
needs rebase

README.md
jsturtevant: Just want to make sure we don't back our selves into a naming thing.  *If* it was a thing, would naming for isolation be something like `hyperv-host` :thinking:

jsturtevant:
Added a logging statement for the who the hostprocess container to see what it thinks it is running as and it is coming back as     

`run_windows_test.go:35: whoami nt authority\system`

@dcantah I would think that since it is system it should be able to start `Local Service`?

dcantah:
> Added a logging statement for the who the hostprocess container to see what it thinks it is running as and it is coming back as
> 
> `run_windows_test.go:35: whoami nt authority\system`
> 
> @dcantah I would think that since it is system it should be able to start `Local Service`?

Yea this is really odd.. The only thing I can think of is there's something blocking logging in users in this environment. The cri-integration tests in Containerd have tests that login as different service accounts https://github.com/containerd/containerd/blob/5162238c7da04d6bdd771d0aa43f8e34c91ea2ca/integration/windows_hostprocess_test.go#L38

README.md
dcantah: If we're able to specify --isolation twice (e.g. `--isolation host --isolation hyper-v`) would that suffice? We'd just need to short circuit on if someone specified process and a second option as that doesn't make sense. Or your idea of a fourth hyperv-host option.

AkihiroSuda:
What's current status?

jsturtevant:
I refreshed the PR. It is still failing when running in CI though it works locally, I thought it might be a Cirrus vs github actions thing but doesn't look like it.  

I will need to dig deeper, as it is working locally when I test it.

dcantah:
Nice. We didn't allow a blank username for CRI either, right? That's probably the right way.. defaulting to SYSTEM is kinda scary

dcantah:
Hmm.. if the username string is empty we should error a bit earlier https://github.com/microsoft/hcsshim/blob/6547959343d65e0cf431dd9d589f45ee2b98172c/internal/jobcontainers/logon.go#L109. Maybe it's filling in container-user/container-admin from the image? 🤷‍♂️

jsturtevant:
That wasn't the case, It is being passed by the main nerdctl options.

I compared the oci spec to one I have working via the cri server and nothing jumps out.  Here is the oci spec that is being generated with nerdctl:

```
{
    "ociVersion": "1.0.2-dev",
    "process": {
        "user": {
            "uid": 0,
            "gid": 0,
            "username": "nt authority\\local system"
        },
        "args": [
            "cmd.exe",
            "/s",
            "/c",
            "cmd"
        ],
        "env": [
            "PATH=C:\\dig\\;C:\\bin;C:\\curl;C:\\Windows\\System32;C:\\Windows;C:\\Program Files\\PowerShell;",
            "ProgramFiles=C:\\Program Files",
            "LOCALAPPDATA=C:\\Users\\ContainerAdministrator\\AppData\\Local",
            "PSModuleAnalysisCachePath=C:\\Users\\ContainerAdministrator\\AppData\\Local\\Microsoft\\Windows\\PowerShell\\docker\\ModuleAnalysisCache",
            "PSCORE=C:\\Program Files\\PowerShell\\pwsh.exe"
        ],
        "cwd": ""
    },
    "root": {
        "path": ""
    },
    "hostname": "1624c258dfdc",
    "hooks": {
        "createRuntime": [
            .... cut this out, hooks are supported in windows so shouldn't have any affect
        ]
    },
    "annotations": {
        "io.containerd.image.config.stop-signal": "SIGTERM",
        "microsoft.com/hostprocess-container": "true",
        "nerdctl/extraHosts": "null",
        "nerdctl/hostname": "1624c258dfdc",
        "nerdctl/name": "busybox-1624c",
        "nerdctl/namespace": "default",
        "nerdctl/networks": "[\"nat\"]",
        "nerdctl/platform": "windows/amd64",
        "nerdctl/state-dir": "C:\\ProgramData\\nerdctl\\052055e3\\containers\\default\\1624c258dfdc478e27fdd3f4cb6a0874486f7e1abd2cef426aaeec2c97b908c2"
    },
    "windows": {
        "layerFolders": null,
        "ignoreFlushesDuringBoot": true,
        "network": {
            "allowUnqualifiedDNSQuery": true
        }
    }
}
```

and the spec generated with the cri call:

```
{
    "ociVersion": "1.0.2-dev",
    "process": {
        "user": {
            "uid": 0,
            "gid": 0,
            "username": "NT AUTHORITY\\SYSTEM"
        },
        "args": [
            "ping",
            "-t",
            "localhost"
        ],
        "env": [
            "PATH=C:\\Windows\\system32;C:\\Windows;"
        ],
        "cwd": "C:\\"
    },
    "annotations": {
        "io.kubernetes.cri.container-name": "busybox",
        "io.kubernetes.cri.container-type": "container",
        "io.kubernetes.cri.image-name": "k8s.gcr.io/pause@sha256:85cfebc79dccc7a0e56680778a614b29a0b1c2ae98d4b1efc746764c04d5656c",
        "io.kubernetes.cri.sandbox-id": "0164ef1c9d9a72e904fe45f12e5698e06338ec913798798f2030773ffa0b1b09",
        "io.kubernetes.cri.sandbox-name": "sandbox-hp",
        "io.kubernetes.cri.sandbox-namespace": "default",
        "io.kubernetes.cri.sandbox-uid": "alskasdfasdf",
        "microsoft.com/hostprocess-container": "true"
    },
    "windows": {
        "layerFolders": null,
        "network": {}
    }
}
```

dcantah:
Can we make a dummy test that prints out the current user that the CI is running as? It feels like a permission issue

jsturtevant:
Already did https://github.com/containerd/nerdctl/pull/795/commits/f3af71a5937eb9cadf36799538fc9a9318955bb7 but thought it doesn't seem to have output what I thought it should. Let me try something else.

configuration complete! Printing configuration...
Service:

Status      : Running
Name        : containerd
DisplayName : containerd


Id      : 6212
Handles : 148
CPU     : 0.140625
SI      : 0
Name    : containerd

jsturtevant:
btw, the test that uses hostprocess containers without the user specified passes in CI

jsturtevant:
I am also able to reproduce this locally

jsturtevant:
I added a bunch of tests with different inputs, some of them passed. It appears to be the way the paramaters are passed to the binary in the test framework. I will clean it up and I think we will have a few tests passing 🤞

jsturtevant:
@AkihiroSuda this is finally ready for review.  I did move the Windows CI from Cirrus to Github actions during my debugging, I can either drop that commit or move it to a separate PR but I think I prefer GitHub UI for debugging and it aligns better with rest of the tests.

dcantah:
If Cirrus supports nested virt it might be worth staying on it, otherwise we won't be able to test hyper-v isolation on GHA (unless they made progress on offering this..)

cmd/nerdctl/run.go
```diff
@@ -274,6 +274,8 @@ func setCreateFlags(cmd *cobra.Command) {
 	})
 	cmd.Flags().String("cosign-key", "", "Path to the public key file, KMS, URI or Kubernetes Secret for --verify=cosign")
 	// #endregion
+
+	cmd.Flags().String("isolation", "default", "Specify isolation technology for container. On Linux valid value is default. On Windows can also be host or process")
```
AkihiroSuda, line 278: Can we register a shell completion?

cmd/nerdctl/run_windows.go
```diff
@@ -71,9 +80,52 @@ func setPlatformOptions(
...
+
+		opts = append(opts, oci.WithAnnotations(hpAnnotations))
+	case "default":
+		// no op
+		// use containerd's default runtime options
+	case "process":
+		opts = append(opts, WithWindowsProcessIsolated())
+	default:
+		// no op
+		// use containerd's default
```
AkihiroSuda, line 111: Shouldn't we raise an error for an unknown value?

README.md
```diff
@@ -395,6 +395,10 @@ Init process flags:
   Please make sure the binary exists in your `PATH`.
   - Default: `tini`
 
+Isolation flags:
+- :whale: :blue_square: :nerd_face: `--isolation=(default|process|host)`: Used on Windows to change process isolation level. `default` will use the runtime options configured in `default_runtime` in the [containerd configuration](https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cri-plugin-config-guide) which is `process` in containerd by default. `process` runs process isolated containers.  `host` runs Host Process containers.  Host process containers inherit permissions from containerd process unless `--user` is specified then will start with user specified and the user specified must be present on the host.  `host` requires Containerd 1.6+. Not implemented for Linux.
+
```
AkihiroSuda, line 400: No support for the Hyper-V mode?

README.md
```diff
@@ -395,6 +395,10 @@ Init process flags:
   Please make sure the binary exists in your `PATH`.
   - Default: `tini`
 
+Isolation flags:
+- :whale: :blue_square: :nerd_face: `--isolation=(default|process|host)`: Used on Windows to change process isolation level. `default` will use the runtime options configured in `default_runtime` in the [containerd configuration](https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cri-plugin-config-guide) which is `process` in containerd by default. `process` runs process isolated containers.  `host` runs Host Process containers.  Host process containers inherit permissions from containerd process unless `--user` is specified then will start with user specified and the user specified must be present on the host.  `host` requires Containerd 1.6+. Not implemented for Linux.
```
marosset, line 399: Can/should we link to https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/ for more info on HostProcess containres?

marosset, line 399: Let's update this to say we require containerd 1.7 so we get the better volume mount behavior.

cmd/nerdctl/run.go
marosset, line 278: Should we add that on Windows this defaults to process isolation?

cmd/nerdctl/run_windows.go
```diff
@@ -71,9 +80,52 @@ func setPlatformOptions(
...
+		user, err := cmd.Flags().GetString("user")
+		if err != nil {
+			return nil, err
+		}
+		if user == "" {
+			hpAnnotations[hostProcessInheritUser] = "true"
+		}
+
+		opts = append(opts, oci.WithAnnotations(hpAnnotations))
+	case "default":
```
marosset, line 104: nit: should we just move the "default" case right above `default:` and use [fallthrough](https://stackoverflow.com/questions/40821855/do-go-switch-cases-fallthrough-or-not)

marosset:
I had a couple of suggestions for the usage but overall LGTM
Thanks @jsturtevant

README.md
dcantah: I think this should be a separate change, reason being is the way Windows launches the VM in the containerd-shim is a bit different than how it was done for docker (in terms of the resources assigned to the VM). It'd be a bit more than just filling in the hyper-v field on the runtime spec I imagine

cmd/nerdctl/run_windows.go
```diff
@@ -71,9 +80,52 @@ func setPlatformOptions(
...
+		// no op
+		// use containerd's default runtime options
+	case "process":
+		opts = append(opts, WithWindowsProcessIsolated())
+	default:
+		// no op
+		// use containerd's default
+	}
```
dcantah, line 112: I think we should combine default and process cases and get rid of the WithWindowsProcessIsolated option as the spec is only going to have that filled in if we set it anyways which would seem like a bug. 

```suggestion
	switch isolation {
	case "host":
		hpAnnotations := map[string]string{
		        hostProcessContainer: "true",
		}

		// If user is set we will attempt to start container with that user (must be present on the host)
		// Otherwise we will inherit permissions from the user that the containerd process is running as
		user, err := cmd.Flags().GetString("user")
		if err != nil {
			return nil, err
		}
		if user == "" {
			hpAnnotations[hostProcessInheritUser] = "true"
		}

		opts = append(opts, oci.WithAnnotations(hpAnnotations))
	case "default", "process":
		// no op
		// use containerd's default runtime options
	default:
		return nil, fmt.Errorf("unknown isolation value %q. valid values are 'host', 'process' or 'default'", isolation)
	}
```

dcantah: Also, on docker this flag was defined globally I believe and would error on Linux for an invalid value, should we do this parsing in run.go instead?

cmd/nerdctl/run_windows.go
jsturtevant: updated

README.md
jsturtevant: I was also thinking this would be a separate changeset as the hyper-v support in containerd was only recently added.

README.md
jsturtevant: It is the most detailed documentation we have today.  I think it makes sense

cmd/nerdctl/run_windows.go
jsturtevant: Ok, I remember why I did it this way now.

If the `default_runtime` is set in the `config.toml` to be hyper-v don't we need to override the default runtime settings?  aka, blank out hyper-v part of the spec?  

Maybe I don't quite understand it properly...

dcantah: Not much we could do it seems as hyperv is set in the shim itself if that's the case https://github.com/microsoft/hcsshim/blob/b295b1a866b1263ddede4f00dddf7a129b96c114/cmd/containerd-shim-runhcs-v1/service_internal.go#L109-L116

jsturtevant: https://github.com/containerd/nerdctl/pull/795#discussion_r810165333

cmd/nerdctl/run.go
jsturtevant: updated

README.md
jsturtevant: done

jsturtevant: done

cmd/nerdctl/run_windows_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func TestRunAsHostProcessContainerAsUser(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+
+	hostuser := "nt authority\\system"
+	base.Cmd("run", "--rm", "--isolation=host", "-u", "NT AUTHORITY\\SYSTEM", testutil.CommonImage, "whoami").AssertOutContains(hostuser)
+}
+
+func TestRunAsHostProcessContainerAsservice(t *testing.T) {
```
dcantah, line 47:
```suggestion
func TestRunAsHostProcessContainerAsService(t *testing.T) {
```

cmd/nerdctl/run_windows_test.go
```diff
@@ -0,0 +1,59 @@
...
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	hostuser := "nt authority\\local service"
+	base.Cmd("run", "--rm", "--isolation=host", "-u", "NT AUTHORITY\\Local Service", testutil.CommonImage, "whoami").AssertOutContains(hostuser)
+}
+
+func TestRunAsHostProcessContainerAslocalservice(t *testing.T) {
```
dcantah, line 54:
```suggestion
func TestRunAsHostProcessContainerAsLocalService(t *testing.T) {
```

cmd/nerdctl/run_windows_test.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		t.Fatalf("unable to get hostname: %s", err)
+	}
+
+	base.Cmd("run", "--rm", "--isolation=host", testutil.CommonImage, "hostname").AssertOutContains(string(hostname))
+	output := base.Cmd("run", "--rm", "--isolation=host", testutil.CommonImage, "whoami").Out()
+	t.Logf("whoami %s", output)
+}
+
+func TestRunAsHostProcessContainerAsUser(t *testing.T) {
```
dcantah, line 39:
```suggestion
func TestRunHostProcessContainerAsUser(t *testing.T) {
```

cmd/nerdctl/run_windows_test.go
```diff
@@ -0,0 +1,59 @@
...
+package main
+
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestRunAsHostProcessContainer(t *testing.T) {
```
dcantah, line 26:
```suggestion
func TestRunHostProcessContainer(t *testing.T) {
```

dcantah (approved):
LGTM after those test renames (and assuming CI is green)

cmd/nerdctl/run_windows.go
dcantah: Oh missed that, yep yep. Yea we can do best effort here as we can't control what happens there really.. Maybe could have docs on this that if your default runtime is runhcs-hypervisor that isolation defaults to hyperv like it would for docker desktop

jsturtevant: I did call this out in the readme section with `default_runtime` set in config.toml. Maybe it could be clearer?

So I should clear out the hyper-v spec and set the windows part? or should I nil both out?

```golang
func WithWindowsProcessIsolated() oci.SpecOpts {
	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *specs.Spec) error {
		if s.Windows == nil {
			s.Windows = &specs.Windows{} <=== should Set this or set it to nil?
		}
		if s.Windows.HyperV != nil {
			s.Windows.HyperV = nil
		}
		return nil
	}
}
```

dcantah: Ahh, missed that 😬 But there's no point in clearing the spec I think, as if they have runhcs-hypervisor as the default runtime the shim will set the hyperv section back anyways. We're setting all of this up before the shims launched so it seems moot

jsturtevant: > But there's no point in clearing the spec I think, as if they have runhcs-hypervisor as the default runtime the shim will set the hyperv section back anyway

So this means if the default runtime is hyperv then you can't run process isolated?  I not sure we can handle this from here.  doing nothing when `process` and `hyperv` is the default runtime would be bad experience as well.

jsturtevant: I've added this back... Let me know what you think

dcantah: Wait now that I'm thinking about it, isn't the default_runtime_name a CRI field? That shouldn't matter or get read here right? @AkihiroSuda

You can change the runtime as a whole via --runtime (although process, hyperv and LCOW are all wrapped up in the same shim on Windows so it doesn't matter really). The shim options would have to change for this to matter, and it seems like the only runtime whose options are configured is runc https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/run_runtime.go#L34-L63. But regardless, nil'ing out hyperv is still not needed, as we're the ones controlling the spec here. If the shim ends up setting it for some reason ([like here](https://github.com/microsoft/hcsshim/blob/b295b1a866b1263ddede4f00dddf7a129b96c114/cmd/containerd-shim-runhcs-v1/service_internal.go#L109-L116)) it's irrelevant what we did here

dcantah:
> It does look like the compute_engine_instance (https://cirrus-ci.org/guide/custom-vms/) does support nested_virtualization:  nested_virtualization: true # optional. Whether to enable Intel VT-x. Defaults to false.

Sweet!

AkihiroSuda (approved):
Thanks, could you consider squashing commits

hack/configure-windows-ci.ps1
```diff
@@ -5,8 +5,9 @@ choco install --limitoutput --no-progress -y git golang
 
 #install containerd
 $version=$env:ctrdVersion
+echo "Installing containerd $version"
 curl.exe -L https://github.com/containerd/containerd/releases/download/v$version/containerd-$version-windows-amd64.tar.gz -o containerd-windows-amd64.tar.gz
-tar xvf containerd-windows-amd64.tar.gz
+tar.exe xvf containerd-windows-amd64.tar.gz
```
AkihiroSuda, line 10: ?

jsturtevant: Not strictly necessary but makes it explicit we are using the tar.exe shipped with Windows vs a powershell alias.

cmd/nerdctl/run_windows.go
jsturtevant: > If the shim ends up setting it for some reason ([like here](https://github.com/microsoft/hcsshim/blob/b295b1a866b1263ddede4f00dddf7a129b96c114/cmd/containerd-shim-runhcs-v1/service_internal.go#L109-L116)) it's irrelevant what we do here

So this means if the default runtime is `hyperv` and the user runs `nerdctl --isolation process` the user won't get a process isolated container?  Or is it only if sim options are set?

AkihiroSuda (approved):
Thanks

Issue #618, opened 2021-12-14
Multiple docker compose files


darkrift:
nerdctl doesn't support to specify multiple compose files.

The local version only specifies port forwarding rules where as the base one defines all the services to start with their main configuration
```
rlavoie@RLAVOIE-MBP$ TAG=0.0.0-development-SNAPSHOT nerdctl compose -f tests/docker-compose-base.yml -f tests/docker-compose-local.yml up
FATA[0000] service "XXXXX" has neither an image nor a build context specified: invalid compose project 
FATA[0000] exit status 1
```

docker compose actually has a merging mechanism that should be reused here as well.

jsoref:
Since I spent a bit of time looking into why `nerdctl compose -f ... -f ...` doesn't work, here are the pointers for the curious:
https://github.com/containerd/nerdctl/blob/fe8bd459b67546eb1bd6b8424ca2ade2b9fcd999/cmd/nerdctl/compose.go#L48
https://github.com/containerd/nerdctl/blob/fe8bd459b67546eb1bd6b8424ca2ade2b9fcd999/cmd/nerdctl/compose.go#L86-L89
https://github.com/containerd/nerdctl/blob/fe8bd459b67546eb1bd6b8424ca2ade2b9fcd999/pkg/composer/composer.go#L41

And the docker-compose entrypoint (note the `Array` in `StringArrayVarP`):
https://github.com/docker/compose/blob/5262d3bbf5295ca4f5ee341365c245a983c064fb/cmd/compose/compose.go#L148

PR #802, opened 2022-02-11
refactor compose and support multi compose files

Junnplus:

Fixes: https://github.com/containerd/nerdctl/issues/618

- [x] Test

### DIFF
--- a/cmd/nerdctl/compose_config_test.go
+++ b/cmd/nerdctl/compose_config_test.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"path/filepath"
 	"testing"
 
 	"github.com/containerd/nerdctl/pkg/testutil"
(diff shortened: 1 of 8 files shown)


navneetkarnani:
Looks like a nit ( spelling mistake ) in `ProjectDirecotry`

cmd/nerdctl/compose_config_test.go
```diff
@@ -75,3 +76,31 @@ services:
...
 	assert.Assert(t, hash != newHash)
 }
+
+func TestComposeConfigWithMultipleFile(t *testing.T) {
+	base := testutil.NewBase(t)
+
+	var dockerComposeYAML = `
+services:
+  hello1:
+    image: alpine:3.13
```
fahedouch, line 86: plz use  testing predefined image https://github.com/containerd/nerdctl/blob/251ddcef7bac29f53241ed36d38038ed852089bb/pkg/testutil/testutil_linux.go#L36

Junnplus: @fahedouch Only the config is tested here, and the image will not be pulled. It may be better to explicitly write the image version.

AkihiroSuda (approved):
Thanks

PR #817, opened 2022-02-15
Fix Kata Containers networking support

sameo:
This PR defines a new annotation label for getting the networking namespace to use when calling the CNI plugins.
When set by the calling runtime (through `spec.State.Annotations`), this label takes precedence over the PID based networking namespace resolution.

This fixes the Kata Containers networking support, when running Kata with the following PR:
https://github.com/kata-containers/kata-containers/pull/3670

### DIFF
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -44,6 +44,16 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	// NetworkNamespace is the network namespace path to be passed to the CNI plugins.
+	// When this annotation is set from the runtime spec.State payload, it takes
+	// precedence over the PID based resolution (/proc/<pid>/ns/net) where pid is
+	// spec.State.Pid.
+	// This is mostly used for VM based runtime, where the spec.State PID does not
+	// necessarily lives in the created container networking namespace.
+	NetworkNamespace = labels.Prefix + "network-namespace"
+)
+
 func Run(stdin io.Reader, stderr io.Writer, event, dataStore, cniPath, cniNetconfPath string) error {
 	if stdin == nil || event == "" || dataStore == "" || cniPath == "" || cniNetconfPath == "" {
 		return errors.New("got insufficient args")
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -262,9 +262,21 @@ func loadSpec(bundle string) (*hookSpec, error) {
 }
 
 func getNetNSPath(state *specs.State) (string, error) {
-	if state.Pid == 0 {
-		return "", errors.New("state.Pid is unset")
+	// If we have a network-namespace annotation we use it over the passed Pid.
+	netNsPath, netNsFound := state.Annotations[NetworkNamespace]
+	if netNsFound {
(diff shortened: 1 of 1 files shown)


pkg/labels/labels.go
```diff
@@ -74,6 +74,14 @@ const (
...
 	// (like "nerdctl/bypass4netns=true" or "nerdctl/bypass4netns=false")
 	Bypass4netns = Prefix + "bypass4netns"
+
+	// NetworkNamespace is the network namespace path to be passed to the CNI plugins.
+	// When this annotation is set from the runtime spec.State payload, it takes
+	// precedence over the PID based resolution (/proc/<pid>/ns/net) where pid is
+	// spec.State.Pid.
+	// This is mostly used for VM based runtime, where the spec.State PID does not
+	// necessarily lives in the created container networking namespace.
+	NetworkNamespace = Prefix + "network-namespace"
```
AkihiroSuda, line 84: IIUC, this doesn't work with `nerdctl run --restart=always` containers when the host OS rebooted, because the specified netns no longer exists

sameo: I am not familiar with that part of the code base, so please correct me if/where I'm wrong:

After a reboot, nerdctl will create a  new containerd task, which will trigger kata to create a new netns and update the `spec.State.Annotations` accordingly. The OCI hooks will be called with that newly created netns, and nerdctl will use it to run the CNI plugins.

What am I missing?

AkihiroSuda: >  nerdctl will create a new containerd task

The new task is created by the restart monitor, not by nerdctl (unless user executes nerdctl manually): https://github.com/containerd/containerd/tree/v1.6.0-rc.4/runtime/restart/monitor

AkihiroSuda: If you tested it with the restart monitor and it works, this PR LGTM, but the annotation probably shouldn't be defined in `pkg/labels`, as it seems pure OCI annotation, not containerd label.

sameo: I started a container with `--restart always` and verified that it was up and running after a reboot, with networking working as well.

> the annotation probably shouldn't be defined in pkg/labels, as it seems pure OCI annotation, not containerd label.

You mean I should use a `org.opencontainers` prefixed annotation?

sameo: > The proposed annotation is not a containerd label, so it shouldn’t be defined in `pkg/labels ` package, but should be probably defined in `pkg/ocihook` package.
> 
> The `nerdctl/` prefix is fine, and it must NOT be `org.opencontainers/`, but `io.katacontainers/` might be even better?

Thanks for the explanation, that makes sense now.
Since the main and single consumer of this annotation is nerdctl, I'd rather keep the `nerdctl/` prefix. I'll move the definition to the `ocihook` package.

sameo: I changed the first commit to move the annotation key under `ocihook.go`. Let me know if that's fine with you.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
> @AkihiroSuda There's one rootless test that's failing. Should I be worried about it?

Probably unrelated

AkihiroSuda:
@liubin LGTY?

liubin (approved):

PR #819, opened 2022-02-16
build: Allow to specify compression type of the result image

ktock:
BuildKit v0.10 will introduce some improvements for image compression. https://github.com/moby/buildkit/releases/tag/v0.10.0-rc1
This includes support for zstd and estargz compression types and support for `--force-compression` option.

It would be great for `nerdctl build` to have support for compression types.

```
nerdctl build -t test --compression=estargz --force-compression /tmp/ctx
```


AkihiroSuda:
Why not use --output

ktock:
Because `--output` doesn't support loading images  to containerd, IIUC.

AkihiroSuda:
> Because `--output` doesn't support loading images to containerd, IIUC.

Can we just implement rootless containerd worker and use `--output type=image`?

ktock:
> Can we just implement rootless containerd worker and use --output type=image?

SGTM

Issue #434, opened 2021-10-14
Unable to use local images with compose


willcohen:
In workflows where `docker-compose.yml` builds images locally and subsequently tries to build other images relying on those earlier ones, `nerdctl` is unable to find the locally built images, and instead appears to try to find them from nonexistent registries.

To reproduce:

Contents of `Dockerfile`:
```
FROM alpine
```

Contents of `Dockerfile-from`:
```
FROM foo as foo
```

Contents of `docker-compose.yml`:
```
version: "3.1"

services:

  foo:
    container_name: foo
    image: foo
    build:
      context: ./
      dockerfile: Dockerfile

  foo2:
    container_name: foo2
    image: foo2
    build:
      context: ./
      dockerfile: Dockerfile-from
```

Running `compose build` (via lima):

```
$ lima nerdctl compose build        
INFO[0000] Building image foo2                          
[+] Building 0.5s (3/3) FINISHED                                                                                                                                                             
 => [internal] load build definition from Dockerfile-from                                                                                                                               0.1s
 => => transferring dockerfile: 58B                                                                                                                                                     0.0s
 => [internal] load .dockerignore                                                                                                                                                       0.0s
 => => transferring context: 2B                                                                                                                                                         0.0s
 => ERROR [internal] load metadata for docker.io/library/foo:latest                                                                                                                     0.4s
------
 > [internal] load metadata for docker.io/library/foo:latest:
------
Dockerfile-from:1
--------------------
   1 | >>> FROM foo as foo
   2 |     
--------------------
error: failed to solve: foo: pull access denied, repository does not exist or may require authorization: authorization status: 401: authorization failed
FATA[0001] unrecognized image format                    
FATA[0001] error while building image foo2: exit status 1

**Comment by willcohen** [comment]
I should note that, additionally, if `docker-compose.yml` is set to have `foo2` depend on `foo` to force the local building of `foo`, it still fails:

Contents of `docker-compose.yml`:

```
version: "3.1"

services:

  foo:
    container_name: foo
    image: foo
    build:
      context: ./
      dockerfile: Dockerfile

  foo2:
    container_name: foo2
    image: foo2
    build:
      context: ./
      dockerfile: Dockerfile-from
    depends_on:
      - foo
```

```
$ lima nerdctl compose build
INFO[0000] Building image foo                           
[+] Building 0.7s (5/5) FINISHED                                                                                                                                                             
 => [internal] load build definition from Dockerfile                                                                                                                                    0.1s
[...]

PR #858, opened 2022-03-03
rootless: Support BuildKit containerd worker

ktock:
Fixes: #434

This commit patches `containerd-rootless-setuptool.sh` to allow configuring nerdctl for BuildKit rootless containerd worker (BuildKit v0.10).

This enables nerdctl to share images into BuildKit which isn't possible with OCI worker (#434).
For example, this enables BuildKit to use images pulled by `nerdctl pull` as base images.

```console
$ nerdctl pull ghcr.io/stargz-containers/ubuntu:20.04-org
$ nerdctl tag ghcr.io/stargz-containers/ubuntu:20.04-org mybase
$ mkdir -p /tmp/ctx && cat <<EOF > /tmp/ctx/Dockerfile
FROM mybase
RUN echo hello > /hello
CMD ["cat", "/hello"]
EOF
$ nerdctl build -t mytest /tmp/ctx
[+] Building 5.4s (6/6) FINISHED                                                                                                                                                                                                                                                         
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                                0.1s
 => => transferring dockerfile: 95B                                                                                                                                                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                                   0.2s
 => => transferring context: 2B                                                                                                                                                                                                                                                     0.0s
 => [internal] load metadata for docker.io/library/mybase:latest                                                                                                                                                                                                                    2.6s
 => [1/2] FROM docker.io/library/mybase@sha256:adf73ca014822ad8237623d388cedf4d5346aa72c270c5acc01431cc93e18e2d                                                                                                                                                                     1.5s
 => => resolve docker.io/library/mybase@sha256:adf73ca014822ad8237623d388cedf4d5346aa72c270c5acc01431cc93e18e2d                                                                                                                                                                     1.3s
 => [2/2] RUN echo hello > /hello                                                                                                                                                                                                                                                   0.2s
 => exporting to oci image format                                                                                                                                                                                                                                                   0.8s
 => => exporting layers                                                                                                                                                                                                                                                             0.2s
 => => exporting manifest sha256:4b8d0218edd5c342e1a7c3bc9cb3adfeec480563ff06535b85c0fd85fc1bbad2                                                                                                                                                                                   0.0s
 => => exporting config sha256:b6cbc7ea26b1fd045d589f7fef3d011165f3b0f146c32ca6bae06d47199a5a7b                                                                                                                                                                                     0.0s
 => => sending tarball                                                                                                                                                                                                                                                              0.5s
unpacking docker.io/library/mytest:latest (sha256:4b8d0218edd5c342e1a7c3bc9cb3adfeec480563ff06535b85c0fd85fc1bbad2)...done
$ nerdctl run --rm -it mytest
hello
```

Note that 1 containerd worker can use 1 containerd namespace so buildkitd needs to be launched for each namespace the user wants to use.

## setup (rootless)

Rootless user can enable containerd worker (in "default" namespace of containerd) like the following:

```console
$ CONTAINERD_NAMESPACE=default containerd-rootless-setuptool.sh install-buildkit-containerd
```

NOTE1: `containerd-rootless-setuptool.sh` is aware of `CONTAINERD_NAMESPACE` and `CONTAINERD_SNPASHOTTER`.
User can launch additional buildkitd processes by executing this script with different `CONTAINERD_NAMESPACE` values.
NOTE2: If `CONTAINERD_NAMESPACE` is not specified, buildkitd installed in "buildkit" namespace which is the default of buildkitd. If a user wants to give buildkitd access to the images in "default" namespace, this command should be run with `CONTAINERD_NAMESPACE=default`.

When $BUILDKIT_HOST is unset, nerdctl (both rootless and rootful) tries to use `buildkit-<NS>/buildkitd.sock` by default and then try to fall back to `buildkit-default/buildkitd.sock`, `buildkit/buildkitd.sock`

~Maybe we can add an option to allow rootless nerdctl automatically detecting BuildKit socket location for a namespace (i.e. `unix://${XDG_RUNTIME_DIR}/buildkit-${CONTAINERD_NAMESPACE}/buildkitd.sock`) without manually configuring `BUILDKIT_HOST`.~ **Added this**

## setup (rootful)

Rootful containerd worker has been available even without this patch.
Use the configuration (`/etc/buildkit/buildkitd.toml`) suggested at https://github.com/containerd/nerdctl/issues/434#issuecomment-998702312.

```toml
[worker.oci]
  enabled = false

[worker.containerd]
  enabled = true
  # namespace should be "k8s.io" for Kubernetes (including Rancher Desktop)
  namespace = "default"
```

### DIFF
--- a/cmd/nerdctl/build.go
+++ b/cmd/nerdctl/build.go
@@ -28,6 +28,7 @@ import (
 
 	"path/filepath"
 
+	dockerreference "github.com/containerd/containerd/reference/docker"
 	"github.com/containerd/nerdctl/pkg/buildkitutil"
 	"github.com/containerd/nerdctl/pkg/defaults"
 	"github.com/containerd/nerdctl/pkg/platformutil"
@@ -73,6 +74,46 @@ func newBuildCommand() *cobra.Command {
 	return buildCommand
 }
 
+func getBuildkitHost(cmd *cobra.Command) (string, error) {
+	if cmd.Flags().Changed("buildkit-host") {
+		// If address is explicitly specified, use it.
+		buildkitHost, err := cmd.Flags().GetString("buildkit-host")
+		if err != nil {
+			return "", err
+		}
+		if err := buildkitutil.PingBKDaemon(buildkitHost); err != nil {
+			return "", err
+		}
+		return buildkitHost, nil
+	}
+	ns, err := cmd.Flags().GetString("namespace")
+	if err != nil {
+		return "", err
+	}
+	return buildkitutil.GetBuildkitHost(ns)
+}
+
+func isImageSharable(buildkitHost string, namespace, uuid string) (bool, error) {
+	labels, err := buildkitutil.GetWorkerLabels(buildkitHost)
+	if err != nil {
+		return false, err
+	}
+	logrus.Debugf("worker labels: %+v", labels)
+	executor, ok := labels["org.mobyproject.buildkit.worker.executor"]
+	if !ok {
+		return false, nil
+	}
+	containerdUUID, ok := labels["org.mobyproject.buildkit.worker.containerd.uuid"]
+	if !ok {
+		return false, nil
+	}
+	containerdNamespace, ok := labels["org.mobyproject.buildkit.worker.containerd.namespace"]
+	if !ok {
+		return false, nil
+	}
+	return executor == "containerd" && containerdUUID == uuid && containerdNamespace == namespace, nil
+}
+
 func buildAction(cmd *cobra.Command, args []string) error {
 	platform, err := cmd.Flags().GetStringSlice("platform")
 	if err != nil {
@@ -80,15 +121,12 @@ func buildAction(cmd *cobra.Command, args []string) error {
 	}
 	platform = strutil.DedupeStrSlice(platform)
 
-	buildkitHost, err := cmd.Flags().GetString("buildkit-host")
+	buildkitHost, err := getBuildkitHost(cmd)
 	if err != nil {
 		return err
 	}
-	if err := buildkitutil.PingBKDaemon(buildkitHost); err != nil {
-		return err
-	}
 
-	buildctlBinary, buildctlArgs, needsLoading, metaFile, cleanup, err := generateBuildctlArgs(cmd, platform, args)
+	buildctlBinary, buildctlArgs, needsLoading, metaFile, cleanup, err := generateBuildctlArgs(cmd, buildkitHost, platform, args)
 	if err != nil {
 		return err
 	}
@@ -164,7 +202,7 @@ func buildAction(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func generateBuildctlArgs(cmd *cobra.Command, platform, args []string) (string, []string, bool, string, func(), error) {
+func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, args []string) (string, []string, bool, string, func(), error) {
 	var needsLoading bool
 	if len(args) < 1 {
 		return "", nil, false, "", nil, errors.New("context needs to be specified")
@@ -184,13 +222,34 @@ func generateBuildctlArgs(cmd *cobra.Command, platform, args []string) (string, 
 		return "", nil, false, "", nil, err
 	}
 	if output == "" {
-		output = "type=docker"
-		if len(platform) > 1 {
-			// For avoiding `error: failed to solve: docker exporter does not currently support exporting manifest lists`
-			// TODO: consider using type=oci for single-platform build too
-			output = "type=oci"
+		client, ctx, cancel, err := newClient(cmd)
+		if err != nil {
+			return "", nil, false, "", nil, err
 		}
-		needsLoading = true
+		defer cancel()
+		info, err := client.Server(ctx)
+		if err != nil {
+			return "", nil, false, "", nil, err
+		}
+		ns, err := cmd.Flags().GetString("namespace")
+		if err != nil {
+			return "", nil, false, "", nil, err
+		}
+		sharable, err := isImageSharable(buildkitHost, ns, info.UUID)
+		if err != nil {
+			return "", nil, false, "", nil, err
+		}
+		if sharable {
+			output = "type=image,unpack=true" // ensure the target stage is unlazied (needed for any snapshotters)
+		} else {
+			output = "type=docker"
+			if len(platform) > 1 {
+				// For avoiding `error: failed to solve: docker exporter does not currently support exporting manifest lists`
+				// TODO: consider using type=oci for single-platform build too
+				output = "type=oci"
+			}
+			needsLoading = true
+		}
 	}
 	tagValue, err := cmd.Flags().GetStringArray("tag")
 	if err != nil {
@@ -200,12 +259,12 @@ func generateBuildctlArgs(cmd *cobra.Command, platform, args []string) (string, 
 		if len(tagSlice) > 1 {
 			return "", nil, false, "", nil, fmt.Errorf("specifying multiple -t is not supported yet")
 		}
-		output += ",name=" + tagSlice[0]
-	}
-
-	buildkitHost, err := cmd.Flags().GetString("buildkit-host")
-	if err != nil {
-		return "", nil, false, "", nil, err
+		ref := tagSlice[0]
+		named, err := dockerreference.ParseNormalizedNamed(ref)
+		if err != nil {
+			return "", nil, false, "", nil, err
+		}
+		output += ",name=" + dockerreference.TagNameOnly(named).String()
 	}
 
 	buildctlArgs := buildkitutil.BuildctlBaseArgs(buildkitHost)
@@ -375,14 +434,19 @@ func getDigestFromMetaFile(path string) (string, error) {
 	}
 	defer os.Remove(path)
 
-	metadata := map[string]string{}
+	metadata := map[string]json.RawMessage{}
 	if err := json.Unmarshal(data, &metadata); err != nil {
 		logrus.WithError(err).Errorf("failed to unmarshal metadata file %s", path)
 		return "", err
 	}
-	digest, ok := metadata["containerimage.digest"]
+	digestRaw, ok := metadata["containerimage.digest"]
 	if !ok {
 		return "", errors.New("failed to find containerimage.digest in metadata file")
+	}
+	var digest string
+	if err := json.Unmarshal(digestRaw, &digest); err != nil {
+		logrus.WithError(err).Errorf("failed to unmarshal digset")
+		return "", err
 	}
 	return digest, nil
 }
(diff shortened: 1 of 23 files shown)


.github/workflows/test.yml
```diff
@@ -84,8 +84,10 @@ jobs:
     strategy:
       matrix:
         containerd: [v1.4.5, v1.6.1, main]
+        buildkit-worker: [oci, containerd]
```
AkihiroSuda, line 87: We seriously have to deflake CI before expanding the test matrix

https://github.com/containerd/nerdctl/issues?q=is%3Aissue+is%3Aopen+label%3Aarea%2Fci

AkihiroSuda: Until that we can just test the containerd worker (after BK v0.10.0 GA)

cmd/nerdctl/build_test.go
```diff
@@ -48,6 +48,104 @@ CMD ["echo", "nerdctl-build-test-string"]
...
+	assert.NilError(t, err)
+	defer os.RemoveAll(buildCtx2)
+
+	base.Cmd("build", "-t", imageName2, buildCtx2).AssertOK()
+	base.Cmd("build", buildCtx2, "-t", imageName2).AssertOK()
+
+	base.Cmd("run", "--rm", imageName2).AssertOutExactly("hello2\n")
+}
+
+func TestBuildUnlazy(t *testing.T) {
```
AkihiroSuda, line 125: What is "Unlazy" here? Doesn't seem related to stargz stuff though

cmd/nerdctl/build_test.go
```diff
@@ -48,6 +48,104 @@ CMD ["echo", "nerdctl-build-test-string"]
 	base.Cmd("run", "--rm", imageName).AssertOutExactly("nerdctl-build-test-string\n")
 }
 
+func TestBuildBaseImage(t *testing.T) {
+	if os.Getenv("TEST_BUILDKIT_WORKER") != "containerd" {
+		t.Skip()
```
AkihiroSuda, line 53: Needs explanation

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -273,6 +273,51 @@ cmd_entrypoint_install_buildkit() {
...
+	ADDRESS="${XDG_RUNTIME_DIR}/buildkit-${NAMESPACE}/buildkitd.sock"
+	ROOT="${XDG_DATA_HOME}/buildkit-${NAMESPACE}"
+	cat <<-EOT | install_systemd_unit "${NAMESPACE}-${SYSTEMD_BUILDKIT_UNIT}"
+		[Unit]
+		Description=BuildKit (Rootless)
+		PartOf=${SYSTEMD_CONTAINERD_UNIT}
+
+		[Service]
+		Environment=PATH=$BIN:/sbin:/usr/sbin:$PATH
+		ExecStart="$REALPATH0" nsenter -- buildkitd --config=${XDG_CONFIG_HOME}/buildkit/buildkitd.toml --addr="unix://${ADDRESS}" --root="${ROOT}" --containerd-worker-namespace="${NAMESPACE}" --containerd-worker-snapshotter="${CONTAINERD_SNAPSHOTTER:-overlayfs}"
```
AkihiroSuda, line 308: No need to specify the toml AFAICS

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -273,6 +273,51 @@ cmd_entrypoint_install_buildkit() {
...
+		ExecReload=/bin/kill -s HUP \$MAINPID
+		RestartSec=2
+		Restart=always
+		Type=simple
+		KillMode=mixed
+
+		[Install]
+		WantedBy=default.target
+	EOT
+	INFO "Set \`export BUILDKIT_HOST=\"unix://${ADDRESS}\"\` to use this configuration of buildkitd"
```
AkihiroSuda, line 318: When `$BUILDKIT_HOST` is unset, nerdctl (both rootless and rootful) should try to use `buildkit-<NS>/buildkitd.sock` by default and then try to fall back to `buildkit-default/buildkitd.sock`, `buildkit/buildkitd.sock`

AkihiroSuda: Also, `nerdctl build` should try to use the containerd `image` exporter when the worker UUID corresponds to the containerd UUID && the worker namespace corresponds to the current namespace

.github/workflows/test.yml
ktock: Enabled the test only with containerd worker

cmd/nerdctl/build_test.go
ktock: Deleted this test because TestBuild already coverts this.

What I wanted to test was the following edge case:

If a build doesn't run any commands nor perform COPY on the base image, BuildKit tries not to download that image (i.e. keeps it lazy). This behaviour happens not only with stargz-snapshotter but also with any type of snapshotter.
`type=oci`/`type=docker` triggers unlazying of that base image but `type=image` doesn't. This leads to "not found" error for the contents of that image e.g. when nerdctl tries to run it. As a workaround, nerdctl adds `unpack=true` to `buildctl build` which triggers unlazy even when `type=image` is specified.

TestBuild is exactly the situation where the base image can be lazy. So this test is enough for testing that this workaround actually works.

cmd/nerdctl/build_test.go
ktock: Now we don't need this envvar because we always run tests with containerd worker.

extras/rootless/containerd-rootless-setuptool.sh
ktock: Fixed this.

extras/rootless/containerd-rootless-setuptool.sh
ktock: Added this feature.

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -273,6 +273,50 @@ cmd_entrypoint_install_buildkit() {
...
+	ADDRESS="${XDG_RUNTIME_DIR}/buildkit-${NAMESPACE}/buildkitd.sock"
+	ROOT="${XDG_DATA_HOME}/buildkit-${NAMESPACE}"
+	cat <<-EOT | install_systemd_unit "${NAMESPACE}-${SYSTEMD_BUILDKIT_UNIT}"
+		[Unit]
+		Description=BuildKit (Rootless)
+		PartOf=${SYSTEMD_CONTAINERD_UNIT}
+
+		[Service]
+		Environment=PATH=$BIN:/sbin:/usr/sbin:$PATH
+		ExecStart="$REALPATH0" nsenter -- buildkitd --addr="unix://${ADDRESS}" --root="${ROOT}" --containerd-worker-namespace="${NAMESPACE}" --containerd-worker-snapshotter="${CONTAINERD_SNAPSHOTTER:-overlayfs}"
```
AkihiroSuda, line 308: Just leave --containerd-worker-snapshotter unset when $CONTAINERD_SNAPSHOTTER is not set

AkihiroSuda: Same applies to other flags too

cmd/nerdctl/system.go
```diff
@@ -33,6 +39,40 @@ func newSystemCommand() *cobra.Command {
 	systemCommand.AddCommand(
 		newEventsCommand(),
 		newInfoCommand(),
+		newBuildPruneCommand(),
```
AkihiroSuda, line 42: This has to be `nerdctl builder prune`
https://docs.docker.com/engine/reference/commandline/builder_prune/

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -273,6 +273,58 @@ cmd_entrypoint_install_buildkit() {
...
+		ERROR "Install containerd first (\`$ARG0 install\`)"
+		exit 1
+	fi
+	UNIT_NAME=${SYSTEMD_BUILDKIT_UNIT}
+	BUILDKITD_FLAG=
+	if [ -n "${CONTAINERD_NAMESPACE:-}" ] ; then
+		UNIT_NAME="${CONTAINERD_NAMESPACE}-${SYSTEMD_BUILDKIT_UNIT}"
+		BUILDKITD_FLAG="${BUILDKITD_FLAG} --addr=unix://${XDG_RUNTIME_DIR}/buildkit-${CONTAINERD_NAMESPACE}/buildkitd.sock --root=${XDG_DATA_HOME}/buildkit-${CONTAINERD_NAMESPACE} --containerd-worker-namespace=${CONTAINERD_NAMESPACE}"
+	else
+		WARNING "buildkitd has access images in \"buildkit\" namespace by default. If you want to give buildkitd access to the images in \"default\" namespace, run this command with CONTAINERD_NAMESPACE=default"
```
AkihiroSuda, line 304:
```suggestion
		WARNING "buildkitd has access to images in \"buildkit\" namespace by default. If you want to give buildkitd access to the images in \"default\" namespace, run this command with CONTAINERD_NAMESPACE=default"
```

AkihiroSuda (approved):

AkihiroSuda:
We need to have docs about this later

AkihiroSuda:
Regression: OCI worker (`/run/user/1001/buildkit/buildkitd.sock`) can no longer be auto detected:

```console
$ nerdctl build -t foo .
ERRO[0000] error: failed to list workers: Unavailable: connection error: desc = "transport: error while dialing: dial unix /run/user/1001/buildkit-default/buildkitd.sock: connect: no such file or directory" 
WARN[0000] `buildctl` needs to be installed and `buildkitd` needs to be running, see https://github.com/moby/buildkit , and `containerd-rootless-setuptool.sh install-buildkit` for OCI worker or `containerd-rootless-setuptool.sh install-buildkit-containerd` for containerd worker: exit status 1  host="unix:///run/user/1001/buildkit-default/buildkitd.sock"
```

cmd/nerdctl/build.go
```diff
@@ -73,22 +74,59 @@ func newBuildCommand() *cobra.Command {
...
+	}
+	containerdUUID, ok := labels["org.mobyproject.buildkit.worker.containerd.uuid"]
+	if !ok {
+		return false, nil
+	}
+	containerdNamespace, ok := labels["org.mobyproject.buildkit.worker.containerd.namespace"]
+	if !ok {
+		return false, nil
+	}
+	return executor == "containerd" && containerdUUID == uuid && containerdNamespace == namespace, nil
```
AkihiroSuda, line 114: Maybe we should compare the snapshotter too?
To avoid some potential issues with unpacked images.

AkihiroSuda: cc @ktock

ktock: SGTM

PR #886, opened 2022-03-10
fix make failed when use golang1.16

ningmingxiao:
fix make failed

### DIFF
--- a/pkg/portutil/port_allocate_others.go
+++ b/pkg/portutil/port_allocate_others.go
@@ -1,4 +1,5 @@
 //go:build !linux
+// +build !linux
 
 /*
    Copyright The containerd Authors.


pkg/portutil/port_allocate_others.go
```diff
@@ -1,3 +1,4 @@
+//+build !linux
```
AkihiroSuda, line 1: //go:build should come first

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/886/files#r823511893

AkihiroSuda:
The commit title is inaccurate because the current master has been compliable with Go 1.17.
Also please add DCO sign

pkg/portutil/port_allocate_others.go
Zheaoli, line 1: https://github.com/containerd/nerdctl/blob/master/go.mod#L3

The Go 1.17 has been recommended version for the nerdctl project. FYI https://github.com/containerd/nerdctl/blob/master/go.mod#L3.

And in Go 1.17 `//go:build ` is the recommended way to specify the build target.

If you want to backport the build target to Go 1.16, you may need to add `//+build` tag to everywhere we use `//go:build` to maintain the code style. FYI https://github.com/containerd/nerdctl/search?q=repo%3Acontainerd%2Fnerdctl+%2F%2Fgo%3Abuild

BTW, I don't think this would be necessary

AkihiroSuda (approved):

PR #894, opened 2022-03-11
Support fallback to plain http in nerdctl login

### LINKED ISSUE #887: http registry login (`http: server gave HTTP response to HTTPS client`)
I found this question #224 but setting the `--insecure-registry` flag only skips tls verfiy and does not affect the https://github.com/containerd/nerdctl/blob/354d03acb11a1649f0f3aec73218b8d8e2d9e062/pkg/imgutil/dockerconfigresolver/dockerconfigresolver.go#L39 option (anymore?).

I figured out plainHTTP is only accepted on localhost. That's totally fine in terms of security not sending credentials in plain text over the network. Nevertheless I have to do it (even though VPN) ...

Can we provide a new flag for http? Or could we parse the protocol form the host input for example `nerdctl login http://registry.com` ?

Zheaoli:
Fix #887 

### DIFF
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -242,6 +242,7 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
 		return "", fmt.Errorf("got empty []docker.RegistryHost for %q", host)
 	}
 	needFallbackToPlainHttp := false
+	var needFallBackRegHosts []docker.RegistryHost
 	for i, rh := range regHosts {
 		err = tryLoginWithRegHost(ctx, rh)
 		identityToken := fetchedRefreshTokens[rh.Host] // can be empty
@@ -250,33 +251,19 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
 		}
 		if insecure && (errutil.IsErrHTTPResponseToHTTPSClient(err) || errutil.IsErrConnectionRefused(err)) {
 			needFallbackToPlainHttp = true
-			break
+			needFallBackRegHosts = append(needFallBackRegHosts, rh)
 		}
 		logrus.WithError(err).WithField("i", i).Error("failed to call tryLoginWithRegHost")
 	}
 	if needFallbackToPlainHttp {
-		dOpts = append(dOpts, dockerconfigresolver.WithPlainHTTP(true))
-		ho, err = dockerconfigresolver.NewHostOptions(ctx, host, dOpts...)
-		if err != nil {
-			return "", err
-		}
-		fetchedRefreshTokens := make(map[string]string) // key: req.URL.Host
-		// onFetchRefreshToken is called when tryLoginWithRegHost calls rh.Authorizer.Authorize()
-		onFetchRefreshToken := func(ctx context.Context, s string, req *http.Request) {
-			fetchedRefreshTokens[req.URL.Host] = s
-		}
-		ho.AuthorizerOpts = append(ho.AuthorizerOpts, docker.WithFetchRefreshToken(onFetchRefreshToken))
-		regHosts, err = dockerconfig.ConfigureHosts(ctx, *ho)(host)
-		if err != nil {
-			return "", err
-		}
-		for i, rh := range regHosts {
+		for _, rh := range needFallBackRegHosts {
+			rh.Scheme = "http"
 			err = tryLoginWithRegHost(ctx, rh)
 			identityToken := fetchedRefreshTokens[rh.Host] // can be empty
 			if err == nil {
 				return identityToken, nil
 			}
-			logrus.WithError(err).WithField("i", i).Error("failed to call tryLoginWithRegHost")
+			logrus.WithError(err).Error("failed to call tryLoginWithRegHost with plain http")
 		}
 	}
 	return "", err
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -31,6 +31,7 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/remotes/docker"
 	dockerconfig "github.com/containerd/containerd/remotes/docker/config"
+	"github.com/containerd/nerdctl/pkg/errutil"
 	"github.com/containerd/nerdctl/pkg/imgutil/dockerconfigresolver"
 	dockercliconfig "github.com/docker/cli/cli/config"
 	clitypes "github.com/docker/cli/cli/config/types"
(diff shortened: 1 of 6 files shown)


pkg/imgutil/imgutil.go
```diff
@@ -145,7 +146,7 @@ func EnsureImage(ctx context.Context, client *containerd.Client, stdout, stderr
 	img, err := PullImage(ctx, client, stdout, stderr, snapshotter, resolver, ref, ocispecPlatforms, unpack, quiet)
 	if err != nil {
 		// In some circumstance (e.g. people just use 80 port to support pure http), the error will contain message like "dial tcp <port>: connection refused".
-		if !IsErrHTTPResponseToHTTPSClient(err) && !IsErrConnectionRefused(err) {
+		if !common.IsErrHTTPResponseToHTTPSClient(err) && !common.IsErrConnectionRefused(err) {
```
AkihiroSuda, line 149: `common` -> `errutil`

Zheaoli: Got it

cmd/nerdctl/login.go
```diff
@@ -240,14 +241,44 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
...
 		if err == nil {
 			return identityToken, nil
 		}
+		if insecure && (errutil.IsErrHTTPResponseToHTTPSClient(err) || errutil.IsErrConnectionRefused(err)) {
+			needFallbackToPlainHttp = true
+			break
+		}
 		logrus.WithError(err).WithField("i", i).Error("failed to call tryLoginWithRegHost")
 	}
+	if needFallbackToPlainHttp {
```
Junnplus, line 257: I think we can retry directly after 246 lines with plain http.

Zheaoli: Actually, I'm not sure about this.

cmd/nerdctl/login.go
```diff
@@ -240,14 +241,44 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
 	if len(regHosts) == 0 {
 		return "", fmt.Errorf("got empty []docker.RegistryHost for %q", host)
 	}
+	needFallbackToPlainHttp := false
 	for i, rh := range regHosts {
 		err = tryLoginWithRegHost(ctx, rh)
```
Junnplus, line 246:
```suggestion
                err = tryLoginWithRegHost(ctx, rh)
		if err != nil && insecure && (errutil.IsErrHTTPResponseToHTTPSClient(err) || errutil.IsErrConnectionRefused(err)) {
		    rh.Scheme = "http"
		    err = tryLoginWithReghost(ctx, rh)
		}
```

cmd/nerdctl/login.go
Junnplus: see https://github.com/containerd/nerdctl/pull/894/files#r827661658

cmd/nerdctl/login_linux_test.go
```diff
@@ -75,3 +76,102 @@ func TestLoginWithSpecificRegHosts(t *testing.T) {
 	}
 
 }
+
+func TestLoginWithPlainHttp(t *testing.T) {
```
Junnplus, line 80: I'm not sure if this test works, it seems that `registry:2` image can't verify this issue.

I tested it locally through nexus3

cmd/nerdctl/login_linux_test.go
```diff
@@ -75,3 +76,102 @@ func TestLoginWithSpecificRegHosts(t *testing.T) {
...
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	reg := testregistry.NewAuthWithHttp(base, "admin", "validTestPassword", 5000)
+	testCasesForPort5000 := []struct {
```
Junnplus, line 84: IMHO, We should merge `testCasesForPort80` with it.

cmd/nerdctl/login.go
Zheaoli: I'm not sure about changing the Scheme directly is the best practice

cmd/nerdctl/login_linux_test.go
Zheaoli: No, if we merge `testCasesForPort80` with `testCasesForPort5000`, we need to start/shut down all images for the test registry for each test case. I'm not sure it's right or not

cmd/nerdctl/login_linux_test.go
```diff
@@ -75,3 +76,102 @@ func TestLoginWithSpecificRegHosts(t *testing.T) {
...
+			regHost := tc.regHost
+			if tc.useRegPort {
+				regHost = fmt.Sprintf("%s:%d", regHost, tc.regPort)
+			}
+			if tc.shouldSuccess {
+				t.Logf("Good password")
+			} else {
+				t.Logf("Bad password")
+			}
+			cmd := base.Cmd("--insecure-registry", "--debug-full", "--hosts-dir", reg.HostsDir, "login", "-u", tc.username, "-p", tc.password, regHost)
```
Junnplus, line 121: We need to verify that it works without `-insecure-registry` flags.

Junnplus (changes requested):

cmd/nerdctl/login_linux_test.go
Junnplus: > we need to start/shut down all images for the test registry for each test case

Not necessary, just start them both at the same time.

cmd/nerdctl/login.go
```diff
@@ -240,14 +241,44 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
...
+		// onFetchRefreshToken is called when tryLoginWithRegHost calls rh.Authorizer.Authorize()
+		onFetchRefreshToken := func(ctx context.Context, s string, req *http.Request) {
+			fetchedRefreshTokens[req.URL.Host] = s
+		}
+		ho.AuthorizerOpts = append(ho.AuthorizerOpts, docker.WithFetchRefreshToken(onFetchRefreshToken))
+		regHosts, err = dockerconfig.ConfigureHosts(ctx, *ho)(host)
+		if err != nil {
+			return "", err
+		}
+		for i, rh := range regHosts {
```
fahedouch, line 273: no need to retry all `regHosts` you should fallback to http only in the registry that faced `(errutil.IsErrHTTPResponseToHTTPSClient(err) || errutil.IsErrConnectionRefused(err))`

cmd/nerdctl/login_linux_test.go
```diff
@@ -75,3 +76,100 @@ func TestLoginWithSpecificRegHosts(t *testing.T) {
...
+		{
+			regHost:           "127.0.0.1",
+			regPort:           80,
+			useRegPort:        false,
+			username:          "admin",
+			password:          "invalidTestPassword",
+			shouldSuccess:     false,
+			registry:          reg80,
+			shouldUseInSecure: true,
+		},
```
fahedouch, line 146: plz add a test `shouldUseInSecure: false` for `registry:  reg80` to check that there is not a regression

Zheaoli: OK

cmd/nerdctl/login.go
Junnplus: I reserve my opinion. 

When it is not a valid HTTPs request and `insecure-registry` flaged, we can fallback directly to plain HTTP. cc @fahedouch

AkihiroSuda:
Seems reopened as https://github.com/containerd/nerdctl/pull/907 (why?)

Zheaoli:
> Seems reopened as #907 (why?)

I have squashed my commit but not synced in the PR because of the GitHub issue. So I reopened the PR

PR #907, opened 2022-03-18
Support fallback to plain http in nerdctl login

### LINKED ISSUE #887: http registry login (`http: server gave HTTP response to HTTPS client`)

Zheaoli:
Fix #887 

### DIFF
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -31,6 +31,7 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/remotes/docker"
 	dockerconfig "github.com/containerd/containerd/remotes/docker/config"
+	"github.com/containerd/nerdctl/pkg/errutil"
 	"github.com/containerd/nerdctl/pkg/imgutil/dockerconfigresolver"
 	dockercliconfig "github.com/docker/cli/cli/config"
 	clitypes "github.com/docker/cli/cli/config/types"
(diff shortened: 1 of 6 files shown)


cmd/nerdctl/login.go
```diff
@@ -240,14 +241,31 @@ func loginClientSide(ctx context.Context, cmd *cobra.Command, auth types.AuthCon
...
 	for i, rh := range regHosts {
 		err = tryLoginWithRegHost(ctx, rh)
 		identityToken := fetchedRefreshTokens[rh.Host] // can be empty
 		if err == nil {
 			return identityToken, nil
 		}
+		if insecure && (errutil.IsErrHTTPResponseToHTTPSClient(err) || errutil.IsErrConnectionRefused(err)) {
+			needFallbackToPlainHttp = true
+			needFallBackRegHosts = append(needFallBackRegHosts, rh)
+		}
```
Junnplus, line 255: I reserve my opinion on https://github.com/containerd/nerdctl/pull/894#discussion_r827661658.

When it is not a valid HTTPs request and insecure-registry flaged, we can fallback directly to plain HTTP. cc @fahedouch

```suggestion
                err = tryLoginWithRegHost(ctx, rh)
		if err != nil && insecure && (errutil.IsErrHTTPResponseToHTTPSClient(err) || errutil.IsErrConnectionRefused(err)) {
		    rh.Scheme = "http"
		    err = tryLoginWithReghost(ctx, rh)
		}
		identityToken := fetchedRefreshTokens[rh.Host] // can be empty
		if err == nil {
			return identityToken, nil
		}
```

Zheaoli: OK, make sense

Junnplus (approved):
LGTM. Thanks!

AkihiroSuda (approved):
Thanks

PR #914, opened 2022-03-19
Build with go 1.18

Junnplus:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,6 @@ require (
 	github.com/ipfs/go-ipfs-files v0.1.1
 	github.com/ipfs/go-ipfs-http-client v0.2.0
 	github.com/ipfs/interface-go-ipfs-core v0.5.2
-	github.com/jaytaylor/go-hostsfile v0.0.0-20211120191712-f53f85d8b98f
 	github.com/mattn/go-isatty v0.0.14
 	github.com/moby/sys/mount v0.3.1
 	github.com/multiformats/go-multiaddr v0.5.0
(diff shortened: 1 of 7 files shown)


pkg/dnsutil/hostsstore/hosts.go
```diff
@@ -0,0 +1,49 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package hostsstore
+
+import "strings"
+
+// ParseHosts takes in hosts file content and returns a map of parsed results.
+// Copy from https://github.com/jaytaylor/go-hostsfile/blob/59e7508e09b9e08c57183ae15eabf1b757328ebf/hosts.go#L18
```
AkihiroSuda, line 22: https://github.com/jaytaylor/go-hostsfile/blob/f53f85d8b98f8f4925bceab7890a431e2f46c38f/LICENSE#L12-L13

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containerd/nerdctl
 
-go 1.17
+go 1.18
```
AkihiroSuda, line 3: Probably we should still keep this 1.17 for several months

Junnplus: Is there any reason?

AkihiroSuda: Because users and packagers may take several months to migrate to 1.18

Junnplus: If the user is not using the latest version, I think it will be fine. 

In other words, if users need to use the latest version (maybe > 0.18), they should migrate to 1.18.

AkihiroSuda: containerd still specifies go 1.17 in go.mod, though CI uses Go 1.18 now

https://github.com/containerd/containerd/blob/ed1a762707cea5055dec47ed4b049159ac9f23a3/go.mod#L3

Junnplus: OK, let's be consistent

AkihiroSuda (approved):
Thanks

Issue #759, opened 2022-01-30
Unable to mount volumes or bind mount on Windows


MikeTheSnowman:
# Issue:
When using nerdctl + containerd on Windows, I am unable to mount volumes or bind mount to a new windows container. I've tried using the following command with no success:
```
PS C:\Users\Administrator\nerdctl\_output> nerdctl.exe --debug-full run -d -v C:\bind_volumes\v1:c:\src hello-world:nanoserver-1809
time="2022-01-30T02:04:37-08:00" level=debug msg="verification process skipped"
time="2022-01-30T02:04:37-08:00" level=debug msg="Calling proc (1)" spanID=7599fa456674789d traceID=6d976231e769a37073cadb49b4a21362
time="2022-01-30T02:04:37-08:00" level=debug msg="Calling proc (2)" spanID=7599fa456674789d traceID=6d976231e769a37073cadb49b4a21362
time="2022-01-30T02:04:37-08:00" level=fatal msg="failed to parse \"C:\\\\bind_volumes\\\\v1:c:\\\\src\""
```

I've also tried experimenting with the the `-v` option, by trying various ways of single/double quoting an escaping special characters, but again I did not have any success.

# Code Investigation:
Now, I do not know how to program in Go (so please excuse me if I'm totally wrong here). But starting at line 56 in [mountutil.go](https://github.com/containerd/nerdctl/blob/0506e39ef89b88f74571c7946d191170672aeaa7/pkg/mountutil/mountutil.go#L56), it looks like that whole switch statement is really only designed to handle Linux paths. I say this because on line 56, the value passed to `-v` is being split on the ":" character and the switch statement doesn't look like it would be able to handle bind mount (eg: `-v c:\hostDir:c:\guestDir`) or just mounting volumes to a windows container (eg: `-v vol1:c:\guestDir`). 

Also, line 56's parent function (generateMountOpts), returns back up to [run_mount.go](https://github.com/containerd/nerdctl/blob/0506e39ef89b88f74571c7946d191170672aeaa7/cmd/nerdctl/run_mount.go#L195) on line 195. Just a few lines below that on [line 203](https://github.com/containerd/nerdctl/blob/0506e39ef89b88f74571c7946d191170672aeaa7/cmd/nerdctl/run_mount.go#L203), it looks like the SecureJoin may encounter issues if a full windows path containing a colon character were to be joined with tempDir on that line.

[...]

PR #822, opened 2022-02-16
Add Mounts field for nerdctl inspect

liubin:
This PR adds `Mounts` field for the output of `nerdctl inspect` command.

**To reviewers**: User-specified mounts are stored in labels like anonymous volumes until finding a suitable place to store them.

For example, start a container:

```bash
$ nerdctl run --name test -d \
  -v /anony-vol \
  --tmpfs /app1:size=64m \
  --mount type=bind,src=/tmp,dst=/app2,bind-propagation=rshared,ro \
  --mount type=volume,src=vol-1,dst=/app3,readonly=false \
  containerstack/alpine-stress:latest top
```

And get container info:

```bash
$ nerdctl inspect test
[
    {
        "Id": "384bdb0bededd9b879a3863bf5faea6a5f4714c37b19e6e7258322092e2d8df4",
        "Created": "2022-02-16T14:56:59.645505751Z",
        "Path": "top",
        "Args": null,
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Pid": 16254,
            "ExitCode": 0,
            "FinishedAt": "0001-01-01T00:00:00Z"
        },
        "Image": "docker.io/containerstack/alpine-stress:latest",
        "ResolvConfPath": "/var/lib/nerdctl/1935db59/containers/default/384bdb0bededd9b879a3863bf5faea6a5f4714c37b19e6e7258322092e2d8df4/resolv.conf",
        "HostnamePath": "/var/lib/nerdctl/1935db59/containers/default/384bdb0bededd9b879a3863bf5faea6a5f4714c37b19e6e7258322092e2d8df4/hostname",
        "LogPath": "/var/lib/nerdctl/1935db59/containers/default/384bdb0bededd9b879a3863bf5faea6a5f4714c37b19e6e7258322092e2d8df4/384bdb0bededd9b879a3863bf5faea6a5f4714c37b19e6e7258322092e2d8df4-json.log",
        "Name": "test",
        "Driver": "overlayfs",
        "Platform": "linux",
        "AppArmorProfile": "nerdctl-default",
        "Mounts": [
            {
                "Type": "volume",
                "Name": "46ccfe12b21b888a4707a129202cae7ab0308ca3ef69e79bc38a70154aece3aa",
                "Source": "/var/lib/nerdctl/1935db59/volumes/default/46ccfe12b21b888a4707a129202cae7ab0308ca3ef69e79bc38a70154aece3aa/_data",
                "Destination": "/anony-vol",
[...]

PR #924, opened 2022-03-23
Fixed volume mount issue(759) for Windows

click2cloud-lamda:
### Fixed issue #759 :
- volume binding issue in Windows containers
```
D:\GolandProjects\github.com\nerdctl2\cmd\nerdctl>nerdctl run -it -v C:\ProgramData\nerdctl\052055e3\volumes\default\test\_data:c:\src mcr.microsoft.com/windows/nanoserver:20H2
Microsoft Windows [Version 10.0.19042.1586]
(c) 2019 Microsoft Corporation. All rights reserved.

C:\Windows\system32>cd ../..

C:\>dir
 Volume in drive C has no label.
 Volume Serial Number is 7643-AF54

 Directory of C:\

03/03/2022  01:27 PM             5,510 License.txt
03/16/2022  07:00 PM    <DIR>          src
03/16/2022  07:50 PM    <DIR>          Users
03/16/2022  07:50 PM    <DIR>          Windows
               1 File(s)          5,510 bytes
               3 Dir(s)  21,302,005,760 bytes free

C:\>
 ```
When using nerdctl + containerd on Windows, I am able to mount volumes or bind mount to a new windows container.

Signed-off-by: Raymond Mathew click2cloud-lamda@click2cloud.net
Signed-off-by: Dasara Nagaraju dasara.nagaraju@click2cloud.net

### CODE REVIEW

File: go.mod [hunk 1]
```diff
@@ -118,6 +118,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/mitchellh/mapstructure v1.4.3 // indirect
 	github.com/moby/locker v1.0.1 // indirect
+	github.com/moby/moby v20.10.13+incompatible // indirect
```
Zheaoli, line 121: I'm not sure should we add a full moby dependency here?

pkg/mountutil/mountutil_freebsd.go
```diff
@@ -66,3 +74,65 @@ func ProcessFlagTmpfs(s string) (*Processed, error) {
 func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, error) {
 	return nil, errdefs.ErrNotImplemented
 }
+func ProcessSplit(s string, volStore volumestore.VolumeStore) (Processed, error) {
```
Zheaoli, line 77: I think the `ProcessSpllit` code is as same as the code in `mountutil_linux.go`. FYI https://github.com/containerd/nerdctl/pull/924/files#diff-ab0f47dd8154315baddf1347290f120ed5e9975a660c4bfb838d17dcd2418c50R434

Maybe we can add a new file named `mountutils_otthers.go` and add `//go:build !windows` at the top of the code and  move `ProcessSpllit` here

pkg/mountutil/mountutil.go
```diff
@@ -17,21 +17,12 @@
...
 
 import (
-	"errors"
-	"fmt"
-	"path/filepath"
-	"runtime"
-	"strings"
-
-	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/oci"
```
Zheaoli, line 20: Fix the lint error plz

pkg/mountutil/mountutil_freebsd.go
click2cloud-lamda: Or maybe we can keep the conditions of ProcessSplit in the mountutil.go and keep the windows one as it is.

Zheaoli: Yes, I think it's a better choice

AkihiroSuda:
>  Commits 25

Please remove irrelevant commits and squash your ones.
Just like other PRs: https://github.com/containerd/nerdctl/pulls?q=is%3Apr+is%3Amerged

Zheaoli (commented):
Maybe we got some rebase issue here?

cmd/nerdctl/build.go
```diff
@@ -75,7 +75,7 @@ func newBuildCommand() *cobra.Command {
 }
 
 func getBuildkitHost(cmd *cobra.Command) (string, error) {
-	if cmd.Flags().Changed("buildkit-host") {
+	if cmd.Flags().Changed("buildkit-host") || os.Getenv("BUILDKIT_HOST") != "" {
```
Zheaoli, line 78: IMHO, The piece of code here is may belong to the https://github.com/containerd/nerdctl/pull/932

Maybe we got some rebase issue here?

Zheaoli (commented):
Maybe we got some rebase issue here?

pkg/mountutil/mountutil_linux.go
```diff
@@ -422,3 +422,7 @@ func getTmpfsSize(size int64) string {
 
 	return fmt.Sprintf("size=%d%s", size, suffix)
 }
+func ProcessSplit(s string, volStore volumestore.VolumeStore, res Processed, src string, dst string, options []string) (string, string, []string, error) {
```
AkihiroSuda, line 425: Needs godoc

pkg/mountutil/mountutil_windows.go
```diff
@@ -68,3 +71,81 @@ func ProcessFlagTmpfs(s string) (*Processed, error) {
 func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, error) {
 	return nil, errdefs.ErrNotImplemented
 }
+
+func ProcessSplit(s string, volStore volumestore.VolumeStore, res Processed, src string, dst string, options []string) (string, string, []string, error) {
```
AkihiroSuda, line 75: Needs godoc.
Also, can we have unit tests?

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -106,7 +106,7 @@ type Container struct {
 	// TODO: SizeRw     *int64 `json:",omitempty"`
 	// TODO: SizeRootFs *int64 `json:",omitempty"`
 
-	// TODO: Mounts          []MountPoint
+	Mounts []specs.Mount
```
AkihiroSuda, line 109: This should be `[]MountPont`, not `[]specs.Mount` for Docker compatibility

https://github.com/moby/moby/blob/f0014860c1b3345e1fcc7ed81c491298de2633fb/api/types/types.go#L416-L427
https://github.com/opencontainers/runtime-spec/blob/6969a0a09ab162a574cafcc9ac814e498962c943/specs-go/config.go#L110-L120

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -38,7 +38,9 @@ import (
 	"github.com/containerd/nerdctl/pkg/imgutil"
 	"github.com/containerd/nerdctl/pkg/inspecttypes/native"
 	"github.com/containerd/nerdctl/pkg/labels"
+	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/go-connections/nat"
+	"github.com/moby/moby/api/types"
```
AkihiroSuda, line 43: We do not import this package. Please see other struct definitions and keep the coding style consistent.

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -195,12 +196,26 @@ type NetworkEndpointSettings struct {
 
 // ContainerFromNative instantiates a Docker-compatible Container from containerd-native Container.
 func ContainerFromNative(n *native.Container) (*Container, error) {
+	var mounts *specs.Spec
+	// for unmarshalling mount details
+	err := json.Unmarshal(n.Container.Spec.Value, &mounts)
```
AkihiroSuda, line 201: We already have `specs.Spec`. Please see the `if sp, ok := n.Spec.(*specs.Spec); ok {` line below.

pkg/inspecttypes/dockercompat/dockercompat.go
click2cloud-lamda: Okay.

click2cloud-lamda: What if I used this: https://github.com/moby/moby/blob/master/api/types/types.go#L513

pkg/inspecttypes/dockercompat/dockercompat.go
AkihiroSuda: Addressed in https://github.com/containerd/nerdctl/pull/822

pkg/mountutil/mountutil_linux.go
```diff
@@ -35,14 +35,6 @@ import (
...
 )
 
-/*
-   Portions from https://github.com/moby/moby/blob/v20.10.5/daemon/oci_linux.go
-   Portions from https://github.com/moby/moby/blob/v20.10.5/volume/mounts/linux_parser.go
-   Copyright (C) Docker/Moby authors.
-   Licensed under the Apache License, Version 2.0
-   NOTICE: https://github.com/moby/moby/blob/v20.10.5/NOTICE
-*/
-
```
AkihiroSuda, line 45: Why?

AkihiroSuda:
@jsturtevant Could you take a look?

pkg/mountutil/mountutil_windows_test.go
```diff
@@ -78,3 +78,45 @@ func TestParseVolumeOptions(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.s, func(t *testing.T) {
+			src, dst, _, err := ProcessSplit(tt.s, nil, tt.res, tt.src, tt.dst, tt.options)
+			if err != nil {
+				t.Errorf("failed to split %q: %v", tt.s, err)
+				return
+			}
+			if !strings.Contains(tt.s, src) {
```
AkihiroSuda, line 112: This check doesn't seem meaningful because src is always ""

pkg/mountutil/mountutil_windows_test.go
```diff
@@ -78,3 +78,45 @@ func TestParseVolumeOptions(t *testing.T) {
...
+			src:     "",
+			dst:     "",
+			s:       "C:\\ProgramData\\nerdctl\\test\\volumes\\default\\testVol\\_data:C:\\src",
+			options: []string{},
+		},
+		{
+			res:     Processed{},
+			src:     "",
+			dst:     "",
+			s:       "C:\\ProgramData\\nerdctl\\test\\volumes\\default\\Volmnt:C:\\src",
```
AkihiroSuda, line 101: Can we test the split result? like `expected: []string{"C:\\ProgramData\\nerdctl\\test\\volumes\\default\\Volmnt", "C:\\src"}`

click2cloud-lamda: The test itself is about splitting the string into source and destination for windows.

pkg/mountutil/mountutil_linux.go
click2cloud-lamda: Let me fix this.

pkg/mountutil/mountutil.go
```diff
@@ -55,72 +55,84 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
...
-
-	split := strings.Split(s, ":")
-	switch len(split) {
-	case 1:
-		dst = s
-		res.AnonymousVolume = idgen.GenerateID()
-		logrus.Debugf("creating anonymous volume %q, for %q", res.AnonymousVolume, s)
-		anonVol, err := volStore.Create(res.AnonymousVolume, []string{})
+	if runtime.GOOS == "windows" {
+		sr, ds, optns, err := ProcessSplit(s, volStore, res, src, dst, options)
```
jsturtevant, line 59: Why are adding a check for goos if we are doing the process split logic in os specific files?  it feels strange to have the case statements tucked away for Windows but for other OS (linux) process split is in this file especially since process split doesn't do anything for linux/freebsd and are exactly the same

pkg/mountutil/mountutil_freebsd.go
```diff
@@ -66,3 +66,8 @@ func ProcessFlagTmpfs(s string) (*Processed, error) {
 func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, error) {
 	return nil, errdefs.ErrNotImplemented
 }
+
+func ProcessSplit(s string, volStore volumestore.VolumeStore, res Processed, src string, dst string, options []string) (string, string, []string, error) {
+	var x []string
+	return "", "", x, nil
+}
```
jsturtevant, line 73: freebsd implementation is the same as linux, why do we have it duplicated? (see other comment about the goos check for processsplit in mountutil.go)

cmd/nerdctl/run_mount_windows_test.go
```diff
@@ -0,0 +1,96 @@
...
+	defer base.Cmd("rm", "-f", containerName).Run()
+	base.Cmd("run",
+		"-d",
+		"--name", containerName,
+		"-v", fmt.Sprintf("%s:C:\\mnt1", rwDir),
+		"-v", fmt.Sprintf("%s:C:\\mnt2ro", roDir),
+		"-v", fmt.Sprintf("%s:C:\\mnt3", rwVolName),
+		"-v", fmt.Sprintf("%s:C:\\mnt4ro", roVolName),
+		imageName).AssertOK()
+	base.Cmd("container", "inspect", containerName).AssertOK()
```
jsturtevant, line 57: this asserts that the container created, do we also want to make sure that the volume is actually there (maybe with something in it)?  How is this tested on Linux?

jsturtevant (commented):
sorry for the delay, been on PTO for a couple weeks.

pkg/mountutil/mountutil.go
click2cloud-lamda: I will just keep the process split logic to respective os files and remove the os logic.

cmd/nerdctl/run_mount_windows_test.go
click2cloud-lamda: TestRunAnonymousVolume and  TestRunNamedVolume checks whether the volume is mounted or not. For linux, we havent added any test as the problem was with the windows part.

jsturtevant: I guess we some coverage that the volume is created. Another thing noticed now, is it seem like we are also testing read/write behavior here. I would think we want to verify that the in the test.

click2cloud-lamda: For verifying the read/write behavior, inspect should print the details regarding it. But in inspect for windows, it doesn't work.

```
D:\GolandProjects\github.com\nerdctlmst\cmd\nerdctl>nerdctl inspect 1f44fa83de
[
    {
        "Id": "1f44fa83de31867ec49d2639a15f4d04fa6003765660afb9e50c51e887e89815",
        "Created": "2022-04-28T11:03:48.0466636Z",
        "Path": "c:\\windows\\system32\\cmd.exe",
        "Args": null,
        "State": {
            "Status": "exited",
            "Running": false,
            "Paused": false,
            "Pid": 25496,
            "ExitCode": 0,
            "FinishedAt": "2022-04-28T11:03:50.2407926Z"
        },
        "Image": "mcr.microsoft.com/windows/nanoserver:20H2",
        "ResolvConfPath": "",
        "HostnamePath": "",
        "LogPath": "",
        "Name": "nanoserver-1f44f",
        "Driver": "windows",
        "Platform": "windows",
        "AppArmorProfile": "",
        "Mounts": [
            {
                "Source": "C:\\ProgramData\\nerdctl\\052055e3\\volumes\\default\\testVolume\\_data",
                "Destination": "C:\\src",
                "Mode": "",
                "RW": true,
                "Propagation": ""
            }
        ],
        "NetworkSettings": {
            "GlobalIPv6Address": "",
            "GlobalIPv6PrefixLen": 0,
            "IPAddress": "",
            "IPPrefixLen": 0,
            "MacAddress": "",
            "Networks": {}
        }
    }
]
```

click2cloud-lamda: This is the output Im getting.

jsturtevant: So the inspect output for the volume is there but the rw/ro info isn't valid?  Does the actual directory in the container have the correct attributes?   If it is just inspect that isn't return properly lets create an issue to resolve later.

click2cloud-lamda: This issue is only with windows where I cant get the output for read/write behavior.

pkg/mountutil/mountutil_freebsd.go
```diff
@@ -66,3 +69,70 @@ func ProcessFlagTmpfs(s string) (*Processed, error) {
 func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, error) {
 	return nil, errdefs.ErrNotImplemented
 }
+
+// ProcessSplit splits volume mount information received through command line into respective source, destination
+// and optsRaw fields according to the volume type chosen and returns them accordingly.
+func ProcessSplit(s string, volStore volumestore.VolumeStore, res *Processed, src string, dst string, options []string) (string, string, []string, error) {
```
jsturtevant, line 75: This function is exactly the same for freebsd and linux.  Can we reduce this so the code isn't copy pasted in two places?

pkg/mountutil/mountutil.go
```diff
@@ -54,77 +46,17 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
...
+		return nil, err
 	}
-
-	fstype := "nullfs"
 	if runtime.GOOS != "freebsd" {
-		fstype = "none"
 		options = append(options, "rbind")
 	}
 	res.Mount = specs.Mount{
-		Type:        fstype,
```
jsturtevant, line 127: Why is this being dropped?  I don't see it moved anywhere.

click2cloud-lamda: Because of this the mounted type during inspect was null. Now I get this output: 

```
D:\GolandProjects\github.com\nerdctl\cmd\nerdctl>nerdctl inspect 3c46c12ed1f1
[
    {
        "Id": "3c46c12ed1f1f6272b4ab3f93996a19e2e52667313c2a2cbbf94f87b08350dad",
        "Created": "2022-05-06T09:54:49.1905099Z",
        "Path": "c:\\windows\\system32\\cmd.exe",
        "Args": null,
        "State": {
            "Status": "exited",
            "Running": false,
            "Paused": false,
            "Restarting": false,
            "Pid": 10928,
            "ExitCode": 0,
            "FinishedAt": "2022-05-06T09:54:51.2735242Z"
        },
        "Image": "mcr.microsoft.com/windows/nanoserver:20H2",
        "ResolvConfPath": "",
        "HostnamePath": "",
        "LogPath": "",
        "Name": "nanoserver-3c46c",
        "RestartCount": 0,
        "Driver": "windows",
        "Platform": "windows",
        "AppArmorProfile": "",
        "Mounts": [
            {
                "Type": "volume",
                "Name": "500907e8f1465b04f132f5a49b76c59c1e41f18aa9063faffa5c2bb7fc2e72a5",
                "Source": "C:\\ProgramData\\nerdctl\\052055e3\\volumes\\default\\500907e8f1465b04f132f5a49b76c59c1e41f18aa9063faffa5c2bb7fc2e72a5\\_data",
                "Destination": "C:\\src",
                "Driver": "local",
                "Mode": "",
                "RW": true,
                "Propagation": ""
            }
        ],
        "NetworkSettings": {
            "GlobalIPv6Address": "",
            "GlobalIPv6PrefixLen": 0,
            "IPAddress": "",
            "IPPrefixLen": 0,
            "MacAddress": "",
            "Networks": {}
        }
    }
]
```
Please check.

pkg/mountutil/mountutil_freebsd.go
click2cloud-lamda: Okay

pkg/mountutil/mountutil.go
jsturtevant: is only printing `null` an issue on Windows?  I am concerned this change affects something on freebsd or linux.  The tests aren't failing but maybe we are missing coverage here

pkg/mountutil/process_split.go
```diff
@@ -0,0 +1,100 @@
+//go:build freebsd || linux
```
jsturtevant, line 1: This file should be named `mountutil_others.go` or `mountutil_linux.go` to stay consistent that this a function that is specific to runtimes related to mountutil.

click2cloud-lamda: If we pass "none" as fstype for windows, it gives this error: "time="2022-05-12T12:13:13+05:30" level=fatal msg="invalid OCI spec - Type 'none' not supported: unknown"".

click2cloud-lamda: why are we passing nullfs or none, when we are already passing mount types like bind, volume or tmps while splitting the source and destination.

pkg/mountutil/mountutil.go
jsturtevant: > If we pass "none" as fstype for windows, it gives this error: "time="2022-05-12T12:13:13+05:30" level=fatal msg="invalid OCI spec - Type 'none' not supported: unknown"".


> why are we passing nullfs or none, when we are already passing mount types like bind, volume or tmps while splitting the source and destination.

I don't know why none/nullfs is passed here.  @AkihiroSuda do you have any insight?

pkg/mountutil/process_split.go
jsturtevant: These comments look to have landed on the wrong thread. I've responded here: https://github.com/containerd/nerdctl/pull/924#discussion_r871565970

pkg/mountutil/mountutil.go
```diff
@@ -54,73 +46,14 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
...
-		var err error
-		var specOpts []oci.SpecOpts
-		options, specOpts, err = parseVolumeOptions(res.Type, src, rawOpts)
-		if err != nil {
-			return nil, err
-		}
-		res.Opts = append(res.Opts, specOpts...)
-	default:
-		return nil, fmt.Errorf("failed to parse %q", s)
+	src, dst, options, err = ProcessSplit(s, volStore, &res, src, dst, options)
```
TBBle, line 52: Looking at the implementations, `src`, `dst`, and `options` are never consumed, so should probably not be parameters, but local vars in `ProcessSplit`. That would allow simplifying this call slightly to `src, dst, options, err := ProcessSplit(s, volStore, &res)`, trimming the top-of-function `var` lines, rather than growing them.

Also, does it need to be public? `processSplit` seems like it should work, all uses are in pkg/mountutil AFAICT, and it seems slightly safer to hide platform-split functions to control the blast radius in case of future divergence.

apostasie:
@click2cloud-lamda @jsturtevant this got replaced by #2661, right?

Should we close this?

PR #936, opened 2022-03-25
Add check of snapshotter when sharing images

ktock:
Following up https://github.com/containerd/nerdctl/pull/858#pullrequestreview-921330268

### DIFF
--- a/cmd/nerdctl/build.go
+++ b/cmd/nerdctl/build.go
@@ -93,7 +93,7 @@ func getBuildkitHost(cmd *cobra.Command) (string, error) {
 	return buildkitutil.GetBuildkitHost(ns)
 }
 
-func isImageSharable(buildkitHost string, namespace, uuid string) (bool, error) {
+func isImageSharable(buildkitHost string, namespace, uuid, snapshotter string) (bool, error) {
 	labels, err := buildkitutil.GetWorkerLabels(buildkitHost)
 	if err != nil {
 		return false, err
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/build.go
```diff
@@ -111,7 +111,11 @@ func isImageSharable(buildkitHost string, namespace, uuid string) (bool, error)
 	if !ok {
 		return false, nil
 	}
-	return executor == "containerd" && containerdUUID == uuid && containerdNamespace == namespace, nil
+	containerdSnapshotter, ok := labels["org.mobyproject.buildkit.worker.snapshotter"]
```
AkihiroSuda, line 114:
```suggestion
	workerSnapshotter, ok := labels["org.mobyproject.buildkit.worker.snapshotter"]
```

cmd/nerdctl/build.go
```diff
@@ -103,15 +103,19 @@ func isImageSharable(buildkitHost string, namespace, uuid string) (bool, error)
 	if !ok {
 		return false, nil
 	}
-	containerdUUID, ok := labels["org.mobyproject.buildkit.worker.containerd.uuid"]
+	workerUUID, ok := labels["org.mobyproject.buildkit.worker.containerd.uuid"]
```
AkihiroSuda, line 106: This should be still containerdUUID because it is under `o.m.b.w.containerd`.

OTOH, workerSnapshotter shouldn't be containerdSnapshotter because it is not under `o.m.b.w.containerd`.

AkihiroSuda (approved):
Thanks

Issue #492, opened 2021-11-02
`nerdctl start` cannot start containers in **Created** or **Exited** status 


jls-0:
**Required behavior:** Allow start Created or Exited containers, either with `nerdctl start` or `nerdctl exec -it`.
**Actual behavior:**
you can run a container, for example: `nerdctl run -it nginx:latest --name nginx-container`.
You can stop it: `nerdctl stop nginx-container`.
**Issue:** 
You cannot start again the container neither `nerdctl start nginx-container` nor `nerdctl exec -it nginx-container /bin/sh`.
**Error messages:**
`nerdctl start nginx-container`:
`FATA[0000] OCI runtime create failed: cannot allocate tty if runc will detach without setting console socket: unknown`
`nerdctl exec -it nginx-container /bin/sh`:
`FATA[0000] no running task found: task 0a88555b90488894c72918665b9b14bd6ae96e89676d65ca772b9abf34727884 not found: not found`
**Environment:**
  - OS: MacOs Big Sur 11.0.1.
  - Backend: Rancher Desktop 0.6.0 (nerdctl 0.12.1).

AkihiroSuda:
-it is currently incomparable with stop/start, please use -d

jls-0:
> -it is currently incomparable with stop/start, please use -d

It didn't work, I created and ran the container with `nerdctl run --name my-container **-d** nginx`, even modified the image to set the ENTRYPOINT to `/bin/bash`; the container exited immediately because there's no interactive nor tty attached, so there's no process to maintain running in background.
Then, I tried to start or exec the exited container but got the following output:
`nerdctl start my-container`: started the container but exited immediately due to  the mentioned above.
`nerdctl exec -it my-container /bin/sh`: `FATA[0000] cannot exec in a stopped state: unknown`, which is logical because we cannot execute command on stopped containers.

jls-0:
I think in the meantime it cannot be achieve until the implementation of the `create` action from docker, because it allows us to define the `-it` parameters at creation time, then we can start the container in background allowing the interactive/tty connection for later.

pvbiesen:
+1 migrating from docker, very annoying `nerdctl start -i <container> /bin/bash` doesn't work.

animatedcreativity:
Yes, cannot start stopped containers.

raktim-sixt:
yes, same here...cannot start the container which is in the `CREATED` state

[...]

PR #950, opened 2022-03-29
fix can't start with -it flag

ningmingxiao:

### DIFF
--- a/cmd/nerdctl/start.go
+++ b/cmd/nerdctl/start.go
@@ -18,8 +18,14 @@ package main
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
+
+	"github.com/containerd/console"
+	"github.com/containerd/typeurl"
+	"github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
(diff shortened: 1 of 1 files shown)


ningmingxiao:
fix https://github.com/containerd/nerdctl/issues/492

AkihiroSuda:
`docker start` doesn't support tty

```
$ docker start --help

Usage:  docker start [OPTIONS] CONTAINER [CONTAINER...]

Start one or more stopped containers

Options:
  -a, --attach               Attach STDOUT/STDERR and forward signals
      --detach-keys string   Override the key sequence for detaching a container
  -i, --interactive          Attach container's STDIN
```

cmd/nerdctl/start.go
```diff
@@ -95,13 +101,50 @@ func startContainer(ctx context.Context, container containerd.Container) error {
...
+		con = console.Current()
+		defer con.Reset()
+		if err := con.SetRaw(); err != nil {
+			return fmt.Errorf("failed to SetRaw,err is: %w", err)
+		}
+		if con == nil {
+			return errors.New("got nil con with flagT=true")
+		}
+		taskCIO = cio.NewCreator(cio.WithStreams(con, con, nil), cio.WithTerminal)
+	}
```
AkihiroSuda, line 119: For consistency with Docker, we have to detach tty

ningmingxiao: what should I do?   change from   `return errors.New("got nil con with flagT=true")` to `return errors.New("got nil con")`?

AkihiroSuda: Probably you should attach a dummy terminal, not the real terminal, to emulate Docker's behavior.
Needs further investigation.

fahedouch:
Fixed by https://github.com/containerd/nerdctl/pull/1642 ?

ningmingxiao:
> Fixed by #1642 ? @ningmingxiao cc
yes

PR #952, opened 2022-03-29
build: fix `content digest ... not found` for built multi-platform image

### LINKED ISSUE #951: `content digest ... not found` for multi-platform image built using `nerdctl build`
### Description

This ocucrs nerdctl with BuildKit containerd worker configuration.
When `nerdctl build` builds a multi-platform image without `RUN` or `COPY` command, nerdctl cannot use some of the blobs of the result image of non-default platform.

Example:

```console
# mkdir -p /tmp/ctx && cat <<EOF > /tmp/ctx/Dockerfile
FROM ubuntu:20.04
CMD echo hello
EOF
# nerdctl build --platform=linux/amd64,linux/arm64 -t foo /tmp/ctx
# nerdctl save --platform=linux/arm64 foo > /dev/null
FATA[0000] failed to get reader: content digest sha256:57d0418fe9dcc5262d8c4fcb06c852ad2d0407b541c64d0f5f2e6ec01fd411ba: not found 
```

This can occur on any snapshotters.

This is because nerdctl exports the result image with `--output=type=image,unpack=true`.
With this option, BuildKit doesn't download blobs of the non-default platform base image.

In the above case, BuildKit only downloads `linux/amd64` of `ubuntu:20.04` but `linux/arm64` isn't downloaded.

To solve this, when `nerdctl build` builds multi-platform image, it should use `type=docker` or `type=oci` until BuildKit supports an export option to forcibly download all blobs of all platforms.

Please see above example

I expect content found but get content not found.

```
nerdctl version
Client:
 Version:	v0.17.0-114-g987c414
 Git commit:	987c41446172b67de017635873163d571a6a85ab

Server:
 containerd:
  Version:	v1.6.0-beta.2
  GitCommit:	bd81f8a39d85bd7275757c719e08e69a5e997132
```

BuildKit v0.10.0

_No response_

### Host information

```
nerdctl info
Client:
 Namespace:	default
 Debug Mode:	false

Server:
 Server Version: v1.6.0-beta.2
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Log: json-file
  Storage: aufs native overlayfs stargz
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.13.0-35-generic
 Operating System: Debian GNU/Linux 11 (bullseye)
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.3GiB
 Name: 90579471cf9f
 ID: 08dc035e-a79f-4964-98fe-12cc58322846

```

ktock:
Fixes  #951

### DIFF
--- a/cmd/nerdctl/build.go
+++ b/cmd/nerdctl/build.go
@@ -93,7 +93,7 @@ func getBuildkitHost(cmd *cobra.Command) (string, error) {
 	return buildkitutil.GetBuildkitHost(ns)
 }
 
-func isImageSharable(buildkitHost string, namespace, uuid, snapshotter string) (bool, error) {
+func isImageSharable(buildkitHost string, namespace, uuid, snapshotter string, platform []string) (bool, error) {
 	labels, err := buildkitutil.GetWorkerLabels(buildkitHost)
 	if err != nil {
 		return false, err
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Probably we need to document the `isImageSharable` condition in `docs/build.md` too

PR #964, opened 2022-04-05
inspect: add type flag to inspect command

89luca89:
Right now nerdctl inspect does not have a --type flag similar to
docker and podman.

This PR aims to add this, allowing for the basic values:

- container
- image

It will default to container to match docker's behaviour

Example of the command:

```sh
./nerdctl inspect --format '{{json .}}' --type image registry.fedoraproject.org/fedora-toolbox:35
./nerdctl inspect --format '{{json .}}' --type container fedora-toolbox-35
```

Signed-off-by: Luca Di Maio <luca.dimaio1@gmail.com>

### DIFF
--- a/cmd/nerdctl/inspect.go
+++ b/cmd/nerdctl/inspect.go
@@ -42,10 +42,19 @@ func newInspectCommand() *cobra.Command {
 	return inspectCommand
 }
 
+var validInspectType = map[string]bool{
+	"container": true,
+	"image":     true,
+}
+
 func addInspectFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP("format", "f", "", "Format the output using the given Go template, e.g, '{{json .}}'")
 	cmd.RegisterFlagCompletionFunc("format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"json"}, cobra.ShellCompDirectiveNoFileComp
+	})
+	cmd.Flags().String("type", "", "Return JSON for specified type")
+	cmd.RegisterFlagCompletionFunc("type", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"image", "container", ""}, cobra.ShellCompDirectiveNoFileComp
 	})
 	cmd.Flags().String("mode", "dockercompat", `Inspect mode, "dockercompat" for Docker-compatible output, "native" for containerd-native output`)
 	cmd.RegisterFlagCompletionFunc("mode", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
@@ -54,6 +63,15 @@ func addInspectFlags(cmd *cobra.Command) {
 }
 
 func inspectAction(cmd *cobra.Command, args []string) error {
+	inspectType, err := cmd.Flags().GetString("type")
+	if err != nil {
+		return err
+	}
+
+	if len(inspectType) > 0 && !validInspectType[inspectType] {
+		return fmt.Errorf("%q is not a valid value for --type", inspectType)
+	}
+
 	client, ctx, cancel, err := newClient(cmd)
 	if err != nil {
 		return err
@@ -76,31 +94,37 @@ func inspectAction(cmd *cobra.Command, args []string) error {
 
 	var errs []error
 	for _, req := range args {
-		ni, err := imagewalker.Walk(ctx, req)
-		if err != nil {
-			return err
+		var ni int
+		var nc int
+
+		if len(inspectType) == 0 || inspectType == "image" {
+			ni, err = imagewalker.Walk(ctx, req)
+			if err != nil {
+				return err
+			}
 		}
 
-		nc, err := containerwalker.Walk(ctx, req)
-		if err != nil {
-			return err
-		}
-
-		if ni != 0 && nc != 0 || nc > 1 {
-			return fmt.Errorf("multiple IDs found with provided prefix: %s", req)
+		if len(inspectType) == 0 || inspectType == "container" {
+			nc, err = containerwalker.Walk(ctx, req)
+			if err != nil {
+				return err
+			}
 		}
 
 		if nc == 0 && ni == 0 {
 			return fmt.Errorf("no such object %s", req)
 		}
+		if nc != 0 {
+			if err := containerInspectAction(cmd, []string{req}); err != nil {
+				errs = append(errs, err)
+			}
+			continue
+		}
 		if ni != 0 {
 			if err := imageInspectActionWithPlatform(cmd, []string{req}, ""); err != nil {
 				errs = append(errs, err)
 			}
-		} else {
-			if err := containerInspectAction(cmd, []string{req}); err != nil {
-				errs = append(errs, err)
-			}
+			continue
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Thanks, but please use real name for `Signed-off-by` line

cmd/nerdctl/inspect.go
```diff
@@ -42,18 +42,35 @@ func newInspectCommand() *cobra.Command {
...
+	"image":     true,
+	"all":       true,
+}
+
 func addInspectFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP("format", "f", "", "Format the output using the given Go template, e.g, '{{json .}}'")
 	cmd.RegisterFlagCompletionFunc("format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"json"}, cobra.ShellCompDirectiveNoFileComp
 	})
+	cmd.Flags().String("type", "all", "Specify inspect-object type ('image', 'container' or 'all') (default 'all')")
```
AkihiroSuda, line 56: The default should be `container` for Docker compatibility

cmd/nerdctl/inspect.go
```diff
@@ -42,18 +42,35 @@ func newInspectCommand() *cobra.Command {
 	return inspectCommand
 }
 
+var validInspectType = map[string]bool{
+	"container": true,
+	"image":     true,
+	"all":       true,
```
AkihiroSuda, line 48: Not sure we want `all`. Not present in Docker.

cmd/nerdctl/inspect.go
AkihiroSuda, line 56: Could you update the Command Reference in `README.md` too?

cmd/nerdctl/inspect.go
89luca89: I thought the same, but this way we also preserve the way nerdctl works currently, what do you think?
Will obviously fix the previous comment, defaulting to `container`

AkihiroSuda: I think the current way was wrong

89luca89: No problem then, I'll remove the "all" option :+1:

cmd/nerdctl/inspect.go
89luca89: Done :+1:

89luca89: Done :+1:

AkihiroSuda:
Please squash commits

cmd/nerdctl/inspect.go
```diff
@@ -42,18 +42,34 @@ func newInspectCommand() *cobra.Command {
...
+	"container": true,
+	"image":     true,
+}
+
 func addInspectFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP("format", "f", "", "Format the output using the given Go template, e.g, '{{json .}}'")
 	cmd.RegisterFlagCompletionFunc("format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"json"}, cobra.ShellCompDirectiveNoFileComp
 	})
+	cmd.Flags().String("type", "container", "Specify inspect-object type ('image' or 'container') (default 'container')")
```
AkihiroSuda, line 55: You don't need to add `(default 'container')` here. The default help text is automatically generated by cobra.

89luca89: Done

89luca89:
Ok I now get what's happening
before with the "all" it would just do image or container, so if unspecified, it would find either of them
now it's either container or you **must** specify image, so that would make test fail

Docker works by default in the "all" style, so probably will need to find another way to put the logic in `inspectAction`

AkihiroSuda:
> Docker works by default in the "all" style

Slightly different from the "all" style.

When both the container and the image exists, Docker only shows the container

```console
$ docker run -d --name nginx nginx
58f268d9a42c0bd861f95d75c396e271be218191e0d94eb3575d58fac26b27f2

$ docker inspect nginx
[
    {
        "Id": "58f268d9a42c0bd861f95d75c396e271be218191e0d94eb3575d58fac26b27f2",
        "Created": "2022-04-06T09:39:48.330069568Z",
        "Path": "/docker-entrypoint.sh",
        "Args": [
            "nginx",
            "-g",
            "daemon off;"
        ],
        "State": {
            "Status": "running",
...
            "MacAddress": "02:42:ac:11:00:02",
            "Networks": {
                "bridge": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "bd938a276106bfa3b7b791976dbc16b6f88520ae7c712ed39b1e3c176a59450f",
                    "EndpointID": "18d9f5f42293017de4db6105b8609703ed9a235ec2fba11486ce9540e9794010",
                    "Gateway": "172.17.0.1",
                    "IPAddress": "172.17.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:42:ac:11:00:02",
                    "DriverOpts": null
                }
            }
        }
    }
]
```

89luca89:
> > Docker works by default in the "all" style
> 
> Slightly different from the "all" style.
> 
> When both the container and the image exists, Docker only shows the container
> 
> ```
> $ docker run -d --name nginx nginx
> 58f268d9a42c0bd861f95d75c396e271be218191e0d94eb3575d58fac26b27f2
> 
> $ docker inspect nginx
> [
>     {
>         "Id": "58f268d9a42c0bd861f95d75c396e271be218191e0d94eb3575d58fac26b27f2",
>         "Created": "2022-04-06T09:39:48.330069568Z",
>         "Path": "/docker-entrypoint.sh",
>         "Args": [
>             "nginx",
>             "-g",
>             "daemon off;"
>         ],
>         "State": {
>             "Status": "running",
> ...
>             "MacAddress": "02:42:ac:11:00:02",
>             "Networks": {
>                 "bridge": {
>                     "IPAMConfig": null,
>                     "Links": null,
>                     "Aliases": null,
>                     "NetworkID": "bd938a276106bfa3b7b791976dbc16b6f88520ae7c712ed39b1e3c176a59450f",
>                     "EndpointID": "18d9f5f42293017de4db6105b8609703ed9a235ec2fba11486ce9540e9794010",
>                     "Gateway": "172.17.0.1",
>                     "IPAddress": "172.17.0.2",
>                     "IPPrefixLen": 16,
>                     "IPv6Gateway": "",
>                     "GlobalIPv6Address": "",
>                     "GlobalIPv6PrefixLen": 0,
>                     "MacAddress": "02:42:ac:11:00:02",
>                     "DriverOpts": null
>                 }
>             }
>         }
>     }
> ]
> ```

I see, I can revert to the "all" flag (maybe call it "any"?) and make it behave this way by default

AkihiroSuda:
> I see, I can revert to the "all" flag (maybe call it "any"?) and make it behave this way by default

Should be just "" (empty)

89luca89:
Ok implemented with empty string and inverted the logic in the last check, so that we prioritize containers over images

AkihiroSuda:
```console
$ nerdctl run -d --name nginx nginx
55185a099977d453b24e5197bae31cfb828a5e3ea5712ea0afef60b7df4a3bc1
$ nerdctl inspect nginx
FATA[0000] multiple IDs found with provided prefix: nginx 
```

OTOH, Docker shows the information of the container.

AkihiroSuda (approved):
Thanks

PR #977, opened 2022-04-12
seccomp: relax restrictions depending on --cap-add

### LINKED ISSUE #976: --cap-add not working in run command
### Description

`--cap-add SYS_PTRACE ` not working as expected.

`pspec.Capabilities.Bounding = []string{"CAP_SYS_PTRACE"}` <- this options should be in code. 

```
nerdctl run --rm --cap-add SYS_PTRACE -it ubuntu bash
$ apt install -y gdb
$ gdb /bin/bash
(gdb) r
Starting program: /usr/bin/bash
warning: Error disabling address space randomization: Operation not permitted
warning: Could not trace the inferior process.
warning: ptrace: Operation not permitted
During startup program exited with code 127.
```

`warning: ptrace: Operation not permitted`

0.18.0

_No response_

### Host information

limactl = qemu-system_x86_64 = ubuntu impish 21.10

AkihiroSuda:
Fix #976

### DIFF
--- a/cmd/nerdctl/run_linux.go
+++ b/cmd/nerdctl/run_linux.go
@@ -72,26 +72,9 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
 		opts = append(opts, cgOpts...)
 	}
 
-	securityOpt, err := cmd.Flags().GetStringArray("security-opt")
-	if err != nil {
-		return nil, err
-	}
-	securityOptsMaps := strutil.ConvertKVStringsToMap(strutil.DedupeStrSlice(securityOpt))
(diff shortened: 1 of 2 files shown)


ktock (approved):

disconnect3d:
@AkihiroSuda out of curiosity, how/why does those changes fix the issue of ptrace failing due to seccomp when we only changed the order of applying certain opts?

AkihiroSuda:
> @AkihiroSuda out of curiosity, how/why does those changes fix the issue of ptrace failing due to seccomp when we only changed the order of applying certain opts?

Because the seccomp profile generator depends on `.Process.Capabilities.Bounding` .
https://github.com/containerd/containerd/blob/eaf286224b0144dd11c0e68c131d5dd9ebf52a23/contrib/seccomp/seccomp_default.go#L524

PR #1014, opened 2022-04-27
Fix `flag -t needs -i to be specified together` restriction

AkihiroSuda:
Now `nerdctl run -i` can be executed without `-t`, and vice versa.
Same applies to `nerdctl exec` too.

Replaces https://github.com/containerd/nerdctl/pull/726


pkg/taskutil/taskutil.go
```diff
@@ -49,6 +62,21 @@ func NewTask(ctx context.Context, client *containerd.Client, container container
 	} else {
 		var in io.Reader
 		if flagI {
+			if sv, err := infoutil.ServerSemVer(ctx, client); err != nil {
+				logrus.Warn(err)
+			} else if sv.LessThan(semver.MustParse("1.6.0-0")) {
+				return nil, fmt.Errorf("containerd v1.6 or later is required for using `nerdctl (run|exec) -i` without `-t`: got containerd %v", sv)
```
junnplus, line 68: > It seems that `run -i` alone in 1.5.10 doesn't hang, may be that other tests trigger the hang.
> 
> Why not just skip the tests with contianerd 1.5.10, we can comment that it is a known issue.

Can we just warn the user?

AkihiroSuda: fixed

junnplus (approved):
LGTM on green.

ktock (approved):

Issue #1039, opened 2022-05-06
Logging drivers enhancement 


fahedouch:
### What is the problem you're trying to solve

Target Nerdctl logging drivers:

- [ ] none
- [x] json-file : https://github.com/containerd/nerdctl/pull/959
- [ ] local
- [x] journald : https://github.com/containerd/nerdctl/pull/1062
- [x] fluentd : https://github.com/containerd/nerdctl/pull/1073
- [ ] fluentbit
- [ ] syslog

### Describe the solution you'd like

Configure logging drivers

Zheaoli:
Maybe I can handle the rest of the task in this issue. @fahedouch would you mind this issue to me?

fahedouch:
thank you for  you request  @Zheaoli . I start writing code for `journald` so let me finish it based on your refacto https://github.com/containerd/nerdctl/pull/1050 . You can enhance the rest of the list

Zheaoli:
@fahedouch Great! Maybe I can write the `fluentd` and `syslog` first. BTW, maybe we need spilt the logger in two parts: the readable logger and the unreadable logger

fahedouch:
@Zheaoli  What do you mean by readable logger and the unreadable logger here ?

Zheaoli:
Yes, because some of the loggers such as `fluentd` can not read the log directly after we wrote the log. So I think maybe we need to split the logger by two types

AkihiroSuda:
Probably we should support using multiple drivers simultaneously. Then we can read/write jsonfiles while writing to fluents.

Zheaoli:
> Probably we should support using multiple drivers simultaneously. Then we can read/write jsonfiles while writing to fluentd.

SGTM.

fahedouch:
> Probably we should support using multiple drivers simultaneously. Then we can read/write jsonfiles while writing to fluentd.

We are creating a dependency between drivers here; `json-file` driver issue will impact `fluentd` driver from nerdctl point of view. This can be voluntary when the user configures jsonfiles as fluentd  `tail` input (https://docs.fluentd.org/input/tail) 
I think we should implement `fluentd` driver in a separated way ( we can implement one of the fluentd input https://docs.fluentd.org/input/tcp )

fahedouch:
@Zheaoli FYI I start working on `local` driver.

Zheaoli:
> @Zheaoli implement the above list is enough at least for the moment

SGTM

AkihiroSuda:
[...]

PR #959, opened 2022-04-03
implement log opt with containerd logging driver

fahedouch:
- using containerd runtime v2 logging plugin mode.
- add log-opt for run subcmd 

Signed-off-by: fahed dorgaa <fahed.dorgaa@gmail.com>

PR #1050, opened 2022-05-10
Refactor the logging module and clean some code

Zheaoli:

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -775,6 +775,7 @@ func generateLogURI(dataStore, logDriver string, logOptMap map[string]string) (*
 	for k, v := range logOptMap {
 		args[k] = v
 	}
+	args[logging.DriverType] = logDriver
 	if runtime.GOOS == "windows" {
 		return nil, nil
 	}
--- a/pkg/logging/json_logger.go
+++ b/pkg/logging/json_logger.go
@@ -1,0 +1,92 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package logging
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/containerd/containerd/runtime/v2/logging"
+	"github.com/containerd/nerdctl/pkg/logging/jsonfile"
+	"github.com/docker/go-units"
+	"github.com/natefinch/lumberjack"
+)
+
+type JsonLogger struct {
+	capVal  int64
+	maxFile int
+}
+
+func (jsonLogger *JsonLogger) init(argsMap map[string]string) error {
+	if argsMap[MagicArgv1] == "" {
+		return errors.New("got empty data store")
+	}
+	var capVal int64
+	capVal = -1
+	if capacity, ok := argsMap[MaxSize]; ok {
+		var err error
+		capVal, err = units.FromHumanSize(capacity)
+		if err != nil {
+			return err
+		}
+		if capVal <= 0 {
+			return fmt.Errorf("max-size must be a positive number")
+		}
+	}
+	jsonLogger.capVal = capVal
+	maxFile := 1
+	if maxFileString, ok := argsMap[MaxFile]; ok {
+		var err error
+		maxFile, err = strconv.Atoi(maxFileString)
+		if err != nil {
+			return err
+		}
+		if maxFile < 1 {
+			return fmt.Errorf("max-file cannot be less than 1")
+		}
+	}
+	jsonLogger.maxFile = maxFile - 1
+	return nil
+}
+
+func (jsonLogger *JsonLogger) getLoggerFunc(argsMap map[string]string) logging.LoggerFunc {
+	return func(_ context.Context, config *logging.Config, ready func() error) error {
+		if config.Namespace == "" || config.ID == "" {
+			return errors.New("got invalid config")
+		}
+		logJSONFilePath := jsonfile.Path(argsMap[MagicArgv1], config.Namespace, config.ID)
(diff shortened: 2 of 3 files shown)


pkg/logging/logging.go
```diff
@@ -17,83 +17,47 @@
...
 //
 // Should be called only if argv1 == MagicArgv1.
 func Main(argsMap map[string]string) error {
-	fn, err := getLoggerFunc(argsMap)
+	if _, ok := argsMap[DriverType]; !ok {
+		return errors.New("the driverType is empty")
+	}
+	var driver Interface
+	// TODO: Impl more drivers such as local, journald, fluentd
+	if argsMap[DriverType] == "json-file" {
```
fahedouch, line 48: preparing to other drivers, it more humain friendly if we use a switch case here

pkg/logging/json_logger.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		if maxFile < 1 {
+			return fmt.Errorf("max-file cannot be less than 1")
+		}
+	}
+	jsonLogger.maxFile = maxFile - 1
+	return nil
+}
+
+func (jsonLogger *JsonLogger) getLoggerFunc(argsMap map[string]string) (logging.LoggerFunc, error) {
```
junnplus, line 70: Why not just return `logging.LoggerFunc`?

pkg/logging/logging.go
```diff
@@ -17,83 +17,55 @@
...
+	getLoggerFunc(argsMap map[string]string) (logging.LoggerFunc, error)
+}
+
 // Main is the entrypoint for the containerd runtime v2 logging plugin mode.
 //
 // Should be called only if argv1 == MagicArgv1.
 func Main(argsMap map[string]string) error {
-	fn, err := getLoggerFunc(argsMap)
+	if _, ok := argsMap[DriverType]; !ok {
+		return errors.New("the driverType is empty")
```
fahedouch, line 48: no need for this;  driver in never, it has a default value `json-file`

pkg/logging/logging.go
```diff
@@ -17,83 +17,55 @@
...
+	driverType := argsMap[DriverType]
+	switch driverType {
+	case jsonFile:
+		driver = &JsonLogger{}
+	default:
+		driver = nil
+	}
+	if driver == nil {
+		return fmt.Errorf("%s is not yet supported", argsMap[DriverType])
+	}
```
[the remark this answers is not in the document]
fahedouch: no need for error message, we can default to nothing : https://github.com/containerd/nerdctl/blob/70bcdd06417a652be535a68301ce1dc2698c9a30/cmd/nerdctl/run.go#L770

pkg/logging/logging.go
Zheaoli: OK

cmd/nerdctl/run.go
```diff
@@ -771,6 +771,7 @@ func generateLogURI(dataStore, logDriver string, logOptMap map[string]string) (*
 	}
 	args := map[string]string{
 		logging.MagicArgv1: dataStore,
+		logging.DriverType: logDriver,
```
AkihiroSuda, line 774: I don't think this is correct.

We have to ensure that MagicArgv1 is always argv[1], but this map may make it argv[3]

pkg/logging/logging.go
```diff
@@ -17,83 +17,48 @@
...
+	driverType := argsMap[DriverType]
+	switch driverType {
+	case jsonFile:
+		driver = &JsonLogger{}
+	default:
+		return fmt.Errorf("%s is not yet supported", argsMap[DriverType])
+	}
+	if driver == nil {
+		return fmt.Errorf("%s is not yet supported", argsMap[DriverType])
+	}
```
junnplus, line 57: Remove it.

pkg/logging/logging.go
```diff
@@ -17,83 +17,48 @@
...
-	fn, err := getLoggerFunc(argsMap)
-	if err != nil {
+	var driver NerdCTLLoggingInterface
+	// TODO: Impl more drivers such as local, journald, fluentd
+	driverType := argsMap[DriverType]
+	switch driverType {
+	case jsonFile:
+		driver = &JsonLogger{}
+	default:
+		return fmt.Errorf("%s is not yet supported", argsMap[DriverType])
```
junnplus, line 53: Like @fahedouch said, we don't need it, just return.

Zheaoli:
@AkihiroSuda 

@fahedouch and I want to enhance the log driver, see #1039 

In the old version of the code, we hard-coded some `json-file` only logic in the logging module, I think it's not a good choice if we want to support multiple logging-driver. So I refactor the hard-coded code to Interface

fahedouch:
@AkihiroSuda, I agree with @Zheaoli , this refacto will help us handling this list of logging driver  https://github.com/containerd/nerdctl/issues/1039

cmd/nerdctl/run.go
fahedouch: @AkihiroSuda  not really, I have already experimented with a change of position of the args in this PR https://github.com/containerd/nerdctl/pull/959. 

the actual code
```
	args := map[string]string{
		logging.MagicArgv1: dataStore,
	}
	for k, v := range logOptMap {
		args[k] = v
	}
```
does not guarantee a fixed order of arguments
for this reason  the args parsing is refactored in this way : https://github.com/containerd/nerdctl/pull/959/files#diff-055f400ea02598bd9b9ef894b5ffc400954e88ad71cc43a31f421222af4da6c7R81

junnplus (approved):
LGTM

pkg/logging/logging.go
```diff
@@ -17,83 +17,43 @@
...
 	MaxSize    = "max-size"
 	MaxFile    = "max-file"
+	DriverType = "logging-driver"
 )
 
+const (
+	jsonFile = "json-file"
+)
+
+type NerdCTLLoggingInterface interface {
```
fahedouch, line 35: I suggest `type NerdctlLogger interface`.

fahedouch (approved):

AkihiroSuda:
I think this kind of refactoring should be done when you actually implement another driver.
Hard to review whether the design is correct otherwise.

Zheaoli:
> I think this kind of refactoring should be done when you actually implement another driver.
> 
> Hard to review whether the design is correct otherwise.
> 
> 
> 
> 

@fahedouch has worked on journal log driver base on this PR. FYI https://github.com/containerd/nerdctl/issues/1039#issuecomment-1123675221

fahedouch:
@Zheaoli no problem, I rebased your refacto on my branch, I will try to push a PR as soon as possible. Waiting to merge https://github.com/containerd/nerdctl/pull/1056 this PR too

Issue #1054, opened 2022-05-12
logging: ensure that MagicArgv1 is always argv1


AkihiroSuda:
> Sorry for going back and forth, but this logic doesn’t seem robust.
> We have to ensure that MagicArgv1 is always argv1.
> Other args should be encoded in argv2 or in the metadata referred by argv2.

_Originally posted by @AkihiroSuda in https://github.com/containerd/nerdctl/pull/959#discussion_r870874366_

PR #1055, opened 2022-05-12
logging: ensure that MagicArgv1 is always argv1

Zheaoli:
fix: #1054 

### DIFF
--- a/pkg/strutil/strutil_test.go
+++ b/pkg/strutil/strutil_test.go
@@ -281,3 +281,11 @@ func TestParseCSVMap(t *testing.T) {
 		})
 	}
 }
+
+func TestOrderedQuery(t *testing.T) {
+	query := NewOrderedQuery()
+	query.Set("abc", "1")
+	query.Set("aa", "2")
+	result := query.Encode()
+	assert.Equal(t, result, "abc=1&aa=2", "")
+}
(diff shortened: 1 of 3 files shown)


pkg/strutil/strutil_test.go
```diff
@@ -281,3 +281,11 @@ func TestParseCSVMap(t *testing.T) {
...
 	}
 }
+
+func TestOrderedQuery(t *testing.T) {
+	query := NewOrderedQuery()
+	query.Set("abc", "1")
+	query.Set("aa", "2")
+	result := query.Encode()
+	assert.Equal(t, result, "abc=1&aa=2", "")
+}
```
AkihiroSuda, line 291: Sorry, this is not correct, as the ordering is ignored on parsing back the string to `*net.URL`


https://github.com/containerd/containerd/blob/e85b5a0b81a610f564c3240c69d8f7324b5ecadb/pkg/process/io_util.go#L31

Zheaoli: May I ask if it is correct if we sort the args when in the nerdctl xmain function?

AkihiroSuda:
Opened an alternative PR https://github.com/containerd/nerdctl/pull/1056

pkg/strutil/strutil_test.go
AkihiroSuda: > May I ask if it is correct if we sort the args when in the nerdctl xmain function?

No, that doesn't work with `nerdctl run busybox echo "_NERDCTL_INTERNAL_LOGGING"`


>  Maybe we don't need to concern about the args order?

Need

PR #1070, opened 2022-05-20
compose: enable entrypoint as stringArray

fahedouch:
I am raising this fix to handle this case https://github.com/containerd/nerdctl/discussions/1069 

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -51,7 +51,6 @@ import (
 	"github.com/containerd/nerdctl/pkg/referenceutil"
 	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/containerd/nerdctl/pkg/taskutil"
-
 	"github.com/docker/cli/opts"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


cmd/nerdctl/run.go
```diff
@@ -693,7 +694,11 @@ func generateRootfsOpts(ctx context.Context, client *containerd.Client, platform
 		}
 		var processArgs []string
 		if entrypoint != "" {
-			processArgs = append(processArgs, entrypoint)
+			s, err := shellwords.Parse(entrypoint)
+			if err != nil {
+				return nil, nil, nil, err
+			}
+			processArgs = append(processArgs, s...)
```
AkihiroSuda, line 701: Seems incompatible with Docker.
Docker doesn't parse the entrypoint.
```
$ docker run --rm --entrypoint "echo foo bar" alpine
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "echo foo bar": executable file not found in $PATH: unknown.
```

AkihiroSuda: I think we should just change the type of the `entrypoint` from `string` to `[]string`.

fahedouch: @AkihiroSuda I am rethinking about it , switching `entrypoint` to slice this means sthg like this 
`docker run --rm  --entrypoint "foo" --entrypoint "echo"  alpine` which is confusing.
FYI `docker-compose --entrypoint` is parsing shell words.
`nerdctl compose` rely on `nerdctl` which is not the same behavior for docker-compose and docker. Keeping 100% compatibility  is tough in this situation

AkihiroSuda: > docker run --rm --entrypoint "foo" --entrypoint "echo" alpine which is confusing.

I think this is fine as an internal implementation detail of `nerdctl compose`.
`nerdctl run` users are not expected to specify multiple `--entrypoint` flags manually.

fahedouch: fixed

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -478,12 +478,6 @@ func newContainer(project *compose.Project, parsed *Service, i int) (*Container,
 	for _, v := range svc.DNS {
 		c.RunArgs = append(c.RunArgs, fmt.Sprintf("--dns=%s", v))
 	}
-
-	if len(svc.Entrypoint) > 1 {
-		return nil, fmt.Errorf("service %s: specifying entrypoint with multiple strings (%v) is not supported yet",
-			svc.Name, svc.Entrypoint)
-	}
```
AkihiroSuda, line 485: Could you also update https://github.com/containerd/nerdctl/blob/master/docs/compose.md#incompatibility ?

AkihiroSuda (approved):
Thanks

pkg/composer/serviceparser/serviceparser.go
AkihiroSuda: Pushed a commit by myself https://github.com/containerd/nerdctl/pull/1070/commits/08924b9c54f4df48ce02090ad6c05f26c1bf4aa1

Issue #1091, opened 2022-06-06
Change default bridge subnet with containerd runtime


obeone:
### Description

The default bridge subnet is 10.4.0.0/24, and it's a big problem for me : I already use this subnet... I can't find how to change it

### Steps to reproduce the issue

1. Just start containerd

### Describe the results you received and expected

No possibility found to edit the default bridge subnet

### What version of nerdctl are you using?

Client:
 Version:	v0.20.0
 OS/Arch:	linux/amd64
 Git commit:	e77e05b5fd252274e3727e0439e9a2d45622ccb9

Server:
 containerd:
  Version:	v1.5.8
  GitCommit:	1e5ef943eb76627a6d3b6de8cd1ef6537f393a71

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Colima

### Host information

MacOS 12.3

colima version 0.4.2
git commit: f112f336d05926d62eb6134ee3d00f206560493b

runtime: containerd
arch: x86_64
client: v0.20.0
server: v1.5.8

PR #1092, opened 2022-06-06
persist default network config

junnplus:

Fixes: https://github.com/containerd/nerdctl/issues/1091

### DIFF
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -44,6 +44,10 @@ func NewCNIEnv(cniPath, cniConfPath string) (*CNIEnv, error) {
 		NetconfPath: cniConfPath,
 	}
 	var err error
+	err = e.ensureDefaultNetworkConfig()
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


pkg/netutil/netutil.go
```diff
@@ -44,6 +44,10 @@ func NewCNIEnv(cniPath, cniConfPath string) (*CNIEnv, error) {
 		NetconfPath: cniConfPath,
 	}
 	var err error
+	err = e.ensureDefaultNetworkConfig()
+	if err != nil {
+		return nil, err
+	}
```
AkihiroSuda, line 50: We have to make sure to lock the CNI dir (`git grep pkg/lockutil`)

junnplus: Seems to already have a lock when create/remove network. 
https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/network_create.go#L149

`ensureDefaultNetworkConfig` just makes sure the default network exists.

AkihiroSuda (approved):
Thanks

PR #1108, opened 2022-06-08
Adding --attach flag to start

mattfarina:
This change adds the -a/--attach flag to attach to stdout/stderr
and os events. It follow the style used in run without -d to do
this.

One decision was if the user uses the -a flag and there is a log
label to choose -a and warn that the label is ignored.

### DIFF
--- a/cmd/nerdctl/start.go
+++ b/cmd/nerdctl/start.go
@@ -109,7 +109,7 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
 			return err
 		}
 		if flagA {
-			logrus.Warn("attaching output instead of using the log-uri")
+			logrus.Debug("attaching output instead of using the log-uri")
 		} else {
 			taskCIO = cio.LogURI(logURI)
 		}
--- a/cmd/nerdctl/start.go
+++ b/cmd/nerdctl/start.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
+	"github.com/containerd/containerd/cmd/ctr/commands"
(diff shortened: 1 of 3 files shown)


mattfarina:
The test failures appear to be unrelated to `start`. Are these flaky tests or am I missing something?

cmd/nerdctl/start.go
```diff
@@ -41,6 +43,10 @@ func newStartCommand() *cobra.Command {
 		SilenceUsage:      true,
 		SilenceErrors:     true,
 	}
+
+	startCommand.Flags().SetInterspersed(false)
+	startCommand.Flags().BoolP("attach", "a", false, "Attach STDOUT/STDERR and forward signals")
```
AkihiroSuda, line 48: Could you update the Command Reference in README.md?
Also, please consider adding tests

cmd/nerdctl/start.go
```diff
@@ -72,19 +92,29 @@ func startAction(cmd *cobra.Command, args []string) error {
...
+		taskCIO = cio.NewCreator(cio.WithStreams(os.Stdin, os.Stdout, os.Stderr))
+	}
 	if logURIStr := lab[labels.LogURI]; logURIStr != "" {
 		logURI, err := url.Parse(logURIStr)
 		if err != nil {
 			return err
 		}
-		taskCIO = cio.LogURI(logURI)
+		if flagA {
+			logrus.Warn("attaching output instead of using the log-uri")
```
AkihiroSuda, line 112: maybe just logrus.Debug

mattfarina: Done

cmd/nerdctl/start.go
mattfarina: Updated the readme and added tests for both start and start with attach. Hope this looks good.

AkihiroSuda (approved):
Thanks

PR #1114, opened 2022-06-09
support print stack

ningmingxiao:


AkihiroSuda:
I’m not sure how this is useful, as nerdctl isn’t a daemon.

Even if it is useful, the default behavior must be to propagate signals to the container.

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/1114#issuecomment-1151744719

fahedouch:
it may be implemented on the containerd side and enabled in the nerdctl.

PR #1119, opened 2022-06-11
fix log-driver  name not check error

ningmingxiao:


ningmingxiao:
nerdctl run -d --log-driver=fluentd1111 busybox  "top" will not show error

fahedouch:
LGTM on Green CI

AkihiroSuda (approved):
Thanks, LGTM.

The list of the logging plugins should be rather fetched from `pkg/logging`, but that can be another PR.

PR #1126, opened 2022-06-14
fix swap memory

junnplus:

```
$ sudo docker run -it --rm -m 42m alpine cat /sys/fs/cgroup/memory.max /sys/fs/cgroup/memory.swap.max
44040192
44040192
$ sudo nerdctl run -it --rm -m 42m alpine cat /sys/fs/cgroup/memory.max /sys/fs/cgroup/memory.swap.max
44040192
max
```

### DIFF
--- a/cmd/nerdctl/run_cgroup_linux_test.go
+++ b/cmd/nerdctl/run_cgroup_linux_test.go
@@ -58,6 +58,7 @@ func TestRunCgroupV2(t *testing.T) {
 	}
 	const expected1 = `42000 100000
 44040192
+44040192
 42
 77
 0-1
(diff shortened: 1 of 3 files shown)


junnplus:
cc @ningmingxiao

ningmingxiao:
LGTM

AkihiroSuda (approved):
Thanks

cmd/nerdctl/run_cgroup_linux_test.go
```diff
@@ -71,29 +72,52 @@ func TestRunCgroupV2(t *testing.T) {
...
+	defer base.Cmd("rm", "-f", testutil.Identifier(t)+"-testUpdate1").Run()
+	update := []string{"update", "--cpu-quota", "42000", "--cpuset-mems", "0", "--cpu-period", "100000",
+		"--memory", "42m",
+		"--pids-limit", "42", "--cpu-shares", "2000", "--cpuset-cpus", "0-1"}
 	if base.Target == testutil.Docker && info.CgroupVersion == "2" && info.SwapLimit {
 		// Workaround for Docker with cgroup v2:
 		// > Error response from daemon: Cannot update container 67c13276a13dd6a091cdfdebb355aa4e1ecb15fbf39c2b5c9abee89053e88fce:
 		// > Memory limit should be smaller than already set memoryswap limit, update the memoryswap at the same time
-		update = append(update, "--memory-swap=-1")
+		update = append(update, "--memory-swap=84m")
```
AkihiroSuda, line 103: Any reason for this change?
Also wondering we should now set `--memory-swap` for `Target == Nerdctl` too

junnplus: > Any reason for this change?
This change is only for compatibility with test results `expected1`.

> Also wondering we should now set `--memory-swap` for `Target == Nerdctl` too
Actually, nerdctl does not limit `memory` and `memory-swap` need to be set at the same time. Of course, it's better to keep the same as docker.

PR #1137, opened 2022-06-17
Validation for nerdctl load to fail when stdin is empty.

manugupt1:
When running nerdctl load gets an empty stdin, it gets stuck.
This adds a validation letting the user immediately know that
stdin is empty and input flag is not specified.

This PR adds validation to return error when stdin is empty and input is not 
specified.

Fixes https://github.com/containerd/nerdctl/issues/1138

### DIFF
--- a/cmd/nerdctl/load.go
+++ b/cmd/nerdctl/load.go
@@ -67,8 +67,16 @@ func loadAction(cmd *cobra.Command, args []string) error {
 		}
 		defer f.Close()
 		in = f
+	} else {
+		// check if stdin is empty.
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Thanks, but something seems wrong with your cosign (?) signature

image

AkihiroSuda (approved):

manugupt1:
> Thanks, but something seems wrong with your cosign (?) signature
> 
> image

This is because I am using gitsign (ref: https://github.com/sigstore/gitsign) and one of the limitations is that the Github shows the commit to be unverified.

Issue #1117, opened 2022-06-10
--output argument doesn't work as expected


Kitanotori:
### Description

**Problem 1: `--tag` overrides `name`**

Example 1: `nerdctl build -t $TAG1 --output type=image,name=$TAG2,push=true .`
Expected behavior: $TAG1 either ignored or error raised due to passing two tags, i.e. same as if two `-t` args passed (the latter would be more consistent with existing existing logic).

Example 2: `docker build -t $TAG1 -t $TAG2 --output type=image,name=$TAG3,push=true .`
Pushes $TAG1 and $TAG2 to registry and stores to local but totally ignores `name` (ignoring name feels buggy and/or weird API design, though).

**Problem 2: no way to load image locally using --output**

Example: neither `nerdctl build --output type=image,name=$TAG,push=true .` or `nerdctl build --output type=image,name=$TAG,push=false .` load the image to local.

### Steps to reproduce the issue

see above

### Describe the results you received and expected

**Problem 1**

As a user, I would expect both tools to have more intuitive behavior. Suggestions (some mutually exclusive) for improvement:

**1. Treat tags and name as equals**

With current logic, in case of tag(s) passed with name, this would mean raising error as multiple tags are not allowed.

**2. Use tags for tagging and loading to local only**

I.e. the same behavior as when --output is not passed. E.g., `nerdctl build -t $TAG1 --output type=image,name=$TAG2,push=true .` would store $TAG1 to local and push $TAG2 to registry. In case push=false, it would be same as `nerdctl build -t $TAG1 .`. So essentially tags would be limited to local use only, and by local I mean visibility via `nerdctl images`.

**Problem 2**

BuildKit has option store=true as default, so it would be logical to have same behavior. Examples:

1. `nerdctl build --output type=image,name=$TAG .`: push to registry and store to local (push=true and store=true as default)
2. `nerdctl build --output type=image,name=$TAG,store=false .` push without storing to local
3. `nerdctl build --output type=image,name=$TAG,push=false .` store to local without pushing
4. `nerdctl build --output type=image,name=$TAG,push=false,store=false .` no-op, or raise error

### What version of nerdctl are you using?

v0.20.0 (just updated to RD 1.4.1)

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Rancher Desktop for macOS

Kitanotori:
Example of Buildkit `--output` values:

[...]

PR #1146, opened 2022-06-20
1117: specify needsLoading when types are specified in output

manugupt1:
When output is non-empty string, needsLoading is always false.
However, someone can specify type=docker in output in which
case you need needsLoading=true. This commit handles the guess.

Fixes: https://github.com/containerd/nerdctl/issues/1117 (Example 1 of the #1117)

### DIFF
--- a/cmd/nerdctl/build.go
+++ b/cmd/nerdctl/build.go
@@ -262,6 +262,10 @@ func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, arg
 			}
 			needsLoading = true
 		}
+	} else {
+		if strings.Contains(output, "type=docker") || strings.Contains(output, "type=oci") {
+			needsLoading = true
+		}
 	}
 	tagValue, err := cmd.Flags().GetStringArray("tag")
 	if err != nil {
--- a/cmd/nerdctl/build_test.go
+++ b/cmd/nerdctl/build_test.go
@@ -44,8 +44,12 @@ CMD ["echo", "nerdctl-build-test-string"]
 
 	base.Cmd("build", "-t", imageName, buildCtx).AssertOK()
 	base.Cmd("build", buildCtx, "-t", imageName).AssertOK()
+	ignoredImageNamed := imageName + "-" + "ignored"
+	outputOpt := fmt.Sprintf("--output=type=docker,name=%s", ignoredImageNamed)
+	base.Cmd("build", buildCtx, "-t", imageName, outputOpt).AssertOK()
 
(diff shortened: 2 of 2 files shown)


cmd/nerdctl/build_test.go
```diff
@@ -44,6 +44,7 @@ CMD ["echo", "nerdctl-build-test-string"]
 
 	base.Cmd("build", "-t", imageName, buildCtx).AssertOK()
 	base.Cmd("build", buildCtx, "-t", imageName).AssertOK()
+	base.Cmd("build", buildCtx, "-t", imageName, fmt.Sprintf("--output=type=docker,name=name")).AssertOK()
```
AkihiroSuda, line 47: We should test `base.Cmd("run", "--rm", imageName).AssertOutExactly("nerdctl-build-test-string\n")` again (but the imageName should be changed)

manugupt1: Thanks for reviewing, Should we not take what the docker behavior is?

```
➜  nerdctl git:(fix-nerdctl-build) ✗ docker build -t nerdctl --output=type=docker,name="nerdctl2"  .                
➜  nerdctl git:(fix-nerdctl-build) ✗ docker images
REPOSITORY   TAG       IMAGE ID   CREATED   SIZE
....
➜  nerdctl git:(fix-nerdctl-build) ✗ docker images | grep nerdctl
nerdctl      latest    514a49968fa2   3 minutes ago   686MB

➜   git:(main) ✗ docker run -it --rm nerdctl2 /bin/sh
Unable to find image 'nerdctl2:latest' locally
```

manugupt1: Can you review again; I added the docker compatibility test to ensure that this fails as well? If you still prefer having a second tag; I will make the code changes appropriately.

cmd/nerdctl/build.go
```diff
@@ -262,6 +262,10 @@ func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, arg
 			}
 			needsLoading = true
 		}
+	} else {
```
fahedouch, line 265: looks like we need `needsLoading=true` for `type=oci` too right ? https://github.com/containerd/nerdctl/pull/1146/files#diff-05bfdfb8af1ef5eb710a573250948e9c6935773725ddd08d162309b367e592bfR261

manugupt1: Right. thanks for catching it.

AkihiroSuda:
Thanks, LGTM but please squash commits

AkihiroSuda (approved):
Thanks

Issue #369, opened 2021-09-16
cannot mount host paths with kata


paleozogt:
When using the kata runtime, mounting host paths doesn't work:
```
~$ sudo nerdctl run --runtime io.containerd.kata.v2 --rm -it -v $PWD:$PWD -w $PWD busybox pwd
FATA[0000] ENOENT: No such file or directory: unknown
```

However, this works with `ctr`:
```
~$ sudo ctr run --runtime io.containerd.kata.v2 -t --rm --mount type=bind,src=$PWD,dst=$PWD,options=rbind:ro --cwd $PWD docker.io/library/busybox:latest foo pwd
/home/admin
```

paleozogt:
Upgrading to containerd `v.1.5.5` shows a slightly different error:
```
~$ sudo nerdctl run --runtime io.containerd.kata.v2 --rm -it -v $PWD:$PWD -w $PWD busybox pwd
FATA[0008] failed to create shim: ENOENT: No such file or directory: unknown
```

chaokw:
met same issue, any update? thanks

PR #1172, opened 2022-06-29
set bind/rbind options for volume/bind-mounts

liubin:
This PR is for fixing #369 

The first commit will set fstype to bind in the options contain bind or rbind, with this commit, below commands will start container successfully:

```bash
# nerdctl run --runtime io.containerd.kata.v2 \
--mount type=bind,src=/tmp/foo,dst=/foo,bind-nonrecursive=true \
containerstack/alpine-stress sh -c "while true; do date; sleep 1; done"

# nerdctl run --runtime io.containerd.kata.v2 \
-v /tmp/foo:/foo:bind containerstack/alpine-stress \
sh -c "while true; do date; sleep 1; done"

# nerdctl run --runtime io.containerd.kata.v2 \
-v /tmp/foo:/foo:rbind containerstack/alpine-stress \
sh -c "while true; do date; sleep 1; done"
```

But `--mount type=bind,src=/tmp/foo,dst=/foo` will not work, the second commit fix this by adding `bind` option.

@AkihiroSuda @junnplus


liubin:
runc works correctly because it has [overwritten the mount type](https://github.com/opencontainers/runc/blob/9524366183517b379a1fca8b9b250002ba3b1be7/libcontainer/specconv/spec_linux.go#L515)

pkg/mountutil/mountutil_linux.go
```diff
@@ -404,6 +404,8 @@ func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, e
 		}
 		if mountType == Bind && bindNonRecursive {
 			options = append(options, "bind")
+		} else if mountType == Bind {
+			options = append(options, "rbind")
 		}
```
AkihiroSuda, line 409:
```go
if mountType == Bind {
  if bindNotRecursive {
    options = append(options, "bind")
  } else {
    options = append(options, "rbind")
  }
}
```

liubin: Updated.

AkihiroSuda (approved):
Thanks

PR #1174, opened 2022-06-29
Support multiple tags in nerdctl

manugupt1:


cmd/nerdctl/build.go
```diff
@@ -207,49 +208,73 @@ func buildAction(cmd *cobra.Command, args []string) error {
...
 	return nil
 }
 
-func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, args []string) (string, []string, bool, string, func(), error) {
-	var needsLoading bool
+func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, args []string) (buildCtlBinary string,
+	buildctlArgs []string, needsLoading bool, metaFile string, tags []string, cleanup func(), err error) {
 	if len(args) < 1 {
-		return "", nil, false, "", nil, errors.New("context needs to be specified")
+		return "", nil, false, "", []string{}, nil, errors.New("context needs to be specified")
```
AkihiroSuda, line 241:
```suggestion
		return "", nil, false, "", nil, nil, errors.New("context needs to be specified")
```

cmd/nerdctl/build.go
```diff
@@ -207,49 +208,73 @@ func buildAction(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if len(tags) > 1 {
```
fahedouch, line 211: why not delegating this action to the buildkid  ? => passing tag to buildctl here https://github.com/containerd/nerdctl/pull/1174/files#diff-05bfdfb8af1ef5eb710a573250948e9c6935773725ddd08d162309b367e592bfR166

PS: here you are building image with the containerd worker but buildkit can build with other worker too like oci so this can be confusing for users

manugupt1: As far as I understand, we need these images in the containerd store even in case of oci and local image exports for nerdctl.  I briefly looked at the code: https://github.com/moby/buildkit/blob/master/exporter/local/export.go but they do not seem to send back the tags to containerd.

I am trying to think through, correct me if I am wrong.
1. Image is built through oci, docker, local
2. nerdctl list queries containerd; 
3. but the image is viewable only when it is in the containerd store.

AkihiroSuda (approved):
Thanks

Issue #1175, opened 2022-06-29
Add static check bot


hwdef:
### What is the problem you're trying to solve

I found some static check errors in code. 

```shell
$ staticcheck ./...
cmd/nerdctl/apparmor_inspect_linux.go:44:11: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
cmd/nerdctl/exec.go:257:2: should replace loop with pspec.Env = append(pspec.Env, strutil.DedupeStrSlice(env)...) (S1011)
cmd/nerdctl/info.go:234:34: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (SA1019)
cmd/nerdctl/ipfs_linux_test.go:146:2: redundant return statement (S1023)
cmd/nerdctl/ipfs_registry_up.go:121:74: should use String() instead of fmt.Sprintf (S1025)
cmd/nerdctl/kill.go:58:17: containerd.ParseSignal is deprecated: Use github.com/moby/sys/signal instead.  (SA1019)
cmd/nerdctl/main_unix.go:47:2: should replace loop with candidates = append(candidates, nsList...) (S1011)
cmd/nerdctl/main_unix.go:68:2: should replace loop with candidates = append(candidates, snapshotterPlugins...) (S1011)
cmd/nerdctl/run.go:55:2: package "github.com/docker/cli/opts" is being imported more than once (ST1019)
        cmd/nerdctl/run.go:56:2: other import of "github.com/docker/cli/opts"
cmd/nerdctl/run.go:1071:3: should check returned error before deferring f.Close() (SA5001)
cmd/nerdctl/run_linux_test.go:120:11: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (S1028)
cmd/nerdctl/run_network.go:125:6: should omit nil check; len() for []string is defined as zero (S1009)
cmd/nerdctl/run_network_linux_test.go:194:23: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (S1007)
cmd/nerdctl/stats.go:317:5: ineffective break statement. Did you mean to break out of the outer loop? (SA4011)
cmd/nerdctl/stop.go:141:18: containerd.ParseSignal is deprecated: Use github.com/moby/sys/signal instead.  (SA1019)
[...]

PR #1179, opened 2022-07-01
Implements static check in nerdctl CI

manugupt1:
Fixes https://github.com/containerd/nerdctl/issues/1175

### DIFF
--- a/cmd/nerdctl/info.go
+++ b/cmd/nerdctl/info.go
@@ -23,6 +23,9 @@ import (
 	"strings"
 	"text/template"
 
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+
 	"github.com/containerd/containerd/api/services/introspection/v1"
 	"github.com/containerd/nerdctl/pkg/infoutil"
 	"github.com/containerd/nerdctl/pkg/inspecttypes/dockercompat"
@@ -231,7 +234,7 @@ func prettyPrintInfoDockerCompat(cmd *cobra.Command, info *dockercompat.Info) er
 			if k == "name" {
 				continue
 			}
-			fmt.Fprintf(w, "   %s: %s\n", strings.Title(k), v)
+			fmt.Fprintf(w, "   %s: %s\n", cases.Title(language.English).String(k), v)
 		}
 	}
 	fmt.Fprintf(w, " Kernel Version: %s\n", info.KernelVersion)
--- a/cmd/nerdctl/stats.go
+++ b/cmd/nerdctl/stats.go
@@ -313,9 +313,6 @@ func statsAction(cmd *cobra.Command, args []string) error {
 			return errors.New("unsupported format: \"raw\"")
 		default:
 			tmpl, err = parseTemplate(format)
-			if err != nil {
-				break
-			}
 		}
 
 		for _, c := range ccstats {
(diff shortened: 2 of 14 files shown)


cmd/nerdctl/info.go
```diff
@@ -231,7 +231,7 @@ func prettyPrintInfoDockerCompat(cmd *cobra.Command, info *dockercompat.Info) er
 			if k == "name" {
 				continue
 			}
-			fmt.Fprintf(w, "   %s: %s\n", strings.Title(k), v)
+			fmt.Fprintf(w, "   %s: %s\n", strings.ToTitle(k), v)
```
AkihiroSuda, line 234: strings.Title != strings.ToTitle

https://pkg.go.dev/strings#Title
https://pkg.go.dev/strings#ToTitle

cmd/nerdctl/stop.go
```diff
@@ -176,12 +177,12 @@ func stopContainer(ctx context.Context, container containerd.Container, timeout
 		}
 	}
 
-	signal, err := containerd.ParseSignal("SIGKILL")
+	sig, err := signal.ParseSignal("SIGKILL")
```
AkihiroSuda, line 180: Why?

manugupt1: Has been deprecated in favor moby: https://pkg.go.dev/github.com/containerd/containerd#ParseSignal and gave a deprecated error warning.

cmd/nerdctl/stats.go
```diff
@@ -313,9 +313,6 @@ func statsAction(cmd *cobra.Command, args []string) error {
 			return errors.New("unsupported format: \"raw\"")
 		default:
 			tmpl, err = parseTemplate(format)
-			if err != nil {
-				break
-			}
```
AkihiroSuda, line 318: What should we do with this err?
Cc @fahedouch

pkg/imgutil/pull/pull.go
```diff
@@ -73,7 +73,8 @@ func Pull(ctx context.Context, client *containerd.Client, ref string, config *Co
 	opts := []containerd.RemoteOpt{
 		containerd.WithResolver(config.Resolver),
 		containerd.WithImageHandler(h),
-		containerd.WithSchema1Conversion,
+		//nolint:staticcheck
+		containerd.WithSchema1Conversion, //lint:ignore SA1019 trying out line-based linter directives
```
AkihiroSuda, line 77: ?

manugupt1: Sorry! I was not sure how to use lint:ignore, so that was a stranded comment, I have fixed it now. We need both ignore to ignore on local runs and also golangci

cmd/nerdctl/stats.go
fahedouch: what is the problem with this err @manugupt1 ?

manugupt1: It does not break out the ticker loop and therefore, it is ineffective. Should I modify the code to stop the ticker?
Here is an example: https://go.dev/play/p/vpSz8gWxyQr

AkihiroSuda: We should parse the template and check the err before starting the ticker, but it can be another PR.

AkihiroSuda (approved):
Thanks

junnplus (approved):
LGTM

PR #1150, opened 2022-06-21
Support config default network name for nerdctl

Zheaoli:
Fix https://github.com/containerd/nerdctl/discussions/1128

Signed-off-by: Zheao.Li <me@manjusaka.me>

PR #1184, opened 2022-07-03
Support default platform config in nerdctl.toml

YangKian:


cmd/nerdctl/main.go
```diff
@@ -87,17 +88,41 @@ func xmain() error {
...
+	InsecureRegistry bool         `toml:"insecure_registry"`
+	HostsDir         []string     `toml:"hosts_dir"`
+	RunConfig        *RunConfig   `toml:"default_run_config"`
+	BuildConfig      *BuildConfig `toml:"default_build_config"`
+}
+
+type RunConfig struct {
+	Platform string   `toml:"platform"`
+	Network  []string `toml:"network"`
+}
```
AkihiroSuda, line 109: Can we reuse the compose structure? https://pkg.go.dev/github.com/compose-spec/compose-go@v1.2.8/types#ServiceConfig

(Some fields such as `image` and `container_name` cannot be supported and should raise an error.)

cmd/nerdctl/main.go
```diff
@@ -87,17 +88,41 @@ func xmain() error {
...
+	Address          string       `toml:"address"`
+	Namespace        string       `toml:"namespace"`
+	Snapshotter      string       `toml:"snapshotter"`
+	CNIPath          string       `toml:"cni_path"`
+	CNINetConfPath   string       `toml:"cni_netconfpath"`
+	DataRoot         string       `toml:"data_root"`
+	CgroupManager    string       `toml:"cgroup_manager"`
+	InsecureRegistry bool         `toml:"insecure_registry"`
+	HostsDir         []string     `toml:"hosts_dir"`
+	RunConfig        *RunConfig   `toml:"default_run_config"`
```
AkihiroSuda, line 102: Shouldn't this be `map[string]...` where the key is a containerd namespace string such as "default"?

YangKian: I think both `map[string]...` and user-defined structs are legal, I find an example from https://pkg.go.dev/github.com/pelletier/go-toml@v1.9.5#section-readme and test it works.

YangKian: Ah, if I can reuse the compose structure, then here I can just use `map[namespace]ServiceConfig` to avoid defining multiple similar structures. I'll have some try, thanks!

cmd/nerdctl/main.go
Zheaoli: I think we should not reuse https://pkg.go.dev/github.com/compose-spec/compose-go@v1.2.8/types#ServiceConfig. 

I think this would be a little bit complicated for user. For instance, like #1128, maybe people just need config like this

```toml
[default_run_config]
network=["test"]
```

cmd/nerdctl/main.go
```diff
@@ -19,6 +19,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"github.com/containerd/nerdctl/pkg/netutil"
```
Zheaoli, line 22: fix the lint plz

cmd/nerdctl/main_test.go
```diff
@@ -88,3 +90,58 @@ version = 2
 	base.Env = append(base.Env, "NERDCTL_TOML="+tomlPath)
 	base.Cmd("info").AssertFail()
 }
+
+// TestNerdctlSpecificConfig validates the configuration precedence [CLI, TOML, Default].
+func TestNerdctlSpecificConfig(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	t.Parallel()
+	tomlPath := filepath.Join(t.TempDir(), "nerdctl.toml")
+	err := os.WriteFile(tomlPath, []byte(`
```
Zheaoli, line 99: remove the file after the test complete

here's the demo:  https://github.com/containerd/nerdctl/pull/1150/files#diff-4843eb8080d6ff754054971c8db0ad20041e350810c18ba82302907eefcf77ecR475-R479

cmd/nerdctl/main_test.go
```diff
@@ -88,3 +90,58 @@ version = 2
 	base.Env = append(base.Env, "NERDCTL_TOML="+tomlPath)
 	base.Cmd("info").AssertFail()
 }
+
+// TestNerdctlSpecificConfig validates the configuration precedence [CLI, TOML, Default].
+func TestNerdctlSpecificConfig(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	t.Parallel()
+	tomlPath := filepath.Join(t.TempDir(), "nerdctl.toml")
```
Zheaoli, line 98: use the `ncdefaults.NerdctlTOML()` method

cmd/nerdctl/main.go
YangKian: I'm afraid we can't reuse the compose structure. I agree with @Zheaoli ’s option, the `ServiceConfig` is complicated for users. Another reason is that even if the parameters have the same name, they may not have the same type in different scenarios, e.g. `platform` only needs to be a string in the `run` command, but it needs to be a []string in the `build` command because it needs to support parsing parameters like `arm64,amd64`

cmd/nerdctl/main.go
YangKian: fixed.

cmd/nerdctl/main_test.go
YangKian: fixed

cmd/nerdctl/main_test.go
YangKian: fixed

cmd/nerdctl/main.go
YangKian: I think it is more appropriate to choose parsing to structs than to maps here, so that only fields defined in structs will be parsed, and unsupported fields will be ignored in parsing even if they are mistakenly written to the configuration file by the user.

cmd/nerdctl/main.go
YangKian: Hi @AkihiroSuda, I'd like to do some double-check to make sure I understand your suggestion correctly:

`[default_run_config]` will be decoded as `ServiceConfig` and `[default_build_config]` will be decoded as `BuildConfig`, right? 

And another question, 
> (Some fields such as image and container_name cannot be supported and should raise an error.)

which means If the user adds these unsupported fields to the `[default_run_config]` section of the `Nerdctl.toml` file, we should raise an error when decoding?

AkihiroSuda: > [default_run_config] will be decoded as ServiceConfig and [default_build_config] will be decoded as BuildConfig, right?

ServiceConfig contains BuildConfig, so no need to have "default_build_config".
A single "default_template" (per namespace) will suffice.

> which means If the user adds these unsupported fields to the [default_run_config] section of the Nerdctl.toml file, we should raise an error when decoding?

Yes. At least a warning should be printed.
Like this: https://github.com/containerd/nerdctl/search?q=reflectutil.UnknownNonEmptyFields

Zheaoli:
@YangKian Is there any update about this PR ?

YangKian:
@Zheaoli I'm sorry I've been a little busy at work lately and haven't had time to continue, but I should have more free time to continue writing starting next week. The ServiceConfig-based update is about 60% complete. I'll finish the rest as soon as I can, and then I think we can start the review first and finish the final update after the upstream pr merged.

Zheaoli:
@YangKian Is there any update on this PR?

Zheaoli:
I may take a hand on this PR base on what the owner has already done because the owner may not have enough time to do this if there is not any update on this PR before this weekend. cc @AkihiroSuda

apostasie:
Hey @Zheaoli is this something you would still like to engage?

apostasie:
@Zheaoli any update on this? ^

PR #1186, opened 2022-07-04
Validate formats before ticking 

manugupt1:
Follow up on:
https://github.com/containerd/nerdctl/pull/1179#discussion_r912340899

Testing details

```
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O         BLOCK I/O     PIDS
d8da42cafb9e   busybox-d8da   0.00%     1.094MiB / 16EiB    0.00%     1.38kB / 752B   1.18MB / 0B   2
```

on every tick after moving the writer outside.

### DIFF
--- a/cmd/nerdctl/stats.go
+++ b/cmd/nerdctl/stats.go
@@ -125,6 +125,16 @@ func statsAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	var w = cmd.OutOrStdout()
+	var tmpl *template.Template
+	switch format {
+	case "", "table":
+		w = tabwriter.NewWriter(cmd.OutOrStdout(), 10, 1, 3, ' ', 0)
+	case "raw":
+		return errors.New("unsupported format: \"raw\"")
+	default:
+		tmpl, err = parseTemplate(format)
+	}
 
 	noTrunc, err := cmd.Flags().GetBool("no-trunc")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/stats.go
```diff
@@ -293,6 +293,21 @@ func statsAction(cmd *cobra.Command, args []string) error {
 	ticker := time.NewTicker(500 * time.Millisecond)
 	defer ticker.Stop()
 
+	var w = cmd.OutOrStdout()
+	var tmpl *template.Template
+	switch format {
+	case "", "table":
+		w = tabwriter.NewWriter(cmd.OutOrStdout(), 10, 1, 3, ' ', 0)
+		fmt.Fprintln(w, "CONTAINER ID\tNAME\tCPU %\tMEM USAGE / LIMIT\tMEM %\tNET I/O\tBLOCK I/O\tPIDS")
```
AkihiroSuda, line 301: The header has to be printed on every tick

cmd/nerdctl/stats.go
```diff
@@ -476,3 +480,15 @@ func statsShellComplete(cmd *cobra.Command, args []string, toComplete string) ([
...
 }
+
+func validateFormatFlag(format string) error {
+	switch format {
+	case "", "table":
+		return nil
+	case "raw":
+		return errors.New("unsupported format: \"raw\"")
+	default:
+		_, err := parseTemplate(format)
```
AkihiroSuda, line 491: No need to discard the first value

manugupt1: This is just validation and hence the first value `tmpl` is not required.
Since, we are already setting and checking again in the ticker, it should be okay. LMK what you think.

Ideally, I would like to avoid allocations every singletime ticker is called; but I was not really sure how to do that. Please let me know if this sounds good to you.

AkihiroSuda: Why parse the template on every tick?

manugupt1: This was the code before. Let me think and see if I can make it better

manugupt1: Can you check this again please?

cmd/nerdctl/stats.go
```diff
@@ -302,17 +312,9 @@ func statsAction(cmd *cobra.Command, args []string) error {
 		}
 		cStats.mu.Unlock()
 
-		w := cmd.OutOrStdout()
-		var tmpl *template.Template
-
-		switch format {
-		case "", "table":
-			w = tabwriter.NewWriter(cmd.OutOrStdout(), 10, 1, 3, ' ', 0)
+		// print header for every tab
```
AkihiroSuda, line 315: s/every tab/every tick/

PR #1189, opened 2022-07-05
docs: add nydus snapshotter description

imeoer:
At present, `nerdctl run` has supported nydus snapshotter, this PR adds description in doc.

Signed-off-by: Yan Song <yansong.ys@antfin.com>

### DIFF
--- a/docs/nydus.md
+++ b/docs/nydus.md
@@ -1,0 +1,29 @@
+# Lazy-pulling using Nydus Snapshotter
+
+| :zap: Requirement | nerdctl >= 0.22 |
+| ----------------- | --------------- |
+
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

docs/nydus.md
```diff
@@ -0,0 +1,26 @@
+# Lazy-pulling using Nydus Snapshotter
+
```
AkihiroSuda, line 2:
```markdown
| :zap: Requirement | nerdctl >= 0.22 |
|-------------------|----------------|
```

AkihiroSuda: (And kernel version, maybe?)

ktock (approved):

docs/nydus.md
imeoer: @AkihiroSuda Thanks for the reply, the kernel version is not mandatory if users don't use the in-kernel EROFS backend (added below).

hsiangkao (approved):
Yet not quite sure the CI fails, could we just ignore it or trigger again?

PR #1194, opened 2022-07-05
support image prune

junnplus:

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,7 @@ It does not necessarily mean that the corresponding features are missing in cont
     - [:whale: nerdctl rmi](#whale-nerdctl-rmi)
     - [:whale: nerdctl image inspect](#whale-nerdctl-image-inspect)
     - [:whale: nerdctl image history](#whale-nerdctl-image-history)
+    - [:whale: nerdctl image prune](#whale-nerdctl-image-prune)
     - [:nerd_face: nerdctl image convert](#nerd_face-nerdctl-image-convert)
     - [:nerd_face: nerdctl image encrypt](#nerd_face-nerdctl-image-encrypt)
     - [:nerd_face: nerdctl image decrypt](#nerd_face-nerdctl-image-decrypt)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -987,6 +988,19 @@ Flags:
 - :whale: `-q, --quiet`: Only display snapshots IDs
 - :whale: `--format`: Format the output using the given Go template, e.g, `{{json .}}`
 
+### :whale: nerdctl image prune
+Remove unused images.
+
+NOTE: The default behavior of this command is to remove all unused images.
+Inconsistent with docker image prune command, which by default removes only dangling iamge unless the --all flag is specified.
```
AkihiroSuda, line 995: Better to require `--all` then?

When `--all` is not specified, probably this command should do NOP with a warning message

junnplus: SGTM

cmd/nerdctl/image_prune.go
```diff
@@ -0,0 +1,116 @@
...
+	defer cancel()
+
+	all, err := cmd.Flags().GetBool("all")
+	if err != nil {
+		return err
+	}
+
+	if !all {
+		logrus.Warn("--all must be specified")
+		// NOP
```
AkihiroSuda, line 58: better to warn that this command is (currently) NOP when --all is not specified

junnplus: Sorry, I do not get your point.

AkihiroSuda: like `logrus.Warn("Currently `nerdctl image prune` requires `--all` to be specified. Skipping pruning.")`

cmd/nerdctl/image_prune.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return err
+	}
+	usedImages := make(map[string]struct{})
+	for _, container := range containerList {
+		usedImages[container.Image] = struct{}{}
+	}
+
+	delOpts := []images.DeleteOpt{images.SynchronousDelete()}
+	for _, image := range imageList {
```
fahedouch, line 98: I believe that an `imagewalker` (https://github.com/containerd/nerdctl/blob/master/pkg/idutil/imagewalker/imagewalker.go#L46)  is more efficient than this loop

junnplus: Why, can't filter here.

fahedouch: cc @AkihiroSuda  containerd filters are just a loop that compare with inputs ?

AkihiroSuda: Not sure what is to be filtered

AkihiroSuda (approved):
Thanks

Issue #421, opened 2021-10-08
Sharing network namespace


ryanbrigden:
Is it possible to share the network namespace between two containers, just like when passing `--network=container:<container>` with `docker run`? I'm looking to share an interface / network namespace between two containers that are in a bridge network with other containers.

ryanbrigden:
Anyone have an idea for a workaround for the moment? Would it be enough to create a named network namespace using `ip netns add` and then passing this path [in this OCI hook handler](https://github.com/containerd/nerdctl/blob/master/pkg/ocihook/ocihook.go#L236)? Appreciate the help and once I can get it working would like to contribute this as a feature to meet docker CLI parity.

AkihiroSuda:
> Would it be enough to create a named network namespace using `ip netns add` and then passing this path at `/var/run/netns` [in this OCI hook handler](https://github.com/containerd/nerdctl/blob/master/pkg/ocihook/ocihook.go#L236)?

I don't think we need `ip netns add`. 

We can just use `/proc/${ANOTHER_CONTAINER_PID}/ns/net`.

Supporting `--restart` in conjunction with this might be complicated ,though.

t1anz0ng:
@ryanbrigden this feature was implemented in https://github.com/containerd/nerdctl/releases/tag/v0.22.1
please use check it out on  https://github.com/containerd/nerdctl/releases/tag/v0.22.2 (v0.22.2 fixed a regression from [v0.22.1](https://github.com/containerd/nerdctl/releases/tag/v0.22.1)) 😄

junnplus:
@t1anz0ng Thanks, and I close this.

PR #1214, opened 2022-07-11
support --network=container:<contaienr> for sharing network namespace

TianZong48:
implement: https://github.com/containerd/nerdctl/issues/421

Signed-off-by: tianyang ni <tianzong48@gmail.com>

```
sudo ls /proc/76239/ns -alh |grep net
lrwxrwxrwx 1 root root 0 Jul 11 23:50 net -> net:[4026532895]
```

```
sudo ls /proc/65159/ns -alh |grep net
lrwxrwxrwx 1 lxd tianyang 0 Jul 11 22:55 net -> net:[4026532895]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,8 +382,9 @@ Init process flags:
   - Default: `tini`
 
 Network flags:
-- :whale: `--net, --network=(bridge|host|none|<CNI>)`: Connect a container to a network
+- :whale: `--net, --network=(bridge|host|none|container:<container>|<CNI>)`: Connect a container to a network.
   - Default: "bridge"
+  - 'container:<name|id>': reuse another container's network stack, container has to be precreated.
   - :nerd_face: Unlike Docker, this flag can be specified multiple times (`--net foo --net bar`)
 - :whale: `-p, --publish`: Publish a container's port(s) to the host
 - :whale: `--dns`: Set custom DNS servers
--- a/cmd/nerdctl/restart.go
+++ b/cmd/nerdctl/restart.go
@@ -66,7 +66,7 @@ func restartAction(cmd *cobra.Command, args []string) error {
 			if err := stopContainer(ctx, found.Container, timeout); err != nil {
 				return err
 			}
-			if err := startContainer(ctx, found.Container, false); err != nil {
+			if err := startContainer(ctx, found.Container, false, client); err != nil {
 				return err
 			}
 			_, err = fmt.Fprintf(cmd.OutOrStdout(), "%s\n", found.Req)
@@ -81,5 +81,6 @@ func restartAction(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("no such container %s", req)
 		}
 	}
+
 	return nil
 }
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -123,7 +123,7 @@ func setCreateFlags(cmd *cobra.Command) {
 
 	// #region network flags
 	// network (net) is defined as StringSlice, not StringArray, to allow specifying "--network=cni1,cni2"
-	cmd.Flags().StringSlice("network", []string{netutil.DefaultNetworkName}, `Connect a container to a network ("bridge"|"host"|"none"|<CNI>)`)
+	cmd.Flags().StringSlice("network", []string{netutil.DefaultNetworkName}, `Connect a container to a network ("bridge"|"host"|"none"|"container:<container>"|<CNI>)`)
 	cmd.RegisterFlagCompletionFunc("network", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return shellCompleteNetworkNames(cmd, []string{})
 	})
@@ -548,12 +548,6 @@ func createContainer(cmd *cobra.Command, ctx context.Context, client *containerd
 	}
 	cOpts = append(cOpts, withStop(stopSignal, stopTimeout, ensuredImage))
 
-	netOpts, netSlice, ipAddress, ports, err := generateNetOpts(cmd, dataStore, stateDir, ns, id)
-	if err != nil {
-		return nil, nil, err
-	}
-	opts = append(opts, netOpts...)
-
 	hostname := id[0:12]
 	customHostname, err := cmd.Flags().GetString("hostname")
 	if err != nil {
@@ -571,6 +565,12 @@ func createContainer(cmd *cobra.Command, ctx context.Context, client *containerd
 		}
 		opts = append(opts, withCustomEtcHostname(hostnamePath))
 	}
+
+	netOpts, netSlice, ipAddress, ports, err := generateNetOpts(cmd, dataStore, stateDir, ns, id)
+	if err != nil {
+		return nil, nil, err
+	}
+	opts = append(opts, netOpts...)
 
 	hookOpt, err := withNerdctlOCIHook(cmd, id, stateDir)
 	if err != nil {
--- a/cmd/nerdctl/run_network.go
+++ b/cmd/nerdctl/run_network.go
@@ -23,12 +23,15 @@ import (
 	"io/fs"
 	"path/filepath"
 	"runtime"
+	"strings"
 
+	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/oci"
 	gocni "github.com/containerd/go-cni"
 	"github.com/containerd/nerdctl/pkg/dnsutil"
 	"github.com/containerd/nerdctl/pkg/dnsutil/hostsstore"
+	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
 	"github.com/containerd/nerdctl/pkg/mountutil"
 	"github.com/containerd/nerdctl/pkg/netutil"
 	"github.com/containerd/nerdctl/pkg/netutil/nettype"
@@ -164,10 +167,87 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
 				ports = append(ports, pm...)
 			}
 		}
+	case nettype.Container:
+		if err := verifyContainerNetwork(cmd, netSlice); err != nil {
+			return nil, nil, "", nil, err
+		}
+		network := strings.Split(netSlice[0], ":")
+		if len(network) != 2 {
+			return nil, nil, "", nil, fmt.Errorf("invalid network: %s, should be \"container:<id|name>\"", netSlice[0])
+		}
+		containerName := network[1]
+		client, ctx, cancel, err := newClient(cmd)
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
+		defer cancel()
+
+		walker := &containerwalker.ContainerWalker{
+			Client: client,
+			OnFound: func(ctx context.Context, found containerwalker.Found) error {
+				if found.MatchCount > 1 {
+					return fmt.Errorf("multiple containers found with prefix: %s", containerName)
+				}
+				containerID := found.Container.ID()
+
+				conStateDir, err := getContainerStateDirPath(cmd, dataStore, containerID)
+				if err != nil {
+					return err
+				}
+
+				s, err := found.Container.Spec(ctx)
+				if err != nil {
+					return err
+				}
+				hostname := s.Hostname
+				hostnamePath := filepath.Join(conStateDir, "hostname")
+				resolvConfPath := filepath.Join(conStateDir, "resolv.conf")
+				etcHostsPath := hostsstore.HostsPath(dataStore, ns, containerID)
+				netNSPath, err := getContainerNetNSPath(ctx, found.Container)
+				if err != nil {
+					return err
+				}
+				opts = append(opts,
+					oci.WithLinuxNamespace(specs.LinuxNamespace{
+						Type: specs.NetworkNamespace,
+						Path: netNSPath,
+					}),
+					withCustomResolvConf(resolvConfPath),
+					withCustomHosts(etcHostsPath),
+					oci.WithHostname(hostname),
+					withCustomEtcHostname(hostnamePath),
+				)
+				// stored in labels with key "nerdctl/networks"
+				netSlice = []string{fmt.Sprintf("container:%s", containerID)}
+				return nil
+			},
+		}
+		n, err := walker.Walk(ctx, containerName)
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
+		if n == 0 {
+			return nil, nil, "", nil, fmt.Errorf("no such container: %s", containerName)
+		}
 	default:
 		return nil, nil, "", nil, fmt.Errorf("unexpected network type %v", netType)
 	}
 	return opts, netSlice, ipAddress, ports, nil
+}
+
+func getContainerNetNSPath(ctx context.Context, c containerd.Container) (string, error) {
+	task, err := c.Task(ctx, nil)
+	if err != nil {
+		return "", err
+	}
+	status, err := task.Status(ctx)
+	if err != nil {
+		return "", err
+	}
+	if status.Status != containerd.Running {
+		return "", fmt.Errorf("invalid target container: %s, should be running", c.ID())
+	}
+	return fmt.Sprintf("/proc/%d/ns/net", task.Pid()), nil
 }
 
 func verifyCNINetwork(cmd *cobra.Command, netSlice []string) error {
@@ -189,6 +269,27 @@ func verifyCNINetwork(cmd *cobra.Command, netSlice []string) error {
 		if !ok {
 			return fmt.Errorf("network %s not found", netstr)
 		}
+	}
+	return nil
+}
+func verifyContainerNetwork(cmd *cobra.Command, netSlice []string) error {
+	if cmd.Flags().Changed("publish") {
+		return fmt.Errorf("conflicting options: port publishing and the container type network mode")
+	}
+	if cmd.Flags().Changed("hostname") {
+		return fmt.Errorf("conflicting options: hostname and the network mode")
+	}
+	if cmd.Flags().Changed("dns") {
+		return fmt.Errorf("conflicting options: dns and the network mode")
+	}
+	if cmd.Flags().Changed("add-host") {
+		return fmt.Errorf("conflicting options: custom host-to-IP mapping and the network mode")
+	}
+	if runtime.GOOS != "linux" {
+		return fmt.Errorf("currently '--network=container:<container>' can only works on linux")
+	}
+	if len(netSlice) > 1 {
+		return fmt.Errorf("only one network allowed using '--network=container:<container>'")
 	}
 	return nil
 }
--- a/cmd/nerdctl/run_network_linux_test.go
+++ b/cmd/nerdctl/run_network_linux_test.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"net"
 	"regexp"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -496,4 +497,42 @@ func TestRunDNS(t *testing.T) {
 	cmd.AssertOutContains("nameserver 8.8.8.8\n")
 	cmd.AssertOutContains("search test\n")
 	cmd.AssertOutContains("options attempts:10\n")
+}
+
+func TestSharedNetworkStack(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		t.Skip("--network=container:<container name|id> only supports linux now")
+	}
+	base := testutil.NewBase(t)
+
+	containerName := testutil.Identifier(t)
+	defer base.Cmd("rm", "-f", containerName).AssertOK()
+	base.Cmd("run", "-d", "--name", containerName,
+		testutil.NginxAlpineImage).AssertOK()
+	base.EnsureContainerStarted(containerName)
+
+	containerNameJoin := testutil.Identifier(t) + "-nework"
+	defer base.Cmd("rm", "-f", containerNameJoin).AssertOK()
+	cmd := base.Cmd("run",
+		"-d",
+		"--name", containerNameJoin,
+		"--network=container:"+containerName,
+		testutil.CommonImage,
+		"sleep", "infinity")
+	cmd.AssertOK()
+
+	base.Cmd("exec", containerNameJoin, "wget", "-qO-", "http://127.0.0.1:80").
+		AssertOutContains(testutil.NginxAlpineIndexHTMLSnippet)
+
+	base.Cmd("restart", containerName).AssertOK()
+	base.Cmd("restart", containerNameJoin).AssertOK()
+	base.Cmd("exec", containerNameJoin, "wget", "-qO-", "http://127.0.0.1:80").
+		AssertOutContains(testutil.NginxAlpineIndexHTMLSnippet)
+
+	base.Cmd("restart", containerName).AssertOK()
+	base.Cmd("stop", containerNameJoin).AssertOK()
+	base.Cmd("start", containerNameJoin).AssertOK()
+	base.Cmd("exec", containerNameJoin, "wget", "-qO-", "http://127.0.0.1:80").
+		AssertOutContains(testutil.NginxAlpineIndexHTMLSnippet)
+
 }
--- a/cmd/nerdctl/start.go
+++ b/cmd/nerdctl/start.go
@@ -18,16 +18,21 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/cmd/ctr/commands"
+	"github.com/containerd/containerd/oci"
 	"github.com/containerd/nerdctl/pkg/formatter"
 	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
 	"github.com/containerd/nerdctl/pkg/labels"
+	"github.com/containerd/nerdctl/pkg/netutil/nettype"
+	"github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -72,7 +77,7 @@ func startAction(cmd *cobra.Command, args []string) error {
 			if found.MatchCount > 1 {
 				return fmt.Errorf("multiple IDs found with provided prefix: %s", found.Req)
 			}
-			if err := startContainer(ctx, found.Container, flagA); err != nil {
+			if err := startContainer(ctx, found.Container, flagA, client); err != nil {
 				return err
 			}
 			if !flagA {
@@ -95,11 +100,16 @@ func startAction(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func startContainer(ctx context.Context, container containerd.Container, flagA bool) error {
+func startContainer(ctx context.Context, container containerd.Container, flagA bool, client *containerd.Client) error {
 	lab, err := container.Labels(ctx)
 	if err != nil {
 		return err
 	}
+
+	if err := reconfigNetContainer(ctx, container, client, lab); err != nil {
+		return err
+	}
+
 	taskCIO := cio.NullIO
 
 	// Choosing the user selected option over the labels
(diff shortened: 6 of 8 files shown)


TianZong48:
@AkihiroSuda PTAL, I'm trying to implement this feature and currently this is just a draft PR as a POC. 
Just wabt to make sure this is the right way to do it. 😄

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +181,44 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+			if err != nil {
+				return nil, nil, "", nil, err
+
+			} else if n == 0 {
+				return nil, nil, "", nil, fmt.Errorf("no such object: %s", containerName)
+			}
+			logrus.Info(f.entries...)
+			container, ok := f.entries[0].(*dockercompat.Container)
+			if !ok {
+				logrus.Fatal("something wrong")
```
AkihiroSuda, line 214:
```suggestion
				logrus.Fatalf("expected *dockercompat.Container, got %T", f.entries[0])
```

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +181,44 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+			}
+			logrus.Infof("sharing network namespace with %d", container.State.Pid)
```
AkihiroSuda, line 216:
```suggestion
			logrus.Debugf("sharing network namespace with %d", container.State.Pid)
```

cmd/nerdctl/run.go
```diff
@@ -123,7 +123,7 @@ func setCreateFlags(cmd *cobra.Command) {
 
 	// #region network flags
 	// network (net) is defined as StringSlice, not StringArray, to allow specifying "--network=cni1,cni2"
-	cmd.Flags().StringSlice("network", []string{netutil.DefaultNetworkName}, `Connect a container to a network ("bridge"|"host"|"none"|<CNI>)`)
+	cmd.Flags().StringSlice("network", []string{netutil.DefaultNetworkName}, `Connect a container to a network ("bridge"|"host"|"none"|"container:<container>"|<CNI>)`)
```
AkihiroSuda, line 126: Please update the corresponding line in the README.md too

AkihiroSuda: It should be also noted that the `container` has to be precreated but `nerdctl compose up` and the restart manager do not care that dependency

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +181,44 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
 				ports = append(ports, pm...)
 			}
 		}
+	case nettype.Container:
+		logrus.Info("NETWORK CONTAINER", netSlice)
```
AkihiroSuda, line 185:
```suggestion
		logrus.Debug("NETWORK CONTAINER", netSlice)
```

cmd/nerdctl/run.go
TianZong48: Thanks for reviewing! and yes, will do all the docs and tests

cmd/nerdctl/run_network.go
```diff
@@ -106,7 +109,21 @@ func withCustomHosts(src string) func(context.Context, oci.Client, *containers.C
 		return nil
 	}
 }
+func withNetworkNamespace(pid int) func(context.Context, oci.Client, *containers.Container, *oci.Spec) error {
+	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *oci.Spec) error {
 
+		for i, v := range s.Linux.Namespaces {
+			if v.Type == specs.NetworkNamespace {
+				s.Linux.Namespaces = append(s.Linux.Namespaces[:i], s.Linux.Namespaces[i+1:]...)
```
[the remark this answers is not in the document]
TianZong48: > If a namespace of the same type already exists in the spec, the existing namespace is replaced by the one provided.

just found there's a func `oci.WithLinuxNamespace` already, will use that one. the quote above is from It's comments as well :D

pkg/ocihook/ocihook.go
```diff
@@ -158,6 +158,8 @@ func newHandlerOpts(state *specs.State, dataStore, cniPath, cniNetconfPath strin
 		if err != nil {
 			return nil, err
 		}
+	case nettype.Container:
```
Zheaoli, line 161: Why should we ignore this type?

pkg/netutil/nettype/nettype.go
```diff
@@ -25,25 +28,31 @@ const (
...
+	Container: "container",
 }
 
 func Detect(names []string) (Type, error) {
 	var res Type
 
 	for _, name := range names {
 		var tmp Type
-		switch name {
+		switch strings.SplitN(name, ":", 2)[0] {
```
Zheaoli, line 47:
```suggestion
                 // In some circumstances, people may use --network=container:<container> to share the network namespace, so we need some special trick.
                 networkName := strings.SplitN(name, ":", 2)[0]
		switch networkName {
```

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +167,48 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
 				ports = append(ports, pm...)
 			}
 		}
+	case nettype.Container:
+		if runtime.GOOS != "linux" {
```
Zheaoli, line 171: I'm not sure whether we need to support this feature in rootless mode or not? cc @AkihiroSuda

README.md
```diff
@@ -383,7 +383,7 @@ Init process flags:
   - Default: `tini`
 
 Network flags:
-- :whale: `--net, --network=(bridge|host|none|<CNI>)`: Connect a container to a network
+- :whale: `--net, --network=(bridge|host|none|container:<container>|<CNI>)`: Connect a container to a network
```
Zheaoli, line 386: Need extra docs to describe the sharing network ns feature

Zheaoli (changes requested):
And we need some e2e test

pkg/ocihook/ocihook.go
TianZong48: I moved this line to LINE 138 following `case nettype.Host, nettype.None`, It just means no special OCIHook need when `createRuntime` and `postStop`

https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/run.go#L891-L913

cmd/nerdctl/run_network.go
AkihiroSuda: Why not?

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +167,51 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+			return nil, nil, "", nil, fmt.Errorf("invalid network: %s, should be \"container:<id|name>\"", netSlice[0])
+		}
+		containerName := network[1]
+		client, ctx, cancel, err := newClient(cmd)
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
+		defer cancel()
+		f := &containerInspector{
+			mode: "dockercompat",
```
junnplus, line 189: why not "native"?

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +167,51 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+			return nil, nil, "", nil, fmt.Errorf("multiple IDs found with provided prefix: %s", containerName)
+		}
+		container, ok := f.entries[0].(*dockercompat.Container)
+		if !ok {
+			return nil, nil, "", nil, fmt.Errorf("expected *dockercompat.Container, got %T", f.entries[0])
+		}
+
+		opts = append(opts, oci.WithLinuxNamespace(specs.LinuxNamespace{
+			Type: specs.NetworkNamespace,
+			Path: fmt.Sprintf("/proc/%d/ns/net", container.State.Pid),
```
junnplus, line 212: IMHO, we shouldn't write directly to the network namespace, it should be joined when oci hooking

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +167,51 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
+		if n == 0 {
+			return nil, nil, "", nil, fmt.Errorf("no such container: %s", containerName)
+		}
+		if n > 1 {
+			return nil, nil, "", nil, fmt.Errorf("multiple IDs found with provided prefix: %s", containerName)
+		}
+		container, ok := f.entries[0].(*dockercompat.Container)
```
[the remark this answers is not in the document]
TianZong48: not quite sure which one I should use, will change to `native` since `dockercompat` is basically converted from `native`

cmd/nerdctl/run_network.go
TianZong48: > why not "native"?

not quite sure which one I should use, will change to native since dockercompat is basically converted from native

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +167,51 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
 		}
+	case nettype.Container:
+		if runtime.GOOS != "linux" {
+			return nil, nil, "", nil, fmt.Errorf("currently '--network=container:<container>' can only works on linux")
+		}
+		if rootlessutil.IsRootless() {
+			return nil, nil, "", nil, fmt.Errorf("currently '--network=container:<container>' can't run rootless'")
+		}
+
+		network := strings.Split(netSlice[0], ":")
```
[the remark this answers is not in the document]
junnplus: I mean specify multi container network, like `--net container:xxx --net container:xxx`

pkg/ocihook/ocihook.go
Zheaoli: SGTM

cmd/nerdctl/run_network.go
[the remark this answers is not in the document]
jschwinger233: > I mean specify multi container network, like `--net container:xxx --net container:xxx`

I think one process has only one netns? A 2nd call of setns(2) will counteract the 1st setns(2).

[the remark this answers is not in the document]
Zheaoli: > I think one process has only one netns? A 2nd call of setns(2) will counteract the 1st setns(2).

Yes, one process should be related with one net an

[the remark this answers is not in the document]
TianZong48: > I mean specify multi container network, like `--net container:xxx --net container:xxx`

Currently, only the first one will be used. Or do you think we should use the last one?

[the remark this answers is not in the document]
junnplus: > Should abort and return error to avoid ambiguity.

Yep, I mean that too.

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +168,57 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+			return nil, nil, "", nil, fmt.Errorf("multiple IDs found with provided prefix: %s", containerName)
+		}
+
+		container, ok := f.entries[0].(*native.Container)
+		if !ok {
+			return nil, nil, "", nil, fmt.Errorf("expected *native.Container, got %T", f.entries[0])
+		}
+		if container.Process.Status.Status != containerd.Running {
+			return nil, nil, "", nil, fmt.Errorf("need container %s running", containerName)
+		}
```
junnplus, line 215: Should be handled inside the walker

cmd/nerdctl/run_network.go
```diff
@@ -164,6 +168,57 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
+		defer cancel()
+		f := &containerInspector{
+			mode: "native",
+		}
+		walker := &containerwalker.ContainerWalker{
+			Client:  client,
+			OnFound: f.Handler,
```
junnplus, line 196: Don't use `containerInspector.Handler`
```suggestion
			OnFound: func(ctx context.Context, found containerwalker.Found) error {
			        if found.MatchCount > 1 {
			                return fmt.Errorf("multiple IDs found with provided prefix: %s.", containerName)
			        }
			        n, err := containerinspector.Inspect(ctx, found.Container)
			        if err != nil {
			                return err
			        }
			        if n.Process.Status.Status != containerd.Running {
			                return fmt.Errorf("cannot join network of a non running container: %s", found.Container.ID()) 
			        }
			        return nil
			},
```

cmd/nerdctl/run_network.go
junnplus: Move to https://github.com/containerd/nerdctl/blob/master/pkg/ocihook/ocihook.go#L161

TianZong48: TBH I don't really understand why we should not specify namespace here. Can you explain a little bit? 🙏

network namespace is deleted in this function as well,  when nettype is Host (https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/run_network.go#L139 ).

AkihiroSuda: OCI hook is preferable for supporting `nerdctl run --restart`

junnplus: If the reused container is restarted, the pid will change, write dead pid will not be able to restart the container

junnplus: After thinking about it, only updating the network namespace during the oci hook cannot make the container join the network.

> OCI hook is preferable for supporting `nerdctl run --restart`

It seems that docker can't do this either

junnplus: In another way, we can support `nerdctl start` join the container network.
- nerdctl run -d --name a
- nerdctl run -d --net container:a --name b
- nerdctl restart a
- nerdctl restart b (update network namespace and restart container)

cmd/nerdctl/restart.go
```diff
@@ -66,6 +73,12 @@ func restartAction(cmd *cobra.Command, args []string) error {
 			if err := stopContainer(ctx, found.Container, timeout); err != nil {
 				return err
 			}
+
+			if ls, err := found.Container.Labels(ctx); err == nil && strings.Contains(ls[labels.Networks], "container") {
+				if err := reconfigureNetContainer(ctx, found.Container, client); err != nil {
+					return err
+				}
+			}
```
junnplus, line 81: Move to `startContainer`, thanks

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +168,72 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+		client, ctx, cancel, err := newClient(cmd)
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
+		defer cancel()
+
+		path, err := getNetworkNamespace(ctx, client, containerName)
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
```
junnplus, line 195: We should save the container's ID in the label instead of the original args.

`container:test` -> `container:<id>`

cmd/nerdctl/restart.go
```diff
@@ -81,5 +94,44 @@ func restartAction(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("no such container %s", req)
 		}
 	}
+
+	return nil
+}
+
+func reconfigureNetContainer(ctx context.Context, c containerd.Container, client *containerd.Client) error {
```
junnplus, line 101: ditto

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +168,72 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+	case nettype.Container:
+		if runtime.GOOS != "linux" {
+			return nil, nil, "", nil, fmt.Errorf("currently '--network=container:<container>' can only works on linux")
+		}
+		if rootlessutil.IsRootless() {
+			return nil, nil, "", nil, fmt.Errorf("currently '--network=container:<container>' can't run rootless")
+		}
+		if len(netSlice) > 1 {
+			return nil, nil, "", nil, fmt.Errorf("only one network allowed using '--network=container:<container>'")
+		}
```
junnplus, line 180: the networked container does not work with port mapping, please check it.

cmd/nerdctl/restart.go
TianZong48: I think If fits better in  between `stop` and `start` here?  `startContainer` is also called when `create` + `start`, which already specified network namespace when the `create` command calls `createContainer`

junnplus: How do?
- nerdctl run -d --name a
- nerdctl run -d --net container:a --name b
- nerdctl restart a
- nerdctl stop b
- nerdctl start b

TianZong48: good catch!  will also add corresponding tests

cmd/nerdctl/run_network.go
TianZong48: 👌

TianZong48: the 5000:5000 need to be published(-o) with the `2f0` container.

checked with docker and prompted `conflicting options: port publishing and the container type network mode.`

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +168,83 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
 				ports = append(ports, pm...)
 			}
 		}
+	case nettype.Container:
+		if cmd.Flags().Changed("publish") {
+			return nil, nil, "", nil, fmt.Errorf("conflicting options: '--publish(-p)' and the '--network=container:<container>'")
```
junnplus, line 173:
```suggestion
			return nil, nil, "", nil, fmt.Errorf("conflicting options: port publishing and the container type network mode.")
```

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +168,83 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+		return "", err
+	}
+	n, err := containerinspector.Inspect(ctx, c)
+	if err != nil {
+		return "", err
+	}
+	if n.Process.Status.Status != containerd.Running {
+		return "", fmt.Errorf("can't join network of a non running container: %s", containerID)
+	}
+	pid := n.Process.Pid
```
junnplus, line 243:
```suggestion
	task, err := c.Task(ctx, nil)
	if err != nil {
	        return "", nil
	}
	status, err := task.Status(ctx)
	if err != nil {
	        return "", err
	}
	if status.Status != containerd.Running {
	        return "", fmt.Errorf("can't join network of a non running container: %s", containerID)
	}
	pid := task.Pid()
```

cmd/nerdctl/run_network_linux_test.go
```diff
@@ -497,3 +500,105 @@ func TestRunDNS(t *testing.T) {
...
+
+	assert.Assert(t, base.InspectContainer(containerNameJoin).State.Pid != pid2)
+	pid2 = base.InspectContainer(containerNameJoin).State.Pid
+	f2, err = os.Stat(fmt.Sprintf("/proc/%d/ns/net", pid2))
+	assert.NilError(t, err)
+
+	stat1, ok = f1.Sys().(*syscall.Stat_t)
+	assert.Equal(t, ok, true)
+	stat2, ok = f2.Sys().(*syscall.Stat_t)
+	assert.Equal(t, ok, true)
```
junnplus, line 572: Why do we need it?

TianZong48: want to check if containers share the same netns

cmd/nerdctl/start.go
```diff
@@ -167,6 +178,44 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
...
+func reconfigureNetContainer(ctx context.Context, c containerd.Container, client *containerd.Client) error {
+	c, err := client.LoadContainer(ctx, c.ID())
+	if err != nil {
+		return err
+	}
+	spec, err := c.Spec(ctx)
+	if err != nil {
+		return err
+	}
+	networksJSON := spec.Annotations[labels.Networks]
```
junnplus, line 190: use `container.Labels`

cmd/nerdctl/start.go
```diff
@@ -95,11 +100,17 @@ func startAction(cmd *cobra.Command, args []string) error {
...
 }
 
-func startContainer(ctx context.Context, container containerd.Container, flagA bool) error {
+func startContainer(ctx context.Context, container containerd.Container, flagA bool, client *containerd.Client) error {
 	lab, err := container.Labels(ctx)
 	if err != nil {
 		return err
 	}
+	if strings.Contains(lab[labels.Networks], "container") {
+		if err := reconfigureNetContainer(ctx, container, client); err != nil {
```
junnplus, line 109: `lab[labels.Networks]` should be as an argument.

cmd/nerdctl/run_network_linux_test.go
junnplus: Checking network connectivity is enough.

TianZong48: 👌

junnplus:
hostname and resolvconfig also need to be considered.

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +167,87 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+			Path: path,
+		}))
+
 	default:
 		return nil, nil, "", nil, fmt.Errorf("unexpected network type %v", netType)
 	}
 	return opts, netSlice, ipAddress, ports, nil
 }
 
+func getNetworkNamespace(ctx context.Context, client *containerd.Client, containerID string) (string, error) {
```
junnplus, line 230: To be honest, this is not a good function name, because it returns a `can't join network` error.

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +167,104 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+		var containerID string
+		walker := &containerwalker.ContainerWalker{
+			Client: client,
+			OnFound: func(ctx context.Context, found containerwalker.Found) error {
+				if found.MatchCount > 1 {
+					return fmt.Errorf("multiple containers found with prefix: %s", containerName)
+				}
+				containerID = found.Container.ID()
+				// share same hostname with the target container
+				cmd.Flags().Set("hostname", containerID[0:12])
```
junnplus, line 202:
```
# docker run --rm -it --domainname baz.cyphar.com --hostname test --net container:nginx alpine sh
docker: Error response from daemon: conflicting options: hostname and the network mode.
```
conflicting options: hostname and the network mode.`

TianZong48: will return an error for this situation.  `--hostname` and `--network=container:***` conflict

cmd/nerdctl/start.go
```diff
@@ -167,6 +177,44 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
...
+	}
+	netType, err := nettype.Detect(networks)
+	if err != nil {
+		return err
+	}
+	if netType == nettype.Container {
+		c, err := client.LoadContainer(ctx, c.ID())
+		if err != nil {
+			return err
+		}
```
junnplus, line 195: Why reload?

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +167,104 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
 		return nil, nil, "", nil, fmt.Errorf("unexpected network type %v", netType)
 	}
 	return opts, netSlice, ipAddress, ports, nil
 }
 
+func getTargetNetNSPath(ctx context.Context, client *containerd.Client, containerID string) (string, error) {
+	c, err := client.LoadContainer(ctx, containerID)
+	if err != nil {
+		return "", err
+	}
```
junnplus, line 251: reload again?

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +167,104 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+
+		path, err := getTargetNetNSPath(ctx, client, containerID)
+		if err != nil {
+			return nil, nil, "", nil, err
+		}
+
+		opts = append(opts, oci.WithLinuxNamespace(specs.LinuxNamespace{
+			Type: specs.NetworkNamespace,
+			Path: path,
+		}))
```
junnplus, line 239: Why not put it to `walker.OnFound`?

cmd/nerdctl/run_network.go
junnplus: Why need it yet?

TianZong48: Not quite sure what you are referring to, the code here is for containers using the same hostname. docker did the same.

```
# docker run -d --name ng nginx:1.19-alpine
# docker run -d -it --name join  --network=container:ng alpine:3.13 sleep infinity
# docker exec ng cat /etc/hostname
7968dccea5f1
# docker exec join cat /etc/hostname
7968dccea5f1

```

junnplus: > will use the target container's hostname instead of Its ID.

Incorrect, they should share the same etchosts file.

TianZong48: my experiment above shows they shared `/etc/hostname` too in docker, and `/etc/hosts` is already using the same file by `withCustomHosts(etcHostsPath)`

junnplus: > Why need it yet?

I don't understand why we need it

junnplus: > my experiment above shows they shared /etc/hostname too in docker, and /etc/hosts is already using(mount) the same file by withCustomHosts(etcHostsPath) in walker 😄

Sorry, using `buildResolvConf` and `AllocHostsFile` confuse me, we just need to share files, we don't need to create they.

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +168,85 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+				if found.MatchCount > 1 {
+					return fmt.Errorf("multiple containers found with prefix: %s", containerName)
+				}
+				containerID := found.Container.ID()
+				// share same hostname with the target container
+				lb, err := found.Container.Labels(ctx)
+				if err != nil {
+					return err
+				}
+				cmd.Flags().Set("hostname", lb[labels.Hostname])
```
junnplus, line 198: What is the use?

cmd/nerdctl/start.go
```diff
@@ -167,6 +177,45 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
...
+	var networks []string
+	if err := json.Unmarshal([]byte(networksJSON), &networks); err != nil {
+		return err
+	}
+	netType, err := nettype.Detect(networks)
+	if err != nil {
+		return err
+	}
+	if netType == nettype.Container {
+		targetConID := strings.Split(networks[0], ":")[1]
```
junnplus, line 192: Please check length

cmd/nerdctl/run_network.go
TianZong48: for setting up the hostname, It's kinda confusing here, maybe should add a comment?
https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/run.go#L557-L573

cmd/nerdctl/start.go
```diff
@@ -167,6 +177,45 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
...
+		targetCon, err := client.LoadContainer(ctx, targetConID)
+		if err != nil {
+			return err
+		}
+		NsPath, err := getContainerNetNSPath(ctx, targetCon)
```
junnplus, line 197:
```suggestion
		netNSPath, err := getContainerNetNSPath(ctx, targetCon)
```

cmd/nerdctl/run_network_linux_test.go
```diff
@@ -497,3 +498,55 @@ func TestRunDNS(t *testing.T) {
...
+		"--network=container:"+containerName,
+		testutil.CommonImage,
+		"sleep", "infinity")
+	cmd.AssertOK()
+
+	cmd = base.Cmd("exec", containerNameJoin, "wget", "-qO-", "http://127.0.0.1:80")
+	cmd.AssertOutContains(testutil.NginxAlpineIndexHTMLSnippet)
+
+	cmd = base.Cmd("exec", containerNameJoin, "nc", "127.0.0.1", "80", "-v")
+	cmd.AssertCombinedOutContains("open")
```
junnplus, line 531: Keep one enough

cmd/nerdctl/run_network.go
```diff
@@ -164,12 +167,89 @@ func generateNetOpts(cmd *cobra.Command, dataStore, stateDir, ns, id string) ([]
...
+					return err
+				}
+				opts = append(opts,
+					oci.WithLinuxNamespace(specs.LinuxNamespace{
+						Type: specs.NetworkNamespace,
+						Path: netNSPath,
+					}),
+					withCustomResolvConf(resolvConfPath),
+					withCustomHosts(etcHostsPath),
+					oci.WithHostname(hostname),
```
junnplus, line 217: Why not use `withCustomEtcHostname`?

TianZong48: used below

junnplus: OK

junnplus:
Please squash commits

junnplus:
@AkihiroSuda @Zheaoli PTAL

junnplus (approved):
LGTM

AkihiroSuda (approved):
Thanks

PR #1223, opened 2022-07-13
refactor container restart command

junnplus:

https://github.com/containerd/nerdctl/pull/1108 introduce attach flag cause restart command failed.
```
$ sudo nerdctl restart test
test
FATA[0000] flag accessed but not defined: attach
```

### DIFF
--- a/cmd/nerdctl/restart.go
+++ b/cmd/nerdctl/restart.go
@@ -17,6 +17,11 @@
 package main
 
 import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/restart_linux_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/containerd/nerdctl/pkg/testutil"
+	"gotest.tools/v3/assert"
+)
+
+func TestRestart(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+	tID := testutil.Identifier(t)
+
+	base.Cmd("run", "-d", "--name", tID, testutil.NginxAlpineImage).AssertOK()
```
manugupt1, line 31: Is it okay to use: `base.EnsureContainerStarted(testContainer)`
https://github.com/containerd/nerdctl/commit/a8ab5cdb3dda7cb64f9e4f964105243b4ab5ef27

cmd/nerdctl/restart_linux_test.go
```diff
@@ -0,0 +1,41 @@
...
+	defer base.Cmd("rm", "-f", tID).AssertOK()
+
+	inspect := base.InspectContainer(tID)
+	pid := inspect.State.Pid
+
+	base.Cmd("restart", tID).AssertOK()
```
manugupt1, line 37: Maybe provide `-t 1` to speed up the test

cmd/nerdctl/restart_linux_test.go
junnplus: make sense

cmd/nerdctl/restart_linux_test.go
junnplus: not essential

manugupt1:
Lgtm

cmd/nerdctl/restart.go
```diff
@@ -35,13 +40,43 @@ func newRestartCommand() *cobra.Command {
...
+	walker := &containerwalker.ContainerWalker{
+		Client: client,
+		OnFound: func(ctx context.Context, found containerwalker.Found) error {
+			if err := stopContainer(ctx, found.Container, timeout); err != nil {
+				return err
+			}
+			if err := startContainer(ctx, found.Container, false); err != nil {
+				return err
+			}
+			_, err = fmt.Fprintf(cmd.OutOrStdout(), "%s\n", found.Req)
```
TianZong48, line 69: what about print container ID? I think there cound be multiple container of from same `found.Req`

junnplus: see https://github.com/containerd/nerdctl/pull/1223/files#diff-5575f185a28781b4e743991089f023ba8d2a6b6385cf3fdfb77c38ba346806a4R63-R65

AkihiroSuda: Still we should print the full ID

junnplus: > Still we should print the full ID

Why not keep the same as docker?

AkihiroSuda: hmm Docker seems wrong, but fine to follow the current Docker's behavior

AkihiroSuda (approved):
Thanks

PR #1244, opened 2022-07-17
Add extra logging warning for log options

Zheaoli:

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -965,6 +965,9 @@ func parseKVStringsMapFromLogOpt(cmd *cobra.Command, logDriver string) (map[stri
 			delete(logOptMap, logging.MaxFile)
 		}
 	}
+	if err := logging.ValidateLogOpts(logDriver, logOptMap); err != nil {
+		return nil, err
+	}
 	return logOptMap, nil
 }
 
--- a/pkg/logging/fluentd_logger.go
+++ b/pkg/logging/fluentd_logger.go
@@ -29,7 +29,9 @@ import (
 	"time"
 
 	"github.com/containerd/containerd/runtime/v2/logging"
+	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/fluent/fluent-logger-golang/fluent"
+	"github.com/sirupsen/logrus"
 )
 
 type FluentdLogger struct {
@@ -47,6 +49,18 @@ const (
 	fluentRequestAck              = "fluentd-request-ack"
 )
 
+var FluentdLogOpts = []string{
+	fluentAddress,
+	fluentdAsync,
+	fluentdBufferLimit,
+	fluentdRetryWait,
+	fluentdMaxRetries,
+	fluentdSubSecondPrecision,
+	fluentdAsyncReconnectInterval,
+	fluentRequestAck,
+	Tag,
+}
+
 const (
 	defaultBufferLimit = 1024 * 1024
 	defaultHost        = "127.0.0.1"
@@ -59,6 +73,18 @@ const (
 	minReconnectInterval = 100 * time.Millisecond
 	maxReconnectInterval = 10 * time.Second
 )
+
+func FluentdLogOptsValidate(logOptMap map[string]string) error {
+	for key := range logOptMap {
+		if !strutil.InStringSlice(FluentdLogOpts, key) {
+			logrus.Warnf("log-opt %s is ignored for fluentd log driver", key)
+		}
+	}
+	if _, ok := logOptMap[fluentAddress]; !ok {
+		logrus.Warnf("%s is missing for fluentd log driver, the default value %s:%d will be used", fluentAddress, defaultHost, defaultPort)
+	}
+	return nil
+}
 
 type fluentdLocation struct {
 	protocol string
--- a/pkg/logging/json_logger.go
+++ b/pkg/logging/json_logger.go
@@ -25,12 +25,28 @@ import (
 
 	"github.com/containerd/containerd/runtime/v2/logging"
 	"github.com/containerd/nerdctl/pkg/logging/jsonfile"
+	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/docker/go-units"
 	"github.com/fahedouch/go-logrotate"
+	"github.com/sirupsen/logrus"
 )
+
+var JSONDriverLogOpts = []string{
+	MaxSize,
+	MaxFile,
+}
 
 type JSONLogger struct {
 	Opts map[string]string
+}
+
+func JSONFileLogOptsValidate(logOptMap map[string]string) error {
+	for key := range logOptMap {
+		if !strutil.InStringSlice(JSONDriverLogOpts, key) {
+			logrus.Warnf("log-opt %s is ignored for json-file log driver", key)
+		}
+	}
+	return nil
 }
 
 func (jsonLogger *JSONLogger) Init(dataStore, ns, id string) error {
--- a/pkg/logging/logging.go
+++ b/pkg/logging/logging.go
@@ -43,11 +43,21 @@ type Driver interface {
 }
 
 type DriverFactory func(map[string]string) (Driver, error)
+type LogOpsValidateFunc func(logOptMap map[string]string) error
 
 var drivers = make(map[string]DriverFactory)
+var driversLogOptsValidateFunctions = make(map[string]LogOpsValidateFunc)
 
-func RegisterDriver(name string, f DriverFactory) {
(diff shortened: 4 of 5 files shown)


pkg/logging/logging.go
```diff
@@ -37,6 +37,26 @@ const (
 	Tag        = "tag"
 )
 
+var FluentdBitLogOpts = []string{
```
fahedouch, line 40: `FluentdLogOpts`

cmd/nerdctl/run.go
```diff
@@ -966,6 +966,35 @@ func parseKVStringsMapFromLogOpt(cmd *cobra.Command, logDriver string) (map[stri
 		if _, ok := logOptMap[logging.MaxSize]; !ok {
 			delete(logOptMap, logging.MaxFile)
 		}
+		for _, value := range logging.FluentdBitLogOpts {
+			if _, ok := logOptMap[value]; ok {
```
fahedouch, line 970: May be it is easier to make the opposite way  by checking the number of drivers and if more than 1 we log that only the first driver is considered

Zheaoli: emmm I'm not get the point of `checking the number of drivers `

cmd/nerdctl/run.go
```diff
@@ -966,6 +966,35 @@ func parseKVStringsMapFromLogOpt(cmd *cobra.Command, logDriver string) (map[stri
 		if _, ok := logOptMap[logging.MaxSize]; !ok {
 			delete(logOptMap, logging.MaxFile)
 		}
+		for _, value := range logging.FluentdLogOpts {
+			if _, ok := logOptMap[value]; ok {
+				logrus.Warnf("log-opt %s is ignored for json-file log driver", value)
+			}
+		}
```
junnplus, line 973: Why not check the other way around?

```suggestion
		for opt := range logOptMap {
			if strutil.InStringSlice(logging.JSONDriverLogOpts, opt) {
			        logrus.Warnf("log-opt %q is ignored for json-file log driver", opt)
			}
		}
```

Zheaoli: The hasmap is a O(1) lookup action(

It's just personal style(

pkg/logging/logging.go
```diff
@@ -37,6 +37,26 @@ const (
...
+
+var JSONDriverLogOpts = []string{
+	MaxSize,
+	MaxFile,
+}
+
+var JournalDriverLogOpts = []string{
+	Tag,
+}
+
```
fahedouch, line 59: each `XXLogOpts` should be moved in `XX_logger.go`

cmd/nerdctl/run.go
```diff
@@ -966,6 +966,35 @@ func parseKVStringsMapFromLogOpt(cmd *cobra.Command, logDriver string) (map[stri
...
+			if _, ok := logOptMap[value]; ok {
+				logrus.Warnf("log-opt %s is ignored for journald log driver", value)
+			}
+		}
+		for _, value := range logging.FluentdLogOpts {
+			if _, ok := logOptMap[value]; ok {
+				logrus.Warnf("log-opt %s is ignored for journald log driver", value)
+			}
+		}
+	}
```
fahedouch, line 991:
```suggestion
	if logDriver == "journald" {
		for opt, _ := range logOptMap {
			if _, ok := logging.JournalDriverLogOpts[opt]; !ok {
				logrus.Warnf("log-opt %s is ignored for journald log driver", opt)
			}
		}
	}
```

this need to convert this (https://github.com/containerd/nerdctl/pull/1244/files#diff-eb19031b20f577302a6bfbcbb42ee60f37256f9ed2044d905a7aa681309bca69R40)  to a map 

```
var FluentdLogOpts  = map[string]struct{}{}
```

cmd/nerdctl/run.go
fahedouch: plz take a look https://github.com/containerd/nerdctl/pull/1244/files#r922888056

cmd/nerdctl/run.go
Zheaoli: Oh, I got your point, but I think this covert would be unnecessary? I guess?

`xxxLogOpts` could just be a slice to store the option key.

But your way seems more friendly to read, I'm not sure about this

pkg/logging/logging.go
Zheaoli: OK

cmd/nerdctl/run.go
fahedouch: An other ++ of my suggestion is that we avoid changing this Block each time we add new driver

junnplus: I agree, `map[string]struct{}{}` or `[]string` are both ok, few parameters will not cause any performance issues

cmd/nerdctl/run.go
junnplus: We can't change it every time we add a new driver

pkg/logging/fluentd_logger.go
```diff
@@ -60,6 +74,18 @@ const (
 	maxReconnectInterval = 10 * time.Second
 )
 
+func FluentdLogOptsValidate(logOptMap map[string]string) error {
```
junnplus, line 77: `func (f *FluentdLogger) Validate`?

pkg/logging/json_logger.go
```diff
@@ -25,14 +25,30 @@ import (
...
+var JSONDriverLogOpts = []string{
+	MaxSize,
+	MaxFile,
+}
+
 type JSONLogger struct {
 	Opts map[string]string
 }
 
+func JSONFileLogOptsValidate(logOptMap map[string]string) error {
```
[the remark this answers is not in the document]
manugupt1: Since we only want to warn should we remove returning the error as it is always nil

pkg/logging/fluentd_logger.go
Zheaoli: I don't think making this function to be a struct method is a good idea. that means that the run.go need to init the struct , I don't think this would be necessary

pkg/logging/json_logger.go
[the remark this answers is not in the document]
Zheaoli: > Why are we not returning error in any of the validate functions?

For now, all three log driver has had fallback logic to maintain the default value, and all three log driver don't validate the argument input format yet. But in the future, some new log drivers may need to return some errors during the validation process. So I think it's good to define `type LogOptValidate func(map[string]string) error`, and we impl the different validation logic for different log drivers. For now we just check the argument key and output some warning (

pkg/logging/logging.go
```diff
@@ -44,6 +44,14 @@ type Driver interface {
 
 type DriverFactory func(map[string]string) (Driver, error)
 
+type LogOptValidate func(map[string]string) error
```
fahedouch, line 47: `LogOptValidateFunc`

cmd/nerdctl/run.go
```diff
@@ -965,6 +965,11 @@ func parseKVStringsMapFromLogOpt(cmd *cobra.Command, logDriver string) (map[stri
 			delete(logOptMap, logging.MaxFile)
 		}
 	}
+	if value, ok := logging.LogOptValidateFuncs[logDriver]; ok {
+		if err := value(logOptMap); err != nil {
+			return nil, err
+		}
+	}
```
fahedouch, line 972: Not really a Fun of putting funcs in Map. I think we can expose this logic by a switch case
I would recommend sthg like this (just a pseudo code) :

```suggestion
switch logDriver {
	case "json-file" :
		LogOptValidateFunc = logging.JSONFileLogOptsValidate
	case "journald":
	        LogOptValidateFunc = logging.JournalLogOptsValidate
}

logging.LogOptValidateFunc(logOptMap)

```

Zheaoli:
@fahedouch Hi, Would you mind to take a look at again if you got some time, thx

pkg/logging/logging.go
```diff
@@ -44,6 +44,18 @@ type Driver interface {
...
 
+func ValidateLogOpts(logDriver string, logOpts map[string]string) error {
+	switch logDriver {
+	case "json-file":
+		return JSONFileLogOptsValidate(logOpts)
+	case "journald":
+		return JournalLogOptsValidate(logOpts)
+	case "fluentd":
+		return FluentdLogOptsValidate(logOpts)
+	}
```
AkihiroSuda, line 55: Driver names shouldn’t be hard coded

see https://github.com/containerd/nerdctl/blob/3f25af7dc8e3843641fce3b1a586c14d6f7aedff/pkg/logging/logging.go#L49

AkihiroSuda (approved):
Thanks

PR #948, opened 2022-03-28
Support --init argument in nerdctl run command

Zheaoli:
FIX: #946 

Signed-off-by: Zheao.Li <me@manjusaka.me>

PR #1256, opened 2022-07-22
rewrite test of run with init flag

junnplus:

`TestRunWithInitAndInitPath` and `TestRunWithInitWithoutInitPath` are invalid test, run image without `--init` alway output `signal: 15`.

```
242         base.Cmd("run", "-d", "--name", containerName, imageName, "--init").AssertOK()
242         base.Cmd("run", "-d", "--name", containerName, imageName).AssertOK()
```

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -783,7 +783,6 @@ func generateRootfsOpts(ctx context.Context, client *containerd.Client, platform
 		initProcessFlag = true
 	}
 	if initProcessFlag {
-		binaryNameInContainer := ""
 		binaryPath, err := exec.LookPath(initBinary)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


fuweid (approved):

AkihiroSuda (approved):

Zheaoli:
In the origin test I write, I build a binary and make a signal forwarding test because we are not sure if the container exit because it is timeout since the stop command or the binary handles the right signal and exists itself. So we need to add a binary to output the signal they received to guarantee the signal forwarding is as we excepted

Zheaoli:
https://github.com/containerd/nerdctl/pull/948#discussion_r837479668

junnplus:
IMHO, just testing the signal forwarding doesn't make sure `--init` is working, signal forwarding is something that needs to be considered after this(or tini itself guarantees).

PR #1260, opened 2022-07-23
add label display support in ps subcommand

minuk-dev:
Signed-off-by: Min Uk Lee <minuk.dev@gmail.com>

Some projects like [kind](https://github.com/kubernetes-sigs/kind) uses labels to distinguish containers.
(e.g. https://github.com/kubernetes-sigs/kind/blob/982df26f902085469d18e109a245965aa81b7f32/pkg/cluster/internal/providers/docker/provision.go#L139,
https://github.com/kubernetes-sigs/kind/blob/982df26f902085469d18e109a245965aa81b7f32/pkg/cluster/internal/providers/docker/provider.go#L105)
Therefore, need to display labels.

Example usage

```bash
$ nerdctl run --label key1=value1,key2=value2 hello-world
# skip output
$ nerdctl ps -a --format=json
# {"Command":"\"/hello\"","CreatedAt":"2022-07-23 18:56:54 +0000 UTC","ID":"bf6dda95ce71","Image":"docker.io/library/hello-world:latest","Platform":"linux/arm64/v8","Names":"hello-world-bf6dd","Ports":"","Status":"Exited (0) 35 seconds ago","Runtime":"io.containerd.runc.v2","Size":"","Labels":"key1=value1,key2=value2,io.containerd.image.config.stop-signal=SIGTERM"}
```

### DIFF
--- a/cmd/nerdctl/ps.go
+++ b/cmd/nerdctl/ps.go
@@ -115,7 +115,8 @@ type containerPrintable struct {
 	Status    string
 	Runtime   string // nerdctl extension
 	Size      string
-	// TODO: "Labels", "LocalVolumes", "Mounts", "Networks", "RunningFor",  "State"
+	Labels    string
+	// TODO: "LocalVolumes", "Mounts", "Networks", "RunningFor", "State"
 }
 
 func printContainers(ctx context.Context, client *containerd.Client, cmd *cobra.Command, containers []containerd.Container, all bool) error {
@@ -216,6 +217,7 @@ func printContainers(ctx context.Context, client *containerd.Client, cmd *cobra.
 			Ports:     formatter.FormatPorts(info.Labels),
 			Status:    cStatus,
 			Runtime:   info.Runtime.Name,
+			Labels:    formatter.FormatLabels(info.Labels),
 		}
 
 		if size || wide {
--- a/cmd/nerdctl/ps_linux_test.go
+++ b/cmd/nerdctl/ps_linux_test.go
@@ -21,12 +21,19 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/containerd/nerdctl/pkg/formatter"
+	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/containerd/nerdctl/pkg/tabutil"
 	"github.com/containerd/nerdctl/pkg/testutil"
 	"gotest.tools/v3/assert"
 )
 
-func prepareTest(t *testing.T) (*testutil.Base, string) {
+type testContainer struct {
+	name   string
+	labels map[string]string
+}
+
+func prepareTest(t *testing.T) (*testutil.Base, testContainer) {
 	base := testutil.NewBase(t)
 
 	base.Cmd("pull", testutil.CommonImage).AssertOK()
@@ -36,11 +43,25 @@ func prepareTest(t *testing.T) (*testutil.Base, string) {
 		base.Cmd("rm", "-f", testContainerName).AssertOK()
 	})
 
+	// A container can have multiple labels.
+	// Therefore, this test container has multiple labels to check it.
+	testLabels := make(map[string]string)
+	keys := []string{
+		testutil.Identifier(t),
+		testutil.Identifier(t),
+	}
+	// fill the value of testLabels
+	for _, k := range keys {
+		testLabels[k] = testutil.Identifier(t)
+	}
+
 	base.Cmd([]string{
 		"run",
 		"-d",
 		"--name",
 		testContainerName,
+		"--label",
+		formatter.FormatLabels(testLabels),
 		testutil.CommonImage,
 		"top",
 	}...).AssertOK()
@@ -50,11 +71,14 @@ func prepareTest(t *testing.T) (*testutil.Base, string) {
 	// let the container occupy 25MiB space.
 	base.Cmd("exec", testContainerName, "dd", "if=/dev/zero", "of=/test_file", "bs=1M", "count=25").AssertOK()
 
-	return base, testContainerName
+	return base, testContainer{
+		name:   testContainerName,
+		labels: testLabels,
+	}
 }
 
 func TestContainerList(t *testing.T) {
-	base, testContainerName := prepareTest(t)
+	base, testContainer := prepareTest(t)
 
 	// hope there are no tests running parallel
 	base.Cmd("ps", "-n", "1", "-s").AssertOutWithFunc(func(stdout string) error {
@@ -74,7 +98,7 @@ func TestContainerList(t *testing.T) {
 		}
 
 		container, _ := tab.ReadRow(lines[1], "NAMES")
-		assert.Equal(t, container, testContainerName)
+		assert.Equal(t, container, testContainer.name)
 
 		image, _ := tab.ReadRow(lines[1], "IMAGE")
 		assert.Equal(t, image, testutil.CommonImage)
@@ -97,7 +121,7 @@ func TestContainerList(t *testing.T) {
 
 func TestContainerListWideMode(t *testing.T) {
 	testutil.DockerIncompatible(t)
-	base, testContainerName := prepareTest(t)
+	base, testContainer := prepareTest(t)
 
 	// hope there are no tests running parallel
 	base.Cmd("ps", "-n", "1", "--format", "wide").AssertOutWithFunc(func(stdout string) error {
@@ -118,7 +142,7 @@ func TestContainerListWideMode(t *testing.T) {
 		}
 
 		container, _ := tab.ReadRow(lines[1], "NAMES")
-		assert.Equal(t, container, testContainerName)
+		assert.Equal(t, container, testContainer.name)
 
 		image, _ := tab.ReadRow(lines[1], "IMAGE")
 		assert.Equal(t, image, testutil.CommonImage)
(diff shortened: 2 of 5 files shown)


Zheaoli:
Thanks, would you mind to add some e
2e test to test if the features works as we expected?

cmd/nerdctl/ps.go
```diff
@@ -216,6 +217,7 @@ func printContainers(ctx context.Context, client *containerd.Client, cmd *cobra.
 			Ports:     formatter.FormatPorts(info.Labels),
 			Status:    cStatus,
 			Runtime:   info.Runtime.Name,
+			Labels:    formatter.FormatLabels(info.Labels),
```
Zheaoli, line 220: I think it would be better if we make these as an argument instead of default behavior such as **nerdctl ps --labels**

If the container has a lot of labels, I think it would be an info disaster in the terminal for user

Zheaoli (changes requested):

cmd/nerdctl/ps.go
minuk-dev: I referenced this doc.
https://github.com/containerd/nerdctl/blob/b5cdd8aa43c6b40502fc740a14682873574470f9/README.md?plain=1#L780
So, I didn't edit table and wide options.
```bash
nerdctl ps -a
CONTAINER ID    IMAGE                                   COMMAND     CREATED        STATUS                    PORTS    NAMES
a03dccba94d1    docker.io/library/hello-world:latest    "/hello"    2 hours ago    Exited (0) 2 hours ago             hello-world-a03dc
bf6dda95ce71    docker.io/library/hello-world:latest    "/hello"    2 hours ago    Exited (0) 2 hours ago             hello-world-bf6dd
c5b8555bad9c    docker.io/library/hello-world:latest    "/hello"    2 hours ago    Exited (0) 2 hours ago             hello-world-c5b85
```
I'm worried that too many options will be created.

My suggestion is
- without any option, display all
- provide an interface to edit it.

minuk-dev:
> Thanks, would you mind to add some e 2e test to test if the features works as we expected?

@Zheaoli Could you review my test code, please?
I'm not sure that the test code should contain all members of `containerPrintable`.
So, just I tested only `--format {{.Label}}" option. (e.g. `ps -n 1 --format {{.Labels}}`)

Test results
- [pass test-integration](https://github.com/containerd/nerdctl/runs/7488464701?check_suite_focus=true#step:6:485)
- [pass test-integration-docker-compatibility](https://github.com/containerd/nerdctl/runs/7488464531?check_suite_focus=true#step:5:395)

I think that other failed tests are caused by timeout.

cmd/nerdctl/ps.go
Zheaoli: SGTM

Zheaoli (approved):
LGTM on all tests green

cmd/nerdctl/ps_linux_test.go
```diff
@@ -36,11 +43,22 @@ func prepareTest(t *testing.T) (*testutil.Base, string) {
 		base.Cmd("rm", "-f", testContainerName).AssertOK()
 	})
 
+	testLabels := make(map[string]string)
+	keys := []string{
+		testutil.Identifier(t),
+		testutil.Identifier(t),
+	}
```
AkihiroSuda, line 50: Why repeat same key twice?

pkg/formatter/formatter.go
```diff
@@ -135,3 +135,16 @@ func FormatPorts(labelMap map[string]string) string {
 func TimeSinceInHuman(since time.Time) string {
 	return fmt.Sprintf("%s ago", units.HumanDuration(time.Since(since)))
 }
+
+func FormatLabels(labelMap map[string]string) string {
+	strs := make([]string, 0, len(labelMap))
+	for i := range labelMap {
+		// exclude labels of nerdctl
```
AkihiroSuda, line 142: Why?

cmd/nerdctl/ps_linux_test.go
minuk-dev: I think multiple labels can exist.
For example, `--label key1=value1,key2=value2`
So, the test makes multiple labels.

pkg/formatter/formatter.go
minuk-dev: When a container has multiple labels, there is no constraint of labels' order.
So, the test should check labels using map.
And, current version cannot guarantee to display only user defined labels. 
For example, my examples display`io.containerd.image.config.stop-signal=SIGTERM` when the containers are terminated.
In other words, the result of ps command can contains internal labels to control them internally.

cmd/nerdctl/ps_linux_test.go
AkihiroSuda: Could you clarify that intent in the comment lines of the code?

pkg/formatter/formatter.go
AkihiroSuda: I'm not sure we want to hide nerdctl labels

minuk-dev: Me, too.
But, I don't think internal important info are exposed throught labels (like `namespace`, `network` and `ports`).
I understood that `Prefix` exists for it.
Is it the right way to expose them?

minuk-dev: I'll try to find other examples and docker running way.

minuk-dev: In [docker cli](https://github.com/docker/cli/blob/9cda6ce0114b64052eb656b4975ab03acb8cb371/cli/command/formatter/container.go#L239), it doesn't filter any labels.
But, the reason why it can do is that [moby](https://github.com/moby/moby/blob/98d8343aa28de0d499464b5529e6b8ccc92e9313/api/types/types.go#L152) containers have their various properties (like `Ports` and `NetworkSettings`).
nerdctl cannot do like it. Because it internally saves these kinds of values into Labels(like [this](https://github.com/containerd/nerdctl/blob/b5cdd8aa43c6b40502fc740a14682873574470f9/cmd/nerdctl/ps.go#L216)).
For these implementations, `Prefix` is defined (we can see from [this commit](https://github.com/containerd/nerdctl/commit/7cccafe5c2f684a959f8f0ee7a9e3b014d239235)).
Eventually, I guess that `Prefix` can be understood for only internal implementation.
Users should not know how to implement functions (like ports, networks and so on).

My suggestions are
- Hide labels with `Prefix`
- Explicitly comment "`Prefix` is for internal labels so they will be hide"

minuk-dev: I really agree we are not sure which is correct.

In [docker cli](https://github.com/docker/cli/blob/9cda6ce0114b64052eb656b4975ab03acb8cb371/cli/command/formatter/container.go#L239), it doesn't filter any labels.
But, the reason why it can do is that [moby](https://github.com/moby/moby/blob/98d8343aa28de0d499464b5529e6b8ccc92e9313/api/types/types.go#L152) containers have their various properties (like `Ports` and `NetworkSettings`).
nerdctl cannot do like it. Because it internally saves these kinds of values into Labels(like [this](https://github.com/containerd/nerdctl/blob/b5cdd8aa43c6b40502fc740a14682873574470f9/cmd/nerdctl/ps.go#L216)).
For these implementations, `Prefix` is defined (we can see from [this commit](https://github.com/containerd/nerdctl/commit/7cccafe5c2f684a959f8f0ee7a9e3b014d239235)).
Eventually, I guess that `Prefix` can be understood for only internal implementation.
Users should not know how to implement functions (like ports, networks and so on).

---
In my opinion, if we don't hide any label, it can be abused to know internal information.
My suggestions are
- Hide labels with `Prefix`
- Explicitly comment "`Prefix` is for internal labels so they will be hide"

cmd/nerdctl/ps_linux_test.go
minuk-dev: I commented.
Could you please check again?

I'm really sorry about it. That's all my fault.

pkg/formatter/formatter.go
Zheaoli: I agree with @minuk-dev 

The final containerd containers created by nerdctl have some labels because the nerdctl needs those labels to interact with the OCI hook. That means that those labels represent some internal status. I think it would not be necessary to explore those labels for the user, and the user shouldn't care about how those labels work.

Zheaoli: For a personal suggestion: maybe we can let people decide what they want to do? Here are the details 

1. The default behavior is that we will filter the internal labels

2. Add an argument --all-labels to allow the user to use this to display all the labels, including the internal labels

AkihiroSuda: Hiding internal labels can be discussed in a separate issue/ PR.

Having --all-labels is probably fine but it has to be consistent with other "inspect" comnands

minuk-dev: Okay, should I remove filtering internal label and push again?

TianZong48: I think we can hide labels by default, and print when `--labels`.

For internal labels, we can keep them for now.
And another follow-up PR to change the behavior for all the `label` related.

also discussed in this PR:
https://github.com/containerd/nerdctl/pull/1265#discussion_r930080531

AkihiroSuda: > Okay, should I remove filtering internal label and push again? 

Yes please 

> or just close this pr?

No

pkg/strutil/strutil.go
```diff
@@ -84,6 +84,7 @@ func DedupeStrSlice(in []string) []string {
 // ParseCSVMap parses a string like "foo=x,bar=y" into a map
 func ParseCSVMap(s string) (map[string]string, error) {
 	csvR := csv.NewReader(strings.NewReader(s))
+	csvR.LazyQuotes = true
```
minuk-dev, line 87: labels can contains "(double quotes)
If we don't use LazyQuotes, some labels cannot be parsed (like [this](https://github.com/containerd/nerdctl/runs/7534998440?check_suite_focus=true#step:6:483))

I'm not sure we just turn on LazyQutoes option to solve this problem.
But there is no reason to avoid it, so i turn it on.

AkihiroSuda: Could you add a comment line to explain that, and add an unit test in `strutil_test.go`

minuk-dev: In this pr?

AkihiroSuda (approved):
Thanks

PR #1267, opened 2022-07-27
fix(hosts): extra-host params should be 'domain:ip'

tosone:


pkg/composer/serviceparser/serviceparser.go
```diff
@@ -500,8 +500,8 @@ func newContainer(project *types.Project, parsed *Service, i int) (*Container, e
 			c.RunArgs = append(c.RunArgs, fmt.Sprintf("-e=%s=%s", k, *v))
 		}
 	}
-	for _, v := range svc.ExtraHosts {
-		c.RunArgs = append(c.RunArgs, fmt.Sprintf("--add-host=%s", v))
+	for k, v := range svc.ExtraHosts {
```
manugupt1, line 503: Nice catch! thanks!
This is the docs: https://docs.docker.com/compose/compose-file/compose-file-v2/

manugupt1 (approved):

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -500,8 +500,8 @@ func newContainer(project *types.Project, parsed *Service, i int) (*Container, e
...
+		c.RunArgs = append(c.RunArgs, fmt.Sprintf("--add-host=\"%s:%s\"", k, v))
```
junnplus, line 504: why do we need quotes?

tosone: Because the colon should be escaped, sometimes it will be treated as an operator.

junnplus (approved):
LGTM

Zheaoli (approved):
LGTM

tosone:
@AkihiroSuda Could you please help me merge this pr? ^_^

pkg/composer/serviceparser/serviceparser.go
AkihiroSuda: What is an “operator” here

AkihiroSuda:
@tosone 
ping https://github.com/containerd/nerdctl/pull/1267#discussion_r933314875

pkg/composer/serviceparser/serviceparser.go
tosone: https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion, I have tested it in my dev. I have many apps startup with `nerdctl compose`.

AkihiroSuda: I don’t think we are using bash here

tosone: seems ok

AkihiroSuda (approved):

PR #1269, opened 2022-07-28
Add the ability to inspect volume sizes

manugupt1:
Introduce a --size flag on volume inspect to inspect a volume
size. The flag is an explicit option because it can be slow and
not everyone might want it.

### DIFF
--- a/pkg/inspecttypes/native/volume.go
+++ b/pkg/inspecttypes/native/volume.go
@@ -21,4 +21,5 @@ type Volume struct {
 	Name       string             `json:"Name"`
 	Mountpoint string             `json:"Mountpoint"`
 	Labels     *map[string]string `json:"Labels,omitempty"`
+	Size       int64              `json:"Size,omitempty"`
 }
--- a/pkg/mountutil/volumestore/volumestore.go
+++ b/pkg/mountutil/volumestore/volumestore.go
@@ -63,6 +63,8 @@ type VolumeStore interface {
 	Create(name string, labels []string) (*native.Volume, error)
 	// Get may return ErrNotFound
 	Get(name string) (*native.Volume, error)
+	// GetWithSize returns a volume with size. May return ErrNotFound. Slow when volume has a lot of files.
+	GetWithSize(name string) (*native.Volume, error)
 	List() (map[string]native.Volume, error)
 	Remove(names []string) (removedNames []string, err error)
 }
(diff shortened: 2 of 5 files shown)


pkg/inspecttypes/native/volume.go
```diff
@@ -21,4 +21,5 @@ type Volume struct {
 	Name       string             `json:"Name"`
 	Mountpoint string             `json:"Mountpoint"`
 	Labels     *map[string]string `json:"Labels,omitempty"`
+	Size       string             `json:"Size,omitempty"`
```
AkihiroSuda, line 24: Shouldn't this be int64?

AkihiroSuda: Human-readable string can be printed in `nerdctl volume ls --size`

manugupt1: I am converting to human readable form. I can change to int64

manugupt1: Will make the change! Thanks

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -154,6 +156,28 @@ func (vs *volumeStore) Get(name string) (*native.Volume, error) {
 	return &entry, nil
 }
 
+func (vs *volumeStore) GetWithSize(name string) (*native.Volume, error) {
+	vol, err := vs.Get(name)
+	if err != nil {
+		return nil, err
+	}
+	walkDirFn := func(_ string, info os.FileInfo, err error) error {
```
Zheaoli, line 164: NOLGTM here. We may have a lot of tiny directories, and the recursive subdirectories may need lots of time.  So maybe it's a good idea to add a context here, and we can add extra timeout control outside the func 

See https://github.com/moby/moby/blob/master/pkg/directory/directory_unix.go#L14-L55

manugupt1: Won't the user be able to cancel the command with ctrl+c?

Zheaoli: Yes, they can cancel by sending SIGINT. but maybe it's an excellent choice to set a default timeout for this potential long-time action. I guess

manugupt1: I would really like to avoid a context combined with --size.
We can have a timeout for this function
```
nerdctl volume inspect --size --timeout 10s <volume-name> and this times out after a second. 
```
If timeout is specified and command takes more than 10s, we can timeout with an error message.

LMK if this sounds good.

Zheaoli: SGTM！

AkihiroSuda: > --timeout

`timeout(1)` might be enough.
https://man7.org/linux/man-pages/man1/timeout.1.html

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@Zheaoli 
Is this ready to merge?
I think `timeout(1)` is enough for controlling timeout but we can revisit it later.

Zheaoli:
@AkihiroSuda Ready to merge, we can discuss the timeout detail later

Issue #1279, opened 2022-07-30
`nerdctl system prune` should prune build caches too


AkihiroSuda:

STRRL:
Hi! I want to take it a try! Could you assign this issue to me? :heart:

fahedouch:
I believe that we should do the same thing for `nat` on windows

PR #1284, opened 2022-07-31
feat: system prune also cleanup build cache

STRRL:
Signed-off-by: Zhou Zhiqiang <im@strrl.dev>

close https://github.com/containerd/nerdctl/issues/1279

Changes:

- call `builderPruneAction()` within command `system prune`

### DIFF
--- a/cmd/nerdctl/system_prune.go
+++ b/cmd/nerdctl/system_prune.go
@@ -72,6 +72,7 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
 		}
 		msg += `
   - all images without at least one container associated to them
+  - all build cache
 `
 		msg += "\nAre you sure you want to continue? [y/N] "
 		fmt.Fprintf(cmd.OutOrStdout(), "WARNING! %s", msg)
@@ -99,5 +100,8 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
-	return imagePrune(cmd, client, ctx)
+	if err := imagePrune(cmd, client, ctx); err != nil {
+		return err
+	}
+	return builderPruneAction(cmd, args)
 }
--- a/cmd/nerdctl/system_prune.go
+++ b/cmd/nerdctl/system_prune.go
@@ -17,7 +17,10 @@
 package main
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
+	"github.com/containerd/nerdctl/pkg/buildkitutil"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -103,5 +106,22 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
 	if err := imagePrune(cmd, client, ctx); err != nil {
 		return err
 	}
-	return builderPruneAction(cmd, args)
+
+	var buf bytes.Buffer
+	outWriter := bufio.NewWriter(&buf)
+	copiedCmd := *cmd
+	copiedCmd.SetOut(outWriter)
+	if err := builderPruneAction(&copiedCmd, args); err != nil {
+		return err
+	}
+	parsedOutput, err := buildkitutil.ParseBuildctlPruneTableOutput(buf.Bytes())
+	if err != nil {
+		return err
+	}
+	// pretty print the output of prune build cache
+	out := cmd.OutOrStdout()
+	for _, row := range parsedOutput.Rows {
+		_, _ = fmt.Fprintf(out, "%s\n", row.ID)
+	}
+	return err
 }
--- a/cmd/nerdctl/system_prune.go
+++ b/cmd/nerdctl/system_prune.go
@@ -19,6 +19,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"github.com/containerd/nerdctl/pkg/buildkitutil"
 	"strings"
@@ -123,5 +124,9 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
 	for _, row := range parsedOutput.Rows {
 		_, _ = fmt.Fprintf(out, "%s\n", row.ID)
 	}
+	j, _ := json.Marshal(parsedOutput)
+	fmt.Fprintf(out, "%s\n", j)
+
+	fmt.Fprintf(out, "%s", buf.Bytes())
 	return err
 }
--- a/pkg/tabutil/tabutil.go
+++ b/pkg/tabutil/tabutil.go
@@ -61,11 +61,21 @@ func (r *TabReader) ReadRow(row, key string) (string, bool) {
 	if !ok {
 		return "", false
 	}
+	if idx.start > len(row) {
+		return "", false
+	}
 	var value string
 	if idx.end == -1 {
 		value = row[idx.start:]
 	} else {
-		value = row[idx.start:idx.end]
+		end := min(idx.end, len(row))
+		value = row[idx.start:end]
 	}
 	return strings.TrimSpace(value), true
+}
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
 }
--- a/pkg/buildkitutil/buildctl_prune_output_parser.go
+++ b/pkg/buildkitutil/buildctl_prune_output_parser.go
@@ -1,8 +1,29 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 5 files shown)


STRRL:
PTAL!

@Zheaoli @AkihiroSuda

AkihiroSuda:
> Signed-off-by: STRRL

Thanks, but please use real name

cmd/nerdctl/system_prune.go
```diff
@@ -99,5 +100,8 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
-	return imagePrune(cmd, client, ctx)
+	if err := imagePrune(cmd, client, ctx); err != nil {
+		return err
+	}
+	return builderPruneAction(cmd, args)
```
junnplus, line 106: NOLGTM

```
sudo nerdctl system prune --all
WARNING! This will remove:
  - all stopped containers
  - all networks not used by at least one container
  - all images without at least one container associated to them
  - all build cache

Are you sure you want to continue? [y/N] y
Deleted Images:
Untagged: docker.io/library/nginx:test
deleted: sha256:af32df749f3fb0040acdb0015af7b7da8f6d600572b22e9137250d3dfbe111f2
deleted: sha256:97b6c514961c6bf7c4e9bf4c14bfdcc36b507b88aad22483d28b2f9d21b2c901
deleted: sha256:1c6d36bc7d70f055c5da944961712237413705cc3674f9fb44a960b29552e8a8
deleted: sha256:8b1cace7b5b65b51ffaccdf08ff3a3628748d083ede7d314dd22cad8f7839c4c
deleted: sha256:31a3b8d1409240f584c30db43d20e0bdba94d36e0316da0ad0da7d7d91159519
deleted: sha256:4e024e3fcf12199b325c226eecb7b8bc5d5cb04ffccd16112bf9e51ea458fc44
deleted: sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef

ID                                                                      RECLAIMABLE     SIZE    LAST ACCESSED
5ng2p4kgza3jsz2o35vdm5xdm                                               true            8.22kB
qpaheqehsy3qit4julsunsdn7*                                              true            4.10kB
zr5ioimrfo9faxb82auoeijir*                                              true            8.19kB
w61ewoia9iv5p5pbm1fpoabbs                                               true            17.78kB
3wp3a2ozz29fqq6slzd6ljw6e                                               true            12.96kB
svf78j2ocb1lfvnjf0ai6lr53                                               true            13.18kB
8z7keywza1z3vxu7w254v6cbr                                               true            8.79kB
abq5892g9c8xjv6xvv3usv59w                                               true            88.87MB
j08w18u4sm5o6s71gdlgl32db                                               true            115.81MB
Total:  204.75MB
```

The output of `builder prune` should only contain the ID instead of the table.

junnplus: Also, `Total` can confuse users, it should be all reclaimed space.

AkihiroSuda: What should we do?
Should we extend `buildctl prune` to support JSON output? Or should we use the Go API of BuildKit?
We could use `nerdctl/pkg/tabutil.TabReader` too, but not robust.

junnplus: Go API and `TabReader` are both ok, `TabReader` is enough for now.

AkihiroSuda: Yes, TabReader is enough for now, but we should switch away to use JSON when this gets merged (BuildKit 0.11)
- https://github.com/moby/buildkit/pull/2992

STRRL:
> > Signed-off-by: STRRL
> 
> Thanks, but please use real name

Updated!

cmd/nerdctl/system_prune.go
STRRL: What should I do now? Use `TabReader`? or wait https://github.com/moby/buildkit/pull/2992 get merged then use JSON output

AkihiroSuda: You can use TabReader for now, and add a `TODO` comment to use https://github.com/moby/buildkit/pull/2992 in the future

STRRL:
Hi @AkihiroSuda @junnplus , I meet some trouble when using `TabReader` to parse the output of `buildctl prune`.

The output of the `buildctl prune` is not soo standard, it would use both `\t` and white-space as the delimiter:

image

I also updated the testcase, so it would be failed with the current `TabReader` implementation because it extracts values based on the character position.

STRRL:
What should I do now? Make a new implementation like `TabReader` but only for `buildctl prune`? or wait https://github.com/moby/buildkit/pull/2992 getting merged?

cmd/nerdctl/system_prune.go
```diff
@@ -99,5 +104,29 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
...
-	return imagePrune(cmd, client, ctx)
+	if err := imagePrune(cmd, client, ctx); err != nil {
+		return err
+	}
+
+	var buf bytes.Buffer
+	outWriter := bufio.NewWriter(&buf)
+	copiedCmd := *cmd
+	copiedCmd.SetOut(outWriter)
+	if err := builderPruneAction(&copiedCmd, args); err != nil {
```
junnplus, line 115: Why not make `builder prune` the same as docker?

```sh
$ sudo docker builder prune
WARNING! This will remove all dangling build cache. Are you sure you want to continue? [y/N] y
Deleted build cache objects:
mzl53zo5788bl0bzvz1j72j1g
yp82akq9eitd5jb5ni6k9pvd4
ik2txugxyg3ik669w77ho4pqq
sd78ank4ncfdhqjilztfbwqik
```

STRRL:
> What should I do now? Make a new implementation like `TabReader` but only for `buildctl prune`? or wait [moby/buildkit#2992](https://github.com/moby/buildkit/pull/2992) getting merged?

I just noticed that PR already get merged! :tada: 

I think I could update code based on the json output!

junnplus:
I think I could update code based on the json output!

> Are we not compatible with older versions?

cmd/nerdctl/system_prune.go
STRRL: I would make it. But I meet some trouble when parsing the output of `buildctl prune`. 🥲

STRRL:
> I think I could update code based on the json output!
> 
> > Are we not compatible with older versions?

Well, if we want to be compatible with the older version of `buildctl`, it seems I must build the parser for `buildctl`. 🤔

Or maybe using the go API would be the better way?

PTAL @junnplus @AkihiroSuda

junnplus:
> The output of the buildctl prune is not soo standard, it would use both \t and white-space as the delimiter:

You can convert `\t` to whitespace.

AkihiroSuda:
Parse the JSON if the JSON output is available, otherwise we can just print the bare output from `buildctl builder prune`

junnplus:
> Parse the JSON if the JSON output is available, otherwise we can just print the bare output from buildctl builder prune

Honestly, I wish we could be compatible with older versions because most users are using them.

pkg/tabutil/tabutil.go
```diff
@@ -61,11 +61,21 @@ func (r *TabReader) ReadRow(row, key string) (string, bool) {
 	if !ok {
 		return "", false
 	}
+	if idx.start > len(row) {
+		return "", false
+	}
```
junnplus, line 66:
```suggestion
	row = strings.ReplaceAll(row, "\t", "    ")
```

junnplus, line 66:
```suggestion
	row = strings.ReplaceAll(row, "\t", "    ")
```

junnplus: Also in `ParseHeader`

cmd/nerdctl/system_prune.go
junnplus: https://github.com/containerd/nerdctl/pull/1284#issuecomment-1203933222

STRRL:
> > The output of the buildctl prune is not soo standard, it would use both \t and white-space as the delimiter:
> 
> You can convert `\t` to whitespace.



> Parse the JSON if the JSON output is available, otherwise we can just print the bare output from `buildctl builder prune`

What about using BuildKit Go API? I prefer to use that instead of parsing the text, because go API seems more stable and easy to use. 🤔

junnplus:
> What about using BuildKit Go API? I prefer to use that instead of parsing the text, because go API seems more stable and easy to use. 🤔

SGTM.

STRRL:
> > What about using BuildKit Go API? I prefer to use that instead of parsing the text, because go API seems more stable and easy to use. thinking
> 
> SGTM.

Hi @junnplus @AkihiroSuda @fahedouch, I took a look at using buildkit go API instead of using `buildctl` for pruning cache. It seems it would take more changes, but I still think that worth it.

I think I would create a tracking issue about:

- migrate the existed `build prune` using buildkit API
- then finished the work about `system prune` 

What do you think about it?

AkihiroSuda:
Using buildkit API sounds good but fearing it may increase the binary footprint


Cc @ktock

STRRL:
It has been a long time since this PR not active, I would continue the work in another PR with new codebase.

Issue #1290, opened 2022-08-01
kata + privileged: `failed to create shim task: QMP command failed: The device is not writable: Permission denied: unknown`


AkihiroSuda:
### Description

```console
$ sudo nerdctl run -it --rm --runtime=io.containerd.kata.v2 --privileged alpine
FATA[0004] failed to create shim task: QMP command failed: The device is not writable: Permission denied: unknown
```

### Steps to reproduce the issue

1. `sudo PATH=/opt/kata/bin:$PATH containerd` (as root)
2.  `sudo nerdctl run -it --rm --runtime=io.containerd.kata.v2 --privileged alpine`

### Describe the results you received and expected

Received: failed with the error shown above
Expected: should work

Non-privileged mode works as expected

### What version of nerdctl are you using?

```console
$ sudo nerdctl version
Client:
 Version:       v0.22.2
 OS/Arch:       linux/amd64
 Git commit:    2899222cb0715f1e5ffe356d10c3439ee8ee3ba4
 buildctl:
  Version:      v0.10.0-380-g874eef9b
  GitCommit:    874eef9b70dbaf4f074d2bc8f4dc64237f8e83a0

Server:
 containerd:
  Version:      v1.6.6
  GitCommit:    10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:      1.1.2
  GitCommit:    v1.1.2-0-ga916309f

$ sudo /opt/kata/bin/kata-runtime -v
kata-runtime  : 2.5.0-rc0
   commit   : ac91fb7a126238b595afa59c17fbb82a12c4f5ad
   OCI specs: 1.0.2-dev
```

### Host information

```console
$ sudo nerdctl info
Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v1.6.6
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Log: fluentd journald json-file
  Storage: native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.15.0-43-generic
 Operating System: Ubuntu 22.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.59GiB
 Name: suda-ws01
 ID: e4b49f8a-ed36-4fab-98a2-bc5d0e51d2ac
```

[...]

PR #1291, opened 2022-08-02
run: add a new `privileged-without-host-devices` option

liubin:
Containers with  --privileged enabled will default pass all
devices on the host side to containers, but for runtimes
like Kata Containers, this action will bring up some issues,
for example permission or naming conflict, add a new command
line option `--privileged-without-host-devices` will not pass
all host devices to containers, if needed, users can pass
devices by `--device` option.

Fixes: #1290

Signed-off-by: bin liu <liubin0329@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,7 @@ Security flags:
 - :whale: `--security-opt seccomp=<PROFILE_JSON_FILE>`: specify custom seccomp profile
 - :whale: `--security-opt apparmor=<PROFILE>`: specify custom AppArmor profile
 - :whale: `--security-opt no-new-privileges`: disallow privilege escalation, e.g., setuid and file capabilities
+- :nerd_face: `--security-opt privileged-without-host-devices`: Don't pass host devices to privileged containers
 - :whale: `--cap-add=<CAP>`: Add Linux capabilities
 - :whale: `--cap-drop=<CAP>`: Drop Linux capabilities
 - :whale: `--privileged`: Give extended privileges to this container
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -188,7 +188,7 @@ func setCreateFlags(cmd *cobra.Command) {
 	// #region security flags
 	cmd.Flags().StringArray("security-opt", []string{}, "Security options")
 	cmd.RegisterFlagCompletionFunc("security-opt", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		return []string{"seccomp=", "seccomp=unconfined", "apparmor=", "apparmor=" + defaults.AppArmorProfileName, "apparmor=unconfined", "no-new-privileges"}, cobra.ShellCompDirectiveNoFileComp
+		return []string{"seccomp=", "seccomp=unconfined", "apparmor=", "apparmor=" + defaults.AppArmorProfileName, "apparmor=unconfined", "no-new-privileges", "privileged-without-host-devices"}, cobra.ShellCompDirectiveNoFileComp
 	})
 	// cap-add and cap-drop are defined as StringSlice, not StringArray, to allow specifying "--cap-add=CAP_SYS_ADMIN,CAP_NET_ADMIN" (compatible with Podman)
 	cmd.Flags().StringSlice("cap-add", []string{}, "Add Linux capabilities")
--- a/cmd/nerdctl/run_linux.go
+++ b/cmd/nerdctl/run_linux.go
@@ -93,23 +93,20 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
 		opts = append(opts, capOpts...)
 	}
 
+	privileged, err := cmd.Flags().GetBool("privileged")
+	if err != nil {
+		return nil, err
+	}
+
 	securityOpt, err := cmd.Flags().GetStringArray("security-opt")
 	if err != nil {
 		return nil, err
 	}
 	securityOptsMaps := strutil.ConvertKVStringsToMap(strutil.DedupeStrSlice(securityOpt))
-	if secOpts, err := generateSecurityOpts(securityOptsMaps); err != nil {
+	if secOpts, err := generateSecurityOpts(privileged, securityOptsMaps); err != nil {
 		return nil, err
 	} else {
 		opts = append(opts, secOpts...)
-	}
-
-	privileged, err := cmd.Flags().GetBool("privileged")
-	if err != nil {
-		return nil, err
-	}
-	if privileged {
-		opts = append(opts, privilegedOpts...)
 	}
 
 	b4nnOpts, err := bypass4netnsutil.GenerateBypass4netnsOpts(securityOptsMaps, labelsMap, id)
--- a/cmd/nerdctl/run_security_linux.go
+++ b/cmd/nerdctl/run_security_linux.go
@@ -18,8 +18,6 @@ package main
 
 import (
 	"errors"
-	"fmt"
-	"strconv"
 	"strings"
 	"sync"
 
@@ -29,6 +27,7 @@ import (
 	"github.com/containerd/containerd/pkg/cap"
 	"github.com/containerd/nerdctl/pkg/apparmorutil"
 	"github.com/containerd/nerdctl/pkg/defaults"
+	"github.com/containerd/nerdctl/pkg/maputil"
 	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/sirupsen/logrus"
 )
@@ -40,10 +39,15 @@ var privilegedOpts = []oci.SpecOpts{
 	oci.WithNewPrivileges,
 }
 
-func generateSecurityOpts(securityOptsMap map[string]string) ([]oci.SpecOpts, error) {
+var privilegedWithoutDevicesOpts = []oci.SpecOpts{
+	oci.WithPrivileged,
+	oci.WithNewPrivileges,
+}
+
+func generateSecurityOpts(privileged bool, securityOptsMap map[string]string) ([]oci.SpecOpts, error) {
 	for k := range securityOptsMap {
 		switch k {
-		case "seccomp", "apparmor", "no-new-privileges":
+		case "seccomp", "apparmor", "no-new-privileges", "privileged-without-host-devices":
 		default:
 			logrus.Warnf("Unknown security-opt: %q", k)
 		}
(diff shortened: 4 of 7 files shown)


cmd/nerdctl/run_security_linux.go
```diff
@@ -35,8 +35,6 @@ import (
 
 var privilegedOpts = []oci.SpecOpts{
 	oci.WithPrivileged,
-	oci.WithAllDevicesAllowed,
```
manugupt1, line 38: Curious: Does this affect network devices as well? How will network connectivity work inside the VM?

cmd/nerdctl/run_linux.go
```diff
@@ -108,8 +108,17 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
...
+
 	if privileged {
 		opts = append(opts, privilegedOpts...)
+
+		withoutHostDevices, err := cmd.Flags().GetBool("privileged-without-host-devices")
+		if err != nil {
+			return nil, err
+		}
+		if !withoutHostDevices {
+			opts = append(opts, oci.WithHostDevices, oci.WithAllDevicesAllowed)
```
manugupt1, line 120: nit: can we keep it as two as `deviceOpts` together in run_security_linux.go

README.md
```diff
@@ -433,6 +433,7 @@ Security flags:
 - :whale: `--cap-add=<CAP>`: Add Linux capabilities
 - :whale: `--cap-drop=<CAP>`: Drop Linux capabilities
 - :whale: `--privileged`: Give extended privileges to this container
+- :whale: `--privileged-without-host-devices`: Don't pass host devices to privileged containers
```
manugupt1, line 436: nerd-face ??

cmd/nerdctl/run_linux.go
```diff
@@ -108,8 +108,17 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
 	if err != nil {
 		return nil, err
 	}
+
```
manugupt1, line 111: Is it possible to have test for this?

cmd/nerdctl/run.go
```diff
@@ -196,6 +196,7 @@ func setCreateFlags(cmd *cobra.Command) {
 	cmd.Flags().StringSlice("cap-drop", []string{}, "Drop Linux capabilities")
 	cmd.RegisterFlagCompletionFunc("cap-drop", capShellComplete)
 	cmd.Flags().Bool("privileged", false, "Give extended privileges to this container")
+	cmd.Flags().Bool("privileged-without-host-devices", false, "Don't pass host devices to privileged containers")
```
AkihiroSuda, line 199: Should this automatically imply `--privileged`?
Or should this fail when `--privileged` is not specified?

cmd/nerdctl/run_security_linux.go
liubin: Kata has its route to set up network between host and guest without /dev/xx device. If additional devices are needed, that can be specified by `--device` option.

cmd/nerdctl/run_linux.go
liubin: Test `TestRunPrivileged` added.

cmd/nerdctl/run_linux.go
liubin: A new struct `privilegedWithoutDevicesOpts` added for the new options after the original `privilegedOpts`

README.md
liubin: Fixed

cmd/nerdctl/run.go
liubin: Updated. IMO it's better to imply `--privileged`, and if both are specified, an error will be returned.

cmd/nerdctl/run_security_linux_test.go
```diff
@@ -190,3 +191,37 @@ func TestRunSeccompCapSysPtrace(t *testing.T) {
...
 	// Docker/Moby 's seccomp profile allows ptrace(2) by default, but containerd does not (yet): https://github.com/containerd/containerd/issues/6802
 }
+
+func TestRunPrivileged(t *testing.T) {
+	// docker does not support --privileged-without-host-devices
+	testutil.DockerIncompatible(t)
+
+	base := testutil.NewBase(t)
+
+	devPath := "/dev/dummy-zero"
```
manugupt1, line 201: Q (for my own understanding): Why not re-use /dev/zero or /dev/urandom? Why do we need to create a new device?

cmd/nerdctl/run_security_linux_test.go
```diff
@@ -190,3 +191,37 @@ func TestRunSeccompCapSysPtrace(t *testing.T) {
...
+	// ensure the file will be removed in case of failed in the test
+	defer func() {
+		exec.Command("rm", devPath).Run()
+	}()
+
+	// get device with host devices
+	base.Cmd("run", "--rm", "--privileged", testutil.AlpineImage, "ls", devPath).AssertOutExactly(devPath + "\n")
+
+	// get device without host devices
+	res := base.Cmd("run", "--rm", "--privileged-without-host-devices", testutil.AlpineImage, "ls", devPath).Run()
```
manugupt1, line 220: Should `ls /dev` return empty list in this case, would that be a better test?

liubin: No, this option is only disable passing all host devices to containers. Containers will have its default devices [automatically created](https://github.com/opencontainers/runc/blob/d11f4d756e85ece5cdba8bb69f8bd4db3cdcbeab/libcontainer/specconv/spec_linux.go#L173-L174).

cmd/nerdctl/run_security_linux_test.go
liubin: Choosing a nonstandard device to ensure it will not be affected by different OSes/Environments that this test running on.

And also as https://github.com/opencontainers/runc/blob/d11f4d756e85ece5cdba8bb69f8bd4db3cdcbeab/libcontainer/specconv/spec_linux.go#L173-L174 has shown, `/dev/zero` and/or `/dev/random` are standard devices that will be created for containers whether this `without-host-device` is specified or not.

AkihiroSuda:
Sorry for going back and forth, but probably we should implement this as a `--privileged --security-opt privileged-without-host-devices` for ease of implementing compose support: https://pkg.go.dev/github.com/compose-spec/compose-go@v1.4.0/types#ServiceConfig.

`--security-opt privileged-without-host-devices` should require specifying `--privileged`

liubin:
> Sorry for going back and forth, but probably we should implement this as a `--privileged --security-opt privileged-without-host-devices` for ease of implementing compose support: https://pkg.go.dev/github.com/compose-spec/compose-go@v1.4.0/types#ServiceConfig.
> 
> `--security-opt privileged-without-host-devices` should require specifying `--privileged`

Ok, I'll update ASAP.

liubin:
@AkihiroSuda @manugupt1 I updated as a `--security-opt` option, please take a look.

pkg/maputil/maputil.go
```diff
@@ -0,0 +1,41 @@
...
+
+package maputil
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// MapBoolValueAsOpt will parse key as a command-line option.
+// If only key is specified will be treated as true,
```
manugupt1, line 25: Why true? Default is false in most cases. Can we handle on the caller side.

liubin: In case of `--security-opt xxx` should be equal with `--security-opt xxx=true`.

pkg/maputil/maputil.go
```diff
@@ -0,0 +1,41 @@
...
+// otherwise, the value will be parsed and returned
```
manugupt1, line 26:
```suggestion
// otherwise, the value will be parsed and returned.
// This is useful when command line flags have options.
// Following examples illustrate this:
// --security-opt xxx returns true
// --security-opt xxx=true returns true
// --security-opt xxx=false returns false
// --security-opt xxx=invalid returns false and error
```

Upto you to take it or leave it.

manugupt1 (approved):

pkg/maputil/maputil.go
liubin: Thank you, @manugupt1, updated as your suggestion.

AkihiroSuda:
CI is failing

liubin:
@AkihiroSuda I restarted the CI jobs, and still two are failed. It seems to be a network issue.

cmd/nerdctl/run_security_linux.go
```diff
@@ -85,21 +89,28 @@ func generateSecurityOpts(securityOptsMap map[string]string) ([]oci.SpecOpts, er
...
+
 	if !nnp {
 		opts = append(opts, oci.WithNewPrivileges)
 	}
+
+	privilegedWithoutHostDevices, err := maputil.MapBoolValueAsOpt(securityOptsMap, "privileged-without-host-devices")
+	if err != nil {
+		return nil, err
+	}
+
```
AkihiroSuda, line 105: Can we return an error when `privilegedWithoutHostDevices && !privileged`

liubin: Fixed.

liubin:
@AkihiroSuda There are still some failed tests, can you please re-trigger them? thanks.

cmd/nerdctl/run_security_linux_test.go
```diff
@@ -190,3 +192,40 @@ func TestRunSeccompCapSysPtrace(t *testing.T) {
...
+	if rootlessutil.IsRootless() {
+		t.Skip("test skipped for rootless privileged containers")
+	}
+
+	base := testutil.NewBase(t)
+
+	devPath := "/dev/dummy-zero"
+
+	// a dummy zero device: mknod /dev/dummy-zero c 1 5
+	helperCmd := exec.Command("mknod", []string{devPath, "c", "1", "5"}...)
```
AkihiroSuda, line 209: You could just use `unix.Mknod`

cmd/nerdctl/run_security_linux_test.go
```diff
@@ -190,3 +192,40 @@ func TestRunSeccompCapSysPtrace(t *testing.T) {
...
+	if out, err := helperCmd.CombinedOutput(); err != nil {
+		err = fmt.Errorf("cannot create %q: %q: %w", devPath, string(out), err)
+		t.Fatal(err)
+	}
+
+	// ensure the file will be removed in case of failed in the test
+	defer func() {
+		exec.Command("rm", devPath).Run()
+	}()
```
AkihiroSuda, line 218: You could just use `os.Remove`

AkihiroSuda (approved):
Thanks

Issue #1278, opened 2022-07-30
The output of `nerdctl stats` disappears after a few minutes


AkihiroSuda:
### Description

The output of `nerdctl stats` disappears after a few minutes

### Steps to reproduce the issue

1. `nerdctl run -d alpine sleep infinity`
2. `nerdctl stats`
3. Wait for a few minute

### Describe the results you received and expected

Received:
```
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O   BLOCK I/O   PIDS
ce15db2ec247   alpine-ce15d   --        -- / --             --        --        --          --
```

Expected: some stats

### What version of nerdctl are you using?

v0.22.0-42-g4c37225

### Host information

```
Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v1.6.6
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Log: fluentd journald json-file
  Storage: native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.0-41-generic
 Operating System: Ubuntu 22.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.59GiB
...
```

AkihiroSuda:
@fahedouch PTAL

fahedouch:
@AkihiroSuda I tested stats for 15 minutes and didn't get this behavior. I got an empty stats only when I killed the container after this 15 minutes of fetching stats 

````
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O       BLOCK I/O   PIDS
24c458aa71b2   alpine-24c45   99.81%    0B / 0B             0.00%     920B / 962B   0B / 0B     3
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O       BLOCK I/O   PIDS
24c458aa71b2   alpine-24c45   100.21%   0B / 0B             0.00%     920B / 962B   0B / 0B     3
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O       BLOCK I/O   PIDS
24c458aa71b2   alpine-24c45   99.40%    0B / 0B             0.00%     920B / 962B   0B / 0B     3
[...]

PR #1298, opened 2022-08-04
Fix nerdctl stats output

fahedouch:
thanks to @liubin  to detect the issue (https://github.com/containerd/nerdctl/issues/1278#issuecomment-1205151926)

fix stats output and ensure error display (https://github.com/containerd/nerdctl/issues/1278)

### DIFF
--- a/cmd/nerdctl/stats.go
+++ b/cmd/nerdctl/stats.go
@@ -283,17 +283,6 @@ func statsAction(cmd *cobra.Command, args []string) error {
 		// make sure each container get at least one valid stat data
 		waitFirst.Wait()
 
-		var errs []string
-		cStats.mu.Lock()
-		for _, c := range cStats.cs {
-			if err := c.GetError(); err != nil {
-				errs = append(errs, err.Error())
-			}
-		}
-		cStats.mu.Unlock()
-		if len(errs) > 0 {
-			return errors.New(strings.Join(errs, "\n"))
-		}
 	}
 
 	cleanScreen := func() {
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/stats.go
```diff
@@ -283,17 +283,6 @@ func statsAction(cmd *cobra.Command, args []string) error {
 		// make sure each container get at least one valid stat data
 		waitFirst.Wait()
 
-		var errs []string
```
manugupt1, line 286: Why remove this block?

fahedouch: useless regarding this https://github.com/containerd/nerdctl/pull/1298/files#diff-8541fa08e70d092bb039071593918dac552c3cbf1088d15e71d0f6117ad4b249R306

the error display should be within the ticker

cmd/nerdctl/stats_linux.go
```diff
@@ -62,11 +62,13 @@ func setContainerStatsAndRenderStatsEntry(previousStats *statsutil.ContainerStat
 		if err != nil {
 			return statsutil.StatsEntry{}, fmt.Errorf("failed to retrieve the statistics in netns %s: %v", ns, err)
 		}
+		defer ns.Close()
```
manugupt1, line 65: Can we wrap this into a function and log errors if Close returns an error; that would help us catch file leaks if there are any if still remaining? Ideally, this should never happen.

junnplus (approved):
Thanks, LGTM

manugupt1 (approved):

AkihiroSuda (approved):
Thanks

PR #1304, opened 2022-08-08
exclude default bridge network in `network prune`

### LINKED ISSUE #1295: `nerdctl network prune` does not need to remove the default `bridge` network
As of v0.22.2, `nerdctl network prune` removes the default `bridge` network when it is not in use, but the default `bridge` network does not need to be removed

tal66:
Hi, this should fix #1295 
(`nerdctl network prune` does not need to remove the default bridge network)

### DIFF
--- a/cmd/nerdctl/network_prune.go
+++ b/cmd/nerdctl/network_prune.go
@@ -26,9 +26,12 @@ import (
 	"github.com/containerd/nerdctl/pkg/labels"
 	"github.com/containerd/nerdctl/pkg/netutil"
 	"github.com/containerd/nerdctl/pkg/netutil/nettype"
+	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
+
+var networkDriversToKeep = []string{"host", "none", DefaultNetworkDriver}
 
 func newNetworkPruneCommand() *cobra.Command {
 	networkPruneCommand := &cobra.Command{
@@ -97,7 +100,7 @@ func networkPrune(cmd *cobra.Command, client *containerd.Client, ctx context.Con
 
 	var removedNetworks []string // nolint: prealloc
 	for _, net := range e.Networks {
-		if net.Name == "host" || net.Name == "none" {
+		if strutil.InStringSlice(networkDriversToKeep, net.Name) {
 			continue
 		}
 		if net.NerdctlID == nil || net.File == "" {


cmd/nerdctl/network_prune.go
```diff
@@ -97,7 +97,7 @@ func networkPrune(cmd *cobra.Command, client *containerd.Client, ctx context.Con
 
 	var removedNetworks []string // nolint: prealloc
 	for _, net := range e.Networks {
-		if net.Name == "host" || net.Name == "none" {
+		if net.Name == "host" || net.Name == "none" || net.Name == "bridge" {
```
manugupt1, line 100: Can you please refactor move all of these to a constant and look them over so that we do not expand this list

manugupt1:
Please sign off with real name.

fahedouch:
plz use `DefaultNetworkDriver`  from package main const so we avoid deleting `bridge` for unix and `nat` for windows

fahedouch:
see https://github.com/containerd/nerdctl/issues/1279#issuecomment-1209646442

cmd/nerdctl/network_prune.go
```diff
@@ -28,8 +28,11 @@ import (
 	"github.com/containerd/nerdctl/pkg/netutil/nettype"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
 )
 
+var dockerDefaultNetworkDrivers = []string{"host", "none", DefaultNetworkDriver} // exist by default
```
fahedouch, line 34: `host` is not a default driver we should look for another adapted name

fahedouch (changes requested):

cmd/nerdctl/network_prune.go
AkihiroSuda: ?

cmd/nerdctl/network_prune.go
```diff
@@ -97,7 +99,14 @@ func networkPrune(cmd *cobra.Command, client *containerd.Client, ctx context.Con
 
 	var removedNetworks []string // nolint: prealloc
 	for _, net := range e.Networks {
-		if net.Name == "host" || net.Name == "none" {
+		keepNetwork := false
+		for _, d := range networkDriversToKeep {
+			if net.Name == d {
+				keepNetwork = true
+				continue
```
fahedouch, line 106: Break

tal66: right:) anyway you can switch this whole thing to `slices.Contains()` when changing go.mod to 1.18

fahedouch:
Thanks, but please squash your commits and  sign the commit for DCO
https://github.com/apps/dco

cmd/nerdctl/network_prune.go
```diff
@@ -97,7 +99,14 @@ func networkPrune(cmd *cobra.Command, client *containerd.Client, ctx context.Con
...
 	var removedNetworks []string // nolint: prealloc
 	for _, net := range e.Networks {
-		if net.Name == "host" || net.Name == "none" {
+		keepNetwork := false
+		for _, d := range networkDriversToKeep {
+			if net.Name == d {
+				keepNetwork = true
+				break
+			}
+		}
```
junnplus, line 108: Please use `strutil.DedupeStrSlice` instead.

junnplus (changes requested):

AkihiroSuda (approved):
Thanks

PR #1342, opened 2022-08-28
Feat: support filter argument for `nerdctl ps` command

### LINKED ISSUE #1339: Support filter argument in `nerdctl container ls` command

Just following the #54 to support filter argument in `nerdctl container ls` command.

FYI https://github.com/moby/moby/blob/master/daemon/list.go#L20

- [ ] before
- [ ] existed
- [ ] id
- [ ] label
- [ ] name
- [ ] status
- [ ] since
- [ ] volume
- [ ] network

### Additional context

_No response_

yuchanns:
Fix #1339 

Signed-off-by: Hanchin Hsieh <me@yuchanns.xyz>

### DIFF
--- a/cmd/nerdctl/ps.go
+++ b/cmd/nerdctl/ps.go
@@ -19,6 +19,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -62,6 +63,7 @@ func newPsCommand() *cobra.Command {
 	psCommand.RegisterFlagCompletionFunc("format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"json", "table", "wide"}, cobra.ShellCompDirectiveNoFileComp
 	})
+	psCommand.Flags().StringSliceP("filter", "f", nil, "Filter matches containers based on given conditions")
 	return psCommand
 }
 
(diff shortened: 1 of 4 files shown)


Zheaoli:
maybe you need checkout about `container ls` command. It's a little bit different with `ps`

cmd/nerdctl/ps.go
```diff
@@ -319,3 +385,355 @@ func getContainerSize(ctx context.Context, client *containerd.Client, c containe
...
+				for _, network := range networks {
+					if network == value {
+						return true
+					}
+				}
+				return false
+			})
+			return nil
+		}},
+	}
```
Zheaoli, line 708: I think refactor these codes to a single function name like `getFilterFuncs` may be a good idea.

It will make the codebase more readable

cmd/nerdctl/ps.go
```diff
@@ -291,6 +306,57 @@ func getPrintableContainerName(containerLabels map[string]string) string {
 	return ""
 }
 
+type containerVolume struct {
```
Zheaoli, line 309: The code change in this PR may be useful when we want to impl the `container ls --filter`. I suggest moving some of the code to a single file to make reusing the code easier

AkihiroSuda:
>  what is the exact difference between ps and container ls.

Same

yuchanns:
> Same

Exactly.

> maybe you need checkout about `container ls` command. It's a little bit different with `ps`

https://github.com/docker/cli/blob/6205b4eae6a116236edc04876e9c6631c823290b/cli/command/container/list.go#L62-L67

I don't see anything different in Docker's implementation.

AkihiroSuda (approved):
Thanks, merging

Issue #1378, opened 2022-09-13
Cannot create an anonymous volume: `nerdctl run --mount type=volume,dst=/foo` returns an error


davidhsingyuchen:
### Description

```sh
➜  ~ lima nerdctl run --mount type=volume,dst=/foo alpine
FATA[0000] malformed name : identifier must not be empty: invalid argument
```

A Github search gives me #67, and `nerdctl run -v /foo` does work for me, so maybe it's a bug.

Relevant doc is in [`nerdctl run`](https://github.com/containerd/nerdctl#whale-blue_square-nerdctl-run) -> `Volume flags:` -> `--mount`:

> 🐳 src, source: Mount source spec for bind and volume. Mandatory for bind.

This implies that for `type=volume`, if `src` is not given, an anonymous volume should be created, but currently an error is returned instead.

### Steps to reproduce the issue

```sh
lima nerdctl run --mount type=volume,dst=/foo alpine
```

### Describe the results you received and expected

Received: an error (`FATA[0000] malformed name : identifier must not be empty: invalid argument`)
Expected: an anonymous volume is created and the container is started

### What version of nerdctl are you using?

```sh
➜  ~ lima nerdctl version
Client:
 Version:	v0.22.2
 OS/Arch:	linux/arm64
 Git commit:	2899222cb0715f1e5ffe356d10c3439ee8ee3ba4
 buildctl:
  Version:	v0.10.3
  GitCommit:	c8d25d9a103b70dc300a4fd55e7e576472284e31

Server:
 containerd:
  Version:	v1.6.6
  GitCommit:	10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:	1.1.3
  GitCommit:	v1.1.3-0-g6724737f
```

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Lima

### Host information

```sh
➜  ~ lima nerdctl info
Client:
 Namespace:	default
 Debug Mode:	false

Server:
 Server Version: v1.6.6
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Log: fluentd journald json-file
  Storage: native overlayfs fuse-overlayfs stargz
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
  rootless
 Kernel Version: 5.15.0-41-generic
 Operating System: Ubuntu 22.04 LTS
 OSType: linux
[...]

PR #1386, opened 2022-09-17
Remove split if first element is empty to create anon volume

manugupt1:
Fixes https://github.com/containerd/nerdctl/issues/1378

### DIFF
--- a/pkg/mountutil/mountutil.go
+++ b/pkg/mountutil/mountutil.go
@@ -56,6 +56,7 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
 		options  []string
 	)
 
+	s = strings.TrimLeft(s, ":")
 	split := strings.Split(s, ":")
 	switch len(split) {
 	case 1:
--- a/cmd/nerdctl/run_mount_linux_test.go
+++ b/cmd/nerdctl/run_mount_linux_test.go
@@ -82,6 +82,13 @@ func TestRunAnonymousVolume(t *testing.T) {
 		"mountpoint", "-q", "/foo").AssertOK()
 }
 
+func TestRunAnonymousVolumeWithTypeMountFlag(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+	base.Cmd("run", "--rm", "--mount", "type=volume,dst=/foo", testutil.AlpineImage,
+		"mountpoint", "-q", "/foo").AssertOK()
+}
+
 func TestRunAnonymousVolumeWithBuild(t *testing.T) {
 	t.Parallel()
 	testutil.RequiresBuild(t)


Zheaoli:
Would you mind to add extra unit test?

pkg/mountutil/mountutil.go
```diff
@@ -57,6 +57,9 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
 	)
 
 	split := strings.Split(s, ":")
+	if split[0] == "" {
+		split = split[1:]
+	}
```
fahedouch, line 62: wrong result  for `nerdctl run  --mount type=volume,dst=/foo,readonly`

AkihiroSuda:
> Would you mind to add extra unit test?

Needs an integration test

manugupt1:
This would require a while; I did not account for the exit code. I will fix it in a few days.

pkg/mountutil/mountutil.go
manugupt1: This does not create for docker as well; however I have made the code changes for this as well.

manugupt1: Sorry for going back and forth; adding incompatible changes did not make much sense as the code there was fragile. Keeping it simple

manugupt1: FYI:
```
➜  nerdctl git:(anon-volume) ✗ docker run --mount type=volume,dst=/foo,readonly alpine
docker: Error response from daemon: invalid mount config for type "volume": must not set ReadOnly mode when using anonymous volumes.
See 'docker run --help'.
➜  nerdctl git:(anon-volume) ✗ docker run --mount type=volume,dst=/foo,rw alpine      
invalid argument "type=volume,dst=/foo,rw" for "--mount" flag: invalid field 'rw' must be a key=value pair
See 'docker run --help'.
➜  nerdctl git:(anon-volume) ✗ 
```

manugupt1:
Ready for review.Ptal

pkg/mountutil/mountutil.go
```diff
@@ -57,9 +57,12 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
 	)
 
 	split := strings.Split(s, ":")
-	switch len(split) {
-	case 1:
+
+	if len(split) == 1 || split[0] == "" {
```
junnplus, line 61: Why not just remove the leading characters ":" ?

manugupt1: Sure, we can do that. I just wanted to isolate under this block. Let me know if you prefer stripping

junnplus (approved):
Thanks, LGTM

pkg/mountutil/mountutil.go
```diff
@@ -56,6 +56,7 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore) (*Processed, error
 		options  []string
 	)
 
+	s = strings.TrimLeft(s, ":")
```
AkihiroSuda, line 59: I'm confused. This code seems for parsing `-v` but the added test uses `--mount`.

junnplus: `--mount` also deal with here
https://github.com/containerd/nerdctl/blob/7dbbc311506c70edcfea9e4f300b4e1309c3ddda/pkg/mountutil/mountutil_linux.go#L418-L427

AkihiroSuda (approved):
Thanks

PR #1393, opened 2022-09-26
Add basic workflow for automatic subcommand testing on Windows.

aznashwan:
This PR adapts some existing command tests to run on both Linux and Windows, as well as add a simple GitHub action to run the command tests on Windows 2019/2022 automatically.


aznashwan:
It might be worth discussing whether `nerdctl push` should also be included in this workflow (and if so, to which public/private registry implementation), but considering the efforts to port BuildKit on Windows are still ongoing it is unlikely to be critical for users at this stage...

The rest of the failing commands currently being tested have clear causes for the most part (`image encrypt/decrypt` have keyfile encoding issues, `pause/unpause/cp` are not yet implemented, and `rename` errors out), and will be addressed in dedicated PRs.

.github/workflows/windows-acceptance-basic.yml
```diff
@@ -0,0 +1,415 @@
...
+  pull_request:
+    branches:
+      - master
+
+env:
+  GO_VERSION: '1.19'
+  CONTAINERD_REF: 'main'
+  TEST_CONTAINER_IMAGE: 'k8s.gcr.io/e2e-test-images/busybox:1.29-2'
+
+jobs:
```
manugupt1, line 20: Would it be better to have a test arg like docker.
```
go test -exec sudo -v ./cmd/nerdctl/... -args -test.target=docker
```

and if a test is windows compatible it will run automatically

AkihiroSuda:
Why not use the same test suite as Linux?

.github/workflows/windows-acceptance-basic.yml
```diff
@@ -0,0 +1,101 @@
+# This workflow runs the command test suite on GitHub-based Windows Server runners.
```
AkihiroSuda, line 1: Could you explain pros/cons compared to the current Cirrus CI ?

aznashwan:
@AkihiroSuda @manugupt1 thanks a lot for the pointers!
I was initially apprehensive on adapting the existing tests since they seem to have been mostly written with Linux in mind but it makes more sense to unify the *nix/Windows testing as much and as soon as possible!

I've updated this PR in the following ways:
* made the workflow setup containerd and just run `go test ./cmd/...` on the GitHub Windows runners (we'd like to have it tested on both 2019 and 2022 on semi-up-to-date Windows images and containerd release so IMO having a per-PR GitHub workflow is more useful than Cirrus here)
* enabled the image encryption/conversion tests on Windows (the commands just worked OOB)
* adapted some tests which used to require image building during the test to work around this on Windows until BuildKit is ported too (namely re-tagging or converting the `testutil.CommonImage`)

There's like a dozen existing tests which are skipped on Windows since they unequivocally require BuildKit as part of the test itself (mostly all of those prefixed with `TestBuild*`), so I have left those as-is for now.

There's a couple of tests which outright fail and I'll be looking into next:
* `TestRename` yields an internal OS error
* `TestRenameUpdate` fails to start the test container (almost surely related to `TestRename`)
* `TestEncryptJWE` (tests `nerdctl image encrypt --jwe=...`) needs a registry running during the test but the current test registry image is Linux-only

Lastly, some points I'd like to bring to your attention now (with separate PRs to work through these later):
* some tests which were cross-platform and executed on Unix too have been "deactivated" by giving them a Windows and Linux versions only (`TestImagePrune` and `TestImagesFilter`), but I've seen that [the test suite isn't even run on FreeBSD in the Cirrus config](https://github.com/containerd/nerdctl/blob/master/.cirrus.yml#L19) so no actual "regression" happened.
* `nerdctl logs` will [always exec the `tail` binary when using simple JSON file logging](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/logs.go#L137), which is impractical to expect to be present on Windows installs. Afaik `docker logs` always just prints/streams to stdout so I was wondering whether it's really worth having `nerdctl` act differently in this case?
* how should I go about updating [the Windows compatibility for the commands in the README](https://github.com/containerd/nerdctl#command-reference)? It *looks* like the list was generated with DocToc but it looks like any updates are manually committed by contributors, correct?

.github/workflows/windows-acceptance-basic.yml
```diff
@@ -0,0 +1,101 @@
...
+      GOTEST: gotestsum --
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-2019, windows-2022]
+
+    defaults:
+      run:
+        shell: bash
```
manugupt1, line 35: Should this be bash or PowerShell?

cmd/nerdctl/build_test.go
```diff
@@ -204,17 +204,6 @@ COPY %s /`,
 	assert.Equal(t, string(data), testContent)
 }
 
-func createBuildContext(dockerfile string) (string, error) {
```
manugupt1, line 207: Why remove this?

cmd/nerdctl/rename_windows_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+package main
+
+import (
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestRename(t *testing.T) {
```
manugupt1, line 25: Just curious: Why not have common tests in rename_test.go and platform specific tests in rename_<os>_test.go using build tags?

.github/workflows/windows-acceptance-basic.yml
aznashwan: Good catch, the initial version of the workflow contained way more steps and mostly used bash in them but none do any more so I'll be updating the workflow to have PS as the default.

cmd/nerdctl/build_test.go
aznashwan: I had moved it to [`image_testing_utils.go`](https://github.com/containerd/nerdctl/pull/1393/files#diff-41f6b435fd39f41851874b66da27db4ec63cfe4a5d7ac9a168bf4acdf0037b8fR29) since it is also called in a couple of non-build-related tests.

cmd/nerdctl/rename_windows_test.go
aznashwan: As far as I could tell all the existing tests were laid out using the filename suffixes for build constraints so I strived to keep it that way.

I also needed to paraemtrize the test image since `testutils.CommonImage` is only defined on Linux/Windows/FreeBSD and that would prevent the tests from compiling on any other future platform we'd like to support.

We could alternatively set a default `testutils.CommonImage` for all platforms and thus have the `TestRename*` tests be 100% shared across all platforms.
I'd personally rather we move the `CommonImage` constants from `testutils` to here since it's only ever referenced in the `./cmd/nerdctl/...` tests anyway...

.github/workflows/windows-acceptance-basic.yml
aznashwan: Updated the PR to use PowerShell as the default shell FWIW.

aznashwan:
Superseded by #2418.

PR #1399, opened 2022-09-28
Add a name for containerd worker when name is not shown.

manugupt1:


cmd/nerdctl/build.go
```diff
@@ -324,6 +324,9 @@ func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, arg
 		return "", nil, false, "", nil, nil, err
 	}
 
+	if strings.Contains(output, "type=image") && len(tags) == 0 {
+		output += ",name=none:none"
+	}
```
AkihiroSuda, line 329: Can we rather fix BuildKit?

manugupt1: LMK see what I can do.

manugupt1: @AkihiroSuda I landed this issue in buildkit; https://github.com/moby/buildkit/issues/3212 (I believe this goes beyond just renaming); Would it be possible for you to suggest there, what can I do to solve it in buildkit itself or evaluate if this PR will be acceptable given that bug?

PR #1402, opened 2022-09-30
Makefile: Set VERSION & REVISION only if unset

sshedi:
This is useful while building nerdctl from git tag. For example, while building rpm from a tagged source tarball.

Signed-off-by: Shreenidhi Shedi <sshedi@vmware.com>


AkihiroSuda (approved):
Thanks

sshedi:
@AkihiroSuda you're welcome and thank you as well.

PR #1403, opened 2022-10-01
Drop dependency on github.com/docker/docker/registry

AkihiroSuda:


yardenshoham (approved):

fahedouch:
@AkihiroSuda Why dropping this dependency ?

dims:
@fahedouch because it causes loops in dependency chains.

fahedouch:
> @fahedouch because it causes loops in dependency chains.

@dims it is need by the main module, this is what you consider a loop ? are you pointing to loop complexity ?
```
FRCOU1WLZ33592:nerdctl u1138593$ go mod why github.com/docker/docker/registry
# github.com/docker/docker/registry
github.com/containerd/nerdctl/cmd/nerdctl
github.com/docker/docker/registry
```

Other point, I thought this value needs to be maintained by the moby project, what if it changes in the future?

AkihiroSuda:
This is not for eliminating loops but for reducing binary footprint of other projects that imports `dockerconfigresolver` pkg.
The binary footprint of `nerdctl` itself doesn't seem affected though.

> Other point, I thought this value needs to be maintained by the moby project, what if it changes in the future?

Just copy the single-line change.

fahedouch (approved):

Issue #1293, opened 2022-08-02
nerdctl run not support --pid=container:xxxx and --ipc=container:xxx but docker is  supported


ibayue:
### Description

nerdctl run not support --pid=container:xxxx and --ipc=container:xxx but docker is  supported

![image](https://user-images.githubusercontent.com/11729209/182340337-a8aeeaeb-d160-43eb-baaf-d90530958c1b.png)

![image](https://user-images.githubusercontent.com/11729209/182340814-05cf8f46-fc6c-447c-b9f2-3c300219ecb7.png)

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

Using nerdctl to create a new container can share pid ipc with other containers

### What version of nerdctl are you using?

 0.22.2

fahedouch:
`joins another container's PID namespace` is not yet supported , same thing for `ipc`

minuk-dev:
I think this issue was resolved by https://github.com/containerd/nerdctl/pull/2757, https://github.com/containerd/nerdctl/pull/1411. Could you check it?

djdongjin:
Close as completed by #2757

PR #1411, opened 2022-10-04
Support --pid=container:xxx for `nerdctl run` cmd

minuk-dev:
#1293 
Signed-off-by: Min Uk Lee <minuk.dev@gmail.com>

### Changes
- support `--pid=container:xxx` for `nerdctl run` command
- extract `generatePIDOpts()` function for pid
- (Updated) update `setPlatformOptions()`
- (Updated) add `setPlatformContainerOptions()` - Details 1
- (Updated) add `PIDContainer` label - Details 2

#### 1. add `setPlatformContainerOptions()`
- Some functions only works on a specific platform (mostly linux). For persistency, they need to save information into container's labels.
- Previously, it was processed on runtime. (e.g. [hostname](https://github.com/containerd/nerdctl/blob/b4c01094b5815e7616c43151086c708ddd57bc37/cmd/nerdctl/run.go#L569), [logURI](https://github.com/containerd/nerdctl/blob/b4c01094b5815e7616c43151086c708ddd57bc37/cmd/nerdctl/run.go#L529)). It caused [too many number of `withInternalLabels()`'s parameters](https://github.com/containerd/nerdctl/blob/b4c01094b5815e7616c43151086c708ddd57bc37/cmd/nerdctl/run.go#L660)
- In my opinion, to keep up with it, if the number of retrun values ​​of `setPlatformOptions()` increase, it would be difficult to maintain the code quality.
- Therefore, I suggest adding `addPlatformContainerOptions()`.

#### 2. add `PIDContainer` label

> Thanks, but we need to deal with shared container restarts, it will change pid namesapce.

- For `--pid` flag's persistency, save the based container's id into the shared container's label.
- When the container restart, `reconfigPIDContainer()` will recover it in `startContainer()`

---
### a difference with docker
- Because docker supports to isolate a container with a user namespace, it should share user namespace between containers when containers share their pid namespace.
  - [docker docs](https://docs.docker.com/engine/security/userns-remap/)
  - [moby/moby/daemon/oci_linux.go#L231-L240](https://github.com/moby/moby/blob/6c1df4f9c53b604866526e0f2aa4d34614c5f7ab/daemon/oci_linux.go#L231-L240)
- But, nerdctl does not support the user ns isolation.

### references
- [moby/moby/daemon/oci_linux.go#L294-L321](https://github.com/moby/moby/blob/6c1df4f9c53b604866526e0f2aa4d34614c5f7ab/daemon/oci_linux.go#L294-L321)
- [man pages - pid_namespaces](https://man7.org/linux/man-pages/man7/pid_namespaces.7.html)

### DIFF
--- a/cmd/nerdctl/restart_linux_test.go
+++ b/cmd/nerdctl/restart_linux_test.go
@@ -17,6 +17,8 @@
 package main
 
 import (
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/containerd/nerdctl/pkg/testutil"
@@ -41,4 +43,29 @@ func TestRestart(t *testing.T) {
 	newInspect := base.InspectContainer(tID)
 	newPid := newInspect.State.Pid
 	assert.Assert(t, pid != newPid)
+}
+
+func TestRestartPIDContainer(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+
+	baseContainerName := testutil.Identifier(t)
+	base.Cmd("run", "-d", "--name", baseContainerName, testutil.AlpineImage, "sleep", "infinity").Run()
+	defer base.Cmd("rm", "-f", baseContainerName).Run()
+
+	sharedContainerName := testutil.Identifier(t)
+	base.Cmd("run", "-d", "--name", sharedContainerName, fmt.Sprintf("--pid=container:%s", baseContainerName), testutil.AlpineImage, "sleep", "infinity").Run()
+	defer base.Cmd("rm", "-f", sharedContainerName).Run()
+
+	base.Cmd("restart", baseContainerName).AssertOK()
+	base.Cmd("restart", sharedContainerName).AssertOK()
+
+	// output format : <inode number> /proc/1/ns/pid
+	// example output: 4026532581 /proc/1/ns/pid
+	basePSResult := base.Cmd("exec", baseContainerName, "ls", "-Li", "/proc/1/ns/pid").Run()
+	baseOutput := strings.TrimSpace(basePSResult.Stdout())
+	sharedPSResult := base.Cmd("exec", sharedContainerName, "ls", "-Li", "/proc/1/ns/pid").Run()
+	sharedOutput := strings.TrimSpace(sharedPSResult.Stdout())
+
+	assert.Equal(t, baseOutput, sharedOutput)
 }
--- a/cmd/nerdctl/run_linux.go
+++ b/cmd/nerdctl/run_linux.go
@@ -21,7 +21,10 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/containerd/containerd"
 	"github.com/containerd/nerdctl/pkg/bypass4netnsutil"
+	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
+	"github.com/containerd/nerdctl/pkg/labels"
 	"github.com/containerd/nerdctl/pkg/rootlessutil"
 	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/docker/go-units"
@@ -55,7 +58,7 @@ func runShellComplete(cmd *cobra.Command, args []string, toComplete string) ([]s
 	}
 }
 
-func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]oci.SpecOpts, error) {
+func setPlatformOptions(ctx context.Context, opts []oci.SpecOpts, cmd *cobra.Command, client *containerd.Client, id string) ([]oci.SpecOpts, error) {
 	opts = append(opts,
 		oci.WithDefaultUnixDevices,
 		WithoutRunMount(), // unmount default tmpfs on "/run": https://github.com/containerd/nerdctl/issues/157)
@@ -128,21 +131,15 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
 		opts = append(opts, oci.WithDevShmSize(shmBytes/1024))
 	}
 
-	pidNs, err := cmd.Flags().GetString("pid")
+	pid, err := cmd.Flags().GetString("pid")
 	if err != nil {
 		return nil, err
 	}
-	pidNs = strings.ToLower(pidNs)
-	if pidNs != "" {
-		if pidNs != "host" {
-			return nil, fmt.Errorf("invalid pid namespace. Set --pid=host to enable host pid namespace")
-		} else {
-			opts = append(opts, oci.WithHostNamespace(specs.PIDNamespace))
-			if rootlessutil.IsRootless() {
-				opts = append(opts, withBindMountHostProcfs)
-			}
-		}
+	pidOpts, err := generatePIDOpts(ctx, client, pid)
+	if err != nil {
+		return nil, err
 	}
+	opts = append(opts, pidOpts...)
 
 	ulimitOpts, err := generateUlimitsOpts(cmd)
 	if err != nil {
@@ -192,6 +189,21 @@ func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, id string) ([]o
 	return opts, nil
 }
 
+func setPlatformContainerOptions(ctx context.Context, cOpts []containerd.NewContainerOpts, cmd *cobra.Command, client *containerd.Client, id string) ([]containerd.NewContainerOpts, error) {
+	pid, err := cmd.Flags().GetString("pid")
+	if err != nil {
+		return nil, err
+	}
+
+	pidCOpts, err := generatePIDCOpts(ctx, client, pid)
+	if err != nil {
+		return nil, err
+	}
+	cOpts = append(cOpts, pidCOpts...)
+
+	return cOpts, nil
+}
+
 func setOOMScoreAdj(opts []oci.SpecOpts, cmd *cobra.Command) ([]oci.SpecOpts, error) {
 	if !cmd.Flags().Changed("oom-score-adj") {
 		return opts, nil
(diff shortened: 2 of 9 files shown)


cmd/nerdctl/run_linux_test.go
```diff
@@ -75,6 +75,18 @@ func TestRunPidHost(t *testing.T) {
 	base.Cmd("run", "--rm", "--pid=host", testutil.AlpineImage, "ps", "auxw").AssertOutContains(strconv.Itoa(pid))
 }
 
+func TestRunPidContainer(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+
+	sharedContainerResult := base.Cmd("run", "-d", testutil.AlpineImage, "sleep", "30").Run()
```
junnplus, line 82: sleep infinity to avoid flaky.

cmd/nerdctl/run_linux.go
```diff
@@ -217,3 +210,90 @@ func withOOMScoreAdj(score int) oci.SpecOpts {
...
+			Client: client,
+			OnFound: func(ctx context.Context, found containerwalker.Found) error {
+				if found.MatchCount > 1 {
+					return fmt.Errorf("multiple IDs found with provided prefix: %s", found.Req)
+				}
+
+				nc, err := containerinspector.Inspect(ctx, found.Container)
+				if err != nil {
+					return err
+				}
```
junnplus, line 254: Avoid using inspect, it will do extra work

cmd/nerdctl/run_linux.go
```diff
@@ -217,3 +210,90 @@ func withOOMScoreAdj(score int) oci.SpecOpts {
...
+		matchedCount, err := walker.Walk(ctx, containerName)
+		if err != nil {
+			return nil, err
+		}
+		if matchedCount < 1 {
+			return nil, fmt.Errorf("no such container: %s", containerName)
+		}
+	}
+
+	return oci.Compose(opts...), nil
```
junnplus, line 298: why compose?

junnplus (changes requested):
We need to deal with shared container restarts, it will change pid namesapce.

And please update the doc.

cmd/nerdctl/run_linux.go
minuk-dev: no need. I'll fix the function returns []oci.Specs.

minuk-dev:
> Thanks, but we need to deal with shared container restarts, it will change pid namesapce.

Could I get an usecase for it? I could not find it because of my little knowledge.
In my opinion, if containers restart, their namespaces should be reset, isn't it?

Here are my manual tests.
In these cases, docker and nerdctl work the same.

- restart base container test
```bash
$ docker run -d ubuntu sleep infinity # run a base container
9bd0d57b85dcbbe7eeabf72f867bb130ddaf35ca716b80f15f6612b92705124f

$ docker run --pid=container:9bd0d -d ubuntu sleep infinity # run shared container
cfe162ae92814624e44f06db81ecba3f016508a72708ee917ad9e60fe0fe663c

$ docker ps # check containers' status
CONTAINER ID   IMAGE     COMMAND            CREATED          STATUS          PORTS     NAMES
cfe162ae9281   ubuntu    "sleep infinity"   7 seconds ago    Up 6 seconds              gallant_galois
9bd0d57b85dc   ubuntu    "sleep infinity"   45 seconds ago   Up 44 seconds             peaceful_mendel

$ docker restart 9bd0d # restart the base container
9bd0d

$ docker ps # only base container survived
CONTAINER ID   IMAGE     COMMAND            CREATED              STATUS        PORTS     NAMES
9bd0d57b85dc   ubuntu    "sleep infinity"   About a minute ago   Up 1 second             peaceful_mendel
```

- restart shared container test
```
$ docker run -d ubuntu sleep infinity # run a base container
57f18c82ab5ebbf49d6adbded478abf85d71172ee06de5e0fb88acf552493957
$ docker run --pid=container:57f18 -d ubuntu sleep infinity # run a shared container
9de0f487563cb709f0065aecc6a02c4eccb2c2c5089ad5bcf885250ac993a4a1

$ docker restart 9de0f # restart the shared container
9de0f

$ docker ps 
CONTAINER ID   IMAGE     COMMAND            CREATED          STATUS         PORTS     NAMES
9de0f487563c   ubuntu    "sleep infinity"   16 seconds ago   Up 2 seconds             gracious_dewdney
57f18c82ab5e   ubuntu    "sleep infinity"   2 minutes ago    Up 2 minutes             beautiful_almeida

# all survived but, not shared.
$ docker exec -it 9de0f bash
root@9de0f487563c:/# ps
    PID TTY          TIME CMD
     18 pts/0    00:00:00 bash
     26 pts/0    00:00:00 ps
```

cmd/nerdctl/run_linux.go
```diff
@@ -217,3 +209,94 @@ func withOOMScoreAdj(score int) oci.SpecOpts {
...
+	switch pid {
+	case "":
+		// do nothing
+	case "host":
+		opts = append(opts, oci.WithHostNamespace(specs.PIDNamespace))
+		if rootlessutil.IsRootless() {
+			opts = append(opts, withBindMountHostProcfs)
+		}
+	default: // container:<id|name>
+		client, ctx, cancel, err := newClient(cmd)
```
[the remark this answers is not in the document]
fahedouch: platform different from linux signature should be like this 
```
func setPlatformOptions(opts []oci.SpecOpts, cmd *cobra.Command, client *containerd.Client, id string) ([]oci.SpecOpts, error) {
      return opts, nil
} 
```

minuk-dev:
> > Could I get an usecase for it? I could not find it because of my little knowledge.
> > In my opinion, if containers restart, their namespaces should be reset, isn't it?
> 
> You can test restart the base container and then restart the shared container.

I fixed it using labels at 7a6d4250619c76eda7d2d2d748d9cf2fbc24313e.

Summary of updates:
- add `setPlatformContainerOptions()` at 82e3d04723bc3bd86f7dd075da5ff091e6432022
- add `reconfigPIDContainer()`
- add `PIDContainer` label
- test this feature in [TestRestartPIDContainer](https://github.com/minuk-dev/nerdctl/blob/7a6d4250619c76eda7d2d2d748d9cf2fbc24313e/cmd/nerdctl/restart_linux_test.go#L48).

Can you review it again? @junnplus 

trivial: I updated the doc of pr. It is marked as updated.

cmd/nerdctl/run_linux.go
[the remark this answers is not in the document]
minuk-dev: Thanks for your kind reply.
I updated it on 8de12b934788734ce480cdc2d2b8e338d2af3697.

cmd/nerdctl/restart_linux_test.go
```diff
@@ -42,3 +44,28 @@ func TestRestart(t *testing.T) {
 	newPid := newInspect.State.Pid
 	assert.Assert(t, pid != newPid)
 }
+
+func TestRestartPIDContainer(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+
+	baseContainerResult := base.Cmd("run", "-d", testutil.AlpineImage, "sleep", "infinity").Run()
+	baseContainerID := strings.TrimSpace(baseContainerResult.Stdout())
```
junnplus, line 53: Using `testutil.Identifier(t)` as container name.

cmd/nerdctl/run_linux.go
```diff
@@ -217,3 +229,85 @@ func withOOMScoreAdj(score int) oci.SpecOpts {
...
+		if matchedCount < 1 {
+			return nil, fmt.Errorf("no such container: %s", containerName)
+		}
+
+		o, err := generateSharingPIDOpts(ctx, targetCon)
+		if err != nil {
+			return nil, err
+		}
+
+		opts = append(opts, o...)
```
junnplus, line 276: Handle in `OnFound`

cmd/nerdctl/run_linux.go
```diff
@@ -217,3 +229,85 @@ func withOOMScoreAdj(score int) oci.SpecOpts {
...
+
+	cOpts := make([]containerd.NewContainerOpts, 0)
+	parsed := strings.Split(pid, ":")
+	if len(parsed) < 2 || parsed[0] != "container" {
+		// no need to save pid options
+		return cOpts, nil
+	}
+
+	containerName := parsed[1]
+	walker := &containerwalker.ContainerWalker{
```
junnplus, line 293: Why do we need to find again?

cmd/nerdctl/run_linux.go
```diff
@@ -217,3 +229,85 @@ func withOOMScoreAdj(score int) oci.SpecOpts {
...
+		Client: client,
+		OnFound: func(ctx context.Context, found containerwalker.Found) error {
+			if found.MatchCount > 1 {
+				return fmt.Errorf("multiple IDs found with provided prefix: %s", found.Req)
+			}
+			cOpts = append(cOpts, containerd.WithAdditionalContainerLabels(map[string]string{
+				labels.PIDContainer: containerName,
+			}))
```
junnplus, line 301: We already have `withInternalLabels`, should be put together

junnplus (changes requested):
Please squash commits

cmd/nerdctl/restart_linux_test.go
minuk-dev: I know that the run command with `-d` should print a container's id.
Why should we name for it?

cmd/nerdctl/run_linux.go
minuk-dev: When a new label is needed, the best way is to add a new parameter of `withInternalLabels()`?
And if so, `--pid` flags is processed in `setPlatformOptions()`. In order to pass into the `withInternalLabels()`, the number of `setPlatformOptions()`'s return value should be increased.

cmd/nerdctl/restart_linux_test.go
junnplus: Ask in turn, why do we need to parse the id from stdout?

AkihiroSuda (approved):
Thanks

minuk-dev:
> Thanks

I'm sorry that I didn't apply reviews from jun. I'm looking forward the better way than setplatformcontaineropts().
In my opinion, that's too hard to me now.
So, could you decide how to deal with this pr (e.g. because of low code quality, just close).
If pr is closed, then I'll try to improve the quality and retry pr.

cmd/nerdctl/run_linux.go
junnplus: > And if so, --pid flags is processed in setPlatformOptions().

I think the `--pid` flags should be platform independent, why not consider moving it out of `setPlatformOptions`?

junnplus (changes requested):
Sorry, I guess the code can be optimized again.

cmd/nerdctl/run_linux.go
minuk-dev: Previsouly, `--pid` only works on linux.
https://github.com/containerd/nerdctl/blob/c6517c7f9a178950b965748f6698f5ca4e76e318/cmd/nerdctl/run_linux.go#L135-L144

The reason, why it only works on the linux, is that it's too sensitive with namespace suppport.

https://github.com/containerd/nerdctl/blob/d15ada7bec3df3a4e2b6d78bdc9f5d518aaf9d44/cmd/nerdctl/run.go#L1184-L1187

```golang
// WithHostNamespace allows a task to run inside the host's linux namespace
func WithHostNamespace(ns specs.LinuxNamespaceType) SpecOpts {
	return func(_ context.Context, _ Client, _ *containers.Container, s *Spec) error {
		setLinux(s)
		for i, n := range s.Linux.Namespaces {
			if n.Type == ns {
				s.Linux.Namespaces = append(s.Linux.Namespaces[:i], s.Linux.Namespaces[i+1:]...)
				return nil
			}
		}
		return nil
	}
}
```

To avoid a compile error, I placed `reconfigPIDContainer()` in `cmd/nerdctl/start.go` but I think it caused confusing.
So, I added a condition to check runtime os now. https://github.com/containerd/nerdctl/blob/8680bdb9a24d0c99fba6682af7473124469cb2fd/cmd/nerdctl/start.go#L231

Anyway, in my opinion, `--pid` currently can work on linux only.
Because of it, `--pid` should be processed in `setPlatformOptions()`. But, I really agree with you. `setPlatformContainerOptions()` can make a fragmentation and cause to find a target container twice.
I tried to find a better way to improve it according to your review, but I couldn't. I'm really sorry.

AkihiroSuda:
@junnplus LGTY?

AkihiroSuda:
Let me merge this, as I don't think the `--pid` flag can be platform-independent in the foreseeable future, but please feel free to submit follow-up issues/PRs.

Thank you @minuk-dev @junnplus 👍

PR #1427, opened 2022-10-12
Use sys/windows to lock file

iyear:
Signed-off-by: iyear <ljyngup@gmail.com>

The official extension library contains a windows implementation of locking file, and boltdb is no longer maintained.


AkihiroSuda (approved):
Thanks

PR #1455, opened 2022-10-19
Dockerfile: demo: configure BuildKit to use containerd worker; enable bash completions

AkihiroSuda:
- configure BuildKit to use containerd worker
- enable bash completions


fahedouch:
LGTM but we should mention somewhere that containerd comes with containerd as default backend

AkihiroSuda:
> LGTM but we should mention somewhere that containerd comes with containerd as default backend

Only the demo image contains the new buildkit.toml.
The nerdctl-full distritbution does not contain the toml, to avoid overriding the existing configuration.

Issue #1222, opened 2022-07-13
TestRunWithJsonFileLogDriver is flaky


manugupt1:
### Description

=== PAUSE TestRunStdin
=== RUN   TestRunWithJsonFileLogDriver
    run_test.go:237: the maximum number of old log files to retain exceeded 2 files
--- FAIL: TestRunWithJsonFileLogDriver (3.42s)

https://github.com/containerd/nerdctl/runs/7312918520?check_suite_focus=true

### Steps to reproduce the issue

N/A

### Describe the results you received and expected

Test should pass.

### What version of nerdctl are you using?

built on master

### Host information

Fedora 36

AkihiroSuda:
=== RUN   TestRunWithJsonFileLogDriver
[...]

PR #1115, opened 2022-06-09
container logs do not invoke logs cmd

fahedouch:
PR fixes :
* `nerdctl compose logs` no longer rely on `nerdctl logs`
* `nerdctl compose up` log output no longer rely on `nerdctl logs`
*  refacto `nerdctl logs`

And 
fixing https://github.com/containerd/nerdctl/issues/1071
Signed-off-by: Fahed DORGAA <fahed.dorgaa@gmail.com>

PR #1468, opened 2022-10-27
Refactor `nerdctl logs` functionality to pkg/logging. 

aznashwan:
NOTE: requires https://github.com/containerd/containerd/pull/7351

This patch refactors log-reading-related logic used by `nerdctl logs`
into the pkg/logging module.

Additionally, checks and workarounds for the `tail` executable not 
available on the system have also been implemented, as it was previously
always called for jsonfile-based container logs.

Signed-off-by: Nashwan Azhari <nazhari@cloudbasesolutions.com>


pkg/logging/log_viewer.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Start/end timestampts to filter logs by.
+	Since string
+	Until string
+}
+
+func (lvo LogViewingOptions) InternalValidate(idsRequired bool) error {
+	if idsRequired && (lvo.ContainerId == "" || lvo.NamespaceId == "") {
+		return fmt.Errorf("log viewing options require a ContainerId and NamespaceId: %#v", lvo)
+	}
```
fahedouch, line 84: we always need `ContainerId` and `Namespace`. This check is useless

pkg/logging/log_viewer.go
```diff
@@ -0,0 +1,141 @@
...
+	lcfg, err := LoadLogConfig(lvopts.DatastoreRootPath, lvopts.NamespaceId, lvopts.ContainerId)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load logging config: %s", err)
+	}
+	lv := &ContainerLogViewer{
+		loggingConfig:     lcfg,
+		logViewingOptions: lvopts,
+		stopChannel:       stopChannel,
+	}
+	lv.loggingConfig = lcfg
```
fahedouch, line 122: redundant https://github.com/containerd/nerdctl/pull/1468/files#diff-eabf36f45659163d25efc51651ed872d0d7d96de5fe03318b01dcec12cc1c726R118

cmd/nerdctl/logs.go
```diff
@@ -96,6 +99,10 @@ func logsAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	stopChannel := make(chan os.Signal, 1)
+	// catch SIGETRM/SIGINT
```
fahedouch, line 103: `SIGKILL` too ?

pkg/logging/jsonfile/jsonfile.go
```diff
@@ -78,7 +79,69 @@ func Encode(w io.WriteCloser, stdout, stderr io.Reader) error {
...
+		ts, err := timetypes.GetTimestamp(since, refTime)
+		if err != nil {
+			return fmt.Errorf("invalid value for \"since\": %w", err)
+		}
+		v := strings.Split(ts, ".")
+		i, err := strconv.ParseInt(v[0], 10, 64)
+		if err != nil {
+			return err
+		}
+		if !e.Time.After(time.Unix(i, 0)) {
```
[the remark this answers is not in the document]
aznashwan: 100% agreed, it was just a move of the [existing entry filtering logic](https://github.com/containerd/nerdctl/blob/master/pkg/logging/jsonfile/jsonfile.go#L104), fixed now!

cmd/nerdctl/logs.go
aznashwan: I did update the code to include `SIGKILL` but luckily the [linter was there to stop us](https://github.com/containerd/nerdctl/actions/runs/3345763466/jobs/5541705573#step:4:29) from even trying. 😆

pkg/logging/log_viewer.go
aznashwan: Do you mean the arg `isRequred` was useless for this function (which I have removed as I agree), or that we should not allow `LogViewOptions` to ever be created without a ContainerID/Namespace? (in which case I'd private its fields and make a function to build them which would perform this check)

cmd/nerdctl/logs.go
```diff
@@ -106,83 +113,45 @@ func logsAction(cmd *cobra.Command, args []string) error {
...
-					if err != nil {
-						return fmt.Errorf("invalid value for \"until\": %w", err)
-					}
-					date, err := prepareJournalCtlDate(ts)
-					if err != nil {
-						return err
+					if waitCh, err := task.Wait(ctx); err == nil {
+						<-waitCh
+						logrus.Debugf("container task has finished, sending kill event to log viewer")
+						stopChannel <- os.Interrupt
```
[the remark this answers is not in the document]
aznashwan: Great catch, seeing as though a [restart is just a sequential stop and start](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/restart.go#L66-L71), this would kill the existing `task` and thus stop the `logs -f` command too. (just tested it myself and can confirm it stops)

We could simply requery the container's current task and status but I'm worried we could catch the container "in between restarts" and before it actually reaches `Up` status, but I'll try to figure something out...

aznashwan:
@fahedouch thanks a lot for the review, your older PR (#1115) was a partial inspiration for it!

One thing worth noting in this PR is [this commit](https://github.com/containerd/nerdctl/pull/1468/commits/6db7b19d54992454cef51d0f09af23fff0e90058) which actually enables the `nerdctl/log-url` to be passed to containerd on Windows too. (without it, no container output logging is done on Windows at all)

It depends on https://github.com/containerd/containerd/pull/7351, without which [the `cio.LogURIGenerator` call right after](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/run.go#L900) will always report `absolute path needed` for any container created on Windows. ([visible in Cirrus too](https://github.com/containerd/nerdctl/pull/1468/checks?check_run_id=9164939680), since it uses containerd 1.6)

Is an [update to the containerd ref in go.mod](https://github.com/containerd/nerdctl/blob/master/go.mod#L11) planned?
I could separate that commit into its own PR until that time comes too...

cmd/nerdctl/logs.go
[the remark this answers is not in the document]
aznashwan: @fahedouch I've made some [changes for working around a container restart](https://github.com/aznashwan/nerdctl/commit/2c980ea72cd76e2f0bf76abb82598d894b28d88c) on a separate branch but unfortunately the JSON file based logging [rotates the logfile](https://github.com/containerd/nerdctl/blob/master/pkg/logging/json_logger.go#L80-L82) by moving `$CONTAINER_ID-json.log` to `$CONTAINER_ID-json.log.1` every time a container is started, which leads to:
* the `tail`-based implementation failing to re-open the new logfile, so `tail` keeps feeding the old `$CONTAINER_ID-json.log.1` file which is no longer being written to until the container is finally stopped or exists
* the native implementation will attempt re-open the file at the original path but preserve the file seek, which could lead to it seeking over the rotated logfile's fill

Although both of the above could be theoretically bypassed by sending a `syscall.SIGHUP` on the `stopChannel` and having the `LogViewer` refresh the log viewing, I don't consider the technical overhead to be worth it for something this edge-casey...

Please note I've also tested `docker logs` from Docker CE 20.10.18 and can report that:
* `docker logs -f` also closes itself if the container `docker restart`ed so in that sense the current behavior is correct
* `docker logs` does **not** seem to rotate logfiles so us using logrotate is deviating a bit from how Docker stores logs

fahedouch:
> @fahedouch thanks a lot for the review, your older PR (#1115) was a partial inspiration for it!
> 
> One thing worth noting in this PR is [this commit](https://github.com/containerd/nerdctl/pull/1468/commits/6db7b19d54992454cef51d0f09af23fff0e90058) which actually enables the `nerdctl/log-url` label to be passed to containerd on Windows too. (without it, no container output logging is done on Windows at all)
> 
> It depends on [containerd/containerd#7351](https://github.com/containerd/containerd/pull/7351), without which [the `cio.LogURIGenerator` call right after](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/run.go#L900) will always report `absolute path needed` for any container created on Windows. ([visible in Cirrus too](https://github.com/containerd/nerdctl/pull/1468/checks?check_run_id=9164939680), since it uses containerd 1.6)
> 
> Is an [update to the containerd ref in go.mod](https://github.com/containerd/nerdctl/blob/master/go.mod#L11) planned? I could separate that commit into its own PR until that time comes too...

I believe you can point to your commit on `containerd/main`

> @fahedouch thanks a lot for the review, your older PR (#1115) was a partial inspiration for it!
> 
> One thing worth noting in this PR is [this commit](https://github.com/containerd/nerdctl/pull/1468/commits/6db7b19d54992454cef51d0f09af23fff0e90058) which actually enables the `nerdctl/log-url` label to be passed to containerd on Windows too. (without it, no container output logging is done on Windows at all)
> 
> It depends on [containerd/containerd#7351](https://github.com/containerd/containerd/pull/7351), without which [the `cio.LogURIGenerator` call right after](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/run.go#L900) will always report `absolute path needed` for any container created on Windows. ([visible in Cirrus too](https://github.com/containerd/nerdctl/pull/1468/checks?check_run_id=9164939680), since it uses containerd 1.6)
> 
> Is an [update to the containerd ref in go.mod](https://github.com/containerd/nerdctl/blob/master/go.mod#L11) planned? I could separate that commit into its own PR until that time comes too...

I beleive you can point `github.com/containerd/containerd` to this commit https://github.com/containerd/containerd/pull/7351

pkg/logging/log_viewer.go
fahedouch: I mean the arg `isRequired` was useless for this function 👍

cmd/nerdctl/logs.go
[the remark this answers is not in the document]
fahedouch: Regarding log file rotation, I confirm that after restart nerdctl writes in the principal log file.

benchmark of mod time after a restart of the container  `nerdctl restart  0e8a3fdbf637` : 

```
root@e4276e97ebd3:/home/containerd/nerdctl# stat /var/lib/nerdctl/1935db59/containers/default/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f-json.log
  File: /var/lib/nerdctl/1935db59/containers/default/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f-json.log
  Size: 150083          Blocks: 296        IO Block: 4096   regular file
Device: 10000eh/1048590d        Inode: 3708060     Links: 1
Access: (0600/-rw-------)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2022-11-01 11:18:47.922219013 +0000
Modify: 2022-11-01 11:17:45.048155011 +0000
Change: 2022-11-01 11:17:45.048155011 +0000
 Birth: 2022-11-01 11:17:45.035155011 +0000
root@e4276e97ebd3:/home/containerd/nerdctl# stat /var/lib/nerdctl/1935db59/containers/default/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f-json.log.1 
  File: /var/lib/nerdctl/1935db59/containers/default/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f/0e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f-json.log.1
  Size: 150128          Blocks: 296        IO Block: 4096   regular file
Device: 10000eh/1048590d        Inode: 3708071     Links: 1
Access: (0600/-rw-------)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2022-11-01 11:17:45.035552345 +0000
Modify: 2022-11-01 11:17:45.035552470 +0000
Change: 2022-11-01 11:17:45.034155011 +0000
 Birth: 2022-11-01 11:12:35.079202007 +0000
````

mtime of `e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f-json.log` > mtime of `e8a3fdbf637c61db53536d9ce4a6334726b432c5e1b6101fd71e9c48d7dc93f-json.log.1`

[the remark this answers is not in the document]
Comment (aznashwan) [reply cmd/nerdctl/logs.go:134]: > I think we can make it simpler by checking wether the container exists or not
>
> The principal action(s) that can be exerted on the container that we follow its logs are:
> 
> nerdctl stop => do not trigger the stopChannel
> nerdctl stop => nerdctl start => do not trigger the stopChannel
> nerdctl restart => do not trigger stopChannel
> nerdctl kill => do not trigger the stopChannel, the container exists by no longer emitting logs
> nerdctl stop => nerdctl rm => do trigger stopChannel
> ==> we Interrupt only when the container is removed

If we were to solely rely on the container's existence, it would make `nerdctl logs -f` on a running container never stop even if the container stops. (neither by it stopping by itself or being stopped by another command)

As mentioned, `docker logs -f` exits if the container stops, and there's also a check which looks like it was meant to [explicitly test that `logs -f` exits when a container does](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/logs_test.go#L136-L137), so it looks we'll need to check the container's running status one way or the other.

> @fahedouch apologies for the confusion, I wasn't saying nerdctl is logging to the wrong file, I was saying that the tail executable will continue reading from the old file (which has been renamed to *-json.log.1) instead of the new file (*-json.log).
>
> Although none of the above are that involved, I honestly believe it'd add more code overhead than it's worth for such an edge-casey situation, but please let me know which approach you think is best.

The question of whether or not these workarounds are worth the risk of a blocking/failing `nerdctl logs -f` command still stands. (IMHO it adds too much overhead for a behaviour `docker logs -f` itself doesn't do)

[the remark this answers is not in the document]
Comment (fahedouch) [reply cmd/nerdctl/logs.go:134]: > If we were to solely rely on the container's existence, it would make nerdctl logs -f on a running container never stop even if the container stops. (neither by it stopping by itself or being stopped by another command)
As mentioned, docker logs -f exits if the container stops, and there's also a check which looks like it was meant to [explicitly test that logs -f exits when a container does](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/logs_test.go#L136-L137), so it looks we'll need to check the container's running status one way or the other.

@aznashwan Agree apologies for the confusion. Yes to be compatible with docker we should consider that following logs is conditioned by container state. so I am ok with this [code](https://github.com/aznashwan/nerdctl/commit/2c980ea72cd76e2f0bf76abb82598d894b28d88c).

I got your point regarding tail. We have delegated the read responsibility to tail. It looks like it has a limitation when the file inode number changes. I believe we can put it aside for investigation in another PR may be. let's keep the actual behavior for the moment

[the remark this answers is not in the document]
Comment (aznashwan) [reply cmd/nerdctl/logs.go]: > Yes to be compatible with docker we should consider that following logs is conditioned by container state. so I am ok with this [code](https://github.com/aznashwan/nerdctl/commit/2c980ea72cd76e2f0bf76abb82598d894b28d88c).

@fahedouch unfortunately that was the commit which initially lead me to discover [the issues caused by the log file being rotated](https://github.com/containerd/nerdctl/pull/1468#discussion_r1010483455) with both `tail` and the native implementation.

> I believe we can put it aside for investigation in another PR may be. let's keep the actual behavior for the moment :)

In this sense, I have ["downgraded" the logic](https://github.com/containerd/nerdctl/pull/1468/files#diff-7fd567884e5d5e48e4d2b715637bc3d446f402579041620814c0da24cd9c271dR136-R138) to immediately send the stop signal when the current container task finishes [as is the case in the current main branch](https://github.com/containerd/nerdctl/blob/master/cmd/nerdctl/logs.go#L144-L146).

**Comment by aznashwan** [comment]
> I beleive you can point github.com/containerd/containerd to this commit https://github.com/containerd/containerd/pull/7351

@fahedouch just pushed all revisions including [the containerd dependency update](https://github.com/containerd/nerdctl/pull/1468/commits/7db88280ee62bfedc8a679f59c0502af7d06d7be).

There seems to be a test failure on this run [on containerd 1.6.8 on 20.04 in TestExecWithUser](https://github.com/containerd/nerdctl/actions/runs/3387341420/jobs/5627959951#step:6:1114) but it seems to be pretty flaky since it worked on 20.04 3 other consecutive runs I tested ([here](https://github.com/aznashwan/nerdctl/actions/runs/3387840576), [here](https://github.com/aznashwan/nerdctl/actions/runs/3388027802), and [here](https://github.com/aznashwan/nerdctl/actions/runs/3388029509)), and other tests seemed generally flaky overall (most notably [TestRunWithJsonFileLogDriver](https://github.com/aznashwan/nerdctl/actions/runs/3387840576/jobs/5629101518#step:5:994) which randomly failed 2/3 runs, and have seen it [fail on apparently unrelated PRs](https://github.com/containerd/nerdctl/actions/runs/3341168965/jobs/5601197166#step:5:991)).

Considering that [one of the test runs was green on al platform/containerd combos](https://github.com/aznashwan/nerdctl/actions/runs/3388027802), I'd suspect the issues are external to `nerdctl`. (maybe some flaky behaviour in the containerd releases the CI jobs are testing against?)

File: cmd/nerdctl/login_linux_test.go [hunk 6]
```diff
@@ -145,12 +148,15 @@ func TestLoginWithPlainHttp(t *testing.T) {
...
-			shouldSuccess:     false,
+			regHost:    "127.0.0.1",
+			regPort:    80,
+			useRegPort: false,
+			username:   "admin",
+			password:   "validTestPassword",
+			// Following the merging of the below, any localhost/loopback registries will
+			// get automatically downgraded to HTTP so this will still succceed:
+			// https://github.com/containerd/containerd/pull/7393
+			shouldSuccess:     true,
```
AkihiroSuda, line 159: @Zheaoli PTAL, is this expected?

Zheaoli: Unexpectedly, I think the PR7393 is not included in v1.6.8 which version is dependent by nerdctl

cmd/nerdctl/logs.go
aznashwan: I've opened a backport request at https://github.com/containerd/containerd/pull/7632 and will update go.mod when it merges, thank you both!

cmd/nerdctl/login_linux_test.go
aznashwan: In that case, I've opened a [backport request for the LogURIGenerator only]( https://github.com/containerd/containerd/pull/7632) and will revert this test and update go.mod once it gets in.
Thank you both!

AkihiroSuda: I'm not sure I get the point. `go.mod` in nerdctl should point to v1.7.0-beta.0 (or later), not v1.6.x.

fahedouch: @aznashwan nerdctl depends on other [containerd commits](https://github.com/containerd/containerd/pull/6744) not present on `v1.6.x`; 
```
git tag -l --contains 3df7674058301f290fc148719d483e47f4118494 
v1.7.0-beta.0
```
We should keep pointing to `main` until `v1.17.x` I guess.

@AkihiroSuda what do you mean by expectation here ?

fahedouch:
> > I beleive you can point github.com/containerd/containerd to this commit [containerd/containerd#7351](https://github.com/containerd/containerd/pull/7351)
> 
> @fahedouch just pushed all revisions including [the containerd dependency update](https://github.com/containerd/nerdctl/pull/1468/commits/7db88280ee62bfedc8a679f59c0502af7d06d7be).
> 
> There seems to be a test failure on this run [on containerd 1.6.8 on 20.04 in TestExecWithUser](https://github.com/containerd/nerdctl/actions/runs/3387341420/jobs/5627959951#step:6:1114) but it seems to be pretty flaky since it worked on 20.04 3 other consecutive runs I tested ([here](https://github.com/aznashwan/nerdctl/actions/runs/3387840576), [here](https://github.com/aznashwan/nerdctl/actions/runs/3388027802), and [here](https://github.com/aznashwan/nerdctl/actions/runs/3388029509)), and other tests seemed generally flaky overall (most notably [TestRunWithJsonFileLogDriver](https://github.com/aznashwan/nerdctl/actions/runs/3387840576/jobs/5629101518#step:5:994) which randomly failed 2/3 runs, and have seen it [fail on apparently unrelated PRs](https://github.com/containerd/nerdctl/actions/runs/3341168965/jobs/5601197166#step:5:991)).
> 
> Considering that [one of the test runs was green on al platform/containerd combos](https://github.com/aznashwan/nerdctl/actions/runs/3388027802), I'd suspect the issues are external to `nerdctl`. (maybe some flaky behaviour in the containerd releases the CI jobs are testing against?)

@aznashwan  hard to live without FlakyTest :p . We have an [opened ticket](https://github.com/containerd/nerdctl/issues/1222) for `TestRunWithJsonFileLogDriver`. Agree it is external regarding you PR

cmd/nerdctl/login_linux_test.go
aznashwan: > I'm not sure I get the point. go.mod in nerdctl should point to v1.7.0-beta.0 (or later), not v1.6.x.

My bad, I got confused by the version string currently in go.mod saying `1.6.1`.

It looks like [this is the containerd commit currently referenced in go.mod](https://github.com/containerd/containerd/pull/7257), which would put it on early August around the `v1.6.8` area.
 
Unfortunately the LogURIGenerator fixes we depend on are not included in `v1.6.9` either so I guess the only option is switching to ` v1.7.0-beta.0`, and keeping the changes to these local registry tests.

Please lemme know if `v1.7.0-beta.0` would be okay so I can update the PR and re-run the test suite a couple more times.

pkg/logging/log_viewer.go
```diff
@@ -0,0 +1,145 @@
...
+func RegisterLogViewer(driverName string, lvfn LogViewerFunc) {
+	if v, ok := logViewers[driverName]; ok {
+		logrus.Warnf("A LogViewerFunc with name %q has already been registered: %#v, overriding with %#v either way", driverName, v, lvfn)
+	}
+	logViewers[driverName] = lvfn
+}
+
+func init() {
+	RegisterLogViewer("json-file", viewLogsJSONFile)
+	RegisterLogViewer("journald", viewLogsJournald)
```
fahedouch, line 46: what about `fluentd` still supported ?

cmd/nerdctl/logs.go
```diff
@@ -96,6 +99,10 @@ func logsAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	stopChannel := make(chan os.Signal, 1)
+	// catch OS signals:
+	signal.Notify(stopChannel, syscall.SIGTERM, syscall.SIGINT)
```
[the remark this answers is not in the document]
aznashwan: Even if we do manage to silence the linter, adding `SIGKILL` to the catch list would not have any effect since the process would never actually receive the SIGKILL, as it's handled entirely on the OS-side. (i.e. the OS kills the process without it notifying the process in any way)

pkg/logging/log_viewer.go
aznashwan: As of the making of this patch, it [doesn't look like viewing logs from `fluentd` is supported](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/logs.go#L120-L150) in `nerdctl logs`.
Specifying `nerdctl run --log-driver fluend` seems to be supported, as [it's even tested](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/run_linux_test.go#L202-L234), but [the tests read the logs from the fluentd files directly](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/run_linux_test.go#L255-L262) instead of validating the output of `nerdctl logs` like the other logging-related tests.

It's probably a bit late for me to mention this, but my immediate goal (before I saw [this TODO](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/logs.go#L119)) was getting containerd logging and `nerdctl logs` to work on Windows, so adding log viewing for fluentd/syslog would be even more beyond my initial scope 😆 

I had considered the need for adding `fluentd` and `syslog` log viewers down the line which is why I tried to make the logging viewers as pluggable and extensible as possible. (hypothetically all the logging options needed to read the logs for a given container [will be saved in the logging config on container creation](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/run.go#L527) and loaded in [`LogViewer.loggingConfig`](https://github.com/containerd/nerdctl/pull/1468/files#diff-eabf36f45659163d25efc51651ed872d0d7d96de5fe03318b01dcec12cc1c726R118-R123) so the LogViewer can go from there...)

fahedouch: SGTM

aznashwan:
@fahedouch @AkihiroSuda just changed go.mod to point to `v1.7.0-beta.0`, first test runs saw a failure in [`TestComposeUpWithBypass4netns` on 22.04](https://github.com/containerd/nerdctl/actions/runs/3412068254/jobs/5677094409#step:7:450) and [`TestIPFSComposeUp/overlayfs` on 20.04](https://github.com/containerd/nerdctl/actions/runs/3412068254/jobs/5677093026#step:6:498), though considering they worked on all other version-containerd combos I'd wager it's unrelated to the logging updates.

fahedouch (approved):

fahedouch:
Thanks

AkihiroSuda:
Sorry needs rebase

go.mod
```diff
@@ -165,17 +161,18 @@ require (
...
+	github.com/moby/sys/sequential v0.0.0-20220829095930-b22ba8a69b30 // indirect
 	golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4 // indirect
 	golang.org/x/tools v0.1.12 // indirect
 )
 
 replace (
-	// containerd main
-	github.com/containerd/containerd => github.com/containerd/containerd v1.6.1-0.20220807041520-6deb1174fbfc
+	// Please see: https://github.com/containerd/nerdctl/pull/1468#discussion_r1014006376
+	github.com/containerd/containerd => github.com/containerd/containerd v1.7.0-beta.0
```
AkihiroSuda, line 175: Probably no longer need to use `replace()`

AkihiroSuda (approved):
Thanks

Issue #1374, opened 2022-09-10
Support read logs from fluentd/journal log driver


Zheaoli:
### What is the problem you're trying to solve

For now, we just support one-way write process for fluentd/journal log driver, but in some circumstances, the user may need to use `nerdctl logs` command to fetch the log.

So we can write to the driver destination and a local JSON file for the log driver simultaneously, and the `nerdctl logs` command can fetch the log from the local log file

### Describe the solution you'd like

- [ ] fluentd
- [ ] journal
- [ ] syslog after https://github.com/containerd/nerdctl/issues/1353

yuchanns:
count me in

fahedouch:
I started a PR around this https://github.com/containerd/nerdctl/pull/1115/files a few days ago. Unfortunately, I didn't have time to finish it.  @yuchanns  this may inspire you :)

t1anz0ng:
I can take the journald driver's job😃

t1anz0ng:
I can take the journald driver job😃

manugupt1:
@Zheaoli Should this be fluentbit and not fluentd? Ref: https://github.com/containerd/nerdctl/commit/8bf043a52b5d64a9a7a38103427c884e4a542271

PR #1470, opened 2022-10-30
Logger: Support write to json file as the same time in fluent logger

Zheaoli:


cmd/nerdctl/run_linux_test.go
```diff
@@ -247,7 +247,7 @@ func TestRunWithFluentdLogDriverWithLogOpt(t *testing.T) {
 	time.Sleep(3 * time.Second)
 
 	testContainerName := containerName + "test"
-	base.Cmd("run", "-d", "--log-driver", "fluentd", "--log-opt", "fluentd-address=127.0.0.1:24225",
+	base.Cmd("run", "-d", "--log-driver", "fluentd", "--log-opt", "fluentd-address=127.0.0.1:24225", "--log-opt", "fluentd-log-to-json=true",
```
manugupt1, line 250: Thanks for the change. I am not sure why this feature is required:
1. It is already supported by fluentd as form of multiple outputs (https://docs.fluentd.org/output/secondary_file, https://docs.fluentd.org/output/file). Is there a use-case?
2. It is not supported in docker.
3. Even stdout is supported by fluentd (https://docs.fluentd.org/output/stdout) whichis what this test is doing.

if you want to still merge the change and maintainers are fine, can you update the readme.

pkg/logging/fluentd_logger.go
```diff
@@ -110,21 +117,52 @@ func (f *FluentdLogger) Process(_ string, config *logging.Config) error {
 	if err != nil {
 		return fmt.Errorf("failed to create fluent client: %w", err)
 	}
+	var jsonLogger *logrotate.Logger
```
manugupt1, line 120: Out of curiosity: Is it not possible to exec the nerdctl log driver or call the go routine for json driver run directl.

cmd/nerdctl/run_linux_test.go
Zheaoli: Sorry for replying late

1. Help people to use `nerdctl logs`  to get the container log(quickly debug, not need to go to the fluent backend server to get the log) 
2. Yep, this is a nerdctl-only feature
3. Yep, but in common use case, the people cannot decide whether the fluentd use stdout or not

manugupt1: Could this PR be more generic then (i.e for all loggers except default and json)? There would be other remote log forwarders that can make use of this functionality.

Zheaoli: emmmm, I want to split it into different PR,https://github.com/containerd/nerdctl/issues/1374

manugupt1: Hmm! Would you copy the same code in every logger then? Is it possible to make the code generic for any logger and have it in a single place so that when new loggers are added; they automatically inherit this logic rather than have more changes in the code base.

apostasie:
@yankay are you still interested in carrying this through?

yankay:
> @yankay are you still interested in carrying this through?

Let me invesgate it :-)

yankay:
The feature is not supported by docker https://docs.docker.com/engine/logging/drivers/fluentd/. 
I think the nerdctl does not must to need it. 
So, keep it open.

AkihiroSuda:
> The feature is not supported by docker https://docs.docker.com/engine/logging/drivers/fluentd/. I think the nerdctl does not must to need it. So, keep it open.

nerdctl may have an optional feature that is not implemented by Docker yet

Issue #1257, opened 2022-07-22
Flaky: TestCopyToContainer


junnplus:
https://github.com/containerd/nerdctl/runs/7464047904?check_suite_focus=true#step:7:894

```
=== CONT  TestCopyToContainer
    cp_linux_test.go:49: assertion failed: error is not nil: expected "1000\n", got ""

--- FAIL: TestCopyToContainer (2.16s)
    --- FAIL: TestCopyToContainer/SRC_PATH_specifies_a_file (1.25s)
        --- PASS: TestCopyToContainer/SRC_PATH_specifies_a_file/DEST_PATH_does_not_exist (0.46s)
        --- PASS: TestCopyToContainer/SRC_PATH_specifies_a_file/DEST_PATH_does_not_exist_and_ends_with_/ (0.15s)
        --- FAIL: TestCopyToContainer/SRC_PATH_specifies_a_file/DEST_PATH_exists_and_is_a_file (0.64s)
```

PR #1479, opened 2022-11-05
Enable service args for `compose build`

### LINKED ISSUE #1450: [compose] Add argument to build/push specific service(s)

Some docker-compose files may have multiple services, but you don't want to build/push all of them

The command should look something like

```bash
nedrctl compose build [SERVICE...]
nerdctl compose push [SERVICE...]
```

(like in original [docker-compose](<https://docs.docker.com/engine/reference/commandline/compose_build/>))

### Additional context

_No response_

djdongjin:
Fix #1450 

The `[Service]...` args are passed down to `compose` pkg and all validation are done via docker's own compose lib. The output (succeed or fail message) is identical to `docker compose build`.

Signed-off-by: Jin Dong [jindon@amazon.com](mailto:jindon@amazon.com)

### DIFF
--- a/cmd/nerdctl/compose_build_linux_test.go
+++ b/cmd/nerdctl/compose_build_linux_test.go
@@ -1,0 +1,69 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestComposeBuild(t *testing.T) {
+	const imageSvc0 = "composebuild_svc0"
+	const imageSvc1 = "composebuild_svc1"
+
+	dockerComposeYAML := fmt.Sprintf(`
+services:
+  svc0:
+    build: .
+    image: %s
+    ports:
+    - 8080:80
+  svc1:
+    build: .
+    image: %s
+    ports:
+    - 8081:80
+`, imageSvc0, imageSvc1)
+
+	dockerfile := fmt.Sprintf(`FROM %s`, testutil.AlpineImage)
+
+	testutil.RequiresBuild(t)
+	base := testutil.NewBase(t)
+	defer base.Cmd("builder", "prune").Run()
+
+	comp := testutil.NewComposeDir(t, dockerComposeYAML)
+	defer comp.CleanUp()
+	comp.WriteFile("Dockerfile", dockerfile)
+
+	defer base.Cmd("rmi", imageSvc0).Run()
+	defer base.Cmd("rmi", imageSvc1).Run()
+
+	// 1. build only 1 service
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "build", "svc0").AssertOK()
+	base.Cmd("images").AssertOutContains(imageSvc0)
+	base.Cmd("images").AssertOutNotContains(imageSvc1)
+	// 2. build multiple services
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "build", "svc0", "svc1").AssertOK()
+	base.Cmd("images").AssertOutContains(imageSvc0)
+	base.Cmd("images").AssertOutContains(imageSvc1)
+	// 3. build all if no args are given
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "build").AssertOK()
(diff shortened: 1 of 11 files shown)


austinvazquez:
@djdongjin for project checks you will need to add a DCO to `332962b`

djdongjin:
> 

fixed

djdongjin:
@austinvazquez fixed

yuchanns:
Please squash commits.

djdongjin:
Thanks for the reminder. Squashed and combined #1480 into this one.

djdongjin:
seems a flaky test which is tracked in #1222

cmd/nerdctl/compose_build_linux_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestComposeBuild(t *testing.T) {
+	testutil.RequiresBuild(t)
+	base := testutil.NewBase(t)
+	defer base.Cmd("builder", "prune").Run()
+
+	const dockerComposeYAML = `
```
manugupt1, line 31: Does this work on macOs as well; curious. Docker does not work for me for different archs for reasons unknown.

cmd/nerdctl/compose_build_linux_test.go
```diff
@@ -0,0 +1,65 @@
...
+	dockerfile := fmt.Sprintf(`FROM %s
+RUN uname -m > /usr/share/nginx/html/index.html
+`, testutil.NginxAlpineImage)
+
+	comp := testutil.NewComposeDir(t, dockerComposeYAML)
+	defer comp.CleanUp()
+	comp.WriteFile("Dockerfile", dockerfile)
+
+	// 1. no service args (all services are built)
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "build").AssertOK()
```
manugupt1, line 58: Can you assert and remove the that the images that are built present by doing nerdctl images; and nerdctl rmi

cmd/nerdctl/compose_build_linux_test.go
djdongjin: Seems it's because docker doesn't support multi-arch build. I'm following the compose file [here](https://github.com/containerd/nerdctl/blob/2f70e700d5352c1d326c653e7659c017e9aaaca4/cmd/nerdctl/multi_platform_linux_test.go#L84).

Added a `DockerIncompatible` to the test as well.

cmd/nerdctl/compose_build_linux_test.go
djdongjin: Updated the test.

cmd/nerdctl/compose_build_linux_test.go
```diff
@@ -0,0 +1,74 @@
...
+    build: .
+    image: %s
+    platform: arm64
+    ports:
+    - 8081:80
+`, imageSvc0, imageSvc1)
+
+	dockerfile := fmt.Sprintf(`FROM %s
+RUN uname -m > /usr/share/nginx/html/index.html
+`, testutil.NginxAlpineImage)
```
manugupt1, line 48: nit: can you use AlpineImage as that is smaller. Does it matter to have uname -m for these tests. Dockerfile with `FROM alpine` should be more than enough imo.

djdongjin: sounds good. fixed, thanks.

cmd/nerdctl/compose_build_linux_test.go
manugupt1: There is no need to add DockerIncompatible imo. You can build for same arch; same image (or perhaps different) as long as tags are differently generated and have the test be docker compatible.

yuchanns: This is a Docker-compatible feature so I don't think a `DockerIncompatible` should be here. You can split two arches' build into two unit tests.

cmd/nerdctl/compose_build_linux_test.go
```diff
@@ -0,0 +1,72 @@
...
+    build: .
+    image: %s
+    platform: arm64
+    ports:
+    - 8081:80
+`, imageSvc0, imageSvc1)
+
+	dockerfile := fmt.Sprintf(`FROM %s`, testutil.AlpineImage)
+
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 48: Why incompatible?

cmd/nerdctl/compose_build_linux_test.go
djdongjin: I see. removed it, thanks

cmd/nerdctl/compose_build_linux_test.go
djdongjin: I saw a similar test [here](https://github.com/containerd/nerdctl/blob/2f70e700d5352c1d326c653e7659c017e9aaaca4/cmd/nerdctl/multi_platform_linux_test.go#L84). Seems not the case, I just removed it. Thanks.

AkihiroSuda (approved):
Thanks

manugupt1:
Lgtm

djdongjin:
@AkihiroSuda @manugupt1 thanks both for review! can you help rerun the failed CI? seems like a flaky test (#1257).

Thanks.

PR #1480, opened 2022-11-05
Make compose help more consistent with docker and README

djdongjin:
This PR updates `nerdctl compose` usage in README and in `nerdctl compose -h`, making the output more consistent.

Some inconsistencies include:

1. `[Service...]` args are supported but not shown:

```bash
$ nerdctl compose ps -h
List containers of services

Usage: nerdctl compose ps [flags]
```

2. Sometimes `[flags]` shows before `args` and sometimes after `args`:

```bash
$ nerdctl compose run -h
Run a one-off command on a service

Usage: nerdctl compose run [flags] SERVICE [COMMAND] [ARGS...]

$ nerdctl compose logs -h
Show logs of a running container

Usage: nerdctl compose logs [SERVICE...] [flags]
```

Together with #1479 , #1450 should be fixed, since the support for `[Service...]` args should be consistent with `docker compose`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1401,7 +1401,7 @@ Unimplemented `docker compose up` (V2) flags: `--environment`
 ### :whale: nerdctl compose logs
 Create and start containers
 
-Usage: `nerdctl compose logs [OPTIONS]`
+Usage: `nerdctl compose logs [OPTIONS] [SERVICE...]`
 
 Flags:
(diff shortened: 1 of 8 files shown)


yuchanns:
Can we merge this PR into #1479? I think there is meaningless to divide it into two PRs

djdongjin:
Updated another PR.

PR #1488, opened 2022-11-08
support cpu-rt-runtime and cpu-rt-period options

merlinran:
also recognize them in compose file.

Manually tested on Ubuntu 20.

### DIFF
--- a/cmd/nerdctl/container_run.go
+++ b/cmd/nerdctl/container_run.go
@@ -183,6 +183,8 @@ func setCreateFlags(cmd *cobra.Command) {
 	cmd.Flags().Uint64("cpu-shares", 0, "CPU shares (relative weight)")
 	cmd.Flags().Int64("cpu-quota", -1, "Limit CPU CFS (Completely Fair Scheduler) quota")
 	cmd.Flags().Uint64("cpu-period", 0, "Limit CPU CFS (Completely Fair Scheduler) period")
+	cmd.Flags().Int64("cpu-rt-runtime", -1, "Limit CPU RT (Realtime scheduling) runtime")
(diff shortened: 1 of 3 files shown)


cmd/nerdctl/run_cgroup_linux.go
```diff
@@ -148,6 +148,23 @@ func generateCgroupOpts(cmd *cobra.Command, id string) ([]oci.SpecOpts, error) {
...
+	cpuRTPeriod, err := cmd.Flags().GetUint64("cpu-rt-period")
+	if err != nil {
+		return nil, err
+	}
+	if cpuRTRuntime != -1 || cpuRTPeriod != 0 {
+		if cpus > 0.0 {
+			return nil, errors.New("cpus and RT runtime/period should be used separately")
+		}
+		if cpuQuota != -1 || cpuPeriod != 0 {
+			return nil, errors.New("quota/period and RT runtime/period should be used separately")
```
merlinran, line 164: I'm not 100% sure if this is true. Also not sure if need to add a check of `cpuRTRuntime < cpuRTRuntime`

merlinran:
Well it's not that simple, because `containerd` creates the `/sys/fs/cgroup/cpu/<namespace>` folder with default parameters (`cpu.rt_runtime_us` being 0), so it can't write container specific value later on. https://github.com/containerd/containerd/blob/main/services/server/server_linux.go#L55

Moby handles this by [recursively creating cgroups with the same `cpu.rt_runtime_us` and `cpu.rt_period_us` value](https://github.com/moby/moby/blob/e7904c5faabfc96638aa26bcf725e90cd5aef768/daemon/daemon_unix.go#L1715), but not sure if it's what containerd wants to do, or do we want to have daemon or namespace specific configuration.

Issue #1502, opened 2022-11-12
Refactor `stop containers` and `remove containers` in compose


djdongjin:
### Description

The usage of `stop/remove containers` in `compose` seems duplicated and have some bugs.

Possible bugs:

1. This `WaitGroup` should be outside of `for`, otherwise it's equivalent to sequential exec without `go func`:
https://github.com/containerd/nerdctl/blob/f596926d413b9d0b193270a7a519a25cd1667513/pkg/composer/run.go#L285

2. Here the log says `stopping` but it's `removing`. (docker compose does `stop` instead of `rm`)
https://github.com/containerd/nerdctl/blob/f596926d413b9d0b193270a7a519a25cd1667513/pkg/composer/up_service.go#L88

Duplications:

`stop containers`: [`compose run`](https://github.com/containerd/nerdctl/blob/main/pkg/composer/run.go#L284-L297), [`compose stop`](https://github.com/containerd/nerdctl/blob/main/pkg/composer/stop.go#L62-L73)

`remove containers`: [`compose run`](https://github.com/containerd/nerdctl/blob/main/pkg/composer/run.go#L300-L314), [`compose up` (should be `stop`)](https://github.com/containerd/nerdctl/blob/main/pkg/composer/up_service.go#L89-L104), [`compose down`](https://github.com/containerd/nerdctl/blob/main/pkg/composer/down.go#L114-L126)

I'd like to see if a refactor is useful and made an initial PR. Still have some questions:

1. When `stop` or `remove` containers, when do we want to parallelize them (use `WaitGroup`) and when not (plain `for` loop)?
2. Some code uses `map[string]serviceparser.Container` and some uses `[]containerd.Container` to represent a set of containers. Do we want to unify them or keep them as separate cases.
3. What does `gracefully stopping`(v.s. `forcibly`) means? (e.g. [here](https://github.com/containerd/nerdctl/blob/main/pkg/composer/run.go#L283)

### Steps to reproduce the issue

Attached code link above.

### Describe the results you received and expected

Reduce code duplication and keep log and behavior consistency.

### What version of nerdctl are you using?

main

djdongjin:
After #1534 is merged, I'll combine/dedup `remove containers` functions into [`composer/rm.go`](https://github.com/djdongjin/nerdctl/blob/compose-rm/pkg/composer/rm.go). Then we should be good to close this :)

[...]

PR #1503, opened 2022-11-12
Refactor stop/rm containers in compose

djdongjin:
See #1502 for more details.

This PR made the following changes:

1. Parallelize `stop containers` in `compose stop`.
2. Refactor `stop containers` in `compose run` and `compose up`.
    1. Fix a bug in `compose run` when stopping containers.
    2. Change `rm containers` in `compose up` to `stop containers`, as its log said.


AkihiroSuda (approved):
Thanks

Issue #1514, opened 2022-11-17
`run --blkio-weight` results in a panic


davidhsingyuchen:
### Description

If `--blkio-weight` is given to `nerdctl run`, `nerdctl` panics.

We may need to change this line:

https://github.com/containerd/nerdctl/blob/c00780a1f5b905b09812722459c54936c9e070e6/pkg/infoutil/infoutil.go#L249

to:

```go
info := &dockercompat.Info{}
```

so that `info` won't be a `nil` pointer on the next line:

https://github.com/containerd/nerdctl/blob/c00780a1f5b905b09812722459c54936c9e070e6/pkg/infoutil/infoutil.go#L250

I'm happy to PR it, and I just want to make sure I'm not missing anything.

Re. why it's not caught by existing tests, my guess is that `BlockIOWeight` does not have unit tests because it's platform-dependent, and re. integration tests, `TestRunBlkioWeightCgroupV2` seems to be skipped because `bfq` module is not loaded. I only checked [a random recent workflow run](https://github.com/containerd/nerdctl/actions/runs/3478937000/jobs/5827626977), so I could be missing something.

### Steps to reproduce the issue

```sh
lima nerdctl run --name ctr1 --blkio-weight 300 -w /sys/fs/cgroup alpine sleep infinity
```

#### Received

```sh
➜  ~ lima nerdctl run --name ctr1 --blkio-weight 300 -w /sys/fs/cgroup alpine sleep infinity
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa69f08]

goroutine 1 [running]:
github.com/containerd/nerdctl/pkg/infoutil.BlockIOWeight({0xdd21ec, 0x7})
	/go/src/github.com/containerd/nerdctl/pkg/infoutil/infoutil.go:250 +0x28
main.generateCgroupOpts(0xc5fd00?, {0x40002f81c0, 0x40})
	/go/src/github.com/containerd/nerdctl/cmd/nerdctl/run_cgroup_linux.go:244 +0x1584
main.setPlatformOptions({0x1100978, 0x40004c6000}, {0x400000e040?, 0x1?, 0x40?}, 0x0?, 0x0?, {0x40002f81c0, 0x40})
	/go/src/github.com/containerd/nerdctl/cmd/nerdctl/run_linux.go:72 +0x294
[...]

PR #1516, opened 2022-11-17
Fix: nerdctl run --blkio-weight results in a panic

yankay:
Fix https://github.com/containerd/nerdctl/issues/1514

Signed-off-by: Kay Yan <kay.yan@daocloud.io>


AkihiroSuda (approved):
Thanks

PR #1519, opened 2022-11-18
Refactor the logging driver process function

Zheaoli:
Split the stdout/stderr `io.Reader` in logging driver process function by using the channel and an adapter function

So, we can add process hooks more easy such as write the local file and remote logging driver at the same time

### DIFF
--- a/pkg/logging/fluentd_logger.go
+++ b/pkg/logging/fluentd_logger.go
@@ -17,9 +17,7 @@
 package logging
 
 import (
-	"bufio"
 	"fmt"
-	"io"
 	"math"
 	"net/url"
 	"runtime"
@@ -35,7 +33,9 @@ import (
 )
 
 type FluentdLogger struct {
-	Opts map[string]string
+	Opts         map[string]string
+	fluentClient *fluent.Fluent
+	config       *logging.Config
 }
 
 const (
@@ -97,7 +97,7 @@ func (f *FluentdLogger) Init(dataStore, ns, id string) error {
 	return nil
 }
 
-func (f *FluentdLogger) Process(_ string, config *logging.Config) error {
+func (f *FluentdLogger) PreProcess(_ string, config *logging.Config) error {
 	if runtime.GOOS == "windows" {
 		// TODO: support fluentd on windows
 		return fmt.Errorf("logging to fluentd is not supported on windows")
@@ -110,28 +110,34 @@ func (f *FluentdLogger) Process(_ string, config *logging.Config) error {
 	if err != nil {
 		return fmt.Errorf("failed to create fluent client: %w", err)
 	}
+	f.fluentClient = fluentClient
+	f.config = config
+	return nil
+}
+func (f *FluentdLogger) Process(stdout <-chan string, stderr <-chan string) error {
 	var wg sync.WaitGroup
 	wg.Add(2)
-	fun := func(wg *sync.WaitGroup, reader io.Reader, id, namespace, source string) {
+	fun := func(wg *sync.WaitGroup, dataChan <-chan string, id, namespace, source string) {
 		defer wg.Done()
 		metaData := map[string]string{
 			"container_id": id,
 			"namespace":    namespace,
 			"source":       source,
 		}
-		scanner := bufio.NewScanner(reader)
-		for scanner.Scan() {
-			if scanner.Err() != nil {
-				return
-			}
-			metaData["log"] = scanner.Text()
-			fluentClient.PostWithTime(f.Opts[Tag], time.Now(), metaData)
+		for log := range dataChan {
+			metaData["log"] = log
+			f.fluentClient.PostWithTime(f.Opts[Tag], time.Now(), metaData)
 		}
 	}
-	go fun(&wg, config.Stdout, config.ID, config.Namespace, "stdout")
-	go fun(&wg, config.Stderr, config.ID, config.Namespace, "stderr")
+	go fun(&wg, stdout, f.config.ID, f.config.Namespace, "stdout")
+	go fun(&wg, stderr, f.config.ID, f.config.Namespace, "stderr")
 
 	wg.Wait()
+	return nil
+}
+
+func (f *FluentdLogger) PostProcess() error {
+	defer f.fluentClient.Close()
 	return nil
 }
 
--- a/pkg/logging/journald_logger.go
+++ b/pkg/logging/journald_logger.go
@@ -17,7 +17,6 @@
 package logging
 
 import (
-	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
@@ -52,6 +51,7 @@ func JournalLogOptsValidate(logOptMap map[string]string) error {
 
 type JournaldLogger struct {
 	Opts map[string]string
+	vars map[string]string
 }
 
 type identifier struct {
@@ -64,7 +64,7 @@ func (journaldLogger *JournaldLogger) Init(dataStore, ns, id string) error {
 	return nil
 }
 
-func (journaldLogger *JournaldLogger) Process(dataStore string, config *logging.Config) error {
+func (journaldLogger *JournaldLogger) PreProcess(dataStore string, config *logging.Config) error {
 	if !journal.Enabled() {
 		return errors.New("the local systemd journal is not available for logging")
 	}
@@ -97,23 +97,28 @@ func (journaldLogger *JournaldLogger) Process(dataStore string, config *logging.
 	vars := map[string]string{
 		"SYSLOG_IDENTIFIER": syslogIdentifier,
 	}
+	journaldLogger.vars = vars
+	return nil
+}
+
+func (journaldLogger *JournaldLogger) Process(stdout <-chan string, stderr <-chan string) error {
 	var wg sync.WaitGroup
 	wg.Add(2)
-	f := func(wg *sync.WaitGroup, r io.Reader, pri journal.Priority, vars map[string]string) {
+	f := func(wg *sync.WaitGroup, dataChan <-chan string, pri journal.Priority, vars map[string]string) {
 		defer wg.Done()
-		s := bufio.NewScanner(r)
-		for s.Scan() {
-			if s.Err() != nil {
-				return
-			}
-			journal.Send(s.Text(), pri, vars)
+		for log := range dataChan {
+			journal.Send(log, pri, vars)
 		}
 	}
 	// forward both stdout and stderr to the journal
-	go f(&wg, config.Stdout, journal.PriInfo, vars)
-	go f(&wg, config.Stderr, journal.PriErr, vars)
+	go f(&wg, stdout, journal.PriInfo, journaldLogger.vars)
+	go f(&wg, stderr, journal.PriErr, journaldLogger.vars)
 
 	wg.Wait()
+	return nil
+}
+
+func (journaldLogger *JournaldLogger) PostProcess() error {
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


pkg/logging/journald_logger.go
```diff
@@ -97,23 +98,32 @@ func (journaldLogger *JournaldLogger) Process(dataStore string, config *logging.
...
+					return
+				}
 			}
-			journal.Send(s.Text(), pri, vars)
 		}
 	}
 	// forward both stdout and stderr to the journal
-	go f(&wg, config.Stdout, journal.PriInfo, vars)
-	go f(&wg, config.Stderr, journal.PriErr, vars)
+	go f(&internalWg, stdout, journal.PriInfo, journaldLogger.vars)
```
yuchanns, line 115: Passing WaitGroup into functions is not necessary IMO, as it can be used directly.

pkg/logging/syslog_logger.go
```diff
@@ -115,34 +113,36 @@ func SyslogOptsValidate(logOptMap map[string]string) error {
...
 	return nil
 }
 
-func (sy *SyslogLogger) Process(dataStore string, config *logging.Config) error {
+func (sy *SyslogLogger) PreProcess(dataStore string, config *logging.Config) error {
 	logger, err := parseSyslog(config.ID, sy.Opts)
 	if err != nil {
 		return err
 	}
 	defer logger.Close()
```
yuchanns, line 129: `logger.Close` should be called after the Process function finished.

pkg/logging/logging.go
```diff
@@ -133,6 +137,28 @@ func LoadLogConfig(dataStore, ns, id string) (LogConfig, error) {
...
+		for scanner.Scan() {
+			if scanner.Err() != nil {
+				return
+			}
+			dataChan <- scanner.Text()
+		}
+	}
+	go processLogFunc(config.Stdout, stdout)
+	go processLogFunc(config.Stderr, stderr)
+	go driver.Process(&wg, stdout, stderr)
```
yuchanns, line 157: It's not a good idea to pass the outer WaitGroup instance into the inner function as delegating the control to Driver implementation.
Suggest:
```go
go func() {
	defer wg.Done()
	driver.Process(stdout, stderr)
} ()
```

pkg/logging/logging.go
```diff
@@ -133,6 +137,33 @@ func LoadLogConfig(dataStore, ns, id string) (LogConfig, error) {
...
 }
 
+func loggingProcessAdapter(driver Driver, config *logging.Config) error {
+	var wg sync.WaitGroup
+	wg.Add(3)
+	stdout := make(chan string, 10000)
+	stderr := make(chan string, 10000)
+	processLogFunc := func(reader io.Reader, dataChan chan string) {
+		defer wg.Done()
+		defer close(dataChan)
```
yuchanns, line 147: `processLogFunc` is called twice downside, should make sure to close the channel on the last call.

pkg/logging/logging.go
```diff
@@ -40,7 +43,8 @@ const (
 
 type Driver interface {
 	Init(dataStore, ns, id string) error
-	Process(dataStore string, config *logging.Config) error
+	PreProcess(dataStore string, config *logging.Config) error
+	Process(stdout chan string, stderr chan string) error
```
yuchanns, line 47: Casting params to read-only directional channels may be better IMO.
```go
Process(stdout <-chan string, stderr <-chan string) error
```

Zheaoli: nice catch, done

pkg/logging/logging.go
```diff
@@ -133,6 +137,33 @@ func LoadLogConfig(dataStore, ns, id string) (LogConfig, error) {
...
+			}
+			dataChan <- scanner.Text()
+		}
+	}
+
+	go processLogFunc(config.Stdout, stdout)
+	go processLogFunc(config.Stderr, stderr)
+	go func() {
+		defer wg.Done()
+		driver.Process(stdout, stderr)
```
yuchanns, line 161: should catch the error returns.

yuchanns: And maybe we don't need  WaitGroup here as `driver.Process` is a syncable function that will process all data reader from channels and block until channels are closed.

Zheaoli: No, we need WaitGroup here when we have multiple write target at the same time

pkg/logging/logging.go
```diff
@@ -133,6 +138,36 @@ func LoadLogConfig(dataStore, ns, id string) (LogConfig, error) {
...
+	wg.Add(3)
+	stdout := make(chan string, 10000)
+	stderr := make(chan string, 10000)
+	processLogFunc := func(reader io.Reader, dataChan chan string) {
+		defer wg.Done()
+		defer close(dataChan)
+		scanner := bufio.NewScanner(reader)
+		for scanner.Scan() {
+			if scanner.Err() != nil {
+				return
```
fahedouch, line 155: we have to log an error here

pkg/logging/logging.go
```diff
@@ -133,6 +138,36 @@ func LoadLogConfig(dataStore, ns, id string) (LogConfig, error) {
...
+		}
+	}
+
+	go processLogFunc(config.Stdout, stdout)
+	go processLogFunc(config.Stderr, stderr)
+	go func() {
+		defer wg.Done()
+		driver.Process(stdout, stderr)
+	}()
+	wg.Wait()
```
fahedouch, line 167: `PreProcess` , `Process` and `PostProcess` should be executed sequentially

pkg/logging/fluentd_logger.go
```diff
@@ -110,31 +110,37 @@ func (f *FluentdLogger) Process(_ string, config *logging.Config) error {
 	if err != nil {
 		return fmt.Errorf("failed to create fluent client: %w", err)
 	}
+	f.fluentClient = fluentClient
+	f.config = config
+	return nil
+}
+func (f *FluentdLogger) Process(stdout <-chan string, stderr <-chan string) error {
```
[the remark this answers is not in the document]
Zheaoli: I think it's better here to use chan instead of io.Reader here, it's more easy to translate the data between the goroutine

pkg/logging/jsonfile/jsonfile.go
```diff
@@ -45,40 +43,6 @@ func Path(dataStore, ns, id string) string {
...
-				return
-			}
-		}
-	}
-	go f(stdout, "stdout")
-	go f(stderr, "stderr")
-	wg.Wait()
-	return nil
-}
-
```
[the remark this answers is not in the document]
Zheaoli: Make sense

pkg/logging/logging.go
Zheaoli: Make sense

pkg/logging/logging.go
Zheaoli: I think it's already executed sequentially here

pkg/logging/fluentd_logger.go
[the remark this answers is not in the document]
fahedouch: @Zheaoli please check this answer from [stackoverflow](https://stackoverflow.com/a/13668275)  . seems like no good reason to use channel here

fahedouch, line 117: @Zheaoli please check this answer from [stackoverflow](https://stackoverflow.com/a/13668275)  . seems like no good reason to use channel here

Zheaoli: Thanks for the tips, I have already read the link you attach to the comment.

The answer from StackOverflow seems to have the same idea as mine.

> Channels are a good fit for communicating between goroutines.

Go back to this PR, please take a look at `loggingProcessAdapter` function here https://github.com/containerd/nerdctl/pull/1519/files#diff-eb19031b20f577302a6bfbcbb42ee60f37256f9ed2044d905a7aa681309bca69R142, you will find that I split scan log message from container's stdout/stderr and process the log message into different goroutine. https://github.com/containerd/nerdctl/pull/1519/files#diff-eb19031b20f577302a6bfbcbb42ee60f37256f9ed2044d905a7aa681309bca69R150-R164

You may wonder why I chose to split this code into different goroutines. Please look at #1470 (still WIP, will update after this PR has been merged); in #1470, I want to allow people to save logs to a local JSON file when they choose fluentd/syslog driver to make debugging easier. If I want to implement this feature, I think it's an excellent choice to use different goroutines to write to the fluentd/syslog and a local JSON file. In this circumstance, I think using a channel is a better way to fit the best practice in Golang.

Zheaoli: @fahedouch ping~

fahedouch: @Zheaoli 
Ok I got the intentation of this refacto!  but  `io` is a good fit too.

Now coming back to  `io`  **vs**  `channel` :

> Channels are a good fit for communicating between goroutines.  

this above sentence is very generic, what is important is the next of the comment which explain what the guy means by `good fit`:  

> When a program does something simple, like for example reading stdin, doing something with the stream and outputting the result to stdout - then using a channel is an overkill, unnecessarily hurting performance.
there's no good reason to model simple operations, like those of io.Reader or io.Writer using channels, respective to have a channel based method set (API).

Looking on your code we are not doing more than playing with stream and outputting it somewhere. So the question here, do we really need to hurt performance by using channel ? 

Other thing worth noting is that by design containerd[ provider io interface for stdout and stdin](https://github.com/containerd/containerd/blob/v1.6.10/runtime/v2/logging/logging.go#L25) So why moving to channels if `io` cloud fit the need

Zheaoli: > Other thing worth noting is that by design containerd[ provider io interface for stdout and stdin](https://github.com/containerd/containerd/blob/v1.6.10/runtime/v2/logging/logging.go#L25) So why moving to channels if io cloud fit the need

The problem is the containerd's stdou/stderr `io.Reader` is seekable, which means the same position in the IO stream can be read just once.

> You may wonder why I chose to split this code into different goroutines. Please look at https://github.com/containerd/nerdctl/pull/1470 (still WIP, will update after this PR has been merged); in https://github.com/containerd/nerdctl/pull/1470, I want to allow people to save logs to a local JSON file when they choose fluentd/syslog driver to make debugging easier. If I want to implement this feature, I think it's an excellent choice to use different goroutines to write to the fluentd/syslog and a local JSON file. In this circumstance, I think using a channel is a better way to fit the best practice in Golang.

In the up circumstance, two loggers need to read the same message twice. So we need different buffers for logger instances and put the message we read from stdout/stderr in the buffer. For the buffer, I think the channel would be a good choice.

pkg/logging/json_logger.go
```diff
@@ -109,7 +110,17 @@ func (jsonLogger *JSONLogger) Process(dataStore string, config *logging.Config)
...
+	return nil
+}
+
+func (jsonLogger *JSONLogger) Process(stdout <-chan string, stderr <-chan string) error {
+	return jsonfile.Encoode(stdout, stderr, jsonLogger.logger)
+}
+
+func (jsonLogger *JSONLogger) PostProcess() error {
+	//TODO implement me
+	panic("implement me")
```
AkihiroSuda, line 123:
```suggestion
	return errors.Errorf("function (*JSONLogger).PostProcess() is not implemented: %w", errdefs.ErrNotImplemented
```

AkihiroSuda (approved):
Thanks

pkg/logging/fluentd_logger.go
fahedouch: makes sense

Issue #1473, opened 2022-11-01
Can not save an image with local registry running


azhouwd:
### Description

Using nerdctl v1.0.0, if running a local registry and remove the registry image, then pull any image such as alpine. Will have error if trying to save the alpine image in this case.

### Steps to reproduce the issue

1.`nerdctl run -d -p 5000:5000 registry`
2.`nerdctl rmi -f registry:latest`
3.`nerdctl pull alpine`
4.`nerdctl save alpine -o alpine.tar` 

### Describe the results you received and expected

Received:
`FATA[0000] failed to get reader: content digest sha256:9b18e9b68314027565b90ff6189d65942c0f7986da80df008b8431276885218e: not found`

Expected:
Could save image successfully

### What version of nerdctl are you using?

v1.0.0

### Host information

Linux lima-default aarch64 Thu Apr 28 15:22:08 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux

fahedouch:
fixed by https://github.com/containerd/nerdctl/pull/1526

PR #1526, opened 2022-11-19
Deny rm image action when this image is used by running container

Zheaoli:
#1473 


AkihiroSuda (approved):
Thanks

PR #1531, opened 2022-11-21
Use service label to check orphan container

djdongjin:
The current implementation of `getOrphanContainers` (return orphan containers based on given services) is a little complicated which has a `goto` and 3 nested `for` loops.

Current impl:

for each container, go through every container of every given service and see if it has the same name with one of them.

New impl: ([`compose spec` requires the service label exists in compose containers](https://docs.docker.com/compose/compose-file/#labels-2))

for each container, check if its `ComposeService` label equals to any name of given services.


AkihiroSuda (approved):
Thanks

PR #1494, opened 2022-11-10
Add `nerdctl compose stop` command

djdongjin:
Fix #666 

This implementation uses `nerdctl stop [flags] container` to stop all containers in given services, similar to how `compose down` works [link](https://github.com/containerd/nerdctl/blob/6bcd80643dc3acc3e5e2aaeccc4d0f7a4bbba799/pkg/composer/down.go#L113).

I didn't use a `WaitGroup` to parallelize the `nerdctl stop` calls, as in [here](https://github.com/containerd/nerdctl/blob/660680b7ddfde1d38a66ec1c7f08f8d89ab92c68/pkg/composer/run.go#L283-L297), since I'm not sure what `forcibly`/`gracefully` means. Thanks for any explanation and I can optimize (seems some refactor/dedup can be done here as well)

Also in the same place, it seems like a bug. This [line](https://github.com/containerd/nerdctl/blob/660680b7ddfde1d38a66ec1c7f08f8d89ab92c68/pkg/composer/run.go#L285) should be outside of `for` loop right?

I'll also try to add test as well.

```bash
$ cat full.yaml
version: '3.1'

services:
  service1:
    image: alpine:3.13
    command: "sleep infinity"
  service2:
    image: alpine:3.13
    command: "sleep infinity"
```

```bash
$ sudoe nerdctl compose -f full.yaml up -d
INFO[0000] Ensuring image alpine:3.13
...
$ sudoe nerdctl ps
CONTAINER ID    IMAGE                            COMMAND             CREATED           STATUS    PORTS    NAMES
6ab9f4de1c07    docker.io/library/alpine:3.13    "sleep infinity"    17 seconds ago    Up                 orphan_service2_1
e046dbf0bdf3    docker.io/library/alpine:3.13    "sleep infinity"    17 seconds ago    Up                 orphan_service1_1
$ sudoe nerdctl compose -f full.yaml stop service1
INFO[0000] Stopping container orphan_service1_1
$ sudoe nerdctl ps -a
CONTAINER ID    IMAGE                              COMMAND             CREATED               STATUS                         PORTS    NAMES
...
6ab9f4de1c07    docker.io/library/alpine:3.13      "sleep infinity"    About a minute ago    Up                                      orphan_service2_1
e046dbf0bdf3    docker.io/library/alpine:3.13      "sleep infinity"    About a minute ago    Exited (137) 13 seconds ago             orphan_service1_1
...
```

Signed-off-by: Jin Dong <jindon@amazon.com>

PR #1533, opened 2022-11-21
Add `compose restart` command

djdongjin:
Similar to #1494, this one adds `nerdctl compose restart` command. It's done by calling `nerdctl restart CONTAINER_ID` for each service containers.

### DIFF
--- a/pkg/composer/restart.go
+++ b/pkg/composer/restart.go
@@ -22,21 +22,20 @@ import (
 	"sync"
 
 	"github.com/containerd/containerd"
-	"github.com/containerd/nerdctl/pkg/composer/serviceparser"
 	"github.com/containerd/nerdctl/pkg/labels"
 	"github.com/containerd/nerdctl/pkg/strutil"
 
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/composer/restart.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/containerd/nerdctl/pkg/labels"
+	"github.com/containerd/nerdctl/pkg/strutil"
+
+	"github.com/sirupsen/logrus"
+)
+
+// RestartOptions stores all option input from `nerdctl compose restart`
+type RestartOptions struct {
+	TimeChanged bool
+	Timeout     uint
```
AkihiroSuda, line 34: Can we remove `TImeChanged` by checking `Timeout > 0`, or `Timeout != nil` ?

djdongjin: Thanks for the suggestion. Changed to use a `*uint` in both `compose restart` and `compose stop` so `nil` can be used to check if `timeout` is passed or not.

AkihiroSuda (approved):
Thanks

PR #1534, opened 2022-11-22
Add `compose rm` command

djdongjin:
Add `compose rm` command and related args ([docker doc](https://docs.docker.com/engine/reference/commandline/compose_rm/)).

High level logic:

```shell
for each given service:
    if `stop` is passed:
        `stop` all containers in the service (use `c.stopContainers(containers, ...)`)
    for each container in the service:
        if `status` == `Up`:
            continue
        `rm` the container using `nerdctl rm -f` (and `-v` if it's passed)
```

### DIFF
--- a/pkg/composer/rm.go
+++ b/pkg/composer/rm.go
@@ -1,0 +1,94 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package composer
+
+import (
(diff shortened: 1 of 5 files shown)


pkg/composer/rm.go
```diff
@@ -0,0 +1,92 @@
...
+		rmWG.Add(1)
+		go func() {
+			defer rmWG.Done()
+			info, _ := container.Info(ctx, containerd.WithoutRefreshedMetadata)
+			// skip running containers.
+			cStatus := formatter.ContainerStatus(ctx, container)
+			if strings.HasPrefix(cStatus, "Up") {
+				logrus.Warnf("Removing container %s failed: container still running.", info.Labels[labels.Name])
+				return
+			}
```
fahedouch, line 81:
```suggestion
              if !opt.Stop {
			info, _ := container.Info(ctx, containerd.WithoutRefreshedMetadata)
			// skip running containers.
			cStatus := formatter.ContainerStatus(ctx, container)
			if strings.HasPrefix(cStatus, "Up") {
				logrus.Warnf("Removing container %s failed: container still running.", info.Labels[labels.Name])
				return
			}
		 }
```

fahedouch (approved):
Thanks

djdongjin:
Rebase to resolve the conflict on the doc with the just merged `compose rm` PR.

AkihiroSuda (approved):
Thanks

PR #1538, opened 2022-11-22
Prevent double-creation of nerdctl default network.

aznashwan:
This patch adds logic to pkg/netutil to prevent nerdctl from accidentally creating a second default network if a non-nerdctl-created network with the same default name already happens to exist.

Additionally, a new `nerdctl/default-network` label definition has been added to indicate whether a network was created by nerdctl with the express purpose of being used as a default network.

Signed-off-by: Nashwan Azhari <nazhari@cloudbasesolutions.com>

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -89,6 +89,12 @@ const (
 
 	// PIDContainer is the `nerdctl run --pid` for restarting
 	PIDContainer = Prefix + "pid-container"
+
+	// NerdctlDefaultNetwork indicates whether a network is the default network
+	// created and owned by Nerdctl.
+	// Boolean value which can be parsed with strconv.ParseBool() is required.
+	// (like "nerdctl/default-network=true" or "nerdctl/default-network=false")
+	NerdctlDefaultNetwork = Prefix + "default-network"
 )
 
 var ShellCompletions = []string{
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -38,6 +38,7 @@ import (
 	subnetutil "github.com/containerd/nerdctl/pkg/netutil/subnet"
 	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/containernetworking/cni/libcni"
+	"github.com/sirupsen/logrus"
 )
 
 type CNIEnv struct {
@@ -121,6 +122,7 @@ func NewCNIEnv(cniPath, cniConfPath string, opts ...CNIEnvOpt) (*CNIEnv, error) 
 		return nil, err
 	}
 	e.Networks = networks
+
 	return &e, nil
 }
 
@@ -130,6 +132,20 @@ func (e *CNIEnv) NetworkMap() map[string]*NetworkConfig { //nolint:revive
 		m[n.Name] = n
 	}
 	return m
+}
+
+func (e *CNIEnv) filterNetworks(filterf func(*networkConfig) bool) []*networkConfig {
+	result := []*networkConfig{}
+	for _, networkConfig := range e.Networks {
+		if filterf(networkConfig) {
+			result = append(result, networkConfig)
+		}
+	}
+	return result
+}
+
+func (e *CNIEnv) getConfigPathForNetworkName(netName string) string {
+	return filepath.Join(e.NetconfPath, "nerdctl-"+netName+".conflist")
 }
 
 func (e *CNIEnv) usedSubnets() ([]*net.IPNet, error) {
@@ -214,16 +230,62 @@ func (e *CNIEnv) RemoveNetwork(net *NetworkConfig) error {
 	return lockutil.WithDirLock(e.NetconfPath, fn)
 }
 
-func (e *CNIEnv) ensureDefaultNetworkConfig() error {
-	filename := filepath.Join(e.NetconfPath, "nerdctl-"+DefaultNetworkName+".conflist")
+// Checks whether the default network exists by first checking if
+// any network bears the `labels.NerdctlDefaultNetwork` label,
+// or falls back to checking whether any network bears the
+// `DefaultNetworkName` name.
+func (e *CNIEnv) getDefaultNetworkConfig() (*networkConfig, error) {
+	// Search for networks bearing the `labels.NerdctlDefaultNetwork` label:
+	defaultLabelFilterF := func(nc *networkConfig) bool {
+		if nc.NerdctlLabels == nil {
+			return false
+		} else if _, ok := (*nc.NerdctlLabels)[labels.NerdctlDefaultNetwork]; ok {
+			return true
+		}
+		return false
+	}
+	labelMatches := e.filterNetworks(defaultLabelFilterF)
+	if len(labelMatches) >= 1 {
+		if len(labelMatches) > 1 {
+			logrus.Warnf("returning the first network bearing the %q label out of the multiple found: %#v", labels.NerdctlDefaultNetwork, labelMatches)
+		}
+		return labelMatches[0], nil
+	}
+
+	// Search for networks bearing the DefaultNetworkName:
+	defaultNameFilterF := func(nc *networkConfig) bool {
+		return nc.Name == DefaultNetworkName
+	}
+	nameMatches := e.filterNetworks(defaultNameFilterF)
+	if len(nameMatches) >= 1 {
+		if len(nameMatches) > 1 {
+			logrus.Warnf("returning the first network bearing the %q default network name out of the multiple found: %#v", DefaultNetworkName, nameMatches)
+		}
+
+		// Warn the user if the default network was not created by nerdctl:
+		match := nameMatches[0]
+		_, statErr := os.Stat(e.getConfigPathForNetworkName(DefaultNetworkName))
+		if match.NerdctlID == nil || statErr != nil {
+			logrus.Warnf("default network named %q does not have an internal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl", DefaultNetworkName)
+		}
+
+		return nameMatches[0], nil
+	}
+
+	return nil, nil
+}
+
+func (e *CNIEnv) createDefaultNetworkConfig() error {
+	filename := e.getConfigPathForNetworkName(DefaultNetworkName)
 	if _, err := os.Stat(filename); err == nil {
-		return nil
+		return fmt.Errorf("already found existing network config at %q, cannot create new network named %q", filename, DefaultNetworkName)
 	}
 	opts := CreateOptions{
 		Name:       DefaultNetworkName,
 		Driver:     DefaultNetworkName,
 		Subnet:     DefaultCIDR,
 		IPAMDriver: "default",
+		Labels:     []string{fmt.Sprintf("%s=true", labels.NerdctlDefaultNetwork)},
 	}
 	_, err := e.CreateNetwork(opts)
 	if err != nil && !errdefs.IsAlreadyExists(err) {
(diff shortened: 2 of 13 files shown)


aznashwan:
Note that this is basically just a workaround and the root cause of the issue is nerdctl using [network names as their identifiers instead of IDs](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L67) for most internal filtering logic.

IMHO it'd be worth creating an issue where we could discuss switching to network IDs further...

fahedouch:
> This patch adds logic to pkg/netutil to prevent nerdctl from accidentally creating a second default network if a non-nerdctl-created network with the same default name already happens to exist.

can you give an example plz?

pkg/netutil/netutil.go
```diff
@@ -50,14 +52,23 @@ func NewCNIEnv(cniPath, cniConfPath string) (*CNIEnv, error) {
...
+	// NOTE: must come after `e.networkConfigList()`.
+	defaultNet, err := e.getDefaultNetworkConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for default network: %s", err)
+	}
+	if defaultNet == nil {
+		if err := e.createDefaultNetworkConfig(); err != nil {
+			return nil, fmt.Errorf("failed to create default network: %s", err)
+		}
+	}
```
fahedouch, line 70: can we  move`ensureDefaultNetworkConfig` after  `e.networkConfigList()` and then put that in  ?

aznashwan:
@fahedouch it happened while I was running nerdctl on a containerd install on Windows 2019 which already had a CNI network defined bearing the name `nat` from [some common setup script in the containerd repo](https://github.com/containerd/containerd/blob/main/script/setup/install-cni-windows#L81), though under a different filename. (`0-containerd-nat.conf`)

`nat` is [the default network name in nerdctl on Windows](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil_windows.go#L28), and because [CNIEnv.ensureDefaultNetworkConfig() always gets called](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L53) and nerdctl it only checks for [the nerdctl-specific CNI network file definition is present before re-creating the network](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L155-L158), it meant that nerdctl created its own network named nat, and now I had two.

Note that the original nat thus became completely inaccessible, as nerdctl would always [sort the CNI config files alphabetically by filename](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L231) and [cram them first-come-from-served into a map](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L66-L67), in which case the original network definition in `0-containerd-nat.conf` would always get superseded by the one created by nerdctl in `nerdctl-nat.conflist`.

In the end this can only be really be solved by having nerdctl use network IDs instead of names for any nerdctl-managed networks, and only use names for networks created externally.

Either way, I felt there needed to be some way to differentiate the default network through a label on it. (am open to any better ideas though)

pkg/netutil/netutil.go
aznashwan: I'm not sure what you mean since `ensureDefaultNetworkConfig()` has now become `createDefaultNetworkConfig()`. (which is indeed now called after `e.networkConfigList()`)

It has however just hit me that creating the default network after the `networkConfigList()` will mean that `e.Networks` will not include the newly-created default network.

While I could just go `e.Networks = append(e.Networks, newDefaultNetwork)`, I still think this shows a deeper underlying problem in how we expose this field:
- [`e.Networks` is a public slice of a private type (`networkConfig`)](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L42)
- `e.Networks` is [only set once, and that is in a constructor function in this same module](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L60), so there's no need for it to be public
- it's used in `cmd/nerdctl/nerwork_*.go` but only to be iterated over, it's not actually set anywhere outside this module
- `e.NetworkMap()` is just a convenience method which has the potential to "miss" networks if they share names by [overriding any names already in the map](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L67) from whatever it finds in `e.Networks`

I'm guessing `e.Networks` might have been added for caching reasons or something but at this stage IMO it just makes the network listing logic less consistent and should be removed entirely, what do you think?

pkg/labels/labels.go
```diff
@@ -89,6 +89,12 @@ const (
 
 	// PIDContainer is the `nerdctl run --pid` for restarting
 	PIDContainer = Prefix + "pid-container"
+
+	// NerdctlDefaultNetwork indicates whether a network is the default network
+	// created and owned by Nerdctl.
+	// Boolean value which can be parsed with strconv.ParseBool() is required.
+	// (like "nerdctl/default-network=true" or "nerdctl/default-network=false")
+	NerdctlDefaultNetwork = Prefix + "default-network"
```
AkihiroSuda, line 97: Does this cover multi-network (`nerdctl run --net=foo --net=bar`, commonly invoke via compose) ?

AkihiroSuda: I'd also like to see an integration test

aznashwan: > Does this cover multi-network (nerdctl run --net=foo --net=bar, commonly invoke via compose)?

The label will only get set on the [default network nerdctl creates](https://github.com/containerd/nerdctl/pull/1538/files#diff-669913e187d4dc726dcb9e01a5666ae0562549545138d25d2c8fc2ceb3e23d30R234), and the label will only be [checked when trying to find the default network](https://github.com/containerd/nerdctl/pull/1538/files#diff-669913e187d4dc726dcb9e01a5666ae0562549545138d25d2c8fc2ceb3e23d30R188).

I'll be adding an integration test shortly.

pkg/netutil/netutil.go
fahedouch: - SGTM regarding introducing `NetworkList()` to expose networkConfigs
- What I propose is to wrap the `e.getDefaultNetworkConfig()` and `createDefaultNetworkConfig()` into `ensureDefaultNetworkConfig()`

stgh like this 
```
func (e *CNIEnv) ensureDefaultNetworkConfig() error{
	defaultNet, err := e.GetDefaultNetworkConfig()
	if err != nil {
		return  fmt.Errorf("failed to check for default network: %s", err)
	}
	if defaultNet == nil {
		if err := e.createDefaultNetworkConfig(); err != nil {
			return fmt.Errorf("failed to create default network: %s", err)
		}
	}
        return nil
}
```

fahedouch:
@aznashwan need some rebase ( because of this https://github.com/containerd/nerdctl/pull/1554)

aznashwan:
@fahedouch just updated the PR, thanks for the heads-up!

One thing worth noting is that due to #1554 making default network creation no longer happening on any `nerdctl network` subcommand, I [had to "downgrade" the integration tests into unit tests](https://github.com/containerd/nerdctl/pull/1538/files#diff-bbb2f1fb0e65a12b2beeb7eb450057a2158402906d9d62c6b997eadc689756a5R133).

pkg/netutil/netutil_test.go
```diff
@@ -97,3 +128,136 @@ func TestParseIPAMRange(t *testing.T) {
 		}
 	}
 }
+
+// Tests whether nerdctl properly creates the default network when required.
+func TestDefaultNetworkCreation(t *testing.T) {
+	testutil.RequiresRoot(t)
```
fahedouch, line 134: not sure if this check (RequiresRoot) is within nerdctl responsibilities  ( cc @AkihiroSuda )

pkg/netutil/netutil_test.go
```diff
@@ -97,3 +128,136 @@ func TestParseIPAMRange(t *testing.T) {
...
+// non-nerdctl-managed network config file.
+func TestNetworkWithDefaultNameAlreadyExists(t *testing.T) {
+	// NOTE: we create a tempdir for the CNI conf path to ensure an empty env for this test:
+	cniConfTestDir := t.TempDir()
+	cniEnv := CNIEnv{
+		Path:        t.TempDir(), // irrelevant for this test
+		NetconfPath: cniConfTestDir,
+	}
+
+	// Ensure no default network config is not present:
```
fahedouch, line 210: `Ensure no default network config is present:`

fahedouch:
@aznashwan PTAL!

pkg/netutil/netutil_test.go
```diff
@@ -97,3 +128,136 @@ func TestParseIPAMRange(t *testing.T) {
...
+		assert.NilError(t, defaultCniEnv.RemoveNetwork(defaultNet))
+	}
+
+	// NOTE: we create a tempdir for the CNI conf path to ensure an empty env for this test:
+	cniConfTestDir := t.TempDir()
+	cniEnv := CNIEnv{
+		Path:        ncdefaults.CNIPath(),
+		NetconfPath: cniConfTestDir,
+	}
+	// Ensure no default network config is not present:
```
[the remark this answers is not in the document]
aznashwan: Sorry for the excessive `:` punctuation, I've gone through the rest of the PR and hope I've found all instances of it and standardised them.

pkg/netutil/netutil_test.go
aznashwan: Added, though worth mentioning that `rootlessutil` isn't windows-compatible yet so this test will incorrectly be skipped. (will leave any updates to `rootlessutil` for a separate PR)

pkg/netutil/netutil.go
```diff
@@ -61,34 +62,65 @@ func NewCNIEnv(cniPath, cniConfPath string, opts ...CNIEnvOpt) (*CNIEnv, error)
...
-	return &e, nil
-}
 
-func (e *CNIEnv) NetworkMap() map[string]*networkConfig { //nolint:revive
-	m := make(map[string]*networkConfig, len(e.Networks))
-	for _, n := range e.Networks {
+	m := make(map[string]*networkConfig, len(networks))
+	for _, n := range networks {
+		if original, exists := m[n.Name]; exists {
+			logrus.Warnf("duplicate network name %q, %#v will get superseded by %#v", n.Name, original, n)
```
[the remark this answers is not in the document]
aznashwan: Nerdctl [uses `libcni.ConfListFromFile`](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L298) to parse the [`libcni.NetworkConfigList`](https://github.com/containernetworking/cni/blob/e629d106b4e02030d0f7794758b0f0438b6a78dd/libcni/api.go#L75) which will holds the actual `Name` field which is loaded from the `name` in the CNI JSON conf file. (`netutil.NetworkConfig` [inherits all fields from the `NetworkConfigList`](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L147))

There are no checks preventing multiple JSON config files having the same `name` field in them, which was [exactly what happened to me](https://github.com/containerd/nerdctl/pull/1538#issuecomment-1325461618) as described in my first comment.

The only real solution would be to switch nerdctl to using network IDs internally (for networks it had created itself), and only resort to the `name` field only for networks which were not defined by nerdctl. (and thus do not have a [nerdctl ID field](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L148) set)

I noticed issue #1568 (and PR #1581) are a step in the right direction but most of the network-listing codebase is still relying on identifying networks only by name. (or using `cniEnv.NetworkMap()` which has the potential of networks "shadowing" one-another)

aznashwan:
> @aznashwan FreeBSD CI is failing

@fahedouch good catch, it seems there is no FreeBSD CNI plugin being set up in the CI so I just made that test only run on Linux and Windows for now.

FreeBSD and Windows checks pass now, and the remaining CI failures seem to be unrelated. ([22.04](https://github.com/containerd/nerdctl/actions/runs/3628675258/jobs/6119969623#step:6:1186) and [20.04 rootless](https://github.com/containerd/nerdctl/actions/runs/3628675258/jobs/6119970636#step:7:1034))

PR #1548, opened 2022-11-23
fix `nerdctl exec` init console size 0 0 problem

ChangJH-Mark:
when execute `nerdctl exec -it <containerID> bash`,  initial tty size is `0,0`,  fix this bug by initialize terminal size when create execute spec


AkihiroSuda:
https://github.com/apps/dco

Also, please squash commits, and consider adding an integration test

Issue #1490, opened 2022-11-09
The CIDR of the bridge can't be changed (`network with name bridge already exists`)


AkihiroSuda:
>   I tried to modify the network config with just `nerdctl`, but it looks like there is no way to edit an existing network, and I can't delete the `bridge` network; it is being recreated right away:
> 
> ```console
> $ nerdctl network rm bridge
> bridge
> $ nerdctl network create bridge --gateway 10.10.10.1 --subnet 10.10.10.0/24
> FATA[0000] network with name bridge already exists
> $ nerdctl network inspect bridge
> [
>     {
>         "Name": "bridge",
>         "Id": "17f29b073143d8cd97b5bbe492bdeffec1c5fee55cc1fe2112c8b9335f8b6121",
>         "IPAM": {
>             "Config": [
>                 {
>                     "Subnet": "10.4.0.0/24",
>                     "Gateway": "10.4.0.1"
>                 }
>             ]
>         },
>         "Labels": {}
>     }
> ]
> ```

_Originally posted by @jandubois in https://github.com/rancher-sandbox/rancher-desktop/issues/3365#issuecomment-1308233426_
      

Seems regression in v0.22.1
- https://github.com/containerd/nerdctl/pull/1245

Zheaoli:
like https://github.com/containerd/nerdctl/issues/1425, If we allow people to change the CIDR of a network configuration, I think it would be better to validate if there is any container existed related with this network

AkihiroSuda:
> like #1425, If we allow people to change the CIDR of a network configuration, I think it would be better to validate if there is any container existed related with this network

Yes, it should be validated

fahedouch:
I believe that we have to exclude  `e.ensureDefaultNetworkConfig()`  From `NewCNIEnv(cniPath, cniConfPath string)`. I take this ticket

fahedouch:
closed by https://github.com/containerd/nerdctl/pull/1554

PR #1554, opened 2022-11-25
fix network creation behavior

fahedouch:
Nerdctl `default network` should be considered like other networks, that means we should be able to `CRUD` the default network. 
in this PR I am eliminating the confusion during a default network CRUD by disabling automatic creation during CRUD.

The default network still be automatically created by a container runtime

fixing https://github.com/containerd/nerdctl/issues/1490

### DIFF
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -42,7 +42,18 @@ type CNIEnv struct {
 	Networks    []*networkConfig
 }
 
-func NewCNIEnv(cniPath, cniConfPath string) (*CNIEnv, error) {
+type CNIEnvOpt func(e *CNIEnv) error
+
+func WithDefaultNetwork() CNIEnvOpt {
+	return func(e *CNIEnv) error {
+		if err := e.ensureDefaultNetworkConfig(); err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+func NewCNIEnv(cniPath, cniConfPath string, opts ...CNIEnvOpt) (*CNIEnv, error) {
 	e := CNIEnv{
 		Path:        cniPath,
 		NetconfPath: cniConfPath,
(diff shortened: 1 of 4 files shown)


pkg/netutil/netutil.go
```diff
@@ -42,17 +42,31 @@ type CNIEnv struct {
 	Networks    []*networkConfig
 }
 
-func NewCNIEnv(cniPath, cniConfPath string) (*CNIEnv, error) {
+type NewCNIEnvOpts func(e *CNIEnv) error
```
djdongjin, line 45: Maybe change `NewCNIEnvOpts` to `NewCNIEnvOpt` since the type itself is not plural (not an array of opts) and only means one `object/opt`.

fahedouch: I prefer `NewCNIEnvOpts` to indicate that this allows the caller to set `options` for the new CNIEnv.

Zheaoli:
Like I have saied in https://github.com/containerd/nerdctl/issues/1490#issuecomment-1312680880, I think we need extra validate step before we CRUD the network

fahedouch:
> Like I have saied in [#1490 (comment)](https://github.com/containerd/nerdctl/issues/1490#issuecomment-1312680880), I think we need extra validate step before we CRUD the network

this PR do modify the actual network CRUD but it fix `default network creation behavior` which is one of the prerequisite of networks CRUD

pkg/netutil/netutil.go
AkihiroSuda, line 45:
```suggestion
type CNIEnvOpt func(e *CNIEnv) error
```

AkihiroSuda: The plural form isn't consistent with the (de facto) standard convention

fahedouch: fixed

djdongjin:
LGTM (though still need other reviewer's approval 😛 )

AkihiroSuda (approved):
Thanks

PR #1557, opened 2022-11-26
Add yaml linter in workflow

aman556:
## Changes in PR:
- Add command in the test workflow for yamllinter
- Files with indentation issues are corrected.

Signed-off-by: Aman Sharma <amansh@vmware.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,47 +13,49 @@ jobs:
     runs-on: ubuntu-22.04
     timeout-minutes: 20
     steps:
-    - uses: actions/setup-go@v3
-      with:
-        go-version: 1.19.x
-    - uses: actions/checkout@v3.1.0
-      with:
-        path: src/github.com/containerd/nerdctl
-        fetch-depth: 100
-    - uses: containerd/project-checks@v1.1.0
-      with:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
+      - uses: actions/checkout@v3.1.0
+        with:
+          path: src/github.com/containerd/nerdctl
+          fetch-depth: 100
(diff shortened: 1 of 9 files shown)


AkihiroSuda:
Can we just use  the`yamllint` command that seems preinstalled on GHA?
https://github.com/lima-vm/lima/blob/43fcf3be924173d04b4a4b4b8823b38bf1a54cd5/.github/workflows/test.yml#L28

Just adding `run: yamllint .` into the GHA YAML would be fine. Probably no need to modify Makefile or create a hack script.

aman556:
Hey, @AkihiroSuda Thanks for following up here. Make sense, I also noticed later how golangci-lint is added. I am making the respective changes.

.github/workflows/test.yml
```diff
@@ -13,47 +13,57 @@ jobs:
...
+
+  yaml-lint:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3.1.0
+        with:
+          fetch-depth: 1
+      - name: Run yamllint
+        run: yamllint .
```
AkihiroSuda, line 53: golangci-lint and yaml-lint can be merged into a single "lint" job.

Also, could you squash the commits to 2 commits or 1 commit?

aman556:
@AkihiroSuda I have made the changes but just a thought we cant check the YAML errors locally if we are adding a GitHub action. Yeah, one can install locally but you know the count is less who checks this thing before pushing in change :D.

.github/workflows/test.yml
aman556: I am squashing just trying to resolve all the YAML errors and correct the configuration.

AkihiroSuda (approved):
Thanks

PR #127, opened 2021-03-21
Setup ipv6 hostip

fahedouch:
fixing bug https://github.com/containerd/nerdctl/issues/78#issue-819102502

PR #1558, opened 2022-11-27
Support IPv6 for nerdctl network

### LINKED ISSUE #1547: network create flag --ipv6 support

I am trying to implement containerd network with ipv6 traffic which is not supported currently is there any plan to address this?

if you support similar kind of feature like docker deamon or support nerdctl network create --ipv6 flag support it will help us resolve ipv6 traffic flow issue within contrainer.

### Additional context

_No response_

yuchanns:
Fix #1547 

FYI: [cni-ipam](https://www.cni.dev/plugins/current/ipam/host-local/)

Tasks:
- [x] Support in the `network create` command.
- [x] Support `--ip6` argument for the `container run` command.
- [x] Integration tests.
- [x] Documentation.
- [ ] ...

Signed-off-by: Hanchin Hsieh me@yuchanns.xyz
Co-authored-by: Zheao Li <me@manjusaka.me>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,6 +104,50 @@ jobs:
       - name: "Run integration tests"
         run: docker run -t --rm --privileged test-integration
 
+  test-integration-ipv6:
+    runs-on: "ubuntu-${{ matrix.ubuntu }}"
+    timeout-minutes: 40
+    strategy:
+      fail-fast: false
+      matrix:
+        # ubuntu-20.04: cgroup v1, ubuntu-22.04: cgroup v2
+        include:
+          - ubuntu: 22.04
+            containerd: v1.7.7
+    env:
+      UBUNTU_VERSION: "${{ matrix.ubuntu }}"
+      CONTAINERD_VERSION: "${{ matrix.containerd }}"
+    steps:
+      - uses: actions/checkout@v4.1.1
+        with:
+          fetch-depth: 1
+      - name: Enable ipv4 and ipv6 forwarding
+        run: |
+          sudo sysctl -w net.ipv6.conf.all.forwarding=1
+          sudo sysctl -w net.ipv4.ip_forward=1
+      - name: Enable IPv6 for Docker
+        run: |
+          sudo mkdir -p /etc/docker
+          echo '{"ipv6": true, "fixed-cidr-v6": "2001:db8:1::/64", "experimental": true, "ip6tables": true}' | sudo tee /etc/docker/daemon.json
+          sudo systemctl restart docker
+      - name: "Prepare integration test environment"
+        run: DOCKER_BUILDKIT=1 docker build -t test-integration-ipv6 --target test-integration-ipv6 --build-arg UBUNTU_VERSION=${UBUNTU_VERSION} --build-arg CONTAINERD_VERSION=${CONTAINERD_VERSION} .
+      - name: "Remove snap loopback devices (conflicts with our loopback devices in TestRunDevice)"
+        run: |
+          sudo systemctl disable --now snapd.service snapd.socket
+          sudo apt-get purge -y snapd
+          sudo losetup -Dv
+          sudo losetup -lv
+      - name: "Register QEMU (tonistiigi/binfmt)"
+        run: docker run --privileged --rm tonistiigi/binfmt --install all
+      - name: "Run integration tests"
+        # The nested IPv6 network inside docker and qemu is complex and needs a bunch of sysctl config.
+        # Therefore it's hard to debug why the IPv6 tests fail in such an isolation layer.
+        # On the other side, using the host network is easier at configuration.
+        # Besides, each job is running on a different instance, which means using host network here
+        # is safe and has no side effects on others.
+        run: docker run --network host -t --rm --privileged test-integration-ipv6
+
   test-integration-rootless:
     runs-on: "ubuntu-${{ matrix.ubuntu }}"
     timeout-minutes: 60
@@ -200,6 +244,8 @@ jobs:
           sudo apt-get install -y expect
       - name: "Ensure that the integration test suite is compatible with Docker"
         run: go test -timeout 20m -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon
+      - name: "Ensure that the IPv6 integration test suite is compatible with Docker"
+        run: go test -timeout 20m -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon -test.ipv6
 
   test-integration-windows:
     # A "larger" runner is used for enabling Hyper-V containers
--- a/Dockerfile
+++ b/Dockerfile
@@ -341,4 +341,8 @@ FROM test-integration-rootless AS test-integration-rootless-port-slirp4netns
 COPY ./Dockerfile.d/home_rootless_.config_systemd_user_containerd.service.d_port-slirp4netns.conf /home/rootless/.config/systemd/user/containerd.service.d/port-slirp4netns.conf
 RUN chown -R rootless:rootless /home/rootless/.config
 
+FROM test-integration AS test-integration-ipv6
+CMD ["gotestsum", "--format=testname", "--rerun-fails=2", "--packages=github.com/containerd/nerdctl/cmd/nerdctl/...", \
+  "--", "-timeout=30m", "-args", "-test.kill-daemon", "-test.ipv6"]
+
 FROM base AS demo
--- a/cmd/nerdctl/container_run_network_linux_test.go
+++ b/cmd/nerdctl/container_run_network_linux_test.go
@@ -19,6 +19,7 @@ package main
 import (
 	"fmt"
 	"io"
+	"net"
 	"regexp"
 	"runtime"
 	"strings"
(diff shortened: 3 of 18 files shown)


fahedouch:
this may inspire you https://github.com/containerd/nerdctl/pull/127/files :-D

yuchanns:
@containerd/nerdctl-maintainers Hey guys I'm encountering issues. 

The CI cannot pass when allocating the IPv6 address but test cases work well locally on my laptop. Later @Zheaoli figured out that the problem is lacking IPv6 support in test runners. And I reproduced the scene on the local host after disabling IPv6 support and confirmed that.

The matter is that GHA seems not to support IPv6 on test runners yet (and not going to) according to actions/runner-images#668.

Any ideas?

cmd/nerdctl/container_run_network_linux_test.go
```diff
@@ -575,3 +575,73 @@ func TestRunContainerWithMACAddress(t *testing.T) {
 		}
 	}
 }
+
+func TestRunContainerWithStaticIP6(t *testing.T) {
```
AkihiroSuda, line 579: What about non-static?

AkihiroSuda, line 579: IPv6 tests should require a new test CLI flag like `-test.ipv6=true`
https://github.com/containerd/nerdctl/blob/1642bef99f0931b79fe75fd54c1566e9e061b178/pkg/testutil/testutil.go#L507

yuchanns:
Blocking by lacking IPv6 test support. There is proposal #2031

AkihiroSuda:
> Blocked by lacking IPv6 test support. There is proposal #2031

I think we can merge this PR without CI as an experimental feature.

pkg/ocihook/ocihook.go
```diff
@@ -221,6 +225,7 @@ type handlerOpts struct {
 	extraHosts        map[string]string // host:ip
 	containerIP       string
 	contianerMAC      string
```
aojea, line 227:
```suggestion
	containerMAC      string
```

aojea:
> Blocked by lacking IPv6 test support. There is proposal #2031

the runners support IPv6 , you can run IPv6 internally in the machine, we have CI on kind running ipv6 only clusters.

What is not supported is IPv6 connectivity, the runners can not communicate externally with IPv6

aojea:
can you rebase and push again the PR to take a look at the logs of the failures?

yuchanns:
Fine. I will take a look later today.

cmd/nerdctl/container_run_network_linux_test.go
yuchanns: Make sense

aojea:
it is a permissions problem 

```
ailed (add): failed to set bridge addr: could not add IP address
```

it looks like a permission problems , I can't remember now but I think that you may need to tune a bit more the way you run the tests inside the container

https://github.com/containerd/nerdctl/blob/ce742c94c68897f07f94c7d68c84d1b52b23fce2/.github/workflows/test.yml#L105

cmd/nerdctl/container_run_network_linux_test.go
yuchanns: It was my mistake to name it `static`. In the test table, an item contains an empty IP string that can be assigned a non-static IP6. 

https://github.com/containerd/nerdctl/blob/f29f5f538ca468397af5d0aeeb5d233e6ac98324/cmd/nerdctl/container_run_network_linux_test.go#L507-L511

Later, it is verified whether it is a valid IP6 address.

https://github.com/containerd/nerdctl/blob/f29f5f538ca468397af5d0aeeb5d233e6ac98324/cmd/nerdctl/container_run_network_linux_test.go#L545-L547

Dockerfile
```diff
@@ -252,6 +252,7 @@ RUN perl -pi -e 's/multi-user.target/docker-entrypoint.target/g' /usr/local/lib/
   mkdir -p /etc/bash_completion.d && \
   nerdctl completion bash >/etc/bash_completion.d/nerdctl && \
   mkdir -p -m 0755 /etc/cni
+RUN sudo sysctl -w net.ipv6.conf.all.forwarding=1 && sudo sysctl -w net.ipv4.ip_forward=1
```
AkihiroSuda, line 255: The applied configuration disappears after the `RUN` line, so you have to specify this via `docker run --sysctl`

.github/workflows/test.yml
```diff
@@ -104,6 +104,51 @@ jobs:
       - name: "Run integration tests"
         run: docker run -t --rm --privileged test-integration
 
+  test-integration-ipv6:
```
AkihiroSuda, line 107: We have too many tests now, can we just merge test-integration-ipv6 to test-integration?

AkihiroSuda: If not, tes-integration-ipv6 should be just executed for Ubuntu 22.04 and containerd v1.7

Dockerfile
```diff
@@ -342,3 +342,7 @@ COPY ./Dockerfile.d/home_rootless_.config_systemd_user_containerd.service.d_port
 RUN chown -R rootless:rootless /home/rootless/.config
 
 FROM base AS demo
+
+FROM test-integration AS test-integration-ipv6
+CMD ["gotestsum", "--format=testname", "--rerun-fails=2", "--packages=github.com/containerd/nerdctl/cmd/nerdctl/...", \
+  "--", "-timeout=30m", "-args", "-test.kill-daemon", "-test.ipv6"]
```
AkihiroSuda, line 348: The final stage must remain “demo”, as it is the default stage

cmd/nerdctl/compose_up_linux_test.go
```diff
@@ -367,7 +367,7 @@ networks:
   default:
     ipam:
       config:
-        - subnet: 10.1.100.0/24
+        - subnet: 10.1.101.0/24
```
AkihiroSuda, line 370: ?

pkg/labels/labels.go
```diff
@@ -98,6 +98,8 @@ const (
 	// Boolean value which can be parsed with strconv.ParseBool() is required.
 	// (like "nerdctl/default-network=true" or "nerdctl/default-network=false")
 	NerdctlDefaultNetwork = Prefix + "default-network"
+
+	IP6Address = Prefix + "ip6"
```
AkihiroSuda, line 102: This line should come after the “IPAddress” line

.github/workflows/test.yml
yuchanns: Previously, it was merged. However, when using the `--network host` option in Docker to run integration tests, many non-IPv6 tests failed. Therefore, I separated it and ran it independently.

cmd/nerdctl/compose_up_linux_test.go
yuchanns: This is unintentionally 😓 I'll fix it

pkg/labels/labels.go
yuchanns: Make sense

.github/workflows/test.yml
```diff
@@ -200,6 +239,8 @@ jobs:
           sudo apt-get install -y expect
       - name: "Ensure that the integration test suite is compatible with Docker"
         run: go test -timeout 20m -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon
+      - name: "Ensure that the IPv6 integration test suite is compatible with Docker"
+        run: go test -timeout 20m -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon -test.ipv6
```
AkihiroSuda, line 243: Doesn't this need reconfiguring sysctl and `/etc/docker/daemon.json` ?

yuchanns: No. Fun fact 😂 
image

yuchanns: Somehow the integration-tests-ipv6 fails to run nested inside docker and qemu, despite sysctl and daemon being configured. Later @Zheaoli  figured out a solution by using `network=host`  to run it successfully. However, this leads to non-IPv6 test failure. So I have to separate the IPv6 test.

.github/workflows/test.yml
```diff
@@ -104,6 +104,45 @@ jobs:
...
+      - name: "Remove snap loopback devices (conflicts with our loopback devices in TestRunDevice)"
+        run: |
+          sudo systemctl disable --now snapd.service snapd.socket
+          sudo apt-get purge -y snapd
+          sudo losetup -Dv
+          sudo losetup -lv
+      - name: "Register QEMU (tonistiigi/binfmt)"
+        run: docker run --privileged --rm tonistiigi/binfmt --install all
+      - name: "Run integration tests"
+        run: docker run --network host -t --rm --privileged test-integration-ipv6
```
AkihiroSuda, line 144: Please add a comment to explain why we need (and why we can safely(?) use) `--network host`

.github/workflows/test.yml
aojea: hmm, it should be able to run nested, but nested networks are complex and hard to debug, maybe you are missing certain sysctl or something, who knows

.github/workflows/test.yml
yuchanns: Done.

AkihiroSuda (approved):
Thanks

yankay:
Thanks @yuchanns @AkihiroSuda
amazing

PR #1562, opened 2022-11-27
Refactor remove containers in compose

### LINKED ISSUE #1502: Refactor `stop containers` and `remove containers` in compose

djdongjin:
Fix #1502.

`remove compose containers` are used in several compose commands:

1. `compose down`: force remove (`-f`) containers and their volumes (`-v`)
2. `compose rm`: remove non-running containers. `-v` is passed based on user input. If `-s` is passed, `stop` containers before `rm`.
3. `compose run`: force remove (`-f`) containers when exit (if `--rm` is passed). Don't remove volumes (no `-v`).
4. `remove orphans` (in `up` and `run`): force remove orphans (`-f`) and volumes (`-v`).

What this PR changes:

1. Move all `remove` related functions/code to `compose/rm.go`.
2. Remove `downContainers` func/calls and change to use `removeContainers`.

### DIFF
--- a/pkg/composer/rm.go
+++ b/pkg/composer/rm.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 
 	"github.com/containerd/containerd"
+	"github.com/containerd/nerdctl/pkg/composer/serviceparser"
 	"github.com/containerd/nerdctl/pkg/formatter"
 	"github.com/containerd/nerdctl/pkg/labels"
 	"github.com/containerd/nerdctl/pkg/strutil"
@@ -29,7 +30,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// RemoveOptions stores all option input from `nerdctl compose rm`
+// RemoveOptions stores all options when removing compose containers:
+// Stop: if true, remove using `rm -f`; if false, check and skip running containers.
+// Volumes: if remove anonymous volumes associated with the container.
 type RemoveOptions struct {
 	Stop    bool
 	Volumes bool
@@ -73,7 +76,7 @@ func (c *Composer) removeContainers(ctx context.Context, containers []containerd
 		go func() {
 			defer rmWG.Done()
 			info, _ := container.Info(ctx, containerd.WithoutRefreshedMetadata)
-			// if no `--stop` is passed, check status and skip running container
+			// if not `Stop`, check status and skip running container
 			if !opt.Stop {
 				cStatus := formatter.ContainerStatus(ctx, container)
 				if strings.HasPrefix(cStatus, "Up") {
@@ -91,4 +94,21 @@ func (c *Composer) removeContainers(ctx context.Context, containers []containerd
 	rmWG.Wait()
 
 	return nil
+}
+
+func (c *Composer) removeContainersFromParsedServices(ctx context.Context, containers map[string]serviceparser.Container) {
+	var rmWG sync.WaitGroup
+	for id, container := range containers {
+		id := id
+		container := container
+		rmWG.Add(1)
+		go func() {
+			defer rmWG.Done()
+			logrus.Infof("Removing container %s", container.Name)
+			if err := c.runNerdctlCmd(ctx, "rm", "-f", id); err != nil {
+				logrus.Warn(err)
+			}
+		}()
+	}
+	rmWG.Wait()
 }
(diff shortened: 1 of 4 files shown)


pkg/composer/rm.go
```diff
@@ -92,3 +95,20 @@ func (c *Composer) removeContainers(ctx context.Context, containers []containerd
 
 	return nil
 }
+
+func (c *Composer) removeContainersFromParsedServices(ctx context.Context, containers map[string]serviceparser.Container) {
```
manugupt1, line 99: Could not this entire logic be removed by calling `nerdctl run --rm`, rm should not be supported when detached is true.

djdongjin: Do you mean calling `nerdctl run --rm` to do the job (run service container) when `nerdctl compose run --rm` is called? This function is directly moved from [here](https://github.com/containerd/nerdctl/blob/main/pkg/composer/run.go#L287C7-L300) which is called by `nerdctl compose run --rm` to process the `--rm` arg when needed.

This function is only a helper for deleting a list of containers (passed as `serviceparser.Container`) and doesn't do extra checks (similar to other compose helpers). The caller needs to do any checks it requires. For example, `compose run` checks `detach` [here](https://github.com/containerd/nerdctl/blob/main/pkg/composer/run.go#L263-L266) and returns directly. So this func won't be called when `detach` is true.

manugupt1: Yep! Also can we keep in in run only as this is very specific to compose run.

djdongjin: I think it's a good idea if it's doable (which I'm not sure :), because this func is after [`stopContainer`](https://github.com/containerd/nerdctl/blob/573d2a7bd661e527ebd69255bd2161da0f242593/pkg/composer/run.go#L284)). If it's doable to move `--rm` to `nerdctl run --rm` we should also take care of the `stopContainer` call above.

I'm okay with putting the function in either `composer/rm.go` or `composer/run.go`, but functionality-wise it's more related to `rm.go` so I put it there :). This func is actually not specific to `compose run` before. `compose up` also used [this function](https://github.com/containerd/nerdctl/blob/v1.0.0/pkg/composer/up_service.go#L89-L100) (which was a duplication) to `rm` containers when `stop` was not yet supported.

manugupt1: Got it! cool! thanks for explaining.

djdongjin: thanks for the review!

AkihiroSuda (approved):
Thanks

PR #1567, opened 2022-11-29
Add compose ps json

### LINKED ISSUE #1564: Support --format also on nerdctl compose ps

Currently, I can do `nerdctl ps --format "{{json .}}"` to have an JSON representation of the data.

I would like to have the same for `nerdctl compose ps`. So I can use the output in another program. 

### Additional context

_No response_

djdongjin:
Fix #1564. This seems require more changes than I thought to match docker output :)

Below is a comparision:

(The format of json output including keys matches. Some values might miss/mismatch (maybe because how nerdctl/docker runs differently. Notice the `Publishers` in service `wordpress`)

```shell
$ docker compose ps --format json | jq
[
  {
    "ID": "aba85f638641033254a353baf0f73ac2bce302d46d843c0037849005171a4238",
    "Name": "compose-ps-json-db-1",
    "Command": "docker-entrypoint.sh mysqld",
    "Project": "compose-ps-json",
    "Service": "db",
    "State": "running",
    "Health": "",
    "ExitCode": 0,
    "Publishers": [
      {
        "URL": "",
        "TargetPort": 3306,
        "PublishedPort": 0,
        "Protocol": "tcp"
      }
    ]
  },
  {
    "ID": "001f54b33d9bd481bda12d98821a93e11f40b92e2e557186c2b66e20c9279196",
    "Name": "compose-ps-json-wordpress-1",
    "Command": "docker-entrypoint.sh apache2-foreground",
    "Project": "compose-ps-json",
    "Service": "wordpress",
    "State": "running",
    "Health": "",
    "ExitCode": 0,
    "Publishers": [
      {
        "URL": "0.0.0.0",
        "TargetPort": 80,
        "PublishedPort": 8080,
        "Protocol": "tcp"
      },
      {
        "URL": "::",
        "TargetPort": 80,
        "PublishedPort": 8080,
        "Protocol": "tcp"
      }
    ]
  }
]
```

```shell
$ sudo nerdctl compose ps --format json | jq
[
  {
    "ID": "73a5c6cb8b2cf0a8c55fb571b9b6c2a4f84f2fc70f6706b76c208e370e18ea00",
    "Name": "compose-ps-json_wordpress_1",
    "Command": "docker-entrypoint.sh apache2-foreground",
    "Project": "compose-ps-json",
    "Service": "wordpress",
    "State": "running",
    "Health": "",
    "ExitCode": 0,
    "Publishers": [
      {
        "URL": "0.0.0.0",
        "TargetPort": 80,
        "PublishedPort": 8080,
        "Protocol": "tcp"
      }
    ]
  },
  {
    "ID": "a5330ffb76808ab679202a2c834c1c0270bdf4536a13c7347332029a1942780c",
    "Name": "compose-ps-json_db_1",
    "Command": "docker-entrypoint.sh mysqld",
    "Project": "compose-ps-json",
    "Service": "db",
    "State": "running",
    "Health": "",
    "ExitCode": 0,
    "Publishers": null
  }
]
```
```shell
$ cat docker-compose.yml
version: '3.1'

services:
  wordpress:
    image: wordpress:5.7
    ports:
      - 8080:80
    environment:
      WORDPRESS_DB_HOST: db
      WORDPRESS_DB_USER: exampleuser
      WORDPRESS_DB_PASSWORD: examplepass
      WORDPRESS_DB_NAME: exampledb
    volumes:
      - wordpress:/var/www/html
  db:
    image: mariadb:10.5
    environment:
      MYSQL_DATABASE: exampledb
      MYSQL_USER: exampleuser
      MYSQL_PASSWORD: examplepass
      MYSQL_RANDOM_ROOT_PASSWORD: '1'
    volumes:
      - db:/var/lib/mysql

volumes:
  wordpress:
  db:
```

----
A question is do we need to keep the json output identical as `docker compose ps --format json`?

The current implmenetation has several differences compared to docker:

1. No `Health`: not sure how to add it in nerdctl/containerd. (somehow I remember containerd doesn't support health check)
2. No `ExitCode`: has a function similar to [`ContainerStatus`](https://github.com/djdongjin/nerdctl/blob/950db8cefca281cb13bc8759d8aa45d7e1eac4bd/pkg/formatter/formatter.go#L41) that returns `ExitCode` (or split `ContainerStatus` into 2 funcs (e.g., `GetContainerStatus` and `ParseContainerStatus`))
3. `Status(nerdctl)` v.s. `State(docker)`: `running, Exited (0) 7 hours ago` v.s. `running, exited`.
4. `Ports(nerdctl)` v.s. `Publishers(docker)`: docker uses [this struct](https://github.com/docker/compose/blob/v2/pkg/api/api.go#L306) for port output in json. `nerdctl` uses `gocni.PortMapping`. (So even we output a slide than a string, the keys are still different)

```go
type PortMapping struct {
	HostPort      int32
	ContainerPort int32
	Protocol      string
	HostIP        string
}
```

### DIFF
--- a/cmd/nerdctl/compose_ps.go
+++ b/cmd/nerdctl/compose_ps.go
@@ -17,12 +17,17 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"text/tabwriter"
+	"time"
 
 	"github.com/containerd/containerd"
+	gocni "github.com/containerd/go-cni"
 	"github.com/containerd/nerdctl/pkg/formatter"
 	"github.com/containerd/nerdctl/pkg/labels"
+	"github.com/containerd/nerdctl/pkg/portutil"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
@@ -34,10 +39,34 @@ func newComposePsCommand() *cobra.Command {
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
+	composePsCommand.Flags().String("format", "", "Format the output. Supported values: [json]")
 	return composePsCommand
 }
 
+type composeContainerPrintable struct {
+	ID       string
+	Name     string
+	Command  string
+	Project  string
+	Service  string
+	State    string
+	Health   string // placeholder, lack containerd support.
+	ExitCode uint32
+	// `Publishers` stores docker-compatible ports and used for json output.
+	// `Ports` stores formatted ports and only used for console output.
+	Publishers []PortPublisher
+	Ports      string `json:"-"`
+}
+
 func composePsAction(cmd *cobra.Command, args []string) error {
+	format, err := cmd.Flags().GetString("format")
+	if err != nil {
+		return err
+	}
+	if format != "json" && format != "" {
+		return fmt.Errorf("unsupported format %s, supported formats are: [json]", format)
+	}
+
 	client, ctx, cancel, err := newClient(cmd)
 	if err != nil {
 		return err
@@ -48,61 +77,46 @@ func composePsAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
-	// TODO: make JSON-printable.
-	// The JSON format must correspond to `docker compose ps --json` (Docker Compose v2)
-	type containerPrintable struct {
-		Name    string
-		Command string
-		Service string
-		Status  string
-		Ports   string
-	}
-
-	containersPrintable := []containerPrintable{}
-
 	serviceNames, err := c.ServiceNames(args...)
 	if err != nil {
 		return err
 	}
-
 	containers, err := c.Containers(ctx, serviceNames...)
 	if err != nil {
 		return err
 	}
-	for _, container := range containers {
-		info, err := container.Info(ctx, containerd.WithoutRefreshedMetadata)
-		if err != nil {
-			return err
-		}
 
-		spec, err := container.Spec(ctx)
+	containersPrintable := []composeContainerPrintable{}
+	var p composeContainerPrintable
+	for _, container := range containers {
+		if format == "json" {
+			p, err = composeContainerPrintableJSON(ctx, container)
+		} else {
+			p, err = composeContainerPrintableTab(ctx, container)
+		}
 		if err != nil {
 			return err
-		}
-		status := formatter.ContainerStatus(ctx, container)
-		if status == "Up" {
-			status = "running" // corresponds to Docker Compose v2.0.1
-		}
-		p := containerPrintable{
-			Name:    info.Labels[labels.Name],
-			Command: formatter.InspectContainerCommandTrunc(spec),
-			Service: info.Labels[labels.ComposeService],
-			Status:  status,
-			Ports:   formatter.FormatPorts(info.Labels),
 		}
 		containersPrintable = append(containersPrintable, p)
 	}
 
+	if format == "json" {
+		outJSON, err := formatter.ToJSON(containersPrintable, "", "")
+		if err != nil {
+			return err
+		}
+		_, err = fmt.Fprint(cmd.OutOrStdout(), outJSON)
+		return err
+	}
+
 	w := tabwriter.NewWriter(cmd.OutOrStdout(), 4, 8, 4, ' ', 0)
 	fmt.Fprintln(w, "NAME\tCOMMAND\tSERVICE\tSTATUS\tPORTS")
-
 	for _, p := range containersPrintable {
 		if _, err := fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\n",
 			p.Name,
 			p.Command,
 			p.Service,
-			p.Status,
+			p.State,
 			p.Ports,
 		); err != nil {
 			return err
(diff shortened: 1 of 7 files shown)


AkihiroSuda:
> A question is do we need to keep the json output identical as `docker compose ps --format json`?

Yes

> 1. No `Health`: not sure how to add it in nerdctl/containerd. (somehow I remember containerd doesn't support health check)

Probably the health checker should be implemented on the daemon side as a new internal plugin that is similar to the restart monitor.
Maybe, it can be just embedded in the restart monitor.

> 2, 3, 4

When the information is missing in the output of the existing `nerdctl SUBCOMMAND`,
you can just call the containerd API from the composer.

Also, `nerdctl SUBCOMMAND --mode=native` might be useful for getting the internal information that is not exposed to the default `--mode=dockercompat` output

djdongjin:
thanks for the explanation @AkihiroSuda 

I will skip `1` in this PR as it requires more work on the daemon side. (Or I can put a `Health` placeholder which alway outputs `""`.)

2 and 3 are easy to handle. Just want to make sure we need to keep the output identical with docker.

For 4, I'll update the json output keys to be consistent with docker.
Something like:
```
Ports -> Publishers

HostIP -> URL
HostPort -> PublishedPort
ContainerPort -> TargetPort
Protocol-> Protocol
```

Zheaoli:
For 1, I perfer put a `Health` placeholder.

cmd/nerdctl/compose_ps.go
```diff
@@ -111,3 +125,114 @@ func composePsAction(cmd *cobra.Command, args []string) error {
...
+	var (
+		state    string
+		exitCode uint32
+	)
+	status, err := containerStatus(ctx, container)
+	if err == nil {
+		// show exitCode only when container is exited/stopped
+		if status.Status == containerd.Stopped {
+			exitCode = status.ExitStatus
+		}
```
djdongjin, line 175: use `exitcode` only if container is stopped/exitted: https://github.com/docker/compose/blob/v2/pkg/compose/ps.go#L86-L88

cmd/nerdctl/compose_ps.go
```diff
@@ -111,3 +125,114 @@ func composePsAction(cmd *cobra.Command, args []string) error {
...
+
+	task, err := c.Task(ctx, nil)
+	if err != nil {
+		return containerd.Status{}, err
+	}
+
+	return task.Status(ctx)
+}
+
+// PortPublisher hold status about published port
```
djdongjin, line 207: I copied this struct from docker to avoid importing a whole package. Put it (and the below `formatPublishers` func) here since they are only relevant to `compose ps --format json`.

cmd/nerdctl/compose_ps_linux_test.go
```diff
@@ -0,0 +1,105 @@
...
+				if !strings.Contains(stdout, field) {
+					return fmt.Errorf("[service: %s]marshalled json output doesn't have expected string (%s): %s", svc, field, stdout)
+				}
+			}
+			return nil
+		}
+	}
+
+	// check other formats are not supported
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "--format", "yaml").AssertFail()
```
djdongjin, line 90: Not sure why this line will fail docker-compatible CI test 🤔 I checked docker code and tested locally, docker does return error (not 0) here.

```shell
=== RUN   TestComposePsJSON
    compose_ps_linux_test.go:63: projectName="nerdctl-compose-test2770844775"
    compose_ps_linux_test.go:90: assertion failed: res.ExitCode is 0
--- FAIL: TestComposePsJSON (13.36s)
```

```
$ docker compose ps --format yaml
format value "yaml" could not be parsed: parsing failed
$ echo $?
1
$ sudo nerdctl compose ps --format yaml
FATA[0000] unsupported format yaml, supported formats are: [json]
$ sudo nerdctl compose echo $?
1
```
docker-compose code: [1](https://github.com/docker/compose/blob/v2/cmd/compose/ps.go#L143), [2](https://github.com/docker/compose/blob/v2/cmd/formatter/formatter.go#L70)

djdongjin:
I suspect the failed docker-compatible CI test is because the test is using `docker-compose v1`, not `v2`.

The GH Action image has both v1(1.29.2) and v2(2.12.2+azure-1) ([link](https://github.com/actions/runner-images/blob/ubuntu22/20221119.2/images/linux/Ubuntu2204-Readme.md)).

I manually tested the two versions and the v1 version does return exit code 0.

```
➜  compose-ps-json /home/ec2-user/.docker/cli-plugins/docker-compose version
docker-compose version 1.29.2, build 5becea4c
docker-py version: 5.0.0
CPython version: 3.7.10
OpenSSL version: OpenSSL 1.1.0l  10 Sep 2019
➜  compose-ps-json /home/ec2-user/.docker/cli-plugins/docker-compose ps --format yaml
List containers.

Usage: ps [options] [--] [SERVICE...]

Options:
    -q, --quiet          Only display IDs
    --services           Display services
    --filter KEY=VAL     Filter services by a property
    -a, --all            Show all stopped containers (including those created by the run command)
➜  compose-ps-json echo $?
0
```

AkihiroSuda:
> I suspect the failed docker-compatible CI test is because the test is using `docker-compose v1`, not `v2`.

The test can be skipped with `docker-compose` v1.
Eventually we should update the test suite to use `docker compose` v2

PR #1588, opened 2022-12-01
allow container network_mode in compose

sschepens:
Allow `container:name` network_mode in compose, it's already supported in nerdctl run.


AkihiroSuda (approved):
Thanks

PR #1593, opened 2022-12-02
Change Run to AssertOK in network test

djdongjin:
The `Run()` instead of `AsserOK` seems ignore `network rm` failure. Tested locally, make a draft PR to test in CI.

```shell
$ go test -v -exec sudo ./cmd/nerdctl/...
....
$ snd network ls
NETWORK ID      NAME                                              FILE
17f29b073143    bridge                                            /etc/cni/net.d/nerdctl-bridge.conflist
a6aa6228f4ab    nerdctl-testnetworkcreate                         /etc/cni/net.d/nerdctl-nerdctl-testnetworkcreate.conflist
80fdca7fb464    nerdctl-testnetworkremovewhenlinkwithcontainer    /etc/cni/net.d/nerdctl-nerdctl-testnetworkremovewhenlinkwithcontainer.conflist
# log
=== CONT  TestNetworkRemoveWhenLinkWithContainer
    network_rm_linux_test.go:68: assertion failed: res.ExitCode is not exitCode: time="2022-12-02T04:58:53Z" level=fatal msg="task 9c39efff9d85b1b235a8b8810bea18cf2fd0568f824baba8c8dd0e086723709f not found: not found"
        
--- FAIL: TestNetworkRemoveWhenLinkWithContainer (6.99s)
=== CONT  TestNetworkRemove
    network_rm_linux_test.go:47: assertion failed: res.ExitCode is not exitCode: time="2022-12-02T04:58:53Z" level=fatal msg="task ee06882a2a0a3e93136a6c28fe400779f1651d57f142eb074173de12ef6864bb not found: not found"
--- FAIL: TestNetworkRemove (3.97s)
```


manugupt1:
I think this looks okay! We could extend it to all the test suites imo.
Maybe something like this:
```
➜  nerdctl git:(filter-label) ✗ find ./ -type f -iname "*_test.go" | xargs -i@ sed -i 's/Run()/AssertOk()/g' @ 
```

fahedouch (approved):
Thanks

manugupt1 (approved):

djdongjin:
> Maybe something like this:
> 
> ```
> ➜  nerdctl git:(filter-label) ✗ find ./ -type f -iname "*_test.go" | xargs -i@ sed -i 's/Run()/AssertOk()/g' @ 
> ```

Thanks for the suggestion! We definitely can use this to get a list of defer calls with `Run()` and then we can check them case-by-case, because some tests are written in a way that a normal `rm` is called first and also a `defer rm` is called to ensure it's indeed `rm`ed. So sometimes the `defer` may fail as expected.

AkihiroSuda (approved):

Issue #1586, opened 2022-12-01
Some compose tests don't clean up network resources


djdongjin:
### Description

After running integration test, there are network resources left.

```shell
➜  nerdctl git:(main) sudo go test -v -exec sudo ./cmd/nerdctl/...
...
➜  nerdctl git:(main) snd network ls
NETWORK ID      NAME                                      FILE
17f29b073143    bridge                                    /etc/cni/net.d/nerdctl-bridge.conflist
05ec29f0ffdf    nerdctl-compose-test2152628242_default    /etc/cni/net.d/nerdctl-nerdctl-compose-test2152628242_default.conflist
                host
                none
```

#1535 resolved some of these, but there are tests left because it didn't log compose `projectName` (e.g., nerdctl-compose-test2152628242) and ignored.

### Steps to reproduce the issue

1. sudo nerdctl network prune
2. go test -v -exec sudo ./cmd/nerdctl/...
3. sudo nerdctl network ls

### Describe the results you received and expected

After test, all test resources are cleaned up.

### What version of nerdctl are you using?

main (b160d8173fb765b76cf0920d26621843d377bc3f)

djdongjin:
Close this as #1595 is merged. #1595 adds test name logging which helps target this kind of issue. I found for this issue it's because (probably due to my network setup) a test (`TestComposeRunWithServicePorts`) took too long in my local(exceeding default 10min go test timeout), panicing `go test` so the network resource is left there.

PR #1535, opened 2022-11-22
Ensure `compose down` is always called in tests to clean up all resources

djdongjin:
This PR fixes a test bug where resources created by `compose` are not cleaned up due to (1) no `(defer) compose down` or (2) `defer` not taking effect (at the end of a test) when failure/panic happens.

This CI failure seems related to this issue: https://github.com/containerd/nerdctl/actions/runs/3520521589/jobs/5901512778

Also I noticed after running test in `main` branch, sometimes I got test network left:

```shell
sudo go test -exec sudo -v ./cmd/nerdctl/...

➜  nerdctl git:(main) sudo nerdctl network ls
NETWORK ID      NAME                                      FILE
17f29b073143    bridge                                    /etc/cni/net.d/nerdctl-bridge.conflist
01bc329b3376    nerdctl-compose-test3150724813_default    /etc/cni/net.d/nerdctl-nerdctl-compose-test3150724813_default.conflist
f3a25c876ed7    nerdctl-compose-test455991460_default     /etc/cni/net.d/nerdctl-nerdctl-compose-test455991460_default.conflist
```

Checking the generated log, the leftover networks are generated here:

```shell
=== RUN   TestComposeRun
    compose_run_linux_test.go:52: projectName="nerdctl-compose-test3150724813"  # no `compose down` is called
--- PASS: TestComposeRun (1.24s)
....
=== RUN   TestComposeRunWithServicePorts
    compose_run_linux_test.go:122: projectName="nerdctl-compose-test455991460" # test failure/timeout and the `defer` is at the end (so not called)
```

Signed-off-by: Jin Dong <jindon@amazon.com>

PR #1595, opened 2022-12-02
Improve compose test logging

djdongjin:
This PR adds `projectName` log to compose tests without it. 

Compose tests create resources like images(compose build/pull) and network(compose up/run). Sometimes resources are left due to bugs/test errors/forgot clean up/etc, and it's difficult to find which testcase causes it. (e.g. #1586, #1535).

It also moves some response body log to only when error occurs. Those logs are quite long and I think it's sufficient to only log them when error occurs. This makes test output more clean and short. 

### DIFF
--- a/cmd/nerdctl/compose_run_linux_test.go
+++ b/cmd/nerdctl/compose_run_linux_test.go
@@ -137,8 +137,8 @@ services:
 		if err != nil {
 			return err
 		}
-		t.Logf("respBody=%q", respBody)
 		if !strings.Contains(string(respBody), testutil.NginxAlpineIndexHTMLSnippet) {
+			t.Logf("respBody=%q", respBody)
 			return fmt.Errorf("respBody does not contain %q", testutil.NginxAlpineIndexHTMLSnippet)
 		}
 		return nil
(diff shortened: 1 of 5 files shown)


cmd/nerdctl/compose_run_linux_test.go
```diff
@@ -125,7 +125,7 @@ services:
 		// unbuffer(1) can be installed with `apt-get install expect`.
 		unbuffer := []string{"unbuffer"}
 		base.ComposeCmdWithHelper(unbuffer, "-f", comp.YAMLFullPath(),
-			"run", "--service-ports", "--name", containerName, "web").Run()
+			"run", "--service-ports", "--name", containerName, "web").AssertOK()
```
djdongjin, line 128: I think it makes sense to `AssertOK()` here. If it fails, the for loop check below will also not work.

cmd/nerdctl/compose_run_linux_test.go
```diff
@@ -184,7 +184,7 @@ services:
 		// unbuffer(1) can be installed with `apt-get install expect`.
 		unbuffer := []string{"unbuffer"}
 		base.ComposeCmdWithHelper(unbuffer, "-f", comp.YAMLFullPath(),
-			"run", "--publish", "8080:80", "--name", containerName, "web").Run()
+			"run", "--publish", "8080:80", "--name", containerName, "web").AssertOK()
```
djdongjin, line 187: Same as above.

AkihiroSuda (approved):
Thanks

PR #1598, opened 2022-12-03
Add `compose exec` command

djdongjin:
Implement `compose exec` command. docker doc: https://docs.docker.com/engine/reference/commandline/compose_exec/

Also help feature request from downstream community that relies on nerdctl (finch: https://github.com/runfinch/finch/issues/79)

### DIFF
--- a/cmd/nerdctl/compose_exec_linux_test.go
+++ b/cmd/nerdctl/compose_exec_linux_test.go
@@ -1,0 +1,135 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestComposeExec(t *testing.T) {
+	// `-i` in `compose run & exec` is only supported in compose v2.
+	// Currently CI is using compose v1.
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	var dockerComposeYAML = fmt.Sprintf(`
+version: '3.1'
+
+services:
+  svc0:
+    image: %s
+  svc1:
+    image: %s
+`, testutil.CommonImage, testutil.CommonImage)
+
+	comp := testutil.NewComposeDir(t, dockerComposeYAML)
+	defer comp.CleanUp()
+	projectName := comp.ProjectName()
+	t.Logf("projectName=%q", projectName)
+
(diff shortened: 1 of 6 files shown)


djdongjin:
While implementing `compose exec`, I notice the `nerdctl exec` command mentions "`-t, --tty` should correspond to `-i`":

https://github.com/containerd/nerdctl/blob/dff398c6a9ca40db8f9c25c70e76455267564f51/cmd/nerdctl/exec.go#L53

However, I didn't see this is checked, and they are more independent:

https://github.com/containerd/nerdctl/blob/dff398c6a9ca40db8f9c25c70e76455267564f51/cmd/nerdctl/exec.go#L143-L149

https://github.com/containerd/nerdctl/blob/a8ab5cdb3dda7cb64f9e4f964105243b4ab5ef27/cmd/nerdctl/exec_linux_test.go#L69

Is this restriction removed? I saw similar code in `run` and `compose run`, and `compose run` is following the document:

https://github.com/containerd/nerdctl/blob/7276e4e3ef3b6fddcb7654c5188e40f1ba34fa85/cmd/nerdctl/compose_run.go#L119

pkg/composer/exec.go
```diff
@@ -0,0 +1,91 @@
...
+	args = append(args, eo.Args...)
+	cmd := c.createNerdctlCmd(ctx, args...)
+
+	if eo.Interactive {
+		cmd.Stdin = os.Stdin
+	}
+	if !eo.Detach {
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+	}
```
djdongjin, line 89: Not sure if this IO setup is too simplified, still looking at how other commands (`compose run`, `run`, `exec`) handle IO (given `-i`/`-d`/`-t`). Appreciate any suggestion :)

djdongjin:
> > Is this restriction removed?
> 
> Yes, probably

Seems fixed in this commit: https://github.com/containerd/nerdctl/commit/18ea81f8f7cb6949b4748349b2782aad19e4d484

I'll take a look!

AkihiroSuda (approved):
Thanks

cmd/nerdctl/compose_exec_linux_test.go
```diff
@@ -0,0 +1,135 @@
...
+		"nobody:users": "uid=65534(nobody) gid=100(users)",
+	}
+
+	for userStr, expected := range testCases {
+		args := []string{"-f", comp.YAMLFullPath(), "exec", "-i=false", "-t=false"}
+		if userStr != "" {
+			args = append(args, "--user", userStr)
+		}
+		args = append(args, "svc0", "id")
+		base.ComposeCmd(args...).AssertOutContains(expected)
```
AkihiroSuda, line 94: AssertOutExactly might be preferable

AkihiroSuda:
Merged, but please take a look: https://github.com/containerd/nerdctl/pull/1598#discussion_r1040579720

cmd/nerdctl/compose_exec_linux_test.go
djdongjin: Thanks for the suggestion.

`id` output (the empty case) has extra strings:

```
➜  compose-exec sndc exec svc0 id
uid=0(root) gid=0(root) groups=1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
➜  compose-exec sndc exec -u 1000 svc0 id
uid=1000 gid=0(root)
```

PR #1600, opened 2022-12-04
Filter images by reference

manugupt1:
Adds support for filtering images by reference.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -156,4 +156,4 @@ jobs:
         run: |
           sudo apt-get install -y expect
       - name: "Ensure that the integration test suite is compatible with Docker"
-        run: go test -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon
+        run: go test -timeout 20m -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon
--- a/cmd/nerdctl/images.go
+++ b/cmd/nerdctl/images.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"strings"
 	"text/tabwriter"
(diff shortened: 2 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -156,4 +156,4 @@ jobs:
         run: |
           sudo apt-get install -y expect
       - name: "Ensure that the integration test suite is compatible with Docker"
-        run: go test -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon
+        run: go test -timeout 30m -v -exec sudo ./cmd/nerdctl/... -args -test.target=docker -test.kill-daemon
```
Zheaoli, line 159: Why do we need this?

Zheaoli (commented):
LGTM for the code part

cmd/nerdctl/images.go
```diff
@@ -144,6 +148,30 @@ func imagesAction(cmd *cobra.Command, args []string) error {
...
 
+func filterByReference(imageList []images.Image, filters []string) ([]images.Image, error) {
+	var filteredImageList []images.Image
+	logrus.Debug(filters)
+	for _, image := range imageList {
+		logrus.Debug(image.Name)
+		var matches int
+		for _, f := range filters {
+			var err error
+			matchFound, err := regexp.MatchString(f, image.Name)
```
AkihiroSuda, line 159: Looks like it is expected to support wildcard, not regexp?
https://docs.docker.com/engine/reference/commandline/images/#filter-images-by-reference

```
docker images --filter=reference='busy*:*libc'
```

AkihiroSuda: We should also make sure that `--filter=reference='busy*:*libc'` matches `docker.io/library/busy*:*libc`

AkihiroSuda (approved):
Thanks

PR #1609, opened 2022-12-06
Add `compose top` command

djdongjin:
https://docs.docker.com/engine/reference/commandline/compose_top/


AkihiroSuda (approved):
Thanks

Issue #1596, opened 2022-12-02
nerdctl load output is messed up for multi-image files


hholierhoek:
### Description

Commit https://github.com/containerd/nerdctl/commit/0bcedac04e22208eda6889ead48f682a62443189 messed up the output of the load command. The print statement 'Loaded image: xxx' needs a newline at the end; currently the next 'unpacking' log is written at the end of the previous 'Loaded image' log.

So the output is formatted as:
`unpacking docker.io/lib/my-app1:1.2.3 (sha256:xxxx)...
Loaded image: docker.io/lib/my-app1:1.2.3unpacking docker.io/lib/my-app2:2.3.4 (sha256:yyyy)...`

### Steps to reproduce the issue

1. create a tar bundle containing multiple images
2. run 'nerdctl load -i <tar file>'

### Describe the results you received and expected

Separate lines in the output, or revert to the old output.

### What version of nerdctl are you using?

1.0.0

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

RHEL8.6

hholierhoek:
Note that, despite checking in preview before submitting, the sample output didn't formatted right. 'Loaded image:' is actually on the next line.

airmelt:
Could you assign this bug to me? I am new here and working on this. Thanks!

fahedouch:
closed by https://github.com/containerd/nerdctl/pull/1613

PR #1613, opened 2022-12-06
fix: multi-image files load output

airmelt:
# Description:

Now the output is formatted as:

```bash
unpacking docker.io/lib/my-app1:1.2.3 (sha256:xxxx)... 
Loaded image: docker.io/lib/my-app1:1.2.3unpacking docker.io/lib/my-app2:2.3.4 (sha256:yyyy)...$
```

There is a missing Line Feed("\n") before next order.

# Issue

#1596 

# Test

- I have bundle some images to a tar file
- use `nerdctl load -i image-name` to load the images

# Testing Example

```bash
$ nerdctl load -i ./docker-app/nginx.tar
unpacking docker.io/library/nginx:1.21 (sha256:56edf1dcb8a68cfafa096182f56e9e1c700806fc60662df9118a32638f66fb8f)...
Loaded image: docker.io/library/nginx:1.21
$
```

Signed-off-by: airmelt <4irmelt@gmail.com>

### DIFF
--- a/cmd/nerdctl/load.go
+++ b/cmd/nerdctl/load.go
@@ -132,7 +132,7 @@ func loadImage(in io.Reader, cmd *cobra.Command, platMC platforms.MatchComparer,
 		if quiet {
 			fmt.Fprintln(cmd.OutOrStdout(), img.Target.Digest)
 		} else {
-			fmt.Fprintf(cmd.OutOrStdout(), "Loaded image: %s", img.Name)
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

The test is good to have, but not the hard requirement in this case

airmelt:
> Would you mind to add some test to validate the output format?

Thanks for your comment, I agree with @AkihiroSuda that it's a easy task and it's enough to test.

PR #1628, opened 2022-12-09
Add `compose port` command

djdongjin:
https://docs.docker.com/engine/reference/commandline/compose_port/

It has 2 changes:
1. Add the command
2. Refactor and move the task logic (print port) to a new pkg (`containerutil`).
    1. This gives better separation between command/arg logic and main task logic.
    2. Later on maybe we can move more main task logic into this package (or other `pkg`), so that they can be reused by other commands (e.g. relevant compose command).

(Please lmk if 2 doesn't look good, or any suggestion if there is better place for this kind of functions (e.g., finish a task given a container).

Also there is a incompatible behavior compared to docker compose (docker itself has incompatible behavior in v1 and v2), when a given port/protocol doesn't exist:
1. nerdctl (in this PR) fails the command and prints an error.
2. docker compose v1 prints nothing (`\n`).
3. docker compose v2 prints `:0\n`.

Please lmk if this is okay.

### DIFF
--- a/cmd/nerdctl/compose_port_linux_test.go
+++ b/cmd/nerdctl/compose_port_linux_test.go
@@ -1,0 +1,85 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


cmd/nerdctl/compose_port_linux_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+)
+
+func TestComposePort(t *testing.T) {
+	// when no port mapping is found, docker compose v1 prints `\n` while v2 prints `:0\n`
+	// both v1 and v2 have exit code 0 (succeess)
+	// nerdctl compose will fail with error (no public port fouond).
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 30: Can we split the incompatible part to another test?

djdongjin: fixed.

AkihiroSuda (approved):
Thanks

PR #1630, opened 2022-12-09
pkg/idutil/netwalker: fix nil pointer dereference

AkihiroSuda:
Fix #1629

PR #1618, opened 2022-12-07
Fix: Deleting multiple networks stops at first delete fail

yzxiu:
fixing https://github.com/containerd/nerdctl/issues/1587

```bash
 ~ nerdctl network create deleteme --gateway 10.10.11.1 --subnet 10.10.11.0/24

f9d6cec6deabf54691804bb7c1bf02ad7f461ab07f31a18398495b9de1cabf28

 ~ nerdctl  network ls                                                         
NETWORK ID      NAME               FILE
                k8s-pod-network    /etc/cni/net.d/10-calico.conflist
17f29b073143    bridge             /etc/cni/net.d/nerdctl-bridge.conflist
f9d6cec6deab    deleteme           /etc/cni/net.d/nerdctl-deleteme.conflist
                host               
                none

 ~ nerdctl network rm host k8s-pod-network test deleteme
ERRO[0000] pseudo network "host" cannot be removed      
ERRO[0000] k8s-pod-network is managed outside nerdctl and cannot be removed 
ERRO[0000] No such network: test                        
f9d6cec6deab
```

Signed-off-by: yaozhenxiu <946666800@qq.com>

PR #1633, opened 2022-12-09
fix network walker filter

fahedouch:
see https://github.com/containerd/nerdctl/pull/1630/files#r1044456293  (but #1618 will be on `v1.1.1`) 


AkihiroSuda:
Thanks, but I merged #1618 as it was updated

PR #1638, opened 2022-12-10
Fix bugs in compose stop and restart

djdongjin:
In both command the timeout arg is `--timeout, -t` while the underlying command (`nerdctl restart/stop`) uses `--time, -t`. Change to use `-t` when constructing the underlying command. Also fix a pointer deference bug.

Fix them and add testcases.

### DIFF
--- a/cmd/nerdctl/compose_stop_linux_test.go
+++ b/cmd/nerdctl/compose_stop_linux_test.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/containerd/nerdctl/pkg/testutil"
 )
@@ -64,23 +63,36 @@ volumes:
 	projectName := comp.ProjectName()
 	t.Logf("projectName=%q", projectName)
 
+	upAssertHandler := func(svc string) func(stdout string) error {
+		return func(stdout string) error {
+			// Docker Compose v1: "Up", v2: "running"
+			if !strings.Contains(stdout, "Up") && !strings.Contains(stdout, "running") {
+				return fmt.Errorf("service \"%s\" must have been still running", svc)
+			}
+			return nil
+		}
+	}
+
+	exitAssertHandler := func(svc string) func(stdout string) error {
+		return func(stdout string) error {
+			// Docker Compose v1: "Exit code", v2: "exited (code)"
+			if !strings.Contains(stdout, "Exit") && !strings.Contains(stdout, "exited") {
+				return fmt.Errorf("service \"%s\" must have exited", svc)
+			}
+			return nil
+		}
+	}
+
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "up", "-d").AssertOK()
 	defer base.ComposeCmd("-f", comp.YAMLFullPath(), "down", "-v").Run()
 
+	// stop should (only) stop the given service.
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "stop", "db").AssertOK()
-	time.Sleep(3 * time.Second)
-	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "db").AssertOutWithFunc(func(stdout string) error {
-		// Docker Compose v1: "Exit code", v2: "exited (code)"
-		if !strings.Contains(stdout, "Exit") && !strings.Contains(stdout, "exited") {
-			return fmt.Errorf("service \"db\" must have exited")
-		}
-		return nil
-	})
-	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "wordpress").AssertOutWithFunc(func(stdout string) error {
-		// Docker Compose v1: "Up", v2: "running"
-		if !strings.Contains(stdout, "Up") && !strings.Contains(stdout, "running") {
-			return fmt.Errorf("service \"wordpress\" must have been still running")
-		}
-		return nil
-	})
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "db").AssertOutWithFunc(exitAssertHandler("db"))
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "wordpress").AssertOutWithFunc(upAssertHandler("wordpress"))
+
+	// `--timeout` arg should work properly.
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "stop", "--timeout", "5", "wordpress").AssertOK()
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "wordpress").AssertOutWithFunc(exitAssertHandler("wordpress"))
+
 }
--- a/pkg/composer/restart.go
+++ b/pkg/composer/restart.go
@@ -56,7 +56,8 @@ func (c *Composer) Restart(ctx context.Context, opt RestartOptions, services []s
 func (c *Composer) restartContainers(ctx context.Context, containers []containerd.Container, opt RestartOptions) error {
 	var timeoutArg string
 	if opt.Timeout != nil {
-		timeoutArg = fmt.Sprintf("--timeout=%d", *opt.Timeout)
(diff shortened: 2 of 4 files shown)


cmd/nerdctl/compose_stop_linux_test.go
```diff
@@ -64,23 +63,36 @@ volumes:
...
-		if !strings.Contains(stdout, "Up") && !strings.Contains(stdout, "running") {
-			return fmt.Errorf("service \"wordpress\" must have been still running")
-		}
-		return nil
-	})
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "db").AssertOutWithFunc(exitAssertHandler("db"))
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "wordpress").AssertOutWithFunc(upAssertHandler("wordpress"))
+
+	// `--timeout` arg should work properly.
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "stop", "-t", "5", "wordpress").AssertOK()
```
manugupt1, line 95: Does this also check if stop happens after 5 seconds if the processs runs for more than 5 secs?

djdongjin: The timeout is for grace shutdown period. At first the command sends a SIGTERM, and wait for (at most) 5 sec (the timeout) before sending a SIGKILL. So in either case the container/service should be stopped. We should be good by checking the expected status after command returns.

I believe the timeout arg only ensures it sends SIGKILL after waiting for the period. It cannot ensure the container is indeed stopped after the period. Sometimes it might take a little longer so the SIGKILL takes effect.

pkg/composer/restart.go
```diff
@@ -56,7 +56,7 @@ func (c *Composer) Restart(ctx context.Context, opt RestartOptions, services []s
 func (c *Composer) restartContainers(ctx context.Context, containers []containerd.Container, opt RestartOptions) error {
 	var timeoutArg string
 	if opt.Timeout != nil {
-		timeoutArg = fmt.Sprintf("--timeout=%d", *opt.Timeout)
+		timeoutArg = fmt.Sprintf("-t=%d", *opt.Timeout)
```
AkihiroSuda, line 59: Any reason for this change?

djdongjin: compose  stop/restart use ‘timeout’, stop,restart use ‘time’.  all use ‘t’ as shortcut.

fahedouch:
@djdongjin why are you pushing to use `-t` for both `compose` and `nerdctl` when it invokes `stop/start` ? 
they are two different commands though, no direct link between them!

djdongjin:
> @djdongjin why are you pushing to use `-t` for both `compose` and `nerdctl` when it invokes `stop/start` ?
> they are two different commands though, no direct link between them!

I'm not pushing to use `-t` for both (or for either `nerdctl` or `compose`) :) . Here I only changed the `compose` part (no change to `nerdctl`) to fix the naming bug.

The same argument in `nerdctl compose` and `nerdctl` (in terms of functionality, i.e., time to wait before sending SIGKILL) has different names (`--timeout, -t` in `nerdctl/docker compose` v.s. `--time, -t` in `nerdctl/docker`) (docker compose: [stop](https://docs.docker.com/engine/reference/commandline/compose_stop/#options), [restart](https://docs.docker.com/engine/reference/commandline/compose_restart/#options); docker: [stop](https://docs.docker.com/engine/reference/commandline/stop/#options), [restart](https://docs.docker.com/engine/reference/commandline/restart/#options), same in `nerdctl`).

I didn't notice this naming difference previously, so if we pass `--timeout` from `nerdctl compose` to `nerdctl` it has errors (`nerdctl stop/restart` don't have such arg). So this PR is to fix this. I changed `--timeout` to `-t` when `nerdctl compose stop/restart` invoke `nerdctl stop/restart`.

That being said, I'm okay with either full name (`--time`) or short name (`-t`) when `compose stop/restart` pass to `nerdctl stop/restart`. I can change to `--time` if that's preferred :)

Please lmk if I miss anything or my understanding is not correct.

Thanks.

cmd/nerdctl/compose_stop_linux_test.go
djdongjin: Thanks for the input. I'm not sure if that can be ensured. I tested on `docker compose` and it's common to take longer than the time given in `--timeout`. Take the 1st container as an example, I suspect the `8.3s ~= 5s(timeout) + 3.3s(send SIGKILL, SIGKILL take effect, new status returned to the cli)`.

```
$ d compose stop -t 5
[+] Running 24/24
 ⠿ Container compose-test-svcc7-1   Stopped               8.3s
 ⠿ Container compose-test-svccc2-1  Stopped               6.8s
 ⠿ Container compose-test-svc5-1    Stopped               7.6s
 ⠿ Container compose-test-svccc1-1  Stopped               8.4s
 ⠿ Container compose-test-svccc5-1  Stopped               6.9s
 ⠿ Container compose-test-svcc3-1   Stopped               7.1s
 ⠿ Container compose-test-svc4-1    Stopped               7.7s
 ⠿ Container compose-test-svccc3-1  Stopped               7.3s
 ⠿ Container compose-test-svc2-1    Stopped               7.8s
....
```

Probably we can run the command in a goroutine and let the main goroutine sleep longer than 5s (e.g., 8s) and then check the status. But in uncommon case it might become flaky (or not?) if the latency longer than 8s (as shown above in docker example).

Please lmk if this sounds good. Thanks for any other suggestions.

fahedouch:
> > @djdongjin why are you pushing to use `-t` for both `compose` and `nerdctl` when it invokes `stop/start` ?
> > they are two different commands though, no direct link between them!
> 
> I'm not pushing to use `-t` for both (or for either `nerdctl` or `compose`) :) . Here I only changed the `compose` part (no change to `nerdctl`) to fix the naming bug.
> 
> The same argument in `nerdctl compose` and `nerdctl` (in terms of functionality, i.e., time to wait before sending SIGKILL) has different names (`--timeout, -t` in `nerdctl/docker compose` v.s. `--time, -t` in `nerdctl/docker`) (docker compose: [stop](https://docs.docker.com/engine/reference/commandline/compose_stop/#options), [restart](https://docs.docker.com/engine/reference/commandline/compose_restart/#options); docker: [stop](https://docs.docker.com/engine/reference/commandline/stop/#options), [restart](https://docs.docker.com/engine/reference/commandline/restart/#options), same in `nerdctl`).
> 
> I didn't notice this naming difference previously, so if we pass `--timeout` from `nerdctl compose` to `nerdctl` it has errors (`nerdctl stop/restart` don't have such arg). So this PR is to fix this. I changed `--timeout` to `-t` when `nerdctl compose stop/restart` invoke `nerdctl stop/restart`.
> 
> That being said, I'm okay with either full name (`--time`) or short name (`-t`) when `compose stop/restart` pass to `nerdctl stop/restart`. I can change to `--time` if that's preferred :)
> 
> Please lmk if I miss anything or my understanding is not correct.
> 
> Thanks.

thank you for the details. I think it is better to change to `--time` to highlight this diff for readers and avoid confusing

cmd/nerdctl/compose_stop_linux_test.go
fahedouch: hmmm , I expected that `ensure that container do not exceed 5 sec with running state` is hard to be tested too  and for sure we have to avoid flaky test, we are trying hard to make the CI more stable :) , so SGTM to keep it that way ;)

djdongjin:
> thank you for the details. I think it is better to change to `--time` to highlight this diff for readers and avoid confusing

Thanks for the suggestion! yeah it's a little confusing 😅. updated to use `--time` and add comment to highlight the diff :)

cmd/nerdctl/compose_stop_linux_test.go
djdongjin: thanks :)

AkihiroSuda (approved):
Thanks

PR #1639, opened 2022-12-11
Refactor the package structure in cmd/nerdctl

### LINKED ISSUE #1631: Move *.go files for subcommand out main package

I wish to import the nerdctl subcommand package to add some new features.

Move *.go files for subcommand out of the main package so we can import it.

### Additional context

In addition, are there any other ways to implement this?

Zheaoli:
Fix #1631 

First, I'm sorry that I make this PR too large to review. I have to make a one-shot PR if I want to  re-structured the directory because the code in the `cmd/nerdctl`  has been mixed,

In this PR, I have made four things below

1. split the code file in the `cmd/nerdctl` into subdirectory which the subcommand has organized
2. make generic code into a common utils package in the `cmd/nerdctl/utils`
3. move the shell-completion feature into completion package in the `cmd/nerdct/completion`
4. move all the e2e test together.


djdongjin (commented):
nit: `integration` might be a better name than `e2e`

cmd/nerdctl/event/events.go
```diff
@@ -26,6 +26,8 @@ import (
 
 	"github.com/containerd/containerd/events"
 	"github.com/containerd/containerd/log"
+	client2 "github.com/containerd/nerdctl/cmd/nerdctl/client"
```
AkihiroSuda, line 29: I'd prefer the package name to be like `ncclient`

Zheaoli: Yes, I will clean dirty name when I pass all CI(lol

cmd/nerdctl/event/events.go
```diff
@@ -26,6 +26,8 @@ import (
...
+	fmt2 "github.com/containerd/nerdctl/cmd/nerdctl/utils"
```
AkihiroSuda, line 30: I'd prefer the package name to be something like `utils/fmtutil`.

cmd/nerdctl/create/create.go
```diff
@@ -14,16 +14,20 @@
    limitations under the License.
 */
 
-package main
+package create
```
AkihiroSuda, line 17: Should this be in the `nerdctl/container` package, or maybe better in the `nerdctl/container/create` pkg?

Same applies to the `run`, `exec`, `cp` commands, etc.

cmd/nerdctl/event/events.go
djdongjin: another suggestion is `ndclient` 😃  since `nd` reads more similar to `nerd`ctl

AkihiroSuda: I'd argue that the contaiNERD client library is also "nerd" 🤓

cmd/nerdctl/compose/compose_down.go
```diff
@@ -30,7 +31,7 @@ func newComposeDownCommand() *cobra.Command {
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
-	composeDownCommand.Flags().BoolP("volumes", "v", false, "Remove named volumes declared in the `volumes` section of the Compose file and anonymous volumes attached to containers.")
+	composeDownCommand.Flags().BoolP("volumes", "v", false, "Remove named volume declared in the `volume` section of the Compose file and anonymous volume attached to containers.")
```
djdongjin, line 34: `volume` can have multiple right? we should use plural

cmd/nerdctl/compose/compose_images.go
```diff
@@ -131,7 +133,7 @@ func printComposeImages(ctx context.Context, cmd *cobra.Command, containers []co
 				return err
 			}
 
-			size, err := unpackedImageSize(ctx, sn, image)
+			size, err := utils.UnpackedImageSize(ctx, sn, image)
```
djdongjin, line 136: I feel `pkg/imgutil` might be a better place for this func since it doesn't depend on `cmd` related inputs (flags etc) and more related to `img`.

https://github.com/containerd/nerdctl/blob/main/pkg/imgutil/imgutil.go

cmd/nerdctl/compose/compose_pause.go
```diff
@@ -65,7 +67,7 @@ func composePauseAction(cmd *cobra.Command, args []string) error {
 	for _, c := range containers {
 		c := c
 		eg.Go(func() error {
-			if err := pauseContainer(ctx, client, c.ID()); err != nil {
+			if err := utils.PauseContainer(ctx, client, c.ID()); err != nil {
```
djdongjin, line 70: I think `PauseContainer` should be put in `pkg/containerutil` since it doesn't depend on `cmd` and it's more related to container.

I would suggest within `containerutil` we can create a new `lifecycle.go` to include all high-level operations around a container/containers (such as `Pause`, `Start`, `Stop`, etc). (We don't necessarily check and move all altogether now :) but gradually we should if that sounds good)

https://github.com/containerd/nerdctl/tree/main/pkg/containerutil

cmd/nerdctl/compose/compose_pause.go
```diff
@@ -123,7 +125,7 @@ func composeUnpauseAction(cmd *cobra.Command, args []string) error {
 	for _, c := range containers {
 		c := c
 		eg.Go(func() error {
-			if err := unpauseContainer(ctx, client, c.ID()); err != nil {
+			if err := utils.UnpauseContainer(ctx, client, c.ID()); err != nil {
```
djdongjin, line 128: Similar to `PauseContainer`

cmd/nerdctl/compose/compose_ps.go
```diff
@@ -44,7 +45,7 @@ func newComposePsCommand() *cobra.Command {
 	return composePsCommand
 }
 
-type composeContainerPrintable struct {
+type ContainerPrintable struct {
```
djdongjin, line 48: nit: not sure if the renaming is necessary. In the future if we want to refactor and move these `struct` to a common pkg like `api/types`, this might conflict with the one used in `nerdctl ps`.

cmd/nerdctl/container/container_prune.go
```diff
@@ -58,29 +62,29 @@ func containerPruneAction(cmd *cobra.Command, _ []string) error {
...
 		return err
 	}
 	defer cancel()
 
-	return containerPrune(ctx, cmd, client)
+	return Prune(ctx, cmd, client)
 }
 
-func containerPrune(ctx context.Context, cmd *cobra.Command, client *containerd.Client) error {
+func Prune(ctx context.Context, cmd *cobra.Command, client *containerd.Client) error {
```
djdongjin, line 74: If we only use `cmd.OutOrStdout` within the `Prune`, maybe we can change `cmd *cobra.Command` to `writer io.Writer` and then pass in `cmd.OutOrStdout`.

This way we can abstract the func to `Prune containers and write output to the writer` and decouple from `cmd`. (probably we can also move it to the `pkg/containerutil`)

cmd/nerdctl/container/exec/exec.go
```diff
@@ -28,6 +28,10 @@ import (
 	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/cmd/ctr/commands"
 	"github.com/containerd/containerd/cmd/ctr/commands/tasks"
+	nerdClient "github.com/containerd/nerdctl/cmd/nerdctl/client"
+	"github.com/containerd/nerdctl/cmd/nerdctl/completion"
+	containerUtils "github.com/containerd/nerdctl/cmd/nerdctl/utils/container"
```
djdongjin, line 33: This name is quite similar to https://github.com/containerd/nerdctl/tree/main/pkg/containerutil :)

cmd/nerdctl/container/kill.go
```diff
@@ -26,17 +26,19 @@ import (
...
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
-func newKillCommand() *cobra.Command {
+func NewKillCommand() *cobra.Command {
 	var killCommand = &cobra.Command{
 		Use:               "kill [flags] CONTAINER [CONTAINER, ...]",
-		Short:             "Kill one or more running containers",
+		Short:             "Kill one or more running Containers",
```
djdongjin, line 41: lower case

cmd/nerdctl/container/pause.go
```diff
@@ -14,34 +14,35 @@
...
 	"github.com/spf13/cobra"
 )
 
-func newPauseCommand() *cobra.Command {
+func NewPauseCommand() *cobra.Command {
 	var pauseCommand = &cobra.Command{
 		Use:               "pause [flags] CONTAINER [CONTAINER, ...]",
 		Args:              cobra.MinimumNArgs(1),
-		Short:             "Pause all processes within one or more containers",
+		Short:             "Pause all processes within one or more Containers",
```
djdongjin, line 35: lower case

cmd/nerdctl/container/restart.go
```diff
@@ -63,10 +66,10 @@ func restartAction(cmd *cobra.Command, args []string) error {
 			if found.MatchCount > 1 {
 				return fmt.Errorf("multiple IDs found with provided prefix: %s", found.Req)
 			}
-			if err := stopContainer(ctx, found.Container, timeout); err != nil {
+			if err := action.StopContainer(ctx, found.Container, timeout); err != nil {
 				return err
 			}
-			if err := startContainer(ctx, found.Container, false, client); err != nil {
+			if err := action.StartContainer(ctx, found.Container, false, client); err != nil {
```
djdongjin, line 72: Similar to other `container(s)` lifecycle related func. Putting them in the `pkg(/containerutil)` side might be better.

cmd/nerdctl/container/rm.go
```diff
@@ -0,0 +1,94 @@
...
+				logrus.Error(err)
+			} else {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func rmShellComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
djdongjin, line 91: you have a `completion` pkg right? :)

cmd/nerdctl/container/update.go
```diff
@@ -385,5 +387,5 @@ func copySpec(spec *runtimespec.Spec) (*runtimespec.Spec, error) {
 }
 
 func updateShellComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
djdongjin, line 389: move to completion pkg

djdongjin (commented):
Reviewed till `events.go` :) will review the rest later.

cmd/nerdctl/utils/run/run_runtime.go
```diff
@@ -24,16 +24,16 @@ import (
 	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/plugin"
-	runcoptions "github.com/containerd/containerd/runtime/v2/runc/options"
-	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
+	options2 "github.com/containerd/containerd/runtime/v2/runc/options"
```
djdongjin, line 27: I would prefer `runcoptions`

cmd/nerdctl/utils/run/run_runtime.go
```diff
@@ -24,16 +24,16 @@ import (
...
+	"github.com/opencontainers/runtime-spec/specs-go"
```
djdongjin, line 28: I would prefer `runtimespec`

cmd/nerdctl/utils/run/run_stargz_linux.go
```diff
@@ -0,0 +1,29 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package run
+
+import "github.com/containerd/nerdctl/pkg/testutil"
+
+func RequiresStargz(base *testutil.Base) {
```
djdongjin, line 21: I would prefer putting in `pkg/testutil`

djdongjin (commented):
Hi @Zheaoli thanks for the refactor/improvements. First please feel free to just reply my comments and keep the change your way since different people has different refactor preference. :)

Some high level comments:

1. The PR is too huge :) some files are relocated, while some files are splitted into different packages/files. Sometimes it's difficult to review / build a metal model on which piece goes where.
2. (just suggestion) probably we can do it iteratively, first do file relocation and only change necessary internal/export to struct/funcs. then we can move func/struct accordingly to better places.
3. It seems we are exporting almost all functions/structs now. Probably we should have some criteria on which types of func/struct we should export.

cmd/nerdctl/volume/volume_prune.go
```diff
@@ -14,18 +14,20 @@
...
-package main
+package volume
 
 import (
 	"context"
 	"fmt"
 	"strings"
 
 	"github.com/containerd/containerd"
+	nerdClient "github.com/containerd/nerdctl/cmd/nerdctl/client"
```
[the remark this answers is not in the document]
AkihiroSuda: Yes, this has to be all lower chars

AkihiroSuda:
> The PR is too huge

Yes, but this is unavoidable 🤓

.cirrus.yml
```diff
@@ -38,4 +38,4 @@ windows_task:
     - powershell hack/configure-windows-ci.ps1
     - refreshenv
     - go install .\cmd\nerdctl\
-    - go test -v ./cmd/...
+    - go test -v ./integration/...
```
AkihiroSuda, line 41: The "Test suite" section in `README.md` has to be updated too
https://github.com/containerd/nerdctl/tree/v1.1.0#test-suite

cmd/nerdctl/ps/ps.go
```diff
@@ -14,12 +14,11 @@
    limitations under the License.
 */
 
-package main
+package ps
```
AkihiroSuda, line 17: Should be `cmd/nerdctl/container/ls`

cmd/nerdctl/load/load.go
```diff
@@ -14,24 +14,20 @@
    limitations under the License.
 */
 
-package main
+package load
```
AkihiroSuda, line 17: Should be `cmd/nerdctl/image/load`

cmd/nerdctl/top/top.go
```diff
@@ -0,0 +1,99 @@
...
+/*
+   Portions from:
+   - https://github.com/moby/moby/blob/v20.10.6/api/types/container/container_top.go
+   - https://github.com/moby/moby/blob/v20.10.6/daemon/top_unix.go
+   Copyright (C) The Moby authors.
+   Licensed under the Apache License, Version 2.0
+   NOTICE: https://github.com/moby/moby/blob/v20.10.6/NOTICE
+*/
+
+package top
```
AkihiroSuda, line 26: Should be `cmd/nerdctl/container/top`

djdongjin:
@Zheaoli no worries at all, please do take some rest and hope you feel better soon!

And thanks for initiating the refactor. We can continue the work after you're fully recovered. Please feel no rush :)

yuchanns:
@Zheaoli Get some rest. Wish you a speedy recovery.

AkihiroSuda:
@Zheaoli Hope you will get well soon, and thank you so much for your hard work!

cmd/nerdctl/push/push.go
```diff
@@ -38,7 +41,7 @@ import (
 	estargzconvert "github.com/containerd/stargz-snapshotter/nativeconverter/estargz"
 	httpapi "github.com/ipfs/go-ipfs-http-client"
 	"github.com/multiformats/go-multiaddr"
-	digest "github.com/opencontainers/go-digest"
+	"github.com/opencontainers/go-digest"
```
fahedouch, line 44: why ?

fahedouch:
thanks for you effort @Zheaoli , it seems like the biggest nerdctl Pull Request 😎.

I suggest to merge  `cmd/utils` into `pkg/utils` (e.g  `pkg/utils/cmdutil`) to avoid confusing and simplify project structure 

I think this PR should take more time then expected, this may introduce a lot of conflicts. So, I agree with @djdongjin we can do it iteratively to avoid extra work.

get well soon 😃

cmd/nerdctl/container/container_prune.go
```diff
@@ -58,29 +62,29 @@ func containerPruneAction(cmd *cobra.Command, _ []string) error {
...
 	var deleted []string
 	for _, container := range containers {
-		err = removeContainer(ctx, cmd, container, false, true)
+		err = containerUtils.RemoveContainer(ctx, cmd, container, false, true)
 		if err == nil {
 			deleted = append(deleted, container.ID())
 			continue
 		}
-		if errors.As(err, &statusError{}) {
+		if errors.As(err, &common.StatusError{}) {
```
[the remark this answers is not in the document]
djdongjin: > I am wondering if we should merge cmd/utils and pkg/util*, WDYT ?

I think it might be okay to have a minimum/lightweight `cmd/utils` that only relates to cobra cli or flag process. But I agree that most task/command specific logics should be on the `pkg` side. (No necessarily done altogether in one PR though :)).

AkihiroSuda:
A bunch of functions seems misplaced in wrong files. (See my follow-up commits above)

I'm trying to fix them, but if I fail to do that by the end of the day, I'll prioritize merging other PRs.
But I fear that this PR is almost unrebaseable after merging other PRs...

AkihiroSuda:
And I also don't want to split the integration test package from the command packages

cmd/nerdctl/utils/fmtutil/fmtutil.go
```diff
@@ -45,13 +45,13 @@ func formatLabels(labels map[string]string) string {
...
-// formatSlice formats the slice with `--format` flag.
+// FormatSlice formats the slice with `--format` flag.
 //
 // --format="" (default): JSON
 // --format='{{json .}}': JSON lines
 //
-// formatSlice is expected to be only used for `nerdctl OBJECT inspect` commands.
-func formatSlice(cmd *cobra.Command, x []interface{}) error {
+// FormatSlice is expected to be only used for `nerdctl OBJECT inspect` commands.
+func FormatSlice(cmd *cobra.Command, x []interface{}) error {
```
ktock, line 54: If the purpose of the function is formatting rather than printing, I think it should receive io.Writer rather than *cobra.Command.

ktock (commented):
> I'm trying to fix them, but if I fail to do that by the end of the day, I'll prioritize merging other PRs.
> But I fear that basically this PR isn't rebaseable after merging other PRs...

Thank you for the following up. Possible plan B might be incrementally separating out cobra-independent functions into `pkg` directory but I'm not sure we need to do so because codes under `cmd/nerdctl` are tightly coupled.

djdongjin:
> I'm trying to fix them, but if I fail to do that by the end of the day, I'll prioritize merging other PRs.
But I fear that basically this PR isn't rebaseable after merging other PRs...

Probably we can prioritize bugfix PRs which usually happens in `pkg` side, and wait a little bit on feature PRs?

Zheaoli:
> Sorry, I have to admit that I'm unable to review/carry this amount of PR 🙇‍♂️ 
> 
> 
> 
> > Possible plan B might be incrementally separating out cobra-independent functions into `pkg` directory but I'm not sure we need to do so because codes under `cmd/nerdctl` are tightly coupled.
> 
> 
> 
> Yes, this should be done in a series of small PRs if we want to continue this.
> 
> e.g.,
> 
> - PR 1: move cobra-independent functions into `pkg/...`
> 
> - PR 2: split other common functions into somewhere like `cmd/nerdctl/cmdpkg/...`
> 
> - PR 3: split the client cmdpkg
> 
> - PR 4: split `nerdctl namespace`
> 
> - PR 5: split `nerdctl apparmor`
> 
> - PR 6: split `nerdctl builder`
> 
> - PR 7: ...
> 
> 
> 
> The git diff has to be kept minimal for ease of reviewing and potential rebasing.
> 
> It mustn't contain unnecessary changes such as exposing private functions.
> 
> Some private functions could be exposed, but it has to be done in a separate PR.
> 
> 
> 
> Also, there shouldn't be the `common` package that is likely to result in being the home of random mutually irrelevant functions; the `common` package has to be split to multiple small cmdpkgs with clear purposes.
> 
> 
> 
> - - -
> 
> 
> 
> We also have to revisit whether we really have to split `cmd/nerdctl` in the first place ( https://github.com/containerd/nerdctl/issues/1631#issuecomment-1354329334 ), as the functions in `cmd/nerdctl` aren't really designed to be imported from other projects.
> 
> There might be some functions that can be safely imported from third parties, but they should be rather decoupled from the CLI and should be moved to `pkg/...` .
> 
> 
> 
> Then there might be no good reason to split `cmd/nerdctl` ?
> 
> I expected that splitting `cmd/nerdctl` will increase the code readability as a side effect, but improper splitting will rather decrease the readability, so it has to be done seriously carefully.
> 
> 
> 
> @Zheaoli 
> 
> Thanks anyway for your hard work, and hope you will get well soon! :+1:
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 

I'm sorry for this huge PR, very sorry. It's my bad, the effort scope is beyond my think when I start this process.

I will close this PR first, and I will make a refactor proposal to design the refactor step.

I will close this PR first.

AkihiroSuda:
Not "damn", don't blame your self ❤️

PR #1642, opened 2022-12-12
Fix: Support NewTerminal & Detached mode (nerdctl -t -d ...)

yzxiu:
Support NewTerminal & Detached mode(nerdctl -t -d ...)

```bash
# 1
nerdctal run -t -d ...

# 2
nerdctl create -t ...
nerdctl start

# 3
nerdctl create -t ...
nerdctl start -a ...
```


AkihiroSuda:
Thanks, but how to test?

yzxiu:
> Thanks, but how to test?

This requires building a container that only works properly with -t -d modes
I have tested it locally, and I will add some integration test

yzxiu:
@AkihiroSuda PTAL

yzxiu:
@fahedouch @djdongjin PTAL

cmd/nerdctl/create_linux_test.go
```diff
@@ -106,3 +110,47 @@ func TestCreateWithMACAddress(t *testing.T) {
 		}
 	}
 }
+
+func TestCreateWithTerminal(t *testing.T) {
+
+	if runtime.GOOS == "windows" {
+		t.Skip("buildkit is not enabled on windows, this feature may work on windows.")
+	}
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 119: Why incompatible?

cmd/nerdctl/run_test.go
```diff
@@ -449,3 +449,76 @@ COPY --from=builder /go/src/logger/logger /
 	assert.Check(t, strings.Contains(log, "foo"))
 	assert.Check(t, strings.Contains(log, "bar"))
 }
+
+// this container only works properly with -t -d (or -t) modes
+const TerminalContainerDockerfile = `
+FROM golang:latest as builder
```
AkihiroSuda, line 455: Please pin the golang version and go.sum.

Also, can we avoid using golang here?

cmd/nerdctl/run_test.go
```diff
@@ -449,3 +449,76 @@ COPY --from=builder /go/src/logger/logger /
...
+			con.Write([]byte(time.Now().String() + "\\n")) \n\
+			time.Sleep(time.Second) \n\
+		} \n\
+	} \n' >> app.go
+
+RUN go mod init
+RUN go mod tidy
+RUN go build .
+
+FROM alpine:3.13
```
AkihiroSuda, line 482: Why 3.13?

cmd/nerdctl/run_test.go
```diff
@@ -449,3 +449,76 @@ COPY --from=builder /go/src/logger/logger /
...
+COPY --from=builder /go/src/app/app /root
+RUN chmod +x /root/app
+CMD ["/root/app"]
+	`
+
+func TestNewTerminalDetachedMode(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("buildkit is not enabled on windows, this feature may work on windows.")
+	}
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 492: Why incompatible?

AkihiroSuda:
Let me merge https://github.com/containerd/nerdctl/pull/1639 before this, sorry

cmd/nerdctl/run_test.go
```diff
@@ -449,3 +449,76 @@ COPY --from=builder /go/src/logger/logger /
...
+
+	// with -t, success
+	base.Cmd("run", "-d", "-t", "--name", containerName, imageName).AssertOK()
+	base.Cmd("ps", "-a").AssertOutWithFunc(func(stdout string) error {
+		if !strings.Contains(stdout, "Up") {
+			return fmt.Errorf("terminal container run fail with -t")
+		}
+		return nil
+	})
+	base.Cmd("container", "rm", "-f", containerName).AssertOK()
```
djdongjin, line 523: This line can be a `defer` and moved up to before or after line 506, so that the `rm -f containerName` is always called.

Right now if either 507 or 517 fails, the container won't be cleaned.

yzxiu: Here are actually two little tests, with "-t" and without "-t", so the container must be deleted before running again. 
Perhaps use different name, or test separately?

cmd/nerdctl/run_test.go
yzxiu: > Please pin the golang version and go.sum.
> 
> Also, can we avoid using golang here?


We can package the container in advance and just provide the container name, what do you think?

cmd/nerdctl/run_test.go
djdongjin: I think you can use different names and `defer` both `rm -f containerName` calls.

Same for the test in [cmd/nerdctl/create_linux_test.go](https://github.com/containerd/nerdctl/pull/1642/files#diff-fcc7da15472e79a3ccbd7e3c71d107994cc6d7b5f3efec8b5fe50e38ba36eae1)

cmd/nerdctl/run_test.go
yzxiu: just reference  `testutil_linux.go`
https://github.com/containerd/nerdctl/blob/18944bc70784dfa83010d37054d75487a58ab581/pkg/testutil/testutil_linux.go#L27

cmd/nerdctl/run_test.go
yzxiu: > I think you can use different names and `defer` both `rm -f containerName` calls.
> 
> Same for the test in [cmd/nerdctl/create_linux_test.go](https://github.com/containerd/nerdctl/pull/1642/files#diff-fcc7da15472e79a3ccbd7e3c71d107994cc6d7b5f3efec8b5fe50e38ba36eae1)

Great, thanks

cmd/nerdctl/run_test.go
AkihiroSuda: I'd prefer to minimize dependency on third party images unless it is unavoidable

yzxiu: I thought of an easier way, just run the container directly (bash or /bin/sh), this requires -t to run properly
No need to build image anymore

cmd/nerdctl/run_test.go
```diff
@@ -449,3 +449,31 @@ COPY --from=builder /go/src/logger/logger /
 	assert.Check(t, strings.Contains(log, "foo"))
 	assert.Check(t, strings.Contains(log, "bar"))
 }
+
+func TestNewTerminalDetachedMode(t *testing.T) {
```
djdongjin, line 453: This test failed on windows.

Also suggest to include `Run` in the func name, something like `TestRunWithTtyAndDetached`.

yzxiu: > This test failed on windows. Not sure if it's due to some issue or not supported on windows.
> 

it seems that the error location is different every time

yzxiu: > This test failed on windows. Not sure if it's due to some issue or not supported on windows.
> 


using `-d` will use the json-file log driver and cause the windows test  fail

cmd/nerdctl/create_linux_test.go
```diff
@@ -106,3 +108,31 @@ func TestCreateWithMACAddress(t *testing.T) {
...
+	assert.Equal(base.T, false, withoutTerminalContainer.State.Running)
+
+	// with -t
+	base.Cmd("create", "-t", "--name", withTerminalContainerName, imageName).AssertOK()
+	// start success
+	base.Cmd("start", withTerminalContainerName).AssertOK()
+	defer base.Cmd("container", "rm", "-f", withTerminalContainerName).AssertOK()
+	withTerminalContainer := base.InspectContainer(withTerminalContainerName)
+	assert.Equal(base.T, "running", withTerminalContainer.State.Status)
+	assert.Equal(base.T, true, withTerminalContainer.State.Running)
```
AkihiroSuda, line 137: The test has to verify that tty is allocated when `-t` is specified.

Probably you can just test `stty`

```console
$ docker run -d --name without-t alpine stty
f6ee65e70c512ddba7846f595f3dc5ce0cdb72110e9f901d1661e05f0a8a1372
$ docker logs without-t 
stty: standard input: Not a tty
$ docker container inspect without-t  | jq .[0].State.ExitCode
1
```

```console
$ docker run -td --name with-t alpine stty
5b7bf5c0b3b07af1ccbe2c049aa34841ea635e10c0b493e056201e16dd38de03
$ docker logs with-t 
speed 38400 baud; line = 0;
intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>; eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R; werase = ^W; lnext = ^V; flush = ^O; min = 1; time = 0;
-brkint -imaxbel
$ docker container inspect with-t  | jq .[0].State.ExitCode
0
```

yzxiu: > The test has to verify that tty is allocated when `-t` is specified.
> 
> Probably you can just test `stty`
> 
> ```
> $ docker run -d --name without-t alpine stty
> f6ee65e70c512ddba7846f595f3dc5ce0cdb72110e9f901d1661e05f0a8a1372
> $ docker logs without-t 
> stty: standard input: Not a tty
> $ docker container inspect without-t  | jq .[0].State.ExitCode
> 1
> ```
> 
> ```
> $ docker run -td --name with-t alpine stty
> 5b7bf5c0b3b07af1ccbe2c049aa34841ea635e10c0b493e056201e16dd38de03
> $ docker logs with-t 
> speed 38400 baud; line = 0;
> intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>; eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R; werase = ^W; lnext = ^V; flush = ^O; min = 1; time = 0;
> -brkint -imaxbel
> $ docker container inspect with-t  | jq .[0].State.ExitCode
> 0
> ```

done

cmd/nerdctl/create_linux_test.go
yzxiu: > Why incompatible?

using `-d` will use the json-file log driver and cause the windows test fail

AkihiroSuda (approved):
Thanks

PR #1660, opened 2022-12-15
Fix compose dependency order

### LINKED ISSUE #1658: 'compose up' should start the containers in the dependency order
### Description

'compose up' does not start the containers in the dependency order. Same to 'compose restart'. While 'compose down' stop the containers in the reverse dependency order which is as expected.

1. Download harbor
2. Run 'nerdctl compose up'
3. Run 'nerdctl compose down'

You will see 'compose down' stop the containers one by one in the reverse dependency order, but 'compose up' does not follow any dependency order.

```
$ nerdctl compose up -d
... ...
INFO[0000] Creating container nginx
INFO[0000] Creating container registry
INFO[0000] Creating container registryctl
INFO[0000] Creating container harbor-portal
INFO[0000] Creating container harbor-db
INFO[0000] Creating container redis
INFO[0000] Creating container harbor-core
INFO[0000] Creating container harbor-jobservice

$ nerdctl compose down
INFO[0000] Removing container nginx
INFO[0000] Removing container harbor-jobservice
INFO[0001] Removing container harbor-portal
INFO[0001] Removing container harbor-core
INFO[0002] Removing container redis
INFO[0003] Removing container registry
INFO[0003] Removing container harbor-db
INFO[0004] Removing container registryctl
INFO[0004] Removing network harbor_harbor
INFO[0004] Removing network harbor_default
```

Related code is https://github.com/containerd/nerdctl/blob/v1.1.0/pkg/composer/up_service.go#L48-L72. It starts all containers in go routines in parallel.

I'd expect the containers are started or restarted in the dependency order defined in docker-compose.yaml.

Refer to https://docs.docker.com/compose/startup-order/
`You can control the order of service startup and shutdown with the [depends_on](https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on) option. Compose always starts and stops containers in dependency order, where dependencies are determined by depends_on, links, volumes_from, and network_mode: "service:..."`

1.1.0

None

### Host information

_No response_

djdongjin:
Fix #1658

In restart, the order is in-dependency instead of reverse dependency order.

In up, we should use `errorgroup` _per service_ so the service dependency is ensured.

```shell
# in dependency order: svc2 -> svc1 -> svc0
$ cat docker-compose.yaml
version: '3.1'

services:
  svc0:
    image: alpine
    command: "sleep infinity"
    depends_on:
    - "svc1"
    - "svc2"
  svc1:
    image: alpine
    command: "sleep infinity"
    depends_on:
    - "svc2"
  svc2:
    image: alpine
    command: "sleep infinity"

$ nerdctl compose up -d
INFO[0000] Creating network compdep_default
INFO[0000] Ensuring image alpine
INFO[0000] Ensuring image alpine
INFO[0000] Ensuring image alpine
INFO[0000] Creating container compdep_svc2_1
INFO[0000] Creating container compdep_svc1_1
INFO[0000] Creating container compdep_svc0_1

$ nerdctl compose restart
INFO[0000] Restarting container compdep_svc2_1
INFO[0010] Restarting container compdep_svc1_1
INFO[0020] Restarting container compdep_svc0_1
```

### DIFF
--- a/pkg/composer/up.go
+++ b/pkg/composer/up.go
@@ -70,8 +70,7 @@ func (c *Composer) Up(ctx context.Context, uo UpOptions, services []string) erro
 	var parsedServices []*serviceparser.Service
 	// use WithServices to sort the services in dependency order
 	if err := c.project.WithServices(services, func(svc types.ServiceConfig) error {
-		replicas, ok := uo.Scale[svc.Name]
-		if ok {
+		if replicas, ok := uo.Scale[svc.Name]; ok {
 			if svc.Deploy == nil {
 				svc.Deploy = &types.DeployConfig{}
 			}
--- a/pkg/composer/restart.go
+++ b/pkg/composer/restart.go
@@ -21,9 +21,9 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/compose-spec/compose-go/types"
 	"github.com/containerd/containerd"
 	"github.com/containerd/nerdctl/pkg/labels"
-	"github.com/containerd/nerdctl/pkg/strutil"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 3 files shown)


pkg/composer/up.go
```diff
@@ -70,8 +70,7 @@ func (c *Composer) Up(ctx context.Context, uo UpOptions, services []string) erro
 	var parsedServices []*serviceparser.Service
 	// use WithServices to sort the services in dependency order
 	if err := c.project.WithServices(services, func(svc types.ServiceConfig) error {
-		replicas, ok := uo.Scale[svc.Name]
-		if ok {
+		if replicas, ok := uo.Scale[svc.Name]; ok {
```
AkihiroSuda, line 73: Cosmetic changes should be probably in a separate commit

AkihiroSuda:
Thanks, the PR has no merge conflict but could you rebase to test this with the current main branch?

pkg/composer/up.go
djdongjin: put in a separate commit

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Maybe we should have an integration test too

djdongjin:
> Maybe we should have an integration test too.

Yeah it makes sense. We should have related integration tests for commands with order assumptions (either in-order or reverse-order).

How about I create an issue to track this first and a follow up PR to add those tests for all these command? Adding test in this same PR also works for me :)

djdongjin:
https://github.com/containerd/nerdctl/issues/1661

PR #1664, opened 2022-12-17
Add RequireExecutable to testutil

djdongjin:
`nydus-image` is required for nydus image convert ([link](https://github.com/containerd/nerdctl/blob/main/docs/nydus.md#build-nydus-image-using-nerdctl-image-convert)).

This PR skips nydus convert test when binary is not there, otherwise the test fails. Also simplify other similar checks (cosign, journalctl) in tests.


AkihiroSuda (approved):
Thanks

PR #1668, opened 2022-12-19
Add compose create

djdongjin:
This PR adds `compose create` command and its flags: https://docs.docker.com/engine/reference/commandline/compose_create/

There are some follow-up TODOs/refactors which I think we can track in issues/follow-up PRs because they're not trivial and involves several dependent commands and pkgs. Including them in the same PR might make it too large/confusing.

- [ ] some compose commands have logic in `cmd` instead of `pkg` (also due to other dependency being in `cmd`), making it hard to reuse.
- [ ] current `compose up` implementation should be visited/decoupled:
  - ideally, `up` should equal to `create` + `start` (logic are resuable, `up` can utilize `created` containers)
  - at least we should decouple and reuse some code from `up` in `create` (`run` as well)
- [ ] (related) refactor compose `up|create|start` (possibly `compose run` as well)

This should be the last compose command 😄 (before some refactor happens). After this, there are two remaining commands:

- compose scale: depcrated in v2, I assume we no longer need this?
- compose events: to make it work needs other compose commands releasing relevant events (e.g., service container create/stop/start/etc). it might be easier to implement after we refactor the code base and simplifies some of logic.

### DIFF
--- a/pkg/composer/create.go
+++ b/pkg/composer/create.go
@@ -1,0 +1,217 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package composer
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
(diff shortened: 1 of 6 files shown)


pkg/composer/create.go
```diff
@@ -0,0 +1,217 @@
...
+	"strings"
+
+	"github.com/compose-spec/compose-go/types"
+	"github.com/containerd/nerdctl/pkg/composer/serviceparser"
+	"github.com/containerd/nerdctl/pkg/labels"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+)
+
+// FYI: https://github.com/docker/compose/blob/v2/pkg/api/api.go#L423
```
AkihiroSuda, line 33:
```suggestion
// FYI: https://github.com/docker/compose/blob/v2.14.1/pkg/api/api.go#L423
```

for permanence.

Same applies to other comment lines too

djdongjin: fixed

AkihiroSuda (approved):
Thanks

PR #1669, opened 2022-12-19
Clean up boilerplate code in test

djdongjin:
Some tests have duplicated handler functions to check if a test command output contains (any|all of) strings.

Maybe we can do some cleanup to avoid those duplications. If a test fails, the whole output (and what is the expected strings) will be printed. I think these should be sufficient as logs/debug info in the case of test failure?


AkihiroSuda (approved):
Thanks

PR #1183, opened 2022-07-03
add update restart policy

ningmingxiao:

PR #1682, opened 2022-12-24
[Carry 1183] Add restart flag in update command

djdongjin:
Carry changes from https://github.com/containerd/nerdctl/pull/1183 and make some refactor and fix a bug. Keep it 2 commits so @ningmingxiao 's change is not override :)


cmd/nerdctl/update.go
```diff
@@ -333,15 +337,21 @@ func updateContainer(ctx context.Context, client *containerd.Client, id string,
 
 	if err := updateContainerSpec(ctx, container, spec); err != nil {
 		log.G(ctx).WithError(err).Errorf("Failed to update spec %+v for container %q", spec, id)
+		// reset spec on error.
+		if err := updateContainerSpec(ctx, container, oldSpec); err != nil {
+			log.G(ctx).WithError(err).Errorf("Failed to update spec %+v for container %q", oldSpec, id)
+		}
 	}
```
djdongjin, line 344: not sure if it's necessary to `defer` the `reset to old spec` logic. And the `retErr` is not used anywhere. So change to normal execution.

AkihiroSuda (approved):
Thanks

PR #1686, opened 2022-12-26
feat: generate accessed files list for workload

sctb512:
We can use `fanotify` to generate an in-order accessed files list for workload. This function is useful for workload analysis and image optimization.

We can set `--fanotify` flag to enable this feature.

- Set `--wait-time` flag to specify the recording time.
```shell=
nerdctl run --rm --net host --fanotify --wait-time 30 nginx
```
- Set `--wait-line` to specify a string line that fanotify wait for.
```shell=
nerdctl run --rm --net host --fanotify --wait-line "ready for start up" nginx
```
The task will be killed when getting the expected string line from the standard output.
- Set `--wait-signal` flag to wait for SIGINT signal.
```shell=
nerdctl run --rm --net host --fanotify --wait-signal nginx
```
The task will be killed when getting `SIGINT` (ctrl + C) signal.
- Set `--output` flag to specify the output path of fanotify.

If we do not set the `--output` flag, the default value is `accessed_files.out`.

This PR references the implementation of [stargz-snapshotter](https://github.com/containerd/stargz-snapshotter) and relates to https://github.com/containerd/nydus-snapshotter/pull/228.

Meanwhile, This PR is related to #1687.

Signed-off-by: Bin Tang <tangbin.bin@bytedance.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,6 +573,13 @@ Verify flags:
 - :nerd_face: `--verify`: Verify the image (none|cosign). See [`docs/cosign.md`](./docs/cosign.md) for details.
 - :nerd_face: `--cosign-key`: Path to the public key file, KMS, URI or Kubernetes Secret for `--verify=cosign`
 
+Fanotify flags:
+- :nerd_face: `--fanotify`: Enable fanotify to record accessed files when task running.
+- :nerd_face: `--wait-time`: Available when set `--fanotify` flag. The run time of the `fanotify server` to record accessed files.
+- :nerd_face: `--wait-line`: Available when set `--fanotify` flag. The fanotify server runs until it detaches the string line from the container standard output.
+- :nerd_face: `--wait-signal`: Available when set `--fanotify` flag. The fanotify server runs until it gets the SIGINT (ctrl + C) signal.
+- :nerd_face: `--output`: The output path for fanotify.
+
 Unimplemented `docker run` flags:
     `--attach`, `--blkio-weight-device`, `--cgroup-parent`, `--cpu-rt-*`, `--detach-keys`, `--device-*`,
     `--disable-content-trust`, `--domainname`, `--expose`, `--health-*`, `--ip6`, `--isolation`, `--no-healthcheck`,
--- a/cmd/nerdctl/fanotify.go
+++ b/cmd/nerdctl/fanotify.go
@@ -1,0 +1,34 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newFanotifyCommand() *cobra.Command {
+	var execCommand = &cobra.Command{
+		Use:           "fanotify TARGET",
+		Args:          cobra.MinimumNArgs(1),
+		Short:         "Run fanotify server, DO NOT EXECUTE MANUALLY",
+		Run:           runFanotifyAction,
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		Hidden:        true,
+	}
+	return execCommand
+}
--- a/cmd/nerdctl/fanotify_freebsd.go
+++ b/cmd/nerdctl/fanotify_freebsd.go
@@ -1,0 +1,26 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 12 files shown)


AkihiroSuda:
Can we rather use `/etc/containers/oci/hooks.d/*.json` ?
https://manpages.ubuntu.com/manpages/impish/man5/oci-hooks.5.html

README.md
```diff
@@ -573,6 +573,13 @@ Verify flags:
 - :nerd_face: `--verify`: Verify the image (none|cosign). See [`docs/cosign.md`](./docs/cosign.md) for details.
 - :nerd_face: `--cosign-key`: Path to the public key file, KMS, URI or Kubernetes Secret for `--verify=cosign`
 
+Fanotify flags:
+- :nerd_face: `--fanotify`: Enable fanotify to record accessed files when task running.
+- :nerd_face: `--wait-time`: Available when set `--fanotify` flag. The run time of the `fanotify server` to record accessed files.
+- :nerd_face: `--wait-line`: Available when set `--fanotify` flag. The fanotify server runs until it detaches the string line from the container standard output.
+- :nerd_face: `--wait-signal`: Available when set `--fanotify` flag. The fanotify server runs until it gets the SIGINT (ctrl + C) signal.
+- :nerd_face: `--output`: The output path for fanotify.
+
```
AkihiroSuda, line 582: The flag set seem really confusing to me.
So I suggest using external OCI hooks as I commented above, but if this is not implementable as an OCI hook, at least the flag should be a CSV like `--record-file-access driver=fanotify,wait-time=30s,wait-line="foo",wait-signal=SIGINT,output=/path/to/out`, and has to be marked as experimental.

AkihiroSuda (changes requested):
Please see comments above

README.md
sctb512: Yes, the flag should be replaced with a CSV like format. I will try it with OCI Hook first.

cmd/nerdctl/fanotify.go
```diff
@@ -0,0 +1,34 @@
...
+*/
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newFanotifyCommand() *cobra.Command {
+	var execCommand = &cobra.Command{
```
djdongjin, line 24: `execCommand` -> `fanotifyCommand`

cmd/nerdctl/fanotify_freebsd.go
```diff
@@ -0,0 +1,26 @@
...
+package main
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func runFanotifyAction(cmd *cobra.Command, args []string) {
+	logrus.Warnf("fanotify is unavailable in freebsd platform")
+}
```
djdongjin, line 26: If it's only available in linux, maybe just create a `fanotify_linux.go` instead of having 4 files. As an example:

https://github.com/containerd/nerdctl/blob/ac25d353a947e8ee7f46e156754acae6c0c8bf04/cmd/nerdctl/apparmor_load_linux.go

pkg/fanotify/fanotify_linux.go
```diff
@@ -0,0 +1,440 @@
...
+	accessedFiles      []string
+	persistentPath     string
+	WaitSignal         bool
+	WaitLine           string
+	WaitTime           time.Duration
+	fanotifierClosed   bool
+	fanotifierClosedMu sync.Mutex
+}
+
+func GenerateFanotifyOpts(cmd *cobra.Command, flagT bool) (*Context, error) {
```
djdongjin, line 58: Can we create a `FanotifyOptions` struct here and construct it from the `cmd` in `cmd/nerdctl/fanotify.go` and pass it to `pkg`, instead of passing a cobra `cmd` to `pkg` side.

djdongjin: Also the flag validation logic should be on `cmd` side as well. 

In general, we should read cobra flags and do necessary validation on `cmd` side, (and also construct the `Option` struct from those flags. Then the main logic should be in `pkg` side, using the passed `Option` Struct.

cmd/nerdctl/fanotify.go
djdongjin: `fanotify` seems more like a flag for `run` command than a standalone command. Can we make it part of the run command. E.g., creating something like `run_fanotify_linux.go` and put fanotify related logic here, similar to other parts of run, (run_security_linux.go, run_stargz_linux.go, etc)

pkg/fanotify/fanotify_linux.go
sctb512: Yes, thanks for your explanation. I will refactor the code.

cmd/nerdctl/run.go
```diff
@@ -321,71 +330,68 @@ func runAction(cmd *cobra.Command, args []string) error {
...
-	}
-	var statusC <-chan containerd.ExitStatus
-	if !flagD {
-		defer func() {
-			if rm {
-				if _, taskDeleteErr := task.Delete(ctx); taskDeleteErr != nil {
-					logrus.Error(taskDeleteErr)
-				}
+	if taskCtx.Task == nil {
+		if flagT && !flagD {
```
fahedouch, line 338: Fanotify cannot play with stdio when setting the current console ?

README.md
fahedouch: To make it with hook, you may rely on the `internal` subcommand to trigger the `fanotify` workflow, we already implemented some hooks [here]( https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/run.go#L938)

sctb512:
We are implementing this with OCI Hooks. Thanks for @AkihiroSuda's advice. It's time to close this PR.

sctb512:
By the way, are there any plans to support loading the OCI Hooks configuration from files? I have found an `internal` subcommand `oci-hook` in which `createRuntime` and `postStop` Hooks are available.

sctb512:
Thanks for all comments.

Issue #1680, opened 2022-12-23
Let's refactor the nerdctl CLI package


Zheaoli:
### What is the problem you're trying to solve

OK, I'm back!

For now, the cmd package has a lot of go files mixed. It takes a lot of work to keep maintainer to maintain this project.

And in the community, people may need to add some features of their own, like #1631, so it's time to refactor the command package!

### Describe the solution you'd like

I have made a previous PR, #1639; it's a nasty case for the community.

We may need to make some changes to make this issue happen.

I want to split this issue into three stages.

## First: we need to decompose the flagging process in the command entry point

For now,  we mix the flagging register and flagging process, like https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/exec.go#L53-L62

It means we can not split the common logic unrelated to `cobra.Command`, and care about the flag value from the current cmd package.

We can take a look at some projects in the community alike https://github.com/prometheus/prometheus/blob/main/cmd/prometheus/main.go#L127-L159 and 
https://github.com/containers/podman/blob/main/cmd/podman/containers/attach.go#L45-L54, we will find out that we may collect the flag value into a centralize struct, so we do not need care about the `cobra.Command ` in the entry point in real.

## Second: split the common function into another package

For now, the cmd package maybe can be grouped into three types.

1. helper function, like https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/client.go
2. shellCompletion , like https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/client.go
3.  the logic function, mixed with the flag process logic, like https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/apparmor_ls_linux.go#L56

In this stage, we need to do three things.

1. make the helper function out of the cmd package
2. make the shellCompletion together
3.  after stage1, we can split the logic function into the `pkg/` (for now, we don't need care about the cobra action in the logic function now!)

## Third: restructed the cmd package by the subcommand

And now, we can restruct the cmd package by subcommand. The code in the cmd package has already been cleaned and will focus on the logic to process the flag. So the restruct process will be easy.

 And the final problem in this stage the integration test would be a problem. For most circumstances, the integration test may be written with the command. But here's some in issue in `go test`, the `go test` command will run the test in different directory parallel,  so there would be some potential conflict in this circumstance. I would prefer to make all the integration tests into a single directory together.

AkihiroSuda:
[...]

PR #1773, opened 2022-12-27
Refactor the volume ls command flagging process

miles170:
See #1680 

- [x] Create a file in pkg/api/types/${cmd}_types.go, and define the CommandOption for this command
- [x] Create some file in pkg/api/cmd/${cmd}, and move the command entry point in real into this package

Please let me know if I'm doing it incorrectly.

### DIFF
--- a/cmd/nerdctl/volume.go
+++ b/cmd/nerdctl/volume.go
@@ -17,7 +17,7 @@
 package main
 
 import (
-	"github.com/containerd/nerdctl/pkg/clientutil"
+	"github.com/containerd/nerdctl/pkg/cmd/volume"
 	"github.com/containerd/nerdctl/pkg/mountutil/volumestore"
 	"github.com/spf13/cobra"
 )
@@ -56,9 +56,5 @@ func getVolumeStore(cmd *cobra.Command) (volumestore.VolumeStore, error) {
 	if err != nil {
 		return nil, err
 	}
-	dataStore, err := clientutil.DataStore(dataRoot, address)
-	if err != nil {
-		return nil, err
-	}
-	return volumestore.New(dataStore, ns)
+	return volume.Store(ns, dataRoot, address)
 }
--- a/cmd/nerdctl/volume_ls.go
+++ b/cmd/nerdctl/volume_ls.go
@@ -17,18 +17,9 @@
 package main
 
 import (
-	"bytes"
-	"errors"
-	"fmt"
-	"strconv"
-	"strings"
-	"text/tabwriter"
-	"text/template"
-
-	"github.com/containerd/containerd/pkg/progress"
-	"github.com/containerd/nerdctl/pkg/formatter"
+	types "github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/cmd/volume"
 	"github.com/containerd/nerdctl/pkg/inspecttypes/native"
-	"github.com/sirupsen/logrus"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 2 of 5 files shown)


docs/reference/api/volume_ls_types.md
```diff
@@ -0,0 +1,8 @@
+# VolumeLsOptions
+
+| Field     | CLI flag        | Description                                                                             |
+| --------- | --------------- | --------------------------------------------------------------------------------------- |
+| `Quiet`   | `--quiet`,`-q`  | Only display volume names                                                               |
+| `Format`  | `--format`      | Format the output using the given go template                                           |
+| `Size`    | `--size`,`-s`   | Display the disk usage of volumes. Can be slow with volumes having loads of directories |
+| `Filters` | `--filter`,`-f` | Filter matches volumes based on given conditions                                        |
```
AkihiroSuda, line 8: Isn't godoc enough?

cc @Zheaoli

pkg/api/cmd/volume_ls.go
```diff
@@ -0,0 +1,36 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cmd
+
+import (
+	"github.com/containerd/nerdctl/pkg/api/types"
+
+	"github.com/spf13/cobra"
```
AkihiroSuda, line 22: `pkg/...` is expected to be isolated from `spf13/cobra`

miles170: I cannot really understand what "Create some file in pkg/api/cmd/${cmd}, and move the command entry point in real into this package" means. Would you mind give more specific guidance?

docs/reference/api/volume_ls_types.md
Zheaoli: I'm unsure if we need to describe the struct field.  If we need to describe the struct field, I think godoc is not enough.

pkg/api/cmd/volume_ls.go
Zheaoli: https://github.com/containerd/nerdctl/pull/1774/files

@miles170 I have made a example PR, this may help you

miles170:
I would like to refactor all volume flagging processes, which might take one or two days.

Zheaoli:
This PR should be merged after #1774

Zheaoli:
Since #1779 has been merged, I think this PR is ok to merge after resolve the code conflict

miles170:
Shouldn't we add a prefix to the struct inside `apparmor_types.go`?
I renamed the `LsCommandOptions` to `VolumeLsCommandOptions` to avoid conflict.

Zheaoli:
> Shouldn't we add a prefix to the struct inside `apparmor_types.go`?
> 
> I renamed the `LsCommandOptions` to `VolumeLsCommandOptions` to avoid conflict.

Nice catch, I will add the prefix

Zheaoli (approved):
LGTM on CI green

pkg/cmd/volume/ls.go
```diff
@@ -0,0 +1,230 @@
...
+			fmt.Fprintf(w, "%s\t%s\n", p.Name, p.Mountpoint)
+		}
+	}
+	if f, ok := w.(formatter.Flusher); ok {
+		return f.Flush()
+	}
+	return nil
+}
+
+func GetVolumes(ns string, dataRoot string, address string, volumeSize bool) (map[string]native.Volume, error) {
```
AkihiroSuda, line 131:
```suggestion
func Volumes(ns string, dataRoot string, address string, volumeSize bool) (map[string]native.Volume, error) {
```

https://go.dev/doc/effective_go#Getters

cmd/nerdctl/volume_ls.go
```diff
@@ -54,220 +45,63 @@ func newVolumeLsCommand() *cobra.Command {
...
-
 func volumeLsAction(cmd *cobra.Command, args []string) error {
+	options := &types.VolumeLsCommandOptions{}
+	options.Writer = cmd.OutOrStdout()
 	quiet, err := cmd.Flags().GetBool("quiet")
 	if err != nil {
 		return err
 	}
-	volumeSize, err := cmd.Flags().GetBool("size")
+	options.Quiet = quiet
```
djdongjin, line 55: nit: instead of storing in local variables everytime, you can do

```golang
options.Quiet, err := cmd.Flags().GetBool("quiet")
...
```

same for other options.

djdongjin: another style I recommand is first storing all flags in local variables and at the end, create the option struct

```golang
format, err := cmd.Flags().GetBool("format")
...
quiet,  err := cmd.Flags().GetBool("quiet")
...

option := types.VolumeLsCommandOptions{
    Format: format,
    Quiet: quiet,
}
```

cmd/nerdctl/volume.go
```diff
@@ -56,9 +56,5 @@ func getVolumeStore(cmd *cobra.Command) (volumestore.VolumeStore, error) {
 	if err != nil {
 		return nil, err
 	}
-	dataStore, err := clientutil.DataStore(dataRoot, address)
-	if err != nil {
-		return nil, err
-	}
-	return volumestore.New(dataStore, ns)
+	return volume.Store(ns, dataRoot, address)
```
djdongjin, line 59: seems this `getVolumeStore` is no longer needed as `volume.Store` should be preferred.

cmd/nerdctl/volume_ls.go
```diff
@@ -54,220 +45,63 @@ func newVolumeLsCommand() *cobra.Command {
...
-	}
-	if f, ok := w.(formatter.Flusher); ok {
-		return f.Flush()
-	}
-	return nil
+	options.Address = address
+	return volume.Ls(options)
 }
 
 func getVolumes(cmd *cobra.Command) (map[string]native.Volume, error) {
```
djdongjin, line 89: is this `getVolumes` func still being used somewhere? This func seems no longer necessary since it only read 4 flags from cmd and call another func.

If it's not common enough (e.g., used a lot by other commands as well), then we should remove this helper and instead prefer calling `volume.Volumes` directly.

djdongjin (changes requested):

AkihiroSuda (approved):
Thanks

PR #1775, opened 2022-12-27
Use current console when attaching a tty container

yzxiu:
n: aliased to sudo nerdctl

before update:
Can't get terminal window size correctly, like this:
image

after update:
Can get terminal window size correctly, like this:
image

### DIFF
--- a/cmd/nerdctl/start.go
+++ b/cmd/nerdctl/start.go
@@ -24,8 +24,10 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/containerd/console"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cmd/ctr/commands"
+	"github.com/containerd/containerd/cmd/ctr/commands/tasks"
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/nerdctl/pkg/formatter"
 	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
@@ -119,6 +121,14 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
 		return err
 	}
 	flagT := process.Process.Terminal
+	var con console.Console
+	if flagA && flagT {
+		con = console.Current()
+		defer con.Reset()
+		if err := con.SetRaw(); err != nil {
+			return err
+		}
+	}
 
 	logURI := lab[labels.LogURI]
 
@@ -135,7 +145,7 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
 			logrus.WithError(err).Debug("failed to delete old task")
 		}
 	}
-	task, err := taskutil.NewTask(ctx, client, container, flagA, false, flagT, true, nil, logURI)
+	task, err := taskutil.NewTask(ctx, client, container, flagA, false, flagT, true, con, logURI)
 	if err != nil {
 		return err
 	}
@@ -154,6 +164,11 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
 
 	if !flagA {
 		return nil
+	}
+	if flagA && flagT {
+		if err := tasks.HandleConsoleResize(ctx, task, con); err != nil {
+			logrus.WithError(err).Error("console resize")
+		}
 	}
 
 	sigc := commands.ForwardAllSignals(ctx, task)
--- a/pkg/taskutil/taskutil.go
+++ b/pkg/taskutil/taskutil.go
@@ -42,9 +42,9 @@ func NewTask(ctx context.Context, client *containerd.Client, container container
 	if flagA {
 		logrus.Debug("attaching output instead of using the log-uri")
 		if flagT {
-			ioCreator = cio.NewCreator(cio.WithStreams(os.Stdin, os.Stdout, os.Stderr), cio.WithTerminal)
+			ioCreator = cio.NewCreator(cio.WithStreams(con, con, nil), cio.WithTerminal)
 		} else {
-			ioCreator = cio.NewCreator(cio.WithStreams(os.Stdin, os.Stdout, os.Stderr))
+			ioCreator = cio.NewCreator(cio.WithStdio)
 		}
 
 	} else if flagT && flagD {


AkihiroSuda:
How to test this?

yzxiu:
> How to test this?

Hello, I have added information, and I have a related question

Currently `start -a` doesn't behave quite the same as docker
run with docker
```bash
 ~ docker create -t --name demo busybox
d5e37e58b51acdb3e5785da4623b53de100f65e6ff95b00ee47d3176ca8242bf
 ~ docker start -a(or -a -i) demo                
/ # ps -ef
_
```
didn't get any output.


run with nerdctl
```bash
 ~ n create -t --name demo busybox
3c0de6ccf8db105f5f9e96f9bd77ede409d09db6d125e495083223677a5a5978
 ~ n start -a demo                
/ # ps -ef
PID   USER     TIME  COMMAND
    1 root      0:00 sh
    7 root      0:00 ps -ef
/ # 
```

yzxiu:
> > How to test this?
> 
> Hello, I have added information, and I have a related question
> 
> Currently `start -a` doesn't behave quite the same as docker run with docker
> 
> ```
>  ~ docker create -t --name demo busybox
> d5e37e58b51acdb3e5785da4623b53de100f65e6ff95b00ee47d3176ca8242bf
>  ~ docker start -a(or -a -i) demo                
> / # ps -ef
> _
> ```
> 
> didn't get any output.
> 
> run with nerdctl (n: aliased to sudo nerdctl)
> 
> ```
>  ~ n create -t --name demo busybox
> 3c0de6ccf8db105f5f9e96f9bd77ede409d09db6d125e495083223677a5a5978
>  ~ n start -a demo                
> / # ps -ef
> PID   USER     TIME  COMMAND
>     1 root      0:00 sh
>     7 root      0:00 ps -ef
> / # 
> ```
> 
> can get the output I haven't read the relevant code of `docker start -a`, but I think the way nerdctl works is more reasonable

Please ignore this question, I found the answer,
When using create -it everything works fine

```bash
[xiu-notebook] ~ docker create -it --name demo busybox
a722dafb2cec68ec4e6696bce85768890a22ab753a147ae3f40f02ad12d42b28
[xiu-notebook] ~ docker start -a -i demo              
/ # ps -ef
PID   USER     TIME  COMMAND
    1 root      0:00 sh
    7 root      0:00 ps -ef
/ #
```

yzxiu:
nerdctl currently forwards input and output when use `start -a`,
It seems that the `-i` flag needs to be added to keep similar with docker

This question involves 
`nerdctl run -itd`
`nerdctl create -it`
`nerdctl start -a -i`

pkg/taskutil/taskutil.go
```diff
@@ -42,9 +42,12 @@ func NewTask(ctx context.Context, client *containerd.Client, container container
 	if flagA {
 		logrus.Debug("attaching output instead of using the log-uri")
 		if flagT {
-			ioCreator = cio.NewCreator(cio.WithStreams(os.Stdin, os.Stdout, os.Stderr), cio.WithTerminal)
+			if con == nil {
+				return nil, errors.New("got nil con when start a container with tty")
```
fahedouch, line 46: `con`  [cannot be nil](https://github.com/containerd/console/blob/v1.0.3/console.go#L77 ), one of the standard std streams (os.Stdin, os.Stdout or  os.Stderr) should always be a console otherwise program will panic

fahedouch: cc @AkihiroSuda

yzxiu: There seems to be no need to use `con` for input/output here，using stdio directly seems to be similar with docker.
`con` seems to be only used to resize the  terminal window

cmd/nerdctl/start.go
```diff
@@ -155,6 +165,11 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
 	if !flagA {
 		return nil
 	}
+	if flagA && flagT {
+		if err := tasks.HandleConsoleResize(ctx, task, con); err != nil {
```
fahedouch, line 169: Why new `if flagA && flagT`, you already have the same condition [above](https://github.com/containerd/nerdctl/pull/1775/files#diff-0cc8445057142657a3937f86dc488b5283257fa971cbd4ac7546affa47fa1d53R125)

fahedouch: My bad we need to catch a `task` before

pkg/taskutil/taskutil.go
fahedouch: you are using `con` here to implement the container stdio : `cio.WithStreams(con, con, nil)`, am I missing something ? `con` is a physical implementation of the terminal =>  So `con` is a terminal with a standard stdio. That means using `con` here you are setting `tty` to the standard stdio (stdout , stderr, stdin) which is good regarding the `-a` flag (you algo is good for me). My problem is with  `con == nil`; looking at [console.Current() code](https://github.com/containerd/console/blob/v1.0.3/console.go#L65-L77) the `con` can never be nil so this condition is useless.

yzxiu: > There seems no need to use `con` for input/output here，using stdio seems to be similar with docker. `con` only used to resize the terminal window

@fahedouch 
sorry i didn't make it clear，What I am talking about here is this pr https://github.com/containerd/nerdctl/pull/1777

fahedouch:
Sorry for the delayed reply, I was off the last week ( and I am starting a new job :) ) 

@yzxiu are you still planning to add some integration test ?

yzxiu:
> Sorry for the delayed reply, I was off the last week ( and I am starting a new job :) )
> 
> @yzxiu are you still planning to add some integration test ?

Hi, It's the end of the year, and I'm a little busy these days.
Because this takes effect when the window size changes, I don't know how to do it in ntegration test.

fahedouch (approved):
Thanks

cmd/nerdctl/start.go
```diff
@@ -119,6 +121,14 @@ func startContainer(ctx context.Context, container containerd.Container, flagA b
 		return err
 	}
 	flagT := process.Process.Terminal
+	var con console.Console
+	if flagA && flagT {
```
fahedouch, line 125: cc @yzxiu 
```suggestion
	if flagT {
```

fahedouch: if `process.Process.Terminal` is true, `nerdctl start demo` will be allocated a term (con)

yzxiu: if `start` without `-a`, after nerdctl starts the container, it exits directly.
https://github.com/containerd/nerdctl/blob/40d341dac245a70a6ee4ac43567f2d2818fe3167/cmd/nerdctl/start.go#L160-L162
`con` doesn't do anything here

yzxiu: @AkihiroSuda cc

fahedouch:
> @yzxiu in v1.2.0 nerdctl attach still not use current console window size(console width). It good for you inspect more on this, Thanks

you have to set both `-a` and `-t` to use current console size.

fahedouch:
> > > @yzxiu in v1.2.0 nerdctl attach still not use current console window size(console width). It good for you inspect more on this, Thanks
> > 
> > 
> > you have to set both `-a` and `-t` to use current console size.
> 
> O Thanks, I got the '-a' option for "nerdctl start". But I still have console size issue when i use "nerdctl exec" (which have no -a arg)

could pleas open a issue ticker for this. thanks

PR #623, opened 2021-12-15
fix can't save image by image id

ningmingxiao:
fix can't save image by image id

PR #1776, opened 2022-12-27
[Carry 623] Fix can't save image by image id

djdongjin:
Carry changes from #623 and also make some refactor to `nerdctl save`.

1. Make `nerdctl save` work when image ids are given; also works if the same image has multiple names (only one copy is saved); return error if an id prefix corresponds to multiple different images.
2. Refactor `nerdctl save`: make `saveImages` not depend on cobra.

### DIFF
--- a/cmd/nerdctl/save_linux_test.go
+++ b/cmd/nerdctl/save_linux_test.go
@@ -48,23 +48,6 @@ func TestSave(t *testing.T) {
 	assert.Assert(t, strings.Contains(etcOSRelease, "Alpine"))
 }
 
-func TestSaveById(t *testing.T) {
-	base := testutil.NewBase(t)
-	base.Cmd("pull", testutil.AlpineImage).AssertOK()
-	inspect := base.InspectImage(testutil.AlpineImage)
-	var id string
-	if testutil.GetTarget() == testutil.Docker {
-		id = inspect.ID
-	} else {
-		id = strings.Split(inspect.RepoDigests[0], ":")[1]
-	}
-	archiveTarPath := filepath.Join(t.TempDir(), "id.tar")
-	base.Cmd("save", "-o", archiveTarPath, id).AssertOK()
-	base.Cmd("rmi", "-f", testutil.AlpineImage).AssertOK()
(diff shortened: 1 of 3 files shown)


cmd/nerdctl/save_linux_test.go
```diff
@@ -48,6 +48,43 @@ func TestSave(t *testing.T) {
 	assert.Assert(t, strings.Contains(etcOSRelease, "Alpine"))
 }
 
+func TestSaveById(t *testing.T) {
+	base := testutil.NewBase(t)
+	base.Cmd("pull", testutil.AlpineImage).AssertOK()
+	inspect := base.InspectImage(testutil.AlpineImage)
```
AkihiroSuda, line 54:
```suggestion
	inspect := base.InspectImage(testutil.Commonmage)
```

And the test can be moved to `save_test.go`

AkihiroSuda (approved):
Thanks

cmd/nerdctl/save_linux_test.go
```diff
@@ -33,17 +33,17 @@ import (
...
 	err := extractDockerArchive(archiveTarPath, rootfsPath)
 	assert.NilError(t, err)
 	etcOSReleasePath := filepath.Join(rootfsPath, "/etc/os-release")
 	etcOSReleaseBytes, err := os.ReadFile(etcOSReleasePath)
 	assert.NilError(t, err)
 	etcOSRelease := string(etcOSReleaseBytes)
-	t.Logf("read %q, extracted from %q", etcOSReleasePath, testutil.AlpineImage)
+	t.Logf("read %q, extracted from %q", etcOSReleasePath, testutil.CommonImage)
 	t.Log(etcOSRelease)
 	assert.Assert(t, strings.Contains(etcOSRelease, "Alpine"))
```
AkihiroSuda, line 48: The test is specific to Alpine, so you can retain `testutil.AlpineImage` for this test

djdongjin: fixed

cmd/nerdctl/save_linux_test.go
djdongjin: fixed

Issue #1685, opened 2022-12-26
`nerdctl run -t` is not greppable (terminal output is not connected to stdout)


AkihiroSuda:
Docker (working as expected):
```console
$ docker run -t --rm alpine sh -c 'echo foo; echo bar' | grep foo
foo
```

nerdctl 1.1:
```console
$ nerdctl run -t --rm alpine sh -c 'echo foo; echo bar' | grep foo
foo
bar
```

Looks like the terminal output (`-t`) is not connected to the stdout.

fahedouch:
closed by https://github.com/containerd/nerdctl/pull/2167

PR #1777, opened 2022-12-28
terminal output connected to stdout

yzxiu:
fixing https://github.com/containerd/nerdctl/issues/1685

When `nerdctl run`, I changed the output from `con` to `os.Stdout`.
Executing the `nerdctl run -t --rm alpine sh -c 'echo foo; echo bar' | grep foo` command locally can get the correct result,
but this makes a lot of tests fail

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -337,7 +337,7 @@ func runAction(cmd *cobra.Command, args []string) error {
 	}
 	logURI := lab[labels.LogURI]
 
-	task, err := taskutil.NewTask(ctx, client, container, false, flagI, flagT, flagD, con, logURI)
+	task, err := taskutil.NewTask(ctx, client, container, false, flagI, flagT, flagD, logURI)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


yzxiu:
When `nerdctl run`, I changed the output from `con` to `os.Stdout`.
Executing the `nerdctl run -t --rm alpine sh -c 'echo foo; echo bar' | grep foo` command locally can get the correct result,
but this makes a lot of tests fail

fahedouch:
`nerdctl compose` is allocating a new `Tty` for the underlying command `nerdctl run`. so the stdout of `nerdctl run` is not attached to the current console (that of `nerdctl compose`) . I tried to attach the underlying command std streams to `console.Current()` but it fails  :/ : 

[pseudo code source](https://github.com/containerd/nerdctl/blob/main/pkg/composer/up_service.go#L168-L171) 
```
	if service.Unparsed.StdinOpen {
		cmd.Stdin = os.Stdin
	}
	
	if !container.Detached {
		cmd.Stdout = console.Current()
		cmd.Stderr = console.Current()
	}
```

cc @AkihiroSuda

PR #1779, opened 2022-12-29
[Refactor]: Move fmtutil from cmd to pkg/formatter

Zheaoli:
Signed-off-by: Zheao.Li <me@manjusaka.me>

PR #1780, opened 2022-12-29
[Refactor] Move containerd client helper from cmd to pkg/clientutil

Zheaoli:

### DIFF
--- a/cmd/nerdctl/build.go
+++ b/cmd/nerdctl/build.go
@@ -31,6 +31,7 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	dockerreference "github.com/containerd/containerd/reference/docker"
 	"github.com/containerd/nerdctl/pkg/buildkitutil"
+	"github.com/containerd/nerdctl/pkg/clientutil"
 	"github.com/containerd/nerdctl/pkg/defaults"
 	"github.com/containerd/nerdctl/pkg/platformutil"
 	"github.com/containerd/nerdctl/pkg/strutil"
@@ -194,10 +195,17 @@ func buildAction(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
-
+	namespace, err := cmd.Flags().GetString("namespace")
+	if err != nil {
+		return err
+	}
+	address, err := cmd.Flags().GetString("address")
+	if err != nil {
+		return err
+	}
+	client, ctx, cancel, err := clientutil.NewClient(cmd.Context(), namespace, address)
 	if len(tags) > 1 {
 		logrus.Debug("Found more than 1 tag")
-		client, ctx, cancel, err := newClient(cmd)
 		if err != nil {
 			return fmt.Errorf("unable to tag images: %s", err)
 		}
@@ -241,8 +249,16 @@ func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, arg
 	if err != nil {
 		return "", nil, false, "", nil, nil, err
 	}
+	namespace, err := cmd.Flags().GetString("namespace")
+	if err != nil {
+		return "", nil, false, "", nil, nil, err
+	}
+	address, err := cmd.Flags().GetString("address")
+	if err != nil {
+		return "", nil, false, "", nil, nil, err
+	}
+	client, ctx, cancel, err := clientutil.NewClient(cmd.Context(), namespace, address)
 	if output == "" {
-		client, ctx, cancel, err := newClient(cmd)
 		if err != nil {
 			return "", nil, false, "", nil, nil, err
 		}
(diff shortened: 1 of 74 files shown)


djdongjin (commented):
This snippet is duplicated a lot

```golang
	namespace, err := cmd.Flags().GetString("namespace")
	if err != nil {
		return err
	}
	address, err := cmd.Flags().GetString("address")
	if err != nil {
		return err
	}
	client, ctx, cancel, err := clientutil.NewClient(cmd.Context(), namespace, address)
```

I would suggest that we have a `cmd/nerdctl/cmdutil.go` which contains this kind of code if they're very common (e.g., read some flags from cmd and call the corresponding pkg functions). So in this case:

```golang
// cmd/nerdctl/cmdutil.go, ignore types....
func newClientFromCmd(cmd) client, ctx, cancel, err {
	namespace, err := cmd.Flags().GetString("namespace")
	if err != nil {
		return err
	}
	address, err := cmd.Flags().GetString("address")
	if err != nil {
		return err
	}
	return clientutil.NewClient(cmd.Context(), namespace, address)
}
```

Personally, I think it still make sense to have 2 functions (newClientFromCmd, and your clientutil.NewClient) compared to the old newClient which does two jobs.

Zheaoli:
> I would suggest that we have a `cmd/nerdctl/cmdutil.go` which contains this kind of code if they're very common (e.g., read some flags from cmd and call the corresponding pkg functions). So in this case:

This shouldn't be a problem. after the #1680 is completed, all the global flagging processes will share a command helper function `getGlobalFlag` (like namespace, address etc)

I want to make the `NewClient` out of the cmd package because most of the command logic we want to split into pkg/cmd packages is related to this function. In another worlds, if we want to split the command logic out of the cmd package, we need to make `NewClient` out of cmd package first.

Like @AkihiroSuda has said in https://github.com/containerd/nerdctl/pull/1779#discussion_r1058717829, the code in `pkg/**` should not be related to `cobra`, so IMHO, I think this PR is necessary

djdongjin:
:) Overall the PR LGTM and I agree both the refactor and the PR is necessary.

My suggestion is more a nit without much context, which can be seen as "add a 10 line helper in `cmd` side to dedup some duplication code".

> Like @AkihiroSuda has said in https://github.com/containerd/nerdctl/pull/1779#discussion_r1058717829, the code in pkg/** should not be related to cobra.

The helper I suggested is also on the `cmd` side (simply read 2 flags from cmd and call your new function).

Just repeat, it's a nit and the PR is not blocked on me 😉 thanks for your work!

pkg/clientutil/client.go
```diff
@@ -83,24 +74,16 @@ func newClientWithPlatform(cmd *cobra.Command, platform string, clientOpts ...co
...
-// getDataStore returns a string like "/var/lib/nerdctl/1935db59".
+// GetDataStore returns a string like "/var/lib/nerdctl/1935db59".
 // "1935db9" is from `$(echo -n "/run/containerd/containerd.sock" | sha256sum | cut -c1-8)`
 // on Windows it will return "%PROGRAMFILES%/nerdctl/1935db59"
-func getDataStore(cmd *cobra.Command) (string, error) {
-	dataRoot, err := cmd.Flags().GetString("data-root")
-	if err != nil {
-		return "", err
-	}
+func GetDataStore(dataRoot, address string) (string, error) {
```
AkihiroSuda, line 83:
```suggestion
func DataStore(dataRoot, address string) (string, error) {
```

https://go.dev/doc/effective_go#Getters

Zheaoli: Done

fahedouch:
I prefer to go through the refacto in small steps even if it introduces unnecessary/redundant code, it would make the process of contributing easier. LGTM to retains redundancy code until progressive  introduction of `getGlobalFlag`

cmd/nerdctl/build.go
```diff
@@ -241,8 +250,16 @@ func generateBuildctlArgs(cmd *cobra.Command, buildkitHost string, platform, arg
...
+		return "", nil, false, "", nil, nil, err
+	}
+	address, err := cmd.Flags().GetString("address")
+	if err != nil {
+		return "", nil, false, "", nil, nil, err
+	}
 	if output == "" {
-		client, ctx, cancel, err := newClient(cmd)
+		client, ctx, cancel, err := clientutil.NewClient(cmd.Context(), namespace, address)
 		if err != nil {
```
fahedouch, line 263: Same as above

Laitr0n:
When I refactor the 'compose log' command, I realized that the 'client helper' should be refactored first. After this PR, can the refactoring of the 'compose xxx' command continue?

AkihiroSuda (approved):
Thanks

PR #1785, opened 2022-12-31
Filter dangling images using image filters.

manugupt1:

### DIFF
--- a/cmd/nerdctl/images_test.go
+++ b/cmd/nerdctl/images_test.go
@@ -116,3 +116,22 @@ LABEL version=0.1`, testutil.CommonImage)
 	base.Cmd("images", "--filter", "reference=busy*:*libc*").AssertOutContains("glibc")
 	base.Cmd("images", "--filter", "reference=busy*:*libc*").AssertOutContains("uclibc")
 }
+
+func TestImagesFilterDangling(t *testing.T) {
+	testutil.RequiresBuild(t)
+	base := testutil.NewBase(t)
+	base.Cmd("images", "prune", "--all").AssertOK()
+
+	dockerfile := fmt.Sprintf(`FROM %s
+CMD ["echo", "nerdctl-build-notag-string"]
+	`, testutil.CommonImage)
+	buildCtx, err := createBuildContext(dockerfile)
+	assert.NilError(t, err)
+
+	defer os.RemoveAll(buildCtx)
+	base.Cmd("build", "-f", buildCtx+"/Dockerfile", buildCtx).AssertOK()
+
+	// dangling image test
+	base.Cmd("images", "--filter", "dangling=true").AssertOutContains("<none>")
+	base.Cmd("images", "--filter", "dangling=false").AssertOutNotContains("<none>")
+}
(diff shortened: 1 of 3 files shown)


cmd/nerdctl/images_test.go
```diff
@@ -116,3 +116,22 @@ LABEL version=0.1`, testutil.CommonImage)
 	base.Cmd("images", "--filter", "reference=busy*:*libc*").AssertOutContains("glibc")
 	base.Cmd("images", "--filter", "reference=busy*:*libc*").AssertOutContains("uclibc")
 }
+
+func TestDanglingImagesFilter(t *testing.T) {
```
AkihiroSuda, line 120:
```suggestion
func TestImagesFilterDangling(t *testing.T) {
```

Similar tests should have a common prefix

pkg/imgutil/imgutil.go
```diff
@@ -383,11 +384,20 @@ func ParseRepoTag(imgName string) (string, string) {
...
+	DanglingTrue
+	DanglingFalse
+)
+
 type Filters struct {
 	Before    []string
 	Since     []string
 	Labels    map[string]string
 	Reference []string
+	Dangling  Dangling
```
AkihiroSuda, line 400:
```suggestion
	Dangling  *bool
```

pkg/imgutil/imgutil.go
```diff
@@ -398,7 +408,15 @@ func ParseFilters(filters []string) (*Filters, error) {
 		case 1:
 			return nil, fmt.Errorf("invalid filter %q", filter)
 		case 2:
-			if tempFilterToken[0] == FilterBeforeType {
+			if tempFilterToken[0] == FilterDanglingType {
+				if strings.EqualFold(tempFilterToken[1], "true") {
```
AkihiroSuda, line 412: Not `EqualFold`

```console
$ docker images -f "dangling=TRUE"
Error response from daemon: invalid filter 'dangling=[TRUE]'
```

manugupt1: I was intentional about it as it still does not break backward compatibility. LMK if you still want me to change it.

pkg/imgutil/imgutil.go
```diff
@@ -369,7 +370,7 @@ func ParseRepoTag(imgName string) (string, string) {
 
 	ref, err := refdocker.ParseDockerRef(imgName)
 	if err != nil {
-		logrus.WithError(err).Warnf("unparsable image name %q", imgName)
+		logrus.WithError(err).Debugf("unparsable image name %q", imgName)
```
AkihiroSuda, line 373: ?

pkg/imgutil/imgutil.go
AkihiroSuda: Docker seems case-sensitive, so nerdctl should be probably case-sensitive too

pkg/imgutil/imgutil.go
manugupt1: Every time there is a <none> image listed out; this Warning is printed out; I think we should not do it; now that we are adding more support for dangling images.

AkihiroSuda (approved):
Thanks

Issue #1425, opened 2022-10-12
`nerdctl compose down` with orphan container should keep the same behavior with docker


stillfox-lee:
### What is the problem you're trying to solve

When we execute `nerdctl compose down`. nerdctl does not check is three any other containers still in the network and just remove it.

### Describe the solution you'd like

While `docker compose down` will return error and not remove the network. We should keep the same behavior with it.

### Additional context

The orphan container [scenario](https://github.com/containerd/nerdctl/issues/1344#issuecomment-1241475088)

Retrospection:
I'd like to try to solve this~

PR #1786, opened 2022-12-31
[WIP]feat: add network in use check to `compose down`

Retrospection:
To solve issue #1425


yuchanns:
Please using a valid DCO Sign Off.

pkg/netutil/netutil.go
```diff
@@ -49,37 +49,7 @@ type CNIEnv struct {
...
-			} else {
-				used[k] = v
-			}
-		}
-	}
-	return used, nil
-}
-
-func namespaceUsedNetworks(ctx context.Context, containers []containerd.Container) (map[string][]string, error) {
+func NamespaceUsedNetworks(ctx context.Context, containers []containerd.Container) (map[string][]string, error) {
```
Zheaoli, line 52: I think you just make `NamespaceUsedNetworks` public here, I think it's necessary to reorder the function

pkg/composer/down.go
```diff
@@ -96,7 +96,18 @@ func (c *Composer) downNetwork(ctx context.Context, shortName string) error {
...
+	}
+
+	if netExists {
+		netUsed, err := c.NetworkUsed(ctx, fullName)
+		if err != nil {
+			return err
+		}
+
+		if netUsed {
+			return fmt.Errorf("failed to remove network %q: Error response from daemon: error while removing network: network %q has active endpoints", fullName, fullName)
```
Zheaoli, line 108: The error message here could be the same with https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/network_rm.go#L76.

Just make the message as simple as we can

cmd/nerdctl/compose.go
```diff
@@ -156,6 +157,38 @@ func getComposer(cmd *cobra.Command, client *containerd.Client) (*composer.Compo
...
+		}
+		used := make([]string, 0)
+		for _, ns := range nsList {
+			nsCtx := namespaces.WithNamespace(ctx, ns)
+			containers, err := client.Containers(nsCtx)
+			if err != nil {
+				return false, err
+			}
+			nsUsedN, err := netutil.NamespaceUsedNetworks(nsCtx, containers)
+			if err != nil {
```
Zheaoli, line 174: I think we can make the code simple here,

you can make the code like this

```go
	usedNetworkInfo, err := netutil.UsedNetworks(ctx, client)
	if err != nil {
		return false, err
	}
        if _,ok:=usedNetworkInfo[netName];ok{
                return true, nil
        }
```

Zheaoli (changes requested):
And we need extra e2e test, and plz sign-off your commit,

pkg/composer/down.go
Retrospection: make sense

> The error message here could be the same with https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/network_rm.go#L76.
> 
> Just make the message as simple as we can

pkg/composer/composer.go
```diff
@@ -41,6 +41,7 @@ type Options struct {
 	EnvFile          string
 	NerdctlCmd       string
 	NerdctlArgs      []string
+	NetworkUsed      func(context.Context, string) (bool, error)
```
djdongjin, line 44: nit: suggest rename to `NetworkInUse` (similarly `netInUse` below)

Retrospection: make sense

PR #1789, opened 2023-01-01
tests: Use AssertOK for deferred cleanup

dcantah:
The standard here seems to be to use `AssertOK` to validate that `nerdctl rm -f` exited successfully for a given test, however there was some stragglers, namely in the cgroup and mount tests. Now these should fail properly if `rm -f` exits with != 0


AkihiroSuda (approved):
Thanks

PR #1793, opened 2023-01-02
[Refactor] Move the Config Struct from cmd/main.go to pkg/config

Zheaoli:
Most of the subcommand is related to this Config struct and global flags. like #1792

So I split it from the cmd package to let us can reuse the definition.

### DIFF
--- a/cmd/nerdctl/main.go
+++ b/cmd/nerdctl/main.go
@@ -25,9 +25,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/defaults"
-	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/nerdctl/pkg/config"
 	ncdefaults "github.com/containerd/nerdctl/pkg/defaults"
 	"github.com/containerd/nerdctl/pkg/logging"
 	"github.com/containerd/nerdctl/pkg/rootlessutil"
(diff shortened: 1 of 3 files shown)


pkg/config/types/config.go
```diff
@@ -0,0 +1,60 @@
+/*
```
AkihiroSuda, line 1: > `pkg/config/types/config.go`

Maybe just `pkg/config/config.go` ?

AkihiroSuda: And then `NewConfig` can be renamed to `New`

Zheaoli: Updated

pkg/api/types/global.go
```diff
@@ -0,0 +1,21 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package types
+
+import "github.com/containerd/nerdctl/pkg/config"
+
+type GlobalCommandOptions config.Config
```
AkihiroSuda, line 21: Do we need this?

pkg/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+	DataRoot         string   `toml:"data_root"`
+	CgroupManager    string   `toml:"cgroup_manager"`
+	InsecureRegistry bool     `toml:"insecure_registry"`
+	HostsDir         []string `toml:"hosts_dir"`
+	Experimental     bool     `toml:"experimental"`
+}
+
+// NewConfig creates a default Config object statically,
+// without interpolating CLI flags, env vars, and toml.
+func NewConfig() *Config {
```
AkihiroSuda, line 45:
```suggestion
func New() *Config {
```

pkg/api/types/global.go
Zheaoli: If we introduce a config object into XXCommandOptions object, it will make people confused.

So, I think using the typing alias may be an excellent way to keep the code semantic.

AkihiroSuda (approved):
Thanks

cmd/nerdctl/main.go
```diff
@@ -134,44 +132,8 @@ func xmain() error {
...
-		CgroupManager:    ncdefaults.CgroupManager(),
-		InsecureRegistry: false,
-		HostsDir:         ncdefaults.HostsDirs(),
-		Experimental:     true,
-	}
-}
-
 func initRootCmdFlags(rootCmd *cobra.Command, tomlPath string) (*pflag.FlagSet, error) {
-	cfg := NewConfig()
+	cfg := config.New()
```
fahedouch, line 136: Is it better to init `GlobalCommandOptions` here and then pass it to `config.New()`
```
cfg := GlobalCommandOptions{...}
config.New(cfg)
```
to keep the code semantic.
Cc @Zheaoli

PR #1795, opened 2023-01-03
Refactor the volume flagging process

miles170:
See https://github.com/containerd/nerdctl/issues/1680

- [x] Create a file in pkg/api/types/${cmd}.go, and define the CommandOption for this command
- [x] Create some file in pkg/cmd/${cmd}, and move the command entry point in real into this package

### DIFF
--- a/cmd/nerdctl/volume_create.go
+++ b/cmd/nerdctl/volume_create.go
@@ -17,10 +17,8 @@
 package main
 
 import (
-	"fmt"
-
-	"github.com/containerd/containerd/identifiers"
-	"github.com/containerd/nerdctl/pkg/strutil"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/cmd/volume"
 
 	"github.com/spf13/cobra"
 )
@@ -43,23 +41,13 @@ func volumeCreateAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	name := args[0]
-	if err := identifiers.Validate(name); err != nil {
-		return fmt.Errorf("malformed name %s: %w", name, err)
-	}
-
-	volStore, err := getVolumeStore(globalOptions)
-	if err != nil {
-		return err
-	}
 	labels, err := cmd.Flags().GetStringArray("label")
 	if err != nil {
 		return err
 	}
-	labels = strutil.DedupeStrSlice(labels)
-	if _, err := volStore.Create(name, labels); err != nil {
-		return err
-	}
-	fmt.Fprintf(cmd.OutOrStdout(), "%s\n", name)
-	return nil
+	return volume.Create(&types.VolumeCreateCommandOptions{
+		GOptions: globalOptions,
+		Name:     args[0],
+		Labels:   labels,
+	}, cmd.OutOrStdout())
 }
--- a/cmd/nerdctl/volume_inspect.go
+++ b/cmd/nerdctl/volume_inspect.go
@@ -17,7 +17,8 @@
 package main
 
 import (
-	"github.com/containerd/nerdctl/pkg/formatter"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/cmd/volume"
 	"github.com/spf13/cobra"
 )
 
@@ -48,26 +49,16 @@ func volumeInspectAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
-	volStore, err := getVolumeStore(globalOptions)
-	if err != nil {
-		return err
-	}
-	result := make([]interface{}, len(args))
-
-	for i, name := range args {
-		var vol, err = volStore.Get(name, volumeSize)
-		if err != nil {
-			return err
-		}
-		result[i] = vol
-	}
 	format, err := cmd.Flags().GetString("format")
 	if err != nil {
 		return err
 	}
-
-	return formatter.FormatSlice(format, cmd.OutOrStdout(), result)
+	return volume.Inspect(&types.VolumeInspectCommandOptions{
+		GOptions: globalOptions,
+		Format:   format,
+		Size:     volumeSize,
+		Volumes:  args,
+	}, cmd.OutOrStdout())
 }
 
 func volumeInspectShellComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
--- a/cmd/nerdctl/volume_ls.go
+++ b/cmd/nerdctl/volume_ls.go
@@ -50,33 +50,29 @@ func volumeLsAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
-	options := &types.VolumeLsCommandOptions{}
-	options.Writer = cmd.OutOrStdout()
 	quiet, err := cmd.Flags().GetBool("quiet")
 	if err != nil {
 		return err
 	}
-	options.Quiet = quiet
 	format, err := cmd.Flags().GetString("format")
 	if err != nil {
 		return err
 	}
-	options.Format = format
 	size, err := cmd.Flags().GetBool("size")
 	if err != nil {
 		return err
 	}
-	options.Size = size
 	filters, err := cmd.Flags().GetStringSlice("filter")
 	if err != nil {
 		return err
 	}
-	options.Filters = filters
-	options.Namespace = globalOptions.Namespace
-	options.DataRoot = globalOptions.DataRoot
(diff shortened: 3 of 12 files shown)


cmd/nerdctl/completion.go
```diff
@@ -144,7 +145,23 @@ func shellCompleteNetworkNames(cmd *cobra.Command, exclude []string) ([]string,
 }
 
 func shellCompleteVolumeNames(cmd *cobra.Command) ([]string, cobra.ShellCompDirective) {
-	vols, err := getVolumes(cmd)
```
AkihiroSuda, line 147: Why?

pkg/api/types/volume_types.go
```diff
@@ -16,11 +16,44 @@
...
-import "io"
+import (
+	"context"
+	"io"
 
-type VolumeLsCommandOptions struct {
+	"github.com/containerd/containerd"
+)
+
+type VolumeCommandCommonOptions struct {
```
AkihiroSuda, line 26:
```suggestion
type VolumeCommandOptions struct {
```

AkihiroSuda:
I think we can retain `getVolumes()` to avoid code clone

miles170:
> I think we can retain `getVolumes()` to avoid code clone

Keep also `getVolumeStore()`?

miles170:
Force-pushed to resolve code conflicts.

miles170:
Force-pushed to resolve code conflicts.

AkihiroSuda (approved):
Thanks

cmd/nerdctl/volume_create.go
```diff
@@ -43,23 +41,20 @@ func volumeCreateAction(cmd *cobra.Command, args []string) error {
...
-	}
-	fmt.Fprintf(cmd.OutOrStdout(), "%s\n", name)
-	return nil
+	return volume.Create(&types.VolumeCreateCommandOptions{
+		VolumeCommandOptions: types.VolumeCommandOptions{
+			GlobalCommandOptions: types.GlobalCommandOptions{
+				Namespace: globalOptions.Namespace,
+				DataRoot:  globalOptions.DataRoot,
+				Address:   globalOptions.Address,
+			},
```
Zheaoli, line 54: I think this should be  `GlobalCommandOptions: *globalOptions`

cmd/nerdctl/volume_inspect.go
```diff
@@ -48,26 +49,23 @@ func volumeInspectAction(cmd *cobra.Command, args []string) error {
...
-	}
 	format, err := cmd.Flags().GetString("format")
 	if err != nil {
 		return err
 	}
-
-	return formatter.FormatSlice(format, cmd.OutOrStdout(), result)
+	return volume.Inspect(&types.VolumeInspectCommandOptions{
+		VolumeCommandOptions: types.VolumeCommandOptions{
+			GlobalCommandOptions: types.GlobalCommandOptions{
```
Zheaoli, line 58: same above

cmd/nerdctl/volume_ls.go
```diff
@@ -50,33 +50,36 @@ func volumeLsAction(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return err
 	}
-	options.Filters = filters
-	options.Namespace = globalOptions.Namespace
-	options.DataRoot = globalOptions.DataRoot
-	options.Address = globalOptions.Address
-	return volume.Ls(options)
+	return volume.Ls(&types.VolumeLsCommandOptions{
+		VolumeCommandOptions: types.VolumeCommandOptions{
```
Zheaoli, line 70: same above

cmd/nerdctl/volume_prune.go
```diff
@@ -67,43 +66,16 @@ func volumePruneAction(cmd *cobra.Command, _ []string) error {
...
-		fmt.Fprintln(cmd.OutOrStdout(), "Deleted Volumes:")
-		for _, name := range removedNames {
-			fmt.Fprintln(cmd.OutOrStdout(), name)
-		}
-		fmt.Fprintln(cmd.OutOrStdout(), "")
-	}
-	return nil
+	return volume.Prune(&types.VolumePruneCommandOptions{
+		VolumeCommandOptions: types.VolumeCommandOptions{
+			GlobalCommandOptions: types.GlobalCommandOptions{
```
Zheaoli, line 71: same above

cmd/nerdctl/volume_rm.go
```diff
@@ -55,68 +49,22 @@ func volumeRmAction(cmd *cobra.Command, args []string) error {
...
-	removedNames, err := volStore.Remove(volumenames)
-	if err != nil {
-		return err
-	}
-	for _, name := range removedNames {
-		fmt.Fprintln(cmd.OutOrStdout(), name)
-	}
-	return err
+	return volume.Rm(&types.VolumeRmCommandOptions{
+		VolumeCommandOptions: types.VolumeCommandOptions{
```
Zheaoli, line 53: same above

pkg/api/types/volume_types.go
```diff
@@ -16,11 +16,39 @@
...
+import (
+	"context"
+	"io"
 
-type VolumeLsCommandOptions struct {
+	"github.com/containerd/containerd"
+)
+
+type VolumeCommandOptions struct {
+	GlobalCommandOptions
```
Zheaoli, line 27: I prefer don't use an anonymous field here. 

Maybe `GOptions *GlobalCommanOptions` would be a good choice

pkg/api/types/volume_types.go
```diff
@@ -29,10 +57,22 @@ type VolumeLsCommandOptions struct {
...
+	// Context is underlying command context
+	Context context.Context
+}
+
+type VolumeRmCommandOptions struct {
+	VolumeCommandOptions
+	// Client is the containerd client
+	Client *containerd.Client
+	// Context is underlying command context
+	Context context.Context
```
Zheaoli, line 75: Putting the context into the struct should not be a good pattern. see discussion here https://github.com/containerd/nerdctl/pull/1792#discussion_r1060810151

pkg/api/types/volume_types.go
```diff
@@ -29,10 +57,22 @@ type VolumeLsCommandOptions struct {
...
+}
+
+type VolumeRmCommandOptions struct {
+	VolumeCommandOptions
+	// Client is the containerd client
+	Client *containerd.Client
```
Zheaoli, line 73: I guess we can init the client in the logic method？

Zheaoli (changes requested):

pkg/api/types/volume_types.go
AkihiroSuda: Either is fine to me, but we have to keep this consistent across every commands

pkg/api/types/volume_types.go
```diff
@@ -16,11 +16,39 @@
...
+)
+
+type VolumeCommandOptions struct {
+	GlobalCommandOptions
 	// Writer is the output writer
 	Writer io.Writer
+}
+
+type VolumeCreateCommandOptions struct {
+	VolumeCommandOptions
```
miles170, line 33: Should I use the anonymous field here?

pkg/api/types/volume_types.go
miles170: Is it preferable to put the `io.Writer` in the arguments rather than inside the struct? See https://github.com/containerd/nerdctl/pull/1792/files#diff-3d28f1d0455b7f054ebac7695bbbe3d9c758b44b12602f7365d404a86011a49eR45.

pkg/api/types/volume_types.go
djdongjin: This reminds me a question. If we're using `GOptions *GlobalCommanOptions` (or pointer in general for other options), we are assuming either:

1. The values in the option struct won't be changed at all, or
2. Changing values won't impact either callers or callees:
    1. If `func A` changes option values and calls `func B`, the callee B won't be impacted (e.g., B expects the changed value);
    2. If `func B` also changes option values (e.g., `B` in a separate go routinue), the caller A won't be impacted.

If we use pointers and use goroutines, we're sharing the data between different goroutines.

pkg/api/types/volume_types.go
Zheaoli: > Is it preferable to put the `io.Writer` in the arguments rather than inside the struct? See https://github.com/containerd/nerdctl/pull/1792/files#diff-3d28f1d0455b7f054ebac7695bbbe3d9c758b44b12602f7365d404a86011a49eR45.

Yes, I think it better

pkg/api/types/volume_types.go
Zheaoli: I think a standard filed could be better

AkihiroSuda (approved):
Thanks

PR #1796, opened 2023-01-03
Add a prefix to apparmor command options to avoid conflict

miles170:
See https://github.com/containerd/nerdctl/pull/1773#issuecomment-1368783993

Add a prefix to the apparmor command options to avoid struct name conflicts.


AkihiroSuda (approved):
Thanks

PR #1791, opened 2023-01-02
Refactor the compose build command flagging process

Laitr0n:
There is no ideal about GlobalFlag(like namespace, address). I just refactor the compose build command.
Signed-off-by: Laitron <meetlq@outlook.com>

PR #1797, opened 2023-01-03
[refactor] Add global flag process helper function

Zheaoli:
I'm so sorry I have to make a huge PR again.

After #1793, we have a `types.GlobalCommandOptions`, and for #1792 and #1791 and many refactor PR int he future, we may need to introduce a new helper function to process the global flag.

In this PR, I introduce the new helper function and use this function to process the global flag in all subcommand.

There may be some duplicated code for some subcommand like `nerdctl run`, I think this should be not a problem, I will merge it when I refactor the run command,

### DIFF
--- a/cmd/nerdctl/flagutil.go
+++ b/cmd/nerdctl/flagutil.go
@@ -1,0 +1,87 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/spf13/cobra"
+)
+
+func processRootCmdFlags(cmd *cobra.Command) (*types.GlobalCommandOptions, error) {
+	debug, err := cmd.Flags().GetBool("debug")
+	if err != nil {
+		return nil, err
+	}
+	debugFull, err := cmd.Flags().GetBool("debug-full")
+	if err != nil {
+		return nil, err
+	}
+	address, err := cmd.Flags().GetString("address")
+	if err != nil {
+		return nil, err
+	}
+	namespace, err := cmd.Flags().GetString("namespace")
+	if err != nil {
+		return nil, err
+	}
+	snapshotter, err := cmd.Flags().GetString("snapshotter")
+	if err != nil {
+		return nil, err
+	}
+	cniPath, err := cmd.Flags().GetString("cni-path")
+	if err != nil {
+		return nil, err
+	}
+	cniConfigPath, err := cmd.Flags().GetString("cni-netconfpath")
+	if err != nil {
+		return nil, err
+	}
+	dataRoot, err := cmd.Flags().GetString("data-root")
+	if err != nil {
+		return nil, err
+	}
+	cgroupManager, err := cmd.Flags().GetString("cgroup-manager")
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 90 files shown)


djdongjin (commented):
(Not request changes)

In the current PR, most commands are refactored into a half-way state (e.g., global flag is processed (sometimes, **repeatedly**) whenever it's needed. Ideally, it should only be processed once at the entry of the command, and let other usage use the processed struct.

As an example, `getVolumeStore` calls `processGlobalFlags`, and `getComposer` calls `getVolumeStore`, which means each compose commands call `processGlobalFlags` at least twice (one in `getVolumeStore`, one in its own command if global flags are used).

Either way, we should highlight that helper functions (e.g., `getVolumeStore`) should be refactored to not call `processGlobalFlags` directly and instead use the struct created by the command.

Zheaoli:
> (Not request changes)
> 
> In the current PR, most commands are refactored into a half-way state (e.g., global flag is processed (sometimes, **repeatedly**) whenever it's needed. Ideally, it should only be processed once at the entry of the command, and let other usage use the processed struct.
> 
> As an example, `getVolumeStore` calls `processGlobalFlags`, and `getComposer` calls `getVolumeStore`, which means each compose commands call `processGlobalFlags` at least twice (one in `getVolumeStore`, one in its own command if global flags are used).
> 
> Either way, we should highlight that helper functions (e.g., `getVolumeStore`) should be refactored to not call `processGlobalFlags` directly and instead use the struct created by the command.

nice catch. I have update all the PR to make sure call processGlobalFlags once for each command

cmd/nerdctl/flagutil.go
```diff
@@ -0,0 +1,87 @@
...
+	options.Namespace = namespace
+	options.Snapshotter = snapshotter
+	options.CNIPath = cniPath
+	options.CNINetConfPath = cniConfigPath
+	options.DataRoot = dataRoot
+	options.CgroupManager = cgroupManager
+	options.InsecureRegistry = insecureRegistry
+	options.HostsDir = hostsDir
+	options.Experimental = experimental
+	return options, nil
```
djdongjin, line 86: nit: I might be too picky but still prefer removing line 25, 74-86 and return the constructed struct directly :). It reduces assignment and (IMHO) cleaner.

```go
return &types.GlobalCommandOptions{
  Debug: debug,
  DebugFull: debugFull,
  ...
  Experimental: experimental, 
}, nil
```

djdongjin (commented):
LGTM aside from the nit. thanks

Zheaoli:
@AkihiroSuda PTAL

djdongjin (approved):
LGTM

cmd/nerdctl/flagutil.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, err
+	}
+	hostsDir, err := cmd.Flags().GetStringSlice("hosts-dir")
+	if err != nil {
+		return nil, err
+	}
+	experimental, err := cmd.Flags().GetBool("experimental")
+	if err != nil {
+		return nil, err
+	}
```
ktock, line 72: Could we add `hosts-dir` and `experimental` flags to README? https://github.com/containerd/nerdctl#global-flags

cmd/nerdctl/flagutil.go
```diff
@@ -0,0 +1,87 @@
...
+*/
+
+package main
+
+import (
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/spf13/cobra"
+)
+
+func processGlobalFlag(cmd *cobra.Command) (*types.GlobalCommandOptions, error) {
```
ktock, line 24: `initRootCmdFlags()` in main.go should be renamed to `initGlobalFlags()` or this function should be renamed to `processRootCmdFlags()` maybe?

cmd/nerdctl/flagutil.go
Zheaoli: I think we can add the docs in another PR.

ktock (approved):

cmd/nerdctl/stop.go
```diff
@@ -49,7 +49,11 @@ func newStopCommand() *cobra.Command {
 }
 
 func stopAction(cmd *cobra.Command, args []string) error {
-	// Time to wait after sending a SIGTERM and before sending a SIGKILL.
+	// Time to wait after sending a SIGTERM and before sending a SIGKILL.]
```
AkihiroSuda, line 52: typo

Issue #1187, opened 2022-07-04
FATA[0000] flag -d and --rm cannot be specified together


james-work-account:
### What is the problem you're trying to solve

When using colima, the command `docker run --rm -d --name mongo -p 27017:27017 mongo:latest` works fine. However I've been required to switch to nerdctl, and the command `lima nerdctl run --rm -d --name mongo -p 27017:27017 mongo:latest` outputs the error:

```
FATA[0000] flag -d and --rm cannot be specified together
```

### Describe the solution you'd like

Ultimately, I would like to repeatedly be able to start mongo through Docker without getting `name "mongo" is already used by ID "XYZ"` errors. Using the same method with Colima and nerdctl give me different results - I am not particularly fluent with Docker stuff so if this isn't the correct way of doing this then I'm open to suggestions!

liubin:
If you are using a shell script, as a workaround, you can delete it first without waiting for nerdtcl to support these two options being specified together :

```bash
nerdctl rm mongo -f || true
nerdctl run -d --name mongo
```

TianZong48:
Is it possible that support `-d` + `--rm` using Hook(https://github.com/opencontainers/runtime-spec/blob/main/specs-go/config.go#L153-L155)?  @AkihiroSuda  @junnplus 
If so, can I try to implement this feature?   😄

manugupt1:
@AkihiroSuda @Zheaoli  I tried this issue a few weeks back, where I got stuck was on how to clean up the resources. The approach that I tried was invoking the nerdctl postStop oci hook but the postStop hook executes before deleting the container which is not what we want today. 

Do you have suggestions?

yuchanns:
> we can execute sh -euc "sleep 3 && nerdctl rm -f <CONTAINER>" & in the postStop hook

That's weird.

How about hanging discontinuous responses until periodically checking has ensured containers are deleted?
```sh
$ nerdctl run --rm -d alpine echo hello world
# 047e8df4994637a8b069e1155df6e55d09326e88f13858ea9571d30ac36d92e8
## hanging periodically check until the container is deleted
## or just fork a daemon to do that
```

AkihiroSuda:
Can we rather just extend the restart monitor?
https://github.com/containerd/containerd/tree/main/runtime/restart

AkihiroSuda:
↑ @junnplus WDYT?

junnplus:
> Can we rather just extend the restart monitor to support removal?
[...]

PR #1817, opened 2023-01-06
Fix the container will be created when `nerdctl run -d --rm`

Iceber:
The error caused by the parameter problem should be returned before the container is created

```bash
$ nerdctl run --rm -d ubuntu:latest ps aux
FATA[0000] flag -d and --rm cannot be specified together

$ nerdctl ps -a
CONTAINER ID    IMAGE                                                                                                                         COMMAND                   CREATED           STATUS     PORTS    NAMES
09d12b1005ba    docker.io/library/ubuntu:latest                                                                                               "ps aux"                  31 seconds ago    Created             ubuntu-09d12

$  ctr containers list
CONTAINER                                                           IMAGE                                                                                                                         RUNTIME         
09d12b1005ba9dbeac8dbfd614f97c79d3a7906befa754a0ebb9feeacd0c0a56    docker.io/library/ubuntu:latest                                                                                               io.containerd.runc.v2
```

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -305,6 +305,15 @@ func runAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	rm, err := cmd.Flags().GetBool("rm")
+	if err != nil {
+		return err
+	}
+
+	if rm && flagD {
+		return errors.New("flags -d and --rm cannot be specified together")
+	}
+
 	container, gc, err := createContainer(ctx, cmd, globalOptions, client, args, platform, flagI, flagT, flagD)
 	if err != nil {
 		if gc != nil {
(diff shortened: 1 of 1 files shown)


dcantah (approved):

djdongjin (approved):

AkihiroSuda (approved):
Thanks

fangn2:
Question not directly related to the PR, docker supports usage of both --rm and -d, shouldn't nerdctl do so too? 
[docker run reference](https://docs.docker.com/engine/reference/run/#detached--d):
```
 If you use -d with --rm, the container is removed when it exits or when the daemon exits, whichever happens first.
```

AkihiroSuda:
> Question not directly related to the PR, docker supports usage of both --rm and -d, shouldn't nerdctl do so too? [docker run reference](https://docs.docker.com/engine/reference/run/#detached--d):
> 
> ```
>  If you use -d with --rm, the container is removed when it exits or when the daemon exits, whichever happens first.
> ```

Yes but difficult
- https://github.com/containerd/nerdctl/issues/1187

fangn2:
@AkihiroSuda  Thanks for the quick response and the pointer.
After went through the details from the link you shared, seems adding removal monitor similar to [restart monitor](https://github.com/containerd/containerd/tree/main/runtime/restart) at containerd side is a feasible solution. The idea is very similar to how docker added the same feature  in [https://github.com/moby/moby/pull/20848](https://github.com/moby/moby/pull/20848).

Iceber:
@fangn2 Unlike docker's logic, the restart monitor is based on a timing loop, with a default interval of 10s, which does not meet the requirements of `rm`.

And docker is based on event to implement, I think the future can consider changing the containerd monitor to event-based

PR #1818, opened 2023-01-06
[Refactor] Refactor the compose command flagging process

Laitr0n:
- [x]  Create a file in pkg/api/types/${cmd}_types.go, and define the CommandOption for this command
- [x] Create some file in pkg/cmd/${cmd}, and move the command entry point in real into this package

### DIFF
--- a/cmd/nerdctl/compose.go
+++ b/cmd/nerdctl/compose.go
@@ -17,24 +17,7 @@
 package main
 
 import (
-	"context"
-	"errors"
-	"fmt"
-
-	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/errdefs"
-	"github.com/containerd/containerd/platforms"
-	"github.com/containerd/nerdctl/pkg/api/types"
 	"github.com/containerd/nerdctl/pkg/composer"
-	"github.com/containerd/nerdctl/pkg/composer/serviceparser"
-	"github.com/containerd/nerdctl/pkg/cosignutil"
-	"github.com/containerd/nerdctl/pkg/imgutil"
-	"github.com/containerd/nerdctl/pkg/ipfs"
-	"github.com/containerd/nerdctl/pkg/netutil"
-	"github.com/containerd/nerdctl/pkg/referenceutil"
-	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/sirupsen/logrus"
-
 	"github.com/spf13/cobra"
 )
 
@@ -81,33 +64,26 @@ func newComposeCommand() *cobra.Command {
 	return composeCommand
 }
 
-func getComposer(cmd *cobra.Command, client *containerd.Client, globalOptions *types.GlobalCommandOptions) (*composer.Composer, error) {
+func getComposeOptions(cmd *cobra.Command, debugFull, experimental bool) (composer.Options, error) {
 	nerdctlCmd, nerdctlArgs := globalFlags(cmd)
 	projectDirectory, err := cmd.Flags().GetString("project-directory")
 	if err != nil {
-		return nil, err
+		return composer.Options{}, err
 	}
 	envFile, err := cmd.Flags().GetString("env-file")
 	if err != nil {
-		return nil, err
+		return composer.Options{}, err
 	}
 	projectName, err := cmd.Flags().GetString("project-name")
 	if err != nil {
-		return nil, err
+		return composer.Options{}, err
 	}
-	debugFull := globalOptions.DebugFull
-	snapshotter := globalOptions.Snapshotter
 	files, err := cmd.Flags().GetStringArray("file")
 	if err != nil {
-		return nil, err
+		return composer.Options{}, err
 	}
-	insecure := globalOptions.InsecureRegistry
-	cniPath := globalOptions.CNIPath
-	cniNetconfpath := globalOptions.CNINetConfPath
-	hostsDirs := globalOptions.HostsDir
-	experimental := globalOptions.Experimental
 
-	o := composer.Options{
+	return composer.Options{
 		Project:          projectName,
 		ProjectDirectory: projectDirectory,
 		ConfigPaths:      files,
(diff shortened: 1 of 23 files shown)


cmd/nerdctl/compose.go
```diff
@@ -82,135 +69,44 @@ func newComposeCommand() *cobra.Command {
...
 
 func getComposer(cmd *cobra.Command, client *containerd.Client, globalOptions *types.GlobalCommandOptions) (*composer.Composer, error) {
-	nerdctlCmd, nerdctlArgs := globalFlags(cmd)
-	projectDirectory, err := cmd.Flags().GetString("project-directory")
+	options, err := getComposeOptions(cmd)
 	if err != nil {
 		return nil, err
 	}
-	envFile, err := cmd.Flags().GetString("env-file")
+	volStore, err := getVolumeStore(globalOptions)
```
djdongjin, line 76: You can directly use `volume.Store(globalOptions.Namespace, globalOptions.DataRoot, globalOptions.Address)` in the `pkg` side so there is no need to get `volStore` in `cmd` side and reduce one argument for `compose.GetComposer`).

Below is the definition of `getVolumeStore` and it only wraps a `volume.Store` call. We should remove this wrapper and use `volume.Store` directly. cc @Zheaoli 

https://github.com/containerd/nerdctl/blob/c71c2884a729d3ff8eec0286e39be70c5562f1e4/cmd/nerdctl/volume.go#L47-L49

cmd/nerdctl/compose.go
```diff
@@ -82,135 +69,44 @@ func newComposeCommand() *cobra.Command {
 }
 
 func getComposer(cmd *cobra.Command, client *containerd.Client, globalOptions *types.GlobalCommandOptions) (*composer.Composer, error) {
```
djdongjin, line 71: After the below suggestion, this `getComposer` becomes a wraper of 2 function calls: `getComposeOptions` in `cmd` side and `GetComposer` in `pkg` side.

Although it seems redundant, I think it's fine for now to keep it to minimize the change. Otherwise, you need to change every `getCompoer` call in compose commands to first call `getComposeOptions` and then call `compose.GetComposer`.

cmd/nerdctl/compose.go
```diff
@@ -82,135 +69,44 @@ func newComposeCommand() *cobra.Command {
...
 		return nil, err
 	}
-	projectName, err := cmd.Flags().GetString("project-name")
+	return compose.GetComposer(options, globalOptions, client, volStore, cmd.OutOrStdout(), cmd.ErrOrStderr())
+}
+
+func getComposeOptions(cmd *cobra.Command) (*composer.Options, error) {
+	options := &composer.Options{}
+	options.NerdctlCmd, options.NerdctlArgs = globalFlags(cmd)
+	var err error
```
djdongjin, line 86: line 86 is unnecessary

pkg/cmd/compose/compose.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/containerd/nerdctl/pkg/ipfs"
+	"github.com/containerd/nerdctl/pkg/mountutil/volumestore"
+	"github.com/containerd/nerdctl/pkg/netutil"
+	"github.com/containerd/nerdctl/pkg/referenceutil"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/sirupsen/logrus"
+)
+
+func GetComposer(options *composer.Options, globalOptions *types.GlobalCommandOptions, client *containerd.Client, volStore volumestore.VolumeStore, stdout, stderr io.Writer) (*composer.Composer, error) {
```
djdongjin, line 42: You can put this function here because most `pkg side compose` logic are here. And we might need some further refactor on `GetComposer` and `composer.New` (`GetComposer` is basically a `New` with `options`).

https://github.com/containerd/nerdctl/blob/49c0e571fec30e98b1265dcd85d15244abea5e19/pkg/composer/composer.go

cmd/nerdctl/compose.go
```diff
@@ -81,53 +82,44 @@ func newComposeCommand() *cobra.Command {
...
-		DebugPrintFull:   debugFull,
-		Experimental:     experimental,
+	options.ConfigPaths, err = cmd.Flags().GetStringArray("file")
+	if err != nil {
+		return nil, err
+	}
+	options.Experimental, err = cmd.Flags().GetBool("experimental")
+	if err != nil {
+		return nil, err
 	}
```
djdongjin, line 112: Sorry for the back and forth, but I think the original format of the first half function should serve us well right? (e.g., read cmd flags, and create `composer.Options` altogether) (so suggest not change the 1st half)

Meanwhile, as you can see in this whole function, it contains two types of code: 

1. cobra related flag process: it's not related to compose implementation details so we should keep it in `cmd` side; 
2. compose implementation details: e.g., those `o.NetworkExists = ..., o.VolumeExists = ...` etc. they are not related to cobra/cmd after we get the option _values_ from the 1st part.

What I meant in my comment (https://github.com/containerd/nerdctl/pull/1791#issuecomment-1370041137) is that we should spit this two parts into two functions (so cobra logic stays in `cmd` and implementation goes to `pkg`): 

1. `getComposeOptions(cmd *cobra.Command) composer.Options`: basically [the 1st half](https://github.com/containerd/nerdctl/blob/c3e6ceb0de52e681f1217d6024af97bedece801d/cmd/nerdctl/compose.go#L85-L119). after this call, `composer.Options` contains all option values needed for the implementation.
2. `GetComposer(opt composer.Options, client *containerd.Client, stdout, stderr io.Writer) (*composer.Composer, error)`: basically [the 2nd half](https://github.com/containerd/nerdctl/blob/c3e6ceb0de52e681f1217d6024af97bedece801d/cmd/nerdctl/compose.go#L121-L215). It first adds (our default) implementation to `option` and then returns a `composer`. If we want to enable callers to provide other implementations via `composeoption`, we should check if the field if nil.

```golang
if o.EnsureImage == nil {
    o.EnsureImage = func(....
}
```

After this change, the workflow for a compose command to obtain a `composer` will be:

```golang
    // get option values from cmd
	options, err := getComposeOptions(cmd, client, globalOptions)
	if err != nil {
		return err
	}
    // get composer
	c, err := composer.GetComposer(options, client, cmd.OutOrStdout(), cmd.ErrOrStdErr())
``` 

Let me know if you have other questions. @Laitr0n 

@AkihiroSuda, @Zheaoli WDYT? (does this make sense :))

djdongjin (changes requested):

cmd/nerdctl/compose.go
```diff
@@ -81,136 +64,33 @@ func newComposeCommand() *cobra.Command {
...
-	o := composer.Options{
-		Project:          projectName,
-		ProjectDirectory: projectDirectory,
-		ConfigPaths:      files,
-		EnvFile:          envFile,
-		NerdctlCmd:       nerdctlCmd,
-		NerdctlArgs:      nerdctlArgs,
-		DebugPrintFull:   debugFull,
-		Experimental:     experimental,
-	}
```
djdongjin, line 119: can we keep this first half not changed (e.g., `return o, nil` directly here)? This part is only syntax changes and we can just do the `split it into 2 functions`. It makes the PR easy to review.

djdongjin: And the return type can be `composer.Options` (i.e., not a pointer)

pkg/cmd/compose/compose.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/containerd/nerdctl/pkg/cosignutil"
+	"github.com/containerd/nerdctl/pkg/imgutil"
+	"github.com/containerd/nerdctl/pkg/ipfs"
+	"github.com/containerd/nerdctl/pkg/netutil"
+	"github.com/containerd/nerdctl/pkg/referenceutil"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/sirupsen/logrus"
+)
+
+func GetComposer(options *composer.Options, client *containerd.Client, globalOptions *types.GlobalCommandOptions, stdout, stderr io.Writer) (*composer.Composer, error) {
```
djdongjin, line 41: The arguments order can be:

`GetComposer(client *containerd.Client, globalOptions *types.GlobalCommandOptions, options composer.Options, stdout, stderr io.Writer)`

Two options should close to each other.

ALso, `composer.Options` doesn't need to be a pointer.

djdongjin (changes requested):

djdongjin (approved):
LGTM on CI green, thanks

pkg/composer/composer.go
```diff
@@ -49,6 +59,101 @@ type Options struct {
 	Experimental     bool // enable experimental features
 }
 
+func GetComposer(client *containerd.Client, globalOptions *types.GlobalCommandOptions, options Options, stdout, stderr io.Writer) (*Composer, error) {
```
AkihiroSuda, line 62:
```suggestion
func New(client *containerd.Client, globalOptions *types.GlobalCommandOptions, options Options, stdout, stderr io.Writer) (*Composer, error) {
```

See also https://go.dev/doc/effective_go#Getters

AkihiroSuda: `New` already exists, so maybe this should be moved to `pkg/cmd/...`

pkg/composer/composer.go
```diff
@@ -21,15 +21,25 @@ import (
...
 	"os/exec"
 
 	composecli "github.com/compose-spec/compose-go/cli"
 	compose "github.com/compose-spec/compose-go/types"
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/identifiers"
+	"github.com/containerd/containerd/platforms"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/cmd/volume"
```
AkihiroSuda, line 34: Probably `pkg/composer` should remain decoupled from `pkg/cmd`

djdongjin (approved):
LGTM, thanks

AkihiroSuda (approved):
Thanks

PR #1820, opened 2023-01-09
[Refactor] Refactor the load command flagging process

Zheaoli:
Checklist:

### DIFF
--- a/cmd/nerdctl/load.go
+++ b/cmd/nerdctl/load.go
@@ -17,19 +17,8 @@
 package main
 
 import (
-	"errors"
-	"fmt"
-	"io"
-	"os"
-
-	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/archive/compression"
-	"github.com/containerd/containerd/images"
-	"github.com/containerd/containerd/images/archive"
-	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/nerdctl/pkg/api/types"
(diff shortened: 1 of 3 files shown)


cmd/nerdctl/load.go
```diff
@@ -56,106 +45,35 @@ func newLoadCommand() *cobra.Command {
...
+		return types.LoadCommandOptions{}, err
 	}
-	platMC, err := platformutil.NewMatchComparer(allPlatforms, platform)
-	if err != nil {
-		return err
-	}
-
-	return loadImage(decompressor, cmd, globalOptions, platMC, false)
+	return types.LoadCommandOptions{
+		GOptions:     *globalOptions,
```
djdongjin, line 66: no change required here, but we might gradually change the pointer usage of global option to non-pointer to avoid unintented shared memory among goroutines (https://github.com/containerd/nerdctl/pull/1795#discussion_r1064256900) and also be consistent with other command option usages.

1. Those global options in function signatures should be fine, as they should be gradually replaced by individual command options (with global option being a field).
2. `processRootCmdFlags` should be changed to return a struct not a struct pointer. Similarly, SubCommandOptions should hold a global option struct (as in `LoadCommandOptions`) not a struct pointer.

I guess the original intent of using global option as a pointer is to avoid unnecessary copies, but I think it should be fine as those option structs are quite small, and also pointers use heap allocation which means more GC work.

AkihiroSuda (approved):
Thanks

PR #1824, opened 2023-01-10
[Refactor] Refactor all the GlobalCommandOptions from pointer to value reference

Zheaoli:
Based on the discussion https://github.com/containerd/nerdctl/pull/1820#discussion_r1064666038

### DIFF
--- a/cmd/nerdctl/run.go
+++ b/cmd/nerdctl/run.go
@@ -316,7 +316,7 @@ func runAction(cmd *cobra.Command, args []string) error {
 		return errors.New("flags -d and --rm cannot be specified together")
 	}
 
-	container, gc, err := createContainer(ctx, cmd, globalOptions, client, args, platform, flagI, flagT, flagD)
+	container, gc, err := createContainer(ctx, cmd, client, globalOptions, args, platform, flagI, flagT, flagD)
 	if err != nil {
 		if gc != nil {
 			defer gc()
@@ -397,7 +397,7 @@ func runAction(cmd *cobra.Command, args []string) error {
 }
 
 // FIXME: split to smaller functions
-func createContainer(ctx context.Context, cmd *cobra.Command, globalOptions *types.GlobalCommandOptions, client *containerd.Client, args []string, platform string, flagI, flagT, flagD bool) (containerd.Container, func(), error) {
+func createContainer(ctx context.Context, cmd *cobra.Command, client *containerd.Client, globalOptions types.GlobalCommandOptions, args []string, platform string, flagI, flagT, flagD bool) (containerd.Container, func(), error) {
 	// simulate the behavior of double dash
 	newArg := []string{}
 	if len(args) >= 2 && args[1] == "--" {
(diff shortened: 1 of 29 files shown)


cmd/nerdctl/volume.go
```diff
@@ -44,6 +44,6 @@ func newVolumeCommand() *cobra.Command {
 
 // getVolumeStore returns a volume store
 // that corresponds to a directory like `/var/lib/nerdctl/1935db59/volumes/default`
-func getVolumeStore(globalOptions *types.GlobalCommandOptions) (volumestore.VolumeStore, error) {
+func getVolumeStore(globalOptions types.GlobalCommandOptions) (volumestore.VolumeStore, error) {
 	return volume.Store(globalOptions.Namespace, globalOptions.DataRoot, globalOptions.Address)
 }
```
djdongjin, line 49: I suggest we remove this function and use `volume.Store` (the 2nd one below) directly (we can change `volume.Store` to accept a globalOptions) (Don't necessarily be part of this PR though)

This adds one unnecessary wrap on volume create:

https://github.com/containerd/nerdctl/blob/40d341dac245a70a6ee4ac43567f2d2818fe3167/cmd/nerdctl/volume.go#L47-L49

https://github.com/containerd/nerdctl/blob/40d341dac245a70a6ee4ac43567f2d2818fe3167/pkg/cmd/volume/volume.go#L26-L32

https://github.com/containerd/nerdctl/blob/40d341dac245a70a6ee4ac43567f2d2818fe3167/pkg/mountutil/volumestore/volumestore.go#L42

cmd/nerdctl/run.go
```diff
@@ -395,7 +395,7 @@ func runAction(cmd *cobra.Command, args []string) error {
 }
 
 // FIXME: split to smaller functions
-func createContainer(ctx context.Context, cmd *cobra.Command, globalOptions *types.GlobalCommandOptions, client *containerd.Client, args []string, platform string, flagI, flagT, flagD bool) (containerd.Container, func(), error) {
+func createContainer(ctx context.Context, cmd *cobra.Command, globalOptions types.GlobalCommandOptions, client *containerd.Client, args []string, platform string, flagI, flagT, flagD bool) (containerd.Container, func(), error) {
```
djdongjin, line 398: nit: some functions have `containerd.Client` argument before `options` and some after `options`. E.g.:

```golang
func generateRootfsOpts(ctx context.Context, client *containerd.Client, platform string, cmd *cobra.Command, globalOptions types.GlobalCommandOptions, args []string, id string) ([]oci.SpecOpts, []containerd.NewContainerOpts, *imgutil.EnsuredImage, error) {
```


We might want to keep consistent style (personally I think functional arguments like `client`, `container`, `image` should be before `options`).

djdongjin (commented):
Overall LGTM, left some minor comments. thanks

djdongjin (approved):

cmd/nerdctl/volume.go
Zheaoli: This should be another PR

cmd/nerdctl/run.go
Zheaoli: SGTM, I will update it

djdongjin (approved):
LGTM on CI green. thanks

PR #1834, opened 2023-01-12
Refactor nerdctl rm

djdongjin:
Part of #1680 

### DIFF
--- a/cmd/nerdctl/rm.go
+++ b/cmd/nerdctl/rm.go
@@ -17,24 +17,8 @@
 package main
 
 import (
-	"context"
-	"encoding/json"
-	"fmt"
-	"os"
-	"syscall"
-
-	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/cio"
-	"github.com/containerd/containerd/errdefs"
-	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/nerdctl/pkg/api/types"
-	"github.com/containerd/nerdctl/pkg/clientutil"
-	"github.com/containerd/nerdctl/pkg/dnsutil/hostsstore"
-	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
-	"github.com/containerd/nerdctl/pkg/labels"
-	"github.com/containerd/nerdctl/pkg/namestore"
-
-	"github.com/sirupsen/logrus"
+	"github.com/containerd/nerdctl/pkg/cmd/container"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/nerdctl/rm.go
```diff
@@ -71,155 +49,7 @@ func rmAction(cmd *cobra.Command, args []string) error {
...
-	var delOpts []containerd.DeleteOpts
-	if _, err := container.Image(ctx); err == nil {
-		delOpts = append(delOpts, containerd.WithSnapshotCleanup)
-	}
-
-	if err := container.Delete(ctx, delOpts...); err != nil {
-		return err
-	}
-	return err
+	return container.Remove(cmd.Context(), args, globalOptions, force, removeAnonVolumes, cmd.OutOrStdout())
```
djdongjin, line 52: I didn't create a `RmCommandOptions` struct since it only has 2 boolean (force/volume) and to make other callers (e.g., prune, run) to easily reuse the `RemoveContainer` func (Instead of constructing a rmoptions).

I can change to add a `RmCommandOptions` if that's still preferred and change `Remove` to accept it (I think it's fine to keep the `RemoveContainer` signature).

pkg/cmd/container/remove.go
```diff
@@ -0,0 +1,218 @@
...
+	return fmt.Sprintf("container %s is in %v status", e.ID, e.Status)
+}
+
+// NewStatusError creates an ErrContainerStatus from container id and status.
+func NewStatusError(id string, status containerd.ProcessStatus) error {
+	return ErrContainerStatus{
+		ID:     id,
+		Status: status,
+	}
+}
```
djdongjin, line 61: This part is new that replaces the old `statusError` since after moving to `pkg`, other `cmd` cannot access the `error` within it.

Zheaoli (approved):
LGTM on CI green

cmd/nerdctl/rm.go
AkihiroSuda: Can we make all commands consistently have the option struct?

Also, s/Rm/Remove/

AkihiroSuda (changes requested):
Thanks, but please consider this
https://github.com/containerd/nerdctl/pull/1834#discussion_r1070200155

cmd/nerdctl/rm.go
djdongjin: updated

djdongjin:
> Thanks, but please consider this
> [#1834 (comment)](https://github.com/containerd/nerdctl/pull/1834#discussion_r1070200155)

updated. thanks

Zheaoli (approved):
LGTM, Thanks

AkihiroSuda (approved):
Thanks

PR #1837, opened 2023-01-12
refactor: consolidate main logic of volume.List into volume.Volumes 

davidhsingyuchen:
`volume.Ls` does 3 things:

1. Sanitize the input according to the nerdctl CLI logic (e.g., `"cannot use --size and --quiet together, ignoring --size"`).
2. The core logic: convert the human-readable filter strings into filter functions, apply them to the volumes, and return the resulting volumes.
3. Print the resulting volumes.

This PR extracts the core logic (i.e., step 2) into an exported function so that it can be reused by other projects, potentially not CLI ones, and they can have their own input validation logic and output format.

Appreciate any suggestion, thanks!

### DIFF
--- a/pkg/cmd/volume/list.go
+++ b/pkg/cmd/volume/list.go
@@ -48,18 +48,49 @@ func List(options types.VolumeListCommandOptions, stdout io.Writer) error {
 		logrus.Warn("cannot use --size and --quiet together, ignoring --size")
 		options.Size = false
 	}
-	labelFilterFuncs, nameFilterFuncs, sizeFilterFuncs, isFilter, err := getVolumeFilterFuncs(options.Filters)
-	if err != nil {
-		return err
-	}
-	if len(sizeFilterFuncs) > 0 && options.Quiet {
+	sizeFilter := hasSizeFilter(options.Filters)
+	if sizeFilter && options.Quiet {
 		logrus.Warn("cannot use --filter=size and --quiet together, ignoring --filter=size")
-		sizeFilterFuncs = nil
+		options.Filters = removeSizeFilters(options.Filters)
 	}
-	if len(sizeFilterFuncs) > 0 && !options.Size {
+	if sizeFilter && !options.Size {
 		logrus.Warn("should use --filter=size and --size together")
 		options.Size = true
 	}
+
+	vols, err := Volumes(
+		options.GOptions.Namespace,
+		options.GOptions.DataRoot,
+		options.GOptions.Address,
+		options.Size,
+		options.Filters,
+	)
+	if err != nil {
+		return err
+	}
+	return lsPrintOutput(vols, options, stdout)
+}
+
+func hasSizeFilter(filters []string) bool {
+	for _, filter := range filters {
+		if strings.HasPrefix(filter, "size") {
+			return true
+		}
+	}
+	return false
+}
+
+func removeSizeFilters(filters []string) []string {
+	var res []string
+	for _, filter := range filters {
+		if !strings.HasPrefix(filter, "size") {
+			res = append(res, filter)
+		}
+	}
+	return res
+}
+
+func lsPrintOutput(vols map[string]native.Volume, options types.VolumeListCommandOptions, stdout io.Writer) error {
 	w := stdout
 	var tmpl *template.Template
 	switch options.Format {
@@ -85,15 +116,7 @@ func List(options types.VolumeListCommandOptions, stdout io.Writer) error {
 		}
 	}
 
-	vols, err := Volumes(options.GOptions.Namespace, options.GOptions.DataRoot, options.GOptions.Address, options.Size)
-	if err != nil {
-		return err
-	}
-
 	for _, v := range vols {
-		if isFilter && !volumeMatchesFilter(v, labelFilterFuncs, nameFilterFuncs, sizeFilterFuncs) {
-			continue
-		}
 		p := volumePrintable{
 			Driver:     "local",
 			Labels:     "",
@@ -112,7 +135,7 @@ func List(options types.VolumeListCommandOptions, stdout io.Writer) error {
 			if err := tmpl.Execute(&b, p); err != nil {
 				return err
 			}
-			if _, err = fmt.Fprintf(w, b.String()+"\n"); err != nil {
+			if _, err := fmt.Fprintf(w, b.String()+"\n"); err != nil {
 				return err
 			}
 		} else if options.Quiet {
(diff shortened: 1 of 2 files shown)


pkg/cmd/volume/ls.go
```diff
@@ -129,6 +152,26 @@ func Ls(options *types.VolumeLsCommandOptions, stdout io.Writer) error {
 	return nil
 }
 
+// List contains the main logic and can be used by projects other than CLIs.
+// If improvements are added (e.g., filtering by dangling=true, driver=local, etc.),
+// other projects can also benefit from this.
+func List(namespace, dataRoot, address string, size bool, filters []string) (map[string]native.Volume, error) {
```
AkihiroSuda, line 158: It's confusing to have `List()` and `Ls()` with different functionalities

AkihiroSuda: I think you should just extend `Volumes()` to take options like `VolumesOpt...` that can be used for specifying the filters

AkihiroSuda (changes requested):
Thanks, but needs to be redesigned

https://github.com/containerd/nerdctl/pull/1837/files#r1068870842

pkg/cmd/volume/ls.go
davidhsingyuchen: Thanks for the review.

Does the following make sense to you?

```go
// Export getVolumeFilterFuncs 

type VolumesOpt func([]native.Volume) []native.Volume

// Caller of FilterVolumesOpt can feed the output of the to-be-exported getVolumeFilterFuncs to this func.
func FilterVolumesOpt(labelFilterFuncs []func(*map[string]string) bool, nameFilterFuncs []func(string) bool, sizeFilterFuncs []func(int64) bool) VolumesOpt {
	return func(vols []native.Volume) []native.Volume {
		var res []native.Volume
		for _, vol := range vols {
			if volumeMatchesFilter(vol, labelFilterFuncs, nameFilterFuncs, sizeFilterFuncs) {
				res = append(res, vol)
			}
		}
		return res
	}
}

// Update Volumes() to go through all the options and return the final resulting volumes.
```

Or maybe `getVolumeFilterFuncs` can be invoked in `FilterVolumesOpt(filters []string)` so that we don't have to export `getVolumeFilterFuncs`, and the user of this library only need to call `FilterVolumesOpt`, but `hasSizeFilter` and `removeSizeFilters` will be needed unless we want to invoke `getVolumeFilterFuncs` twice.

WDYT?

djdongjin: This reminds me a question: usually a command does 

1. preprocess step: process cli flag and construct options (which we're refactoring);
2. flag/option validation step: check either flags (previously) or options (after refactor), to resolve conflict options or do some filtering (e.g., here --quiet and --filter=size; in `run` we have some `-i -d -t` conflict).
3. **main logic step**: which I feel we should return the structs if it's a `GET/LIST`-kind func, or a `success or fail` indicator (bool, err, etc), or a `PUT/DELETE`-kind func.
4. postprocess step: after receiving results from 3, print as a table or json, etc.

**I feel each step should be more isolated instead of mixed** (by isolated I don't mean each should in a different func, but their logic/code should be colocated/not mixed). For example, [here](https://github.com/containerd/nerdctl/blob/27b1fef2cc834c55ac02540c6d7cc2635f80a07c/pkg/cmd/volume/list.go#L47-L62) we're mixing option validation (--filter=size, --quiet) with main logic (getVolumeFilterFuncs); [here](https://github.com/containerd/nerdctl/blob/27b1fef2cc834c55ac02540c6d7cc2635f80a07c/pkg/cmd/volume/list.go#L63-L96) we're mixing output with main logic.

Meanwhile, I guess **3 is the main step that we want to modularize and ideally be reused if someone want to use nerdctl as a dependency.**

Take this (specific file) as an example, I think the main logic is **List (return volume structs) Volumes based on given size and filters**. So we should have

1. a function (e.g., `ListCmd`) similar to the `Ls` that (1). preprocess the options, (2). call a `List` func (below) by passing related options (filters, size), (3). postprocess/output the returned volume struct from (2). (each step may be a sub func).
2. a function (e.g., `List`) that has the main logic, i.e., list and filter (and return) volumes.

It's just my perspective and some (or all :D) may not make sense. Feel free to lmk if it doesn't make sense at all :)

WDYT @AkihiroSuda @Zheaoli

djdongjin: > I think you should just extend Volumes() to take options like VolumesOpt... that can be used for specifying the filters.

I think we can just extend `Volumes()` to accept the `options.Filter` []string and handle the `filter construction` and the actual `filtering` steps, instead of doing filtering-related logic in (the current) `List` together with output logic. (I think we mean the same thing :))

Zheaoli: Mostly agree with @djdongjin 

We should keep the semantics consistent with reaching the code readability

TL;DR;

1. I think if we put the `List` and `Ls` in the same file, I think it would make people confuse.(LsCmd or other names are not keeping the same style as other command)
2. we filter the volumes in `List` method, the only difference  between `namespace, dataRoot, address string, size bool`  and `filter []string`  is that `namespace, dataRoot, address string, size bool` is the basic filter and `filter []string` is not. So I think we can just extend the `Volumes`

davidhsingyuchen: Thanks a lot for the suggestions. Added `filters []string` to the parameter list of `Volumes()` in 
0dd9ddb.

> I think if we put the List and Ls in the same file, I think it would make people confuse.(LsCmd or other names are not keeping the same style as other command)

That's true. Regarding #1855, following the naming in this PR, we can probably add a new exported function named `Images()` in `pkg/cmd/image`,  but if we want to extract the "**main logic step**" for other functions that are not "list-ish", what should the name be? For example, should we come up with a name on an ad-hoc basis? Or maybe we can use patterns like "ListCmd and List" consistently as essentially the "**main logic step**" of every functionality can be separated out? Or even another package could be created to hold the "**main logic step**". Need some guidance re. this part :)

Maybe we can continue the discussions in #1855, or maybe I can create an another issue to discuss the high-level approach as the scope is not limited to what's currently stated in #1855. Please let me know what works for you, thanks a lot!

pkg/cmd/volume/list.go
```diff
@@ -48,18 +48,49 @@ func List(options *types.VolumeListCommandOptions, stdout io.Writer) error {
...
-	labelFilterFuncs, nameFilterFuncs, sizeFilterFuncs, isFilter, err := getVolumeFilterFuncs(options.Filters)
-	if err != nil {
-		return err
-	}
-	if len(sizeFilterFuncs) > 0 && options.Quiet {
+	sizeFilter := hasSizeFilter(options.Filters)
+	if sizeFilter && options.Quiet {
 		logrus.Warn("cannot use --filter=size and --quiet together, ignoring --filter=size")
-		sizeFilterFuncs = nil
+		options.Filters = removeSizeFilters(options.Filters)
```
djdongjin, line 54: not related to this PR, but curious why this two flags conflict (`--filter=size` is for functionality/which volumes to `list`, while `--quiet` is for formatting/how to print). can we remove this restriction now? @AkihiroSuda

pkg/cmd/volume/list.go
```diff
@@ -48,18 +48,49 @@ func List(options *types.VolumeListCommandOptions, stdout io.Writer) error {
...
+	var res []string
+	for _, filter := range filters {
+		if !strings.HasPrefix(filter, "size") {
+			res = append(res, filter)
+		}
+	}
+	return res
+}
+
+func lsPrintOutput(options *types.VolumeListCommandOptions, stdout io.Writer, vols map[string]native.Volume) error {
```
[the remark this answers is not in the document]
davidhsingyuchen: Makes sense, fixed.

pkg/cmd/volume/list.go
AkihiroSuda: Maybe yes, but should be another PR

pkg/cmd/volume/list.go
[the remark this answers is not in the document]
davidhsingyuchen: The tests are failing due to [this line](https://github.com/davidhsingyuchen/nerdctl/commit/0b37d7203fcf7932c1c84d1d7e0c42526d797efe#diff-619baec41b9bc22c95bd3e1fcc9e70bd8b9ca46a400a5814f351c8a30de4dcf7L46). Rebased on `main` and updated the signature of `lsPrintOutput` in this PR to fix them.

pkg/cmd/volume/list.go
```diff
@@ -129,12 +152,32 @@ func List(options types.VolumeListCommandOptions, stdout io.Writer) error {
 	return nil
 }
 
-func Volumes(ns string, dataRoot string, address string, volumeSize bool) (map[string]native.Volume, error) {
+// Volumes contains the main logic and can be used by projects other than CLIs.
+// If improvements are added (e.g., filtering by dangling=true, driver=local, etc.),
+// other projects can also benefit from this.
```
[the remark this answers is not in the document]
davidhsingyuchen: My bad, forgot to update this. Updated the function comment according to https://github.com/containerd/nerdctl#whale-nerdctl-volume-ls by removing the shell-related part.

djdongjin:
LGTM, CI error looks not related, @davidhsingyuchen can you rebase to pick up new changes in main

djdongjin (approved):

davidhsingyuchen:
> LGTM, CI error looks not related, @davidhsingyuchen can you rebase to pick up new changes in main

Done, thanks!

AkihiroSuda (approved):
Thanks

Issue #519, opened 2021-11-16
Add Windows Hyperv support


jsturtevant:
Add the correct fields to the spec at https://github.com/containerd/nerdctl/blob/f316402913c4239a2388b7467bcffc9735dd8327/cmd/nerdctl/run_windows.go#L42

Something like:

```
isolation := cmd.Flags().GetString("isolation")
if isolation == "hyperv" {
	opts = append(opts, oci.WithWindowsHyperV)
}
```

See https://docs.docker.com/engine/reference/commandline/run/#specify-isolation-technology-for-container---isolation

dcantah:
Planning to look at this next week

dcantah:
LOL Yes, I was in a hyperbolic time chamber apparently. The work is done it looks like

dcantah:
LOL Yes, possibly the longest week ever. It seems the work is done though

PR #1848, opened 2023-01-13
Add Windows HyperV container mode

dardelean:
Signed-off-by: Dan Ardelean <dardelean@cloudbasesolutions.com>

Implements [519](https://github.com/containerd/nerdctl/issues/519)

### DIFF
--- a/cmd/nerdctl/run_windows.go
+++ b/cmd/nerdctl/run_windows.go
@@ -19,13 +19,13 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/containerd/containerd/containers"
-	"github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/nerdctl/pkg/api/types"
 	"github.com/docker/go-units"
+	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/spf13/cobra"
 )
 
@@ -35,6 +35,8 @@ const (
 
 	// HostProcessContainer will launch a host process container
 	hostProcessContainer = "microsoft.com/hostprocess-container"
+	uvmMemorySizeInMB    = "io.microsoft.virtualmachine.computetopology.memory.sizeinmb"
+	uvmCPUCount          = "io.microsoft.virtualmachine.computetopology.processor.count"
 )
 
 func capShellComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
@@ -75,12 +77,30 @@ func setPlatformOptions(
 		opts = append(opts, oci.WithMemoryLimit(uint64(mem64)))
 	}
 
-	// TODO implement hyper-v isolation
 	isolation, err := cmd.Flags().GetString("isolation")
 	if err != nil {
 		return nil, err
 	}
 	switch isolation {
+	case "hyperv":
+		if memStr != "" {
+			mem64, err := units.RAMInBytes(memStr)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse memory bytes %q: %w", memStr, err)
+			}
+			UVMMemmory := map[string]string{
+				uvmMemorySizeInMB: fmt.Sprintf("%v", mem64),
+			}
+			opts = append(opts, oci.WithAnnotations(UVMMemmory))
+		}
+
+		if cpus > 0.0 {
+			UVMCPU := map[string]string{
+				uvmCPUCount: fmt.Sprintf("%v", cpus),
+			}
+			opts = append(opts, oci.WithAnnotations(UVMCPU))
+		}
+		opts = append(opts, oci.WithWindowsHyperV)
 	case "host":
 		hpAnnotations := map[string]string{
 			hostProcessContainer: "true",
--- a/cmd/nerdctl/run_windows_test.go
+++ b/cmd/nerdctl/run_windows_test.go
@@ -18,9 +18,12 @@ package main
 
 import (
 	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/containerd/nerdctl/pkg/testutil"
+	"golang.org/x/sys/windows/svc/mgr"
+	"gotest.tools/v3/assert"
 )
 
 func TestRunHostProcessContainer(t *testing.T) {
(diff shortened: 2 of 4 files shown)


pkg/testutil/testutil.go
```diff
@@ -232,6 +232,14 @@ func (b *Base) InspectContainer(name string) dockercompat.Container {
 	return dc[0]
 }
 
+// Check if HyperV is set in the container
+// The flag is not yet part of the dockercompat spec
+func (b *Base) InspectContainerModeNative(name string) string {
```
AkihiroSuda, line 237: This should return the struct from https://github.com/containerd/nerdctl/tree/main/pkg/inspecttypes/native

cmd/nerdctl/run_windows_test.go
```diff
@@ -65,3 +67,28 @@ func TestRunProcessIsolated(t *testing.T) {
 	containerUser := "ContainerUser"
 	base.Cmd("run", "--rm", "--isolation=process", "-u", containerUser, testutil.WindowsNano, "whoami").AssertOutContains(containerUser)
 }
+
+func TestRunHyperVContainer(t *testing.T) {
```
AkihiroSuda, line 71: `t.Skip` if Hyper-V is unavailable

dardelean: Added

pkg/testutil/testutil.go
dardelean: Removed the function, doing the inspect output test in the test function.

jsturtevant:
Code changes look fine to me.

I was expecting we would need to turn on `nested_virtualization` in the cirrus configuration to get this to work (https://github.com/containerd/nerdctl/pull/795#issuecomment-1377588311) as well as bump containerd version to 1.7+ (https://github.com/containerd/nerdctl/blob/27b1fef2cc834c55ac02540c6d7cc2635f80a07c/.cirrus.yml#L33) as hyper-v support was only recently enabled in containerd.

@dcantah could you take a look at the test to see if that is the right check?

README.md
```diff
@@ -397,7 +397,7 @@ Init process flags:
...
+- :whale: :blue_square: :nerd_face: `--isolation=(default|process|host|hyperv)`: Used on Windows to change process isolation level. `default` will use the runtime options configured in `default_runtime` in the [containerd configuration](https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cri-plugin-config-guide) which is `process` in containerd by default. `process` runs process isolated containers.  `host` runs [Host Process containers](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/).  Host process containers inherit permissions from containerd process unless `--user` is specified then will start with user specified and the user specified must be present on the host.  `host` requires Containerd 1.7+. Not implemented for Linux. `hyperv` runs Hyper-V hypervisor partition-based isolated containers.
```
jsturtevant, line 400: nit: keep the options together
```suggestion
- :whale: :blue_square: :nerd_face: `--isolation=(default|process|host|hyperv)`: Used on Windows to change process isolation level. `default` will use the runtime options configured in `default_runtime` in the [containerd configuration](https://github.com/containerd/containerd/blob/master/docs/cri/config.md#cri-plugin-config-guide) which is `process` in containerd by default. `process` runs process isolated containers.  `host` runs [Host Process containers](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/).  Host process containers inherit permissions from containerd process unless `--user` is specified then will start with user specified and the user specified must be present on the host.  `host` requires Containerd 1.7+. `hyperv`  runs Hyper-V hypervisor partition-based isolated containers. Not implemented for Linux. 
```

cmd/nerdctl/run_windows_test.go
```diff
@@ -65,3 +68,42 @@ func TestRunProcessIsolated(t *testing.T) {
...
+	}
+	defer s.Close()
+
+	containerName := testutil.Identifier(t)
+
+	base.Cmd("run", "--isolation", "hyperv", "--name", containerName, testutil.WindowsNano).Out()
+	defer base.Cmd("rm", "-f", containerName).AssertOK()
+	inspectOutput := base.Cmd("container", "inspect", "--mode", "native", containerName).Out()
+
+	assert.Check(t, strings.Contains(inspectOutput, "hyperv"))
```
AkihiroSuda, line 96: This is always true because the container name contains "hyperv"

cmd/nerdctl/run_windows_test.go
```diff
@@ -65,3 +68,42 @@ func TestRunProcessIsolated(t *testing.T) {
...
+	base := testutil.NewBase(t)
+	hypervServiceName := "vmcompute"
+
+	m, err := mgr.Connect()
+	if err != nil {
+		t.Fatalf("unable to access Windows service control manager: %s", err)
+	}
+	defer m.Disconnect()
+
+	s, err := m.OpenService(hypervServiceName)
```
dcantah, line 83: I think (emphasis on think) vmcompute should exist with just the containers feature on. Don't have a windows machine for a little bit but I can verify

dcantah: (if so) we'd need a better check

dcantah:
> as well as bump containerd version to 1.7+ (

We should be fine without 1.7 as we're manually toggling the hyper-v field. You'd need 1.7 for Hyper-V through CRI however 😬

cmd/nerdctl/run_windows_test.go
dardelean: This is correct @dcantah, the service is present with only the containers feature on.

dardelean: There is another service called "vmms", will parse for that.

dcantah:
Cc @kevpar

cmd/nerdctl/run_windows.go
```diff
@@ -35,6 +35,8 @@ const (
 
 	// HostProcessContainer will launch a host process container
 	hostProcessContainer = "microsoft.com/hostprocess-container"
+	UVMMemoryAnnotations = "MemorySizeInMB"
```
dcantah, line 38:
```suggestion
	uvmMemorySizeInMB = "io.microsoft.virtualmachine.computetopology.memory.sizeinmb"
```

dcantah: I hate having to redefine these here, but otherwise we'd need to pull them from hcsshim, which means we need an hcsshim dep 😬. That'd feel like a waste also

cmd/nerdctl/run_windows.go
```diff
@@ -75,12 +77,27 @@ func setPlatformOptions(
...
 
-	// TODO implement hyper-v isolation
 	isolation, err := cmd.Flags().GetString("isolation")
 	if err != nil {
 		return nil, err
 	}
 	switch isolation {
+	case "hyperv":
+		if memStr != "" {
+			mem64, _ := units.RAMInBytes(memStr)
```
manugupt1, line 87: Can we handle errors please?

cmd/nerdctl/run_windows_test.go
```diff
@@ -65,3 +68,43 @@ func TestRunProcessIsolated(t *testing.T) {
...
+func TestRunProcessContainer(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	containerName := testutil.Identifier(t)
+
+	base.Cmd("run", "--isolation", "process", "--name", containerName, testutil.WindowsNano).Out()
+	defer base.Cmd("rm", "-f", containerName).AssertOK()
+	inspectOutput := base.Cmd("container", "inspect", "--mode", "native", containerName).Out()
+
+	assert.Check(t, !strings.Contains(inspectOutput, "hyperv"))
```
manugupt1, line 109: I think its safer to use assert.Assert.
https://pkg.go.dev/testing#T.FailNow (will also call defer)

AkihiroSuda (approved):
Thanks

dcantah (approved):

manugupt1 (approved):

dardelean:
I had a merge conflict with #1880 and I had to rebase.

cmd/nerdctl/run_windows.go
```diff
@@ -75,12 +77,30 @@ func setPlatformOptions(
...
 	}
 
-	// TODO implement hyper-v isolation
 	isolation, err := cmd.Flags().GetString("isolation")
 	if err != nil {
 		return nil, err
 	}
 	switch isolation {
+	case "hyperv":
+		if memStr != "" {
```
jsturtevant, line 86: Are there defaults (are those in the runtime config?)

jsturtevant:
lgtm

cmd/nerdctl/run_windows.go
dcantah: Related, if no default was supplied I'm forgetting if the HCS v1 behavior was to expose the hosts cpu and memory size to the VM (if anyone wants to try this, otherwise I can report back tonight). With this change if you didn't supply any values the VM would have 2 cores and a similar low amount of ram. Could change this in a follow up though

AkihiroSuda (approved):
Thanks

PR #1850, opened 2023-01-13
Refactor container inspect

weikequ:
Hey! 👋 It's my first time submitting a PR to this project, hopefully not too rough. Feedback very much encouraged!

Refactoring project from https://github.com/containerd/nerdctl/issues/1680

I also wasn't sure if I should add any unit tests to pkg/cmd/ for this refactor. Let me know if I should.

- [x] Create a file in pkg/api/types/${cmd}.go, and define the CommandOption for this command
- [x]  Create some file in pkg/cmd/${cmd}, and move the command entry point in real into this package

### DIFF
--- a/cmd/nerdctl/container_inspect.go
+++ b/cmd/nerdctl/container_inspect.go
@@ -17,15 +17,10 @@
 package main
 
 import (
-	"context"
 	"fmt"
-	"time"
 
-	"github.com/containerd/nerdctl/pkg/clientutil"
-	"github.com/containerd/nerdctl/pkg/containerinspector"
-	"github.com/containerd/nerdctl/pkg/formatter"
-	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
-	"github.com/containerd/nerdctl/pkg/inspecttypes/dockercompat"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/cmd/container"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


weikequ:
Added sign off to fix project checks.

pkg/cmd/container/inspect.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/clientutil"
+	"github.com/containerd/nerdctl/pkg/containerinspector"
+	"github.com/containerd/nerdctl/pkg/formatter"
+	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
+	"github.com/containerd/nerdctl/pkg/inspecttypes/dockercompat"
+)
+
+func Inspect(ctx context.Context, options *types.ContainerInspectCommandOptions, stdout io.Writer) error {
```
djdongjin, line 33: can you change the `options` to a non-pointer struct?

pkg/api/types/container_types.go
```diff
@@ -0,0 +1,28 @@
...
+
+type ContainerInspectCommandOptions struct {
+	// GOptions is the global options
+	GOptions GlobalCommandOptions
+	// Format of the output
+	Format string
+	// Inspect mode, either dockercompat or native
+	Mode string
+	// Targets are the containers to be inspected
+	Targets []string
```
djdongjin, line 27: Maybe just `Containers`?

weikequ: Makes sense, I started off with trying to refactor inspect and used Targets as it could refer to both images or containers. Containers makes sense in this context though. Thanks!

pkg/cmd/container/inspect.go
weikequ: Changes made ✅

pkg/api/types/container_types.go
weikequ: Changes made ✅

cmd/nerdctl/container_inspect.go
```diff
@@ -52,74 +47,33 @@ func newContainerInspectCommand() *cobra.Command {
 	return containerInspectCommand
 }
 
+var validModeType = map[string]bool{
+	"native":       true,
+	"dockercompat": true,
+}
```
djdongjin, line 53: nit: probably we need to find a better place and define these mode as const. I remember I saw them in other places as well.

pkg/api/types/container_types.go
```diff
@@ -33,3 +33,15 @@ type ContainerRemoveCommandOptions struct {
...
+
+// ContainerInspectCommandOptions specifies options for `nerdctl container inspect`
+type ContainerInspectCommandOptions struct {
+	// GOptions is the global options
+	GOptions GlobalCommandOptions
+	// Format of the output
+	Format string
+	// Inspect mode, either dockercompat or native
+	Mode string
+	// Targets are the containers to be inspected
```
djdongjin, line 45: Update comments to `// Containers to be inspected`

pkg/cmd/container/inspect.go
```diff
@@ -0,0 +1,91 @@
...
+		OnFound: f.Handler,
+	}
+
+	var errs []error
+	for _, req := range options.Containers {
+		n, err := walker.Walk(ctx, req)
+		if err != nil {
+			errs = append(errs, err)
+		} else if n == 0 {
+			errs = append(errs, fmt.Errorf("no such object: %s", req))
```
djdongjin, line 55: nit: `no such object` -> `no such container`

cmd/nerdctl/container_inspect.go
weikequ: Agreed. Should I mark these out as an issue to be refactored later? The place that I saw these were in `cmd/nerdctl/inspect.go` and I made these to match the style found there.

djdongjin: Just noticethere is a copy in `cmd/nerdctl/inspect.go`. Then you can reuse it here instead making a new copy, since they're in the same package.

pkg/api/types/container_types.go
weikequ: Oops, good catch!

cmd/nerdctl/container_inspect.go
weikequ: It is a bit different, mine is to verify the mode (native/dockercompat), `inspect.go` is for the type (image/container).

djdongjin: I see, then no change requested here :)

djdongjin (approved):
LGTM, thanks

AkihiroSuda (approved):
Thanks

PR #1857, opened 2023-01-16
[Refactor] Move exit_coder from cmd to pkg/errutil

Laitr0n:
Checklist:
- [x]  move `ExitCodeError` from `cmd` to ~~`pkg/exitcoder`~~`pkg/errutil`

https://github.com/containerd/nerdctl/blob/0b051a5ddbb29d6673b853f7b1b1863d331f1f3d/cmd/nerdctl/start.go#L187-L190
When I refactor the container start command,  I have discovered  that `ExitCodeError` from `cmd` is blocking my work.

https://github.com/containerd/nerdctl/blob/0b051a5ddbb29d6673b853f7b1b1863d331f1f3d/cmd/nerdctl/main.go#L343-L367
Maybe we should refactor `ExitCodeError` first, by moving it to `pkg/exitcoder`. Then we can continue with our other tasks.

### DIFF
--- a/cmd/nerdctl/main.go
+++ b/cmd/nerdctl/main.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/containerd/nerdctl/pkg/config"
 	ncdefaults "github.com/containerd/nerdctl/pkg/defaults"
+	"github.com/containerd/nerdctl/pkg/errutil"
 	"github.com/containerd/nerdctl/pkg/logging"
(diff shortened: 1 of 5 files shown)


pkg/exitcoder/exit_coder.go
```diff
@@ -0,0 +1,49 @@
+/*
```
AkihiroSuda, line 1: Should be in `pkg/errutil`

PR #1861, opened 2023-01-17
refactor: login flagging process

davidhsingyuchen:

Notes:

- I leave `verifyLoginOptions` in `cmd/nerdctl` for now because moving it to `pkg/cmd/login` requires adding `PasswordStdin` as a field of `LoginCommandOptions`, but that may complicate the comments of its fields a bit (i.e., to explain the relationship between `PasswordStdin` and `Password` [as well as `Username`](https://github.com/containerd/nerdctl/blob/e9195f50f32c0d560ac602e0201f33da977e3d1e/cmd/nerdctl/login.go#L166)). Please let me know if I should move `verifyLoginOptions` to `pkg/cmd/login` instead, thanks!
- The only major modification should be updating the signature of `ConfigureAuthentication` to accept `username, password string` instead of `options *loginOptions` as it only depends on those 2 strings.

### DIFF
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -17,27 +17,12 @@
 package main
 
 import (
-	"bufio"
-	"context"
 	"errors"
-	"fmt"
 	"io"
-	"net/http"
-	"net/url"
-	"os"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


AkihiroSuda (approved):
Thanks

Issue #1811, opened 2023-01-04
CI: test-integration often hits the 30min limit


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/bbe5c052660000e4ef65c715c6b9e6bba25a9fdb/.github/workflows/test.yml#L61-L63

We could just increase the timeout, but optimizing the tests to shorten the time might be more preferable

yuchanns:
I've found that some tests can be run in parallel with `t.Parallel()`, this might be helpful.

amarlearning:
I'd like to take on this issue. I'll start working on it and keep you updated on my progress.

djdongjin:
@amarlearning thanks, please go ahead :)

fahedouch:
> I've found that some tests can be run in parallel with `t.Parallel()`, this might be helpful.

Please be careful when setting `t.Parallel()` this may introduce indirect deps between tests which can lead to test failing.

djdongjin:
Some sample time consuming tests observed from [CI](https://github.com/containerd/nerdctl/actions/runs/3904774683/jobs/6670984959):

```
PASS cmd/nerdctl.TestIPFSComposeUp/overlayfs (65.52s)
PASS cmd/nerdctl.TestIPFSComposeUp/stargz (227.56s)
PASS cmd/nerdctl.TestIPFSComposeUp/ipfs-address (33.36s)
PASS cmd/nerdctl.TestIPFSComposeUp (331.73s)
PASS cmd/nerdctl.TestIPFSComposeUpBuild (22.58s)
PASS cmd/nerdctl.TestLoginWithPlainHttp (28.60s)
PASS cmd/nerdctl.TestSyslogFacilities (26.98s)
PASS cmd/nerdctl.TestRunPort (28.14s)
PASS cmd/nerdctl.TestRunRestartWithOnFailure (20.43s)
PASS cmd/nerdctl.TestRunRestartWithUnlessStopped (20.78s)
PASS cmd/nerdctl.TestUpdateRestartPolicy (20.71s)
```

amarlearning:
@djdongjin Even I saw the CI to find out which test was taking a long time and based on that I tried parallelizing `TestIPFSComposeUp` on my PR.

Thanks for sharing other tests which were taking longer time 👍🏻

amarlearning:
As of now, our test-integration job takes ~8-10 mins just to prepare the environment. I am also finding out if there is a way to cache common docker images across all OS and Containerd versions.

[...]

PR #1867, opened 2023-01-17
Increase CI timeout to reduce failure due to timeout

djdongjin:
Temp workaroud of https://github.com/containerd/nerdctl/issues/1811 before we make progress on reducing testing time, also unifies the workflow and go test timeout:

| | workflow | go test |
| ----- | ----- | ----- | 
| test-integration | 30 -> 40m | 20 -> 30m |
| test-integration-rootless | 60m (no change) | 20 -> 30m |
| test-integration-docker-compatibility | 20 -> 30m | 20m (no change) |

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,7 +62,7 @@ jobs:
 
   test-integration:
     runs-on: "ubuntu-${{ matrix.ubuntu }}"
-    timeout-minutes: 30
+    timeout-minutes: 40
     strategy:
       fail-fast: false
       matrix:
@@ -145,7 +145,7 @@ jobs:
 
   test-integration-docker-compatibility:
     runs-on: ubuntu-22.04
-    timeout-minutes: 20
+    timeout-minutes: 30
     steps:
       - uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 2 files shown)


fahedouch:
test-integration-docker-compatibility:  is it really necessary to double the timeout, 40 minutes seems enormous! I have rarely seen a waiting period in this job (even go test we may keep 20 minutes)

djdongjin:
> `test-integration-docker-compatibility`: is it really necessary to double the timeout, 40 minutes seems enormous! I have rarely seen a waiting period in this job (even go test we may keep 20 minutes)

SGTM. Changing this one is mainly to keep it consistent with other tests :). 

I can change the `workflow` to 30 and `go test` back to 20. Changing workflow 20->30 because otherwise it seems not quite make sense (the whole workflow and `go test`, which is a step of the workflow, have the same timeout). WDYT? @fahedouch

.github/workflows/test.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-      - 'release/**'
+      - "release/**"
```
AkihiroSuda, line 7: ?

AkihiroSuda (approved):
Thanks

.github/workflows/test.yml
djdongjin: auto-correction 😅 not sure if it's needed so removed this change

AkihiroSuda (approved):
Thanks

fahedouch:
> > `test-integration-docker-compatibility`: is it really necessary to double the timeout, 40 minutes seems enormous! I have rarely seen a waiting period in this job (even go test we may keep 20 minutes)
> 
> SGTM. Changing this one is mainly to keep it consistent with other tests :).
> 
> I can change the `workflow` to 30 and `go test` back to 20. Changing workflow 20->30 because otherwise it seems not quite make sense (the whole workflow and `go test`, which is a step of the workflow, have the same timeout). WDYT? @fahedouch

`40` workflow and `30` go test SGTM, 
Also, I would prefer adding a comment here too as a reminder that we should refacto test as this timeout is huge

PR #1873, opened 2023-01-18
Network walker return -1 when error occurs

yzxiu:


fahedouch:
I think this is useless , why ? : 

1. `the number of the found entries` has no effect when `err != nil` .  Here 0 is nil value of int. 
2.  the func comment says `Returns the number of the found entries.`, logically this should be `unit`

AkihiroSuda:
I agree with @fahedouch , let me close this

yzxiu:
@fahedouch @AkihiroSuda  thanks for review

PR #1885, opened 2023-01-19
[VMware Photon OS] Add dbus-user-session related files

sshedi:
In Photon OS or in Redhat derivatives, there is no dbus-user-session package. It is better to keep them in the repo and users can use these files on the need basis on cgroup v2 systems.

I have tested this on Photon OS with 6.x kernel and cgroup v2 enabled.

Update the FAQ for the same.

Signed-off-by: Shreenidhi Shedi <sshedi@vmware.com>

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -348,6 +348,12 @@ sudo apt-get install -y dbus-user-session
 systemctl --user start dbus
 ```
 
+To manually install the dbus-user-session files, consume them from:
+
+https://github.com/sshedi/nerdctl/tree/main/extras/dbus-user-session
+
+And do: `cp -pr extras/dbus-user-session/* /usr/lib/systemd/user/`
+
 ### How to uninstall ? / Can't remove `~/.local/share/containerd`
 
 Run the following commands:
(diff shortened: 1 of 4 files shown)


sshedi:
@AkihiroSuda please take these changes. It took a while for me to find these files to get nerdctl working on my machine.
Hope this helps.

AkihiroSuda:
> Photon OS

The distro should provide its own suppport for dbus-user-session.
Maybe you can negotiate with your colleagues 🙂 

If Photon OS purposely omits the support for dbus-user-session, let me know the reason and I can reconsider this PR.

> Redhat derivatives

```console
$ rpm -qf /usr/lib/systemd/user/dbus.socket 
dbus-common-1.12.20-6.el9.noarch
```

sshedi:
We have dbus in our system just the `dbus-user-session` service files are a bit different.
I added the user dbus related files to Photon sometime back and it will be available soon.
But this is for general audience where they want to just get nerdctl working in cgroup v2 machines without much trouble.

Yeah, you are right, Fedora already has it.
https://src.fedoraproject.org/rpms/dbus/blob/rawhide/f/dbus.spec#_38

Feel free to close this PR. Sorry for the trouble.

Issue #1778, opened 2022-12-28
nerdctl build helptext is incorrect


Liqs-v2:
# Problem
I am still new to using nerdctl and often rely on the documentation and helptext that comes with the commandline tools.
## Expectation
When I follow the usage pattern suggested by the helptext of a commandline tool, the program is supplied with all **required** parameters and will run.
## Current state
When checking the helptext for build `nerdctl build --help` I get the following output:
![grafik](https://user-images.githubusercontent.com/9247500/209812891-0dcbde0c-266e-48fb-8bff-d789b8a6185d.png)

Clearly the **usage pattern suggests** that `nerdctl build -f local.Dockerfile` should run without issue, but it results in:
`FATA[0000] context needs to be specified`

Running: `nerdctl build . --file local.Dockerfile` runs as expected.

## Fix
I would suggest updating the usage pattern to:
`nerdctl build <path_to_folder> [flags]`

to accurately reflect the parameters actually required for the program.

## Metainformation

- Nerdctl Version: 1.0.0
- Windows 10
- Powershell Version: 5.1.19041.2364

maddy020:
Hey @AkihiroSuda  
I would like to work over this issue

vvejell1:
Hi @AkihiroSuda , @djdongjin
I have been looking into this for a while now, and I'll modify the Build usage pattern.

AkihiroSuda:
This has been fixed in v1.1.0:
- https://github.com/containerd/nerdctl/pull/1523

PR #1886, opened 2023-01-19
Update usage pattern for build

vvejell1:
**Issue:** #1778 
The build usage pattern enables us in building and loading the image from the Dockerfile. If we put flags before the PATH likely `nerdctl build -f local.Dockerfile` then it will return an error. However, if we add PATH to the Dockerfile before the flags, it will run as expected. 

**Description of changes:**  
Adding the PATH to Dockerfile before the `[flags]` in the Build usage pattern. 

Signed-off-by: Vaishnavi Vejella <vvejella@amazon.com>

### DIFF
--- a/cmd/nerdctl/build.go
+++ b/cmd/nerdctl/build.go
@@ -33,7 +33,7 @@ import (
 
 func newBuildCommand() *cobra.Command {
 	var buildCommand = &cobra.Command{
-		Use:   "build [flags] PATH",
+		Use:   "build PATH [flags]",
 		Short: "Build an image from a Dockerfile. Needs buildkitd to be running.",
 		Long: `Build an image from a Dockerfile. Needs buildkitd to be running.
 If Dockerfile is not present and -f is not specified, it will look for Containerfile and build with it. `,


vvejell1:
Maintainers, Can I request the rerun of failing checks?

AkihiroSuda:
>  If we put flags before the PATH likely `nerdctl build -f local.Dockerfile` then it will return an error. However, if we add PATH to the Dockerfile before the flags, it will run as expected.

What error?

AkihiroSuda:
`build [flags] PATH` is the correct form (as well as `build PATH [flags]`), so I'm closing this.

If you are facing an error, please open a separate issue or a question.

vvejell1:
I was able to reproduce this line `FATA[0000] context needs to be specified` which I referred above as error.
While I was trying to extract the image from the local.Dockerfile

vvejell1:
I was able to extract the image from the Dockerfile by using this command `nerdctl build . --file local.Dockerfile` and here I am pointing `.` as current directory PATH for the Dockerfile.

AkihiroSuda:
`nerdctl build --file local.Dockerfile .` is correct too

PR #1864, opened 2023-01-17
[Refactor] Refactor the container start command flagging process

Laitr0n:
Part of https://github.com/containerd/nerdctl/issues/1680

Checklist:
- [x] Move startContainer from cmd to pkg/containerutil
- [x] Create a file in pkg/api/types/${cmd}_types.go, and define the CommandOption for this command
- [x] Create some file in pkg/cmd/${cmd}, and move the command entry point in real into this package

Signed-off-by: Laitron <meetlq@outlook.com>

PR #1890, opened 2023-01-20
refactor: move updateContainerStoppedLabel to containerutil

davidhsingyuchen:
PR is a blocker of refactoring `container stop` and `container start` as part of #1680.

Thought process:

1. I removed `Container` from the function name to [avoid stuttering](https://stackoverflow.com/a/45886055) with the package name (i.e., `containerutil`). Also, one of the parameters is `container containerd.Container`, so the semantics should be already clear.
2. I noticed that `stopped bool` is a bit misleading as it seems to say whether the container is stopped, but the point is to convey whether the container is *explicitly* stopped so that containerd can decide if the container should be restarted. As a result, I renamed the parameter and the function accordingly.

Please let me know if I should do this in a follow-up PR instead, or it just doesn't make sense and shouldn't be done at all, thanks!

### DIFF
--- a/cmd/nerdctl/run_restart.go
+++ b/cmd/nerdctl/run_restart.go
@@ -19,7 +19,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/containerd/containerd"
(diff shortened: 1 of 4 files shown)


Zheaoli (approved):
LGTM

davidhsingyuchen:
> We're doing some works that are the same.

Oh I see, I checked the open PRs yesterday, and it seems that #1864 was just reopened today.

> Once one PR is merged, I will proceed with the next one.

Maybe they can be done in parallel. I think the point is to break the giant PR into smaller/independent ones instead of doing them serially.

PR #1893, opened 2023-01-20
[Refactor] Move generateSharingPIDOpts from cmd to pkg/containerutil

Laitr0n:
Splitting pr #1864 
Same with pr #1857, prepare for refactor the container start command.

Checklist:

- [x] Move `generateSharingPIDOpts` from cmd to pkg/containerutil

### DIFF
--- a/pkg/containerutil/containerutil.go
+++ b/pkg/containerutil/containerutil.go
@@ -30,6 +30,7 @@ import (
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/runtime/restart"
 	"github.com/containerd/nerdctl/pkg/portutil"
+	"github.com/containerd/nerdctl/pkg/rootlessutil"
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/containerutil/containerutil.go
```diff
@@ -129,3 +130,46 @@ func WithBindMountHostProcfs(_ context.Context, _ oci.Client, _ *containers.Cont
 	s.Linux.ReadonlyPaths = newROP
 	return nil
 }
+
+// GenerateSharingPIDOpts returns the spec specific information.
```
AkihiroSuda, line 134: The comment isn't descriptive

pkg/containerutil/containerutil.go
```diff
@@ -129,3 +130,48 @@ func WithBindMountHostProcfs(_ context.Context, _ oci.Client, _ *containers.Cont
 	s.Linux.ReadonlyPaths = newROP
 	return nil
 }
+
+// GenerateSharingPIDOpts returns the oci.SpecOpts for shared container
+// that is in host namespace running inside rootless mode or not
+// or in linux namespace.
```
djdongjin, line 136: Sorry I don't quite understand this part 

```
// that is in host namespace running inside rootless mode or not
// or in linux namespace.`.
```

If I read the func correctly, is it doing something like this?
```
// GenerateSharingPIDOpts returns the oci.SpecOpts that shares the host linux namespace from `targetCon` 
// If `targetCon` doesn't have a `PIDNamespace`, a new one is generated from its `Pid`)
```

djdongjin (approved):
LGTM, thanks

AkihiroSuda (approved):
Thanks

PR #1910, opened 2023-01-23
Optimize inspect commands

### LINKED ISSUE #1909: `nerdctl inspect` creates duplicated client
`nerdctl inspect` can inspect both `image` and `container`, but the code for them are inconsistency and create unnecessary `containerd.client`:

1. When inspect a container, `containerInspectAction(cmd, []string{req})` is called (the entry in `cmd` side); when inspect an image, `image.Inspect(cmd.Context(), []string{req}, imageInspectOptions)` is called (the entry in `pkg/cmd` side).
2. In both cases, a new `container.client` is created for each inspect object. (https://github.com/containerd/nerdctl/issues/1907 can resolve this for image inspect, but not container inspect). Plus, another client is created from image|container walker.

I think we need to do the following refactor for it:

(After #1907 is fixed)
1. call `container.Inspect` instead of `containerInspectAction` to avoid unnecessary client creation and cmd flag parsing.
2. create one `client` by [`NewClient`](https://github.com/containerd/nerdctl/blob/67a065a80fceb7602a9092dc5dec52a959454853/pkg/clientutil/client.go#L36) for container inspect, and one `client` by [`NewClientWithPlatform`](https://github.com/containerd/nerdctl/blob/67a065a80fceb7602a9092dc5dec52a959454853/pkg/clientutil/client.go#L59) for image inspect (if no `platform` specified, it can reuse the `client` for container inspect). So in total only 1 or 2 client are needed for container/image inspects.

https://github.com/containerd/nerdctl/blob/67a065a80fceb7602a9092dc5dec52a959454853/cmd/nerdctl/inspect.go#L82-L142

djdongjin:
Fix #1909

1. Refactor `container|image inspect`: move client creation to `./cmd`.
2. Optimize `inspect`:
    1. Reduce duplicated client/options creation by moving them out of for loop.
    2. Continue to the next arg if error due to no object is found.

### DIFF
--- a/cmd/nerdctl/container_inspect.go
+++ b/cmd/nerdctl/container_inspect.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/clientutil"
 	"github.com/containerd/nerdctl/pkg/cmd/container"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


AkihiroSuda:
> 1. Refactor container|image inspect: move client creation to ./cmd.
> 2. Optimize inspect:

Could you split commits

djdongjin:
> > 1. Refactor container|image inspect: move client creation to ./cmd.
> > 2. Optimize inspect:
> 
> Could you split commits

Done

AkihiroSuda (approved):
Thanks

Issue #1907, opened 2023-01-23
`pkg/cmd`: `NewClient` should be called outside `pkg/cmd`


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/5b8453334d06d176a6938e0222aece9fd8ec148a/pkg/cmd/container/inspect.go#L33

`NewClient` should be called outside `pkg/cmd`, and the `*Client` object should be passed to `pkg/cmd` functions as an argument (right after `ctx`, probably)

PR #1916, opened 2023-01-24
refactor: system events flagging process

davidhsingyuchen:

Major changes:

- Move `_ "github.com/containerd/containerd/api/events"` to the `/pkg` file.
- Move `// Events is from https://github.com/containerd/containerd/blob/v1.4.3/cmd/ctr/commands/events/events.go` to the `/pkg` function.

### DIFF
--- a/pkg/cmd/system/events.go
+++ b/pkg/cmd/system/events.go
@@ -1,0 +1,105 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package system
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"text/template"
+	"time"
+
+	"github.com/containerd/containerd"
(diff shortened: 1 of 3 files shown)


suyanhanx:
`NewClient` should be called outside `pkg/cmd`. #1907 PTAL.

pkg/cmd/system/events.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/containerd/containerd"
+	_ "github.com/containerd/containerd/api/events" // Register grpc event types
+	"github.com/containerd/containerd/events"
+	"github.com/containerd/containerd/log"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/formatter"
+	"github.com/containerd/typeurl"
+)
+
+type Out struct {
```
djdongjin, line 37: Can you add some comment here? Something like `// Out contains information about an event.`

davidhsingyuchen:
> NewClient should be called outside pkg/cmd.

Fixed, thanks!

pkg/cmd/system/events.go
davidhsingyuchen: Fixed, thanks!

Zheaoli (approved):
LGTM, Thanks

pkg/cmd/system/events.go
```diff
@@ -0,0 +1,105 @@
...
+	_ "github.com/containerd/containerd/api/events" // Register grpc event types
+	"github.com/containerd/containerd/events"
+	"github.com/containerd/containerd/log"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/formatter"
+	"github.com/containerd/typeurl"
+)
+
+// Out contains information about an event.
+type Out struct {
```
AkihiroSuda, line 38:
```suggestion
type EventOut struct {
```

djdongjin (approved):

AkihiroSuda (approved):
Thanks

PR #1918, opened 2023-01-24
Add support for calculating virtual image size as well.

manugupt1:
Testing details

```
➜  nerdctl git:(virtual-size) ✗ nerdctl image list --format 'table ID:{{.ID}};Repo:{{.Repository}};Tag:{{.Tag}};Size:{{.Size}};VirtSize:{{.VirtualSize}}'

 table ID:f271e74b17ce;Repo:alpine;Tag:latest;Size:8.3 MiB;VirtSize:8.3 MiB table ID:3487c98481d1;Repo:fedora;Tag:latest;Size:195.1 MiB;VirtSize:195.1 MiB
```

### DIFF
--- a/pkg/cmd/image/list.go
+++ b/pkg/cmd/image/list.go
@@ -119,6 +119,7 @@ type imagePrintable struct {
 	Name         string // image name
 	Size         string // the size of the unpacked snapshots.
 	BlobSize     string // the size of the blobs in the content store (nerdctl extension)
+	VirtualSize  string
 	// TODO: "SharedSize", "UniqueSize", "VirtualSize"
 	Platform string // nerdctl extension
 }
@@ -231,7 +232,7 @@ func (x *imagePrinter) printImageSinglePlatform(ctx context.Context, img images.
 		logrus.WithError(err).Warnf("failed to get blob size of image %q for platform %q", img.Name, platforms.Format(ociPlatform))
 	}
 
-	size, err := imgutil.UnpackedImageSize(ctx, x.snapshotter, image)
+	size, err := imgutil.UnpackedImageSize(ctx, image)
 	if err != nil {
 		logrus.WithError(err).Warnf("failed to get unpacked size of image %q for platform %q", img.Name, platforms.Format(ociPlatform))
 	}
@@ -245,6 +246,7 @@ func (x *imagePrinter) printImageSinglePlatform(ctx context.Context, img images.
 		Tag:          tag,
 		Name:         img.Name,
 		Size:         progress.Bytes(size).String(),
+		VirtualSize:  progress.Bytes(size).String(),
 		BlobSize:     progress.Bytes(blobSize).String(),
 		Platform:     platforms.Format(ociPlatform),
 	}
--- a/pkg/imgutil/imgutil.go
+++ b/pkg/imgutil/imgutil.go
@@ -29,7 +29,6 @@ import (
 	"github.com/containerd/containerd/platforms"
 	refdocker "github.com/containerd/containerd/reference/docker"
 	"github.com/containerd/containerd/remotes"
-	"github.com/containerd/containerd/snapshots"
 	"github.com/containerd/imgcrypt"
 	"github.com/containerd/imgcrypt/images/encryption"
 	"github.com/containerd/nerdctl/pkg/errutil"
@@ -37,7 +36,6 @@ import (
 	"github.com/containerd/nerdctl/pkg/imgutil/dockerconfigresolver"
 	"github.com/containerd/nerdctl/pkg/imgutil/pull"
 	"github.com/docker/docker/errdefs"
-	"github.com/opencontainers/image-spec/identity"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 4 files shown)


AkihiroSuda:
> Size:8.3 MiB;VirtSize:8.3 MiB

This shouldn't be same?

pkg/cmd/image/list.go
```diff
@@ -405,3 +412,21 @@ func unpackedImageSize(ctx context.Context, s snapshots.Snapshotter, img contain
 	}
 	return usage.Size, nil
 }
+
+// https://github.com/moby/moby/blob/65c371ed6e48d0014a4826a6933f9231f688327a/daemon/containerd/image_list.go#L184-L194
+// What is VirtualSize: https://github.com/docker/docs/issues/1520#issuecomment-305179362
+func computeVirtualSize(ctx context.Context, s snapshots.Snapshotter, img containerd.Image) (int64, error) {
+	chainIDs, err := img.RootFS(ctx)
```
AkihiroSuda, line 419: diffIDs, not chainID 
https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/config.md#layer-chainid

pkg/cmd/image/list.go
```diff
@@ -405,3 +412,21 @@ func unpackedImageSize(ctx context.Context, s snapshots.Snapshotter, img contain
...
+	if err != nil {
+		return 0, err
+	}
+	var virtualSize int64
+	for _, chainID := range chainIDs {
+		usage, err := s.Usage(ctx, chainID.String())
```
AkihiroSuda, line 425: Can we cache this to avoid calculating the usage twice

manugupt1:
> https://github.com/docker/docs/issues/1520#issuecomment-305179362

See: https://github.com/docker/docs/issues/16315
I think for images it is the same. For containers when we have a writeable layer; it will change.

manugupt1:
@AkihiroSuda I made the change equivalent to moby's containerd implementation. Should we keep blob size now or remove it?
This is sort of a breaking change; will you like to keep the previous behavior or maintain compatibility with moby.

manugupt1:
@fahedouch Can you please have a look? Let me know what do you think?

pkg/cmd/image/list.go
```diff
@@ -177,7 +175,8 @@ type imagePrintable struct {
 	Name         string // image name
 	Size         string // the size of the unpacked snapshots.
```
fahedouch, line 176: `// the size of the unpacked snapshots.` => wrong

pkg/cmd/image/list.go
```diff
@@ -304,7 +303,8 @@ func (x *imagePrinter) printImageSinglePlatform(ctx context.Context, img images.
 		Tag:          tag,
 		Name:         img.Name,
 		Size:         progress.Bytes(size).String(),
-		BlobSize:     progress.Bytes(blobSize).String(),
+		VirtualSize:  progress.Bytes(virtualSize).String(),
+		BlobSize:     progress.Bytes(size).String(), // Was introduced since docker had a different behavior earlier.
```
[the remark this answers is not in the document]
manugupt1: I think we should keep Size (needed your opinion on it).
This will make tests docker incompatible until docker moves to containerd.

pkg/cmd/image/list.go
```diff
@@ -353,55 +353,20 @@ func (x *imagePrinter) printImageSinglePlatform(ctx context.Context, img images.
...
-
-	info, err := s.Stat(ctx, chainID)
-	if err != nil {
-		return 0, err
-	}
-
-	//add ChainID's parent usage to the total usage
-	if err := snapshotKey(info.Parent).add(ctx, s, &usage); err != nil {
-		return 0, err
+		virtualSize += usage.Size
```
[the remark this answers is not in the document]
manugupt1: For images; it is zero; so it really does not matter.
For containers; the writable layer size will be added.

pkg/cmd/image/list.go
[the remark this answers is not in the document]
fahedouch: `VirtualSize` = `blob size` + `writable layer size`

but here you are not counting the writable layer size, before your modification it was counted by add the parent of most top `DiffId` size to the total size

What I did in the past is 

1. calculate the `chainID` of the readonly layer 
2. add `chainID` size to the total size
3. calculate the `chainID` parent size ( writable layer)
4. add the `chainID` parent size to the total size

```
//add ChainID's parent usage to the total usage
	if err := snapshotKey(info.Parent).add(ctx, s, &usage); err != nil {
```

pkg/cmd/image/list.go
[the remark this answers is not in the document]
fahedouch: @manugupt1 

My understanding is that `VirtualSize` = `blob size` + `writable layer size`

but here you are not counting the writable layer size (before your modification it was counted by adding the parent of most top `DiffId` size to the total size)

What I did in the past is 

1. calculate the `chainID` of the readonly layer 
2. add `chainID` size to the total size
3. calculate the `chainID` parent size ( writable layer)
4. add the `chainID` parent size to the total size

```
info, err := s.Stat(ctx, chainID)
if err != nil {
   return 0, err
}
//add ChainID's parent usage to the total usage
	if err := snapshotKey(info.Parent).add(ctx, s, &usage); err != nil {
```

pkg/cmd/image/list.go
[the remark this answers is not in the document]
fahedouch: SGTM

fahedouch (commented):
Could you please check my comment above.

pkg/cmd/image/list.go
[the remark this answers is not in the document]
manugupt1: @fahedouch Sorry for the late response. I have updated this PR; the previous behavior was correct and when I read through the moby issue; this is the behavior that we want. 
However, now containerd provides withSnapshotUsage so we can re-use it.  I have refactored the code and also added the field for VirtualSize.
Please let me know if this sounds good to you. I can write tests if this sounds good to you.

pkg/cmd/image/list.go
```diff
@@ -245,6 +246,7 @@ func (x *imagePrinter) printImageSinglePlatform(ctx context.Context, img images.
 		Tag:          tag,
 		Name:         img.Name,
 		Size:         progress.Bytes(size).String(),
+		VirtualSize:  progress.Bytes(size).String(),
```
AkihiroSuda, line 249: Needs comment lines to explain why these are same

AkihiroSuda: Also needs integration tests

pkg/cmd/image/list.go
```diff
@@ -119,6 +119,7 @@ type imagePrintable struct {
 	Name         string // image name
 	Size         string // the size of the unpacked snapshots.
 	BlobSize     string // the size of the blobs in the content store (nerdctl extension)
+	VirtualSize  string
```
AkihiroSuda, line 122: Needs a comment line

pkg/imgutil/imgutil.go
```diff
@@ -376,55 +374,8 @@ func ParseRepoTag(imgName string) (string, string) {
...
-		return 0, err
-	}
-
-	//add ChainID's parent usage to the total usage
-	if err := snapshotKey(info.Parent).add(ctx, s, &usage); err != nil {
-		return 0, err
-	}
-	return usage.Size, nil
+func UnpackedImageSize(ctx context.Context, img containerd.Image) (int64, error) {
+	return img.Usage(ctx, containerd.WithSnapshotUsage())
```
AkihiroSuda, line 380: Is this just a refactoring, or does this change the computed size?
Could you consider making this a separate commit/PR ?

manugupt1:
@fahedouch Ping

pkg/imgutil/imgutil.go
manugupt1: It does not change the computed size at all. I verified locally.

pkg/cmd/image/list.go
manugupt1: yep! Wanted to get an initial look. Will add by the end of this week.
Thanks!

manugupt1:
Even the sizes are inconistent now with docker and nerdctl.
See: https://github.com/moby/moby/issues/45180
Closing this.

fahedouch:
@manugupt1 why closing this ?

manugupt1:
I mostly closed this because of https://github.com/moby/moby/issues/45180
I will open a new one for VirtualSize only.

PR #1923, opened 2023-01-24
refactor: extract main logic of image.List

### LINKED ISSUE #1855: how to reuse only the "get images" logic of image.List

Currently [`image.List`](https://github.com/containerd/nerdctl/blob/cfe96a9d17b293463ad754ebdae6f0b933d7ae39/pkg/cmd/image/list.go#L49) contains 2 steps:

1. Get the images according to the filters.
2. Print the images.

It'd be nice to have an exported function that only does the first step. The detailed rationale is very similar to https://github.com/containerd/nerdctl/pull/1837#discussion_r1069763156, and this time instead of directly creating a PR, I think opening an issue first probably makes more sense.

Looking at the fields of [`types.ImageListCommandOptions`](https://github.com/containerd/nerdctl/blob/cfe96a9d17b293463ad754ebdae6f0b933d7ae39/pkg/api/types/image_type.go#L19), many of them are only needed by `2. Print the images`, so maybe we can create a new function and make it only depend on the things it needs:

```go
func ListTheNewFunc(ctx context.Context, namespace, address string, Filters, NameAndRefFilter []string) ([]images.Image, error)
```

or maybe it can just take a `*containerd.Client` as the caller of this function can be potentially a long-running process, and it can call `clientutil.NewClient` itself to instantiate the client only once during process initialization:

```go
func ListTheNewFunc(ctx context.Context, client *containerd.Client, Filters, NameAndRefFilter []string) ([]images.Image, error)
```

I don't have a perfect name in mind for the new function, so I just use `ListTheNewFunc` as a placeholder, and I'm not even sure if creating a new function makes sense to you, and I appreciate any suggestion, thanks!

### Additional context

I'm more than happy to work on this as long as the community reaches a consensus on the approach.

davidhsingyuchen:
Fix #1855.

The naming pattern of `ListCommandHandler` (i.e., called by `cmd/nerdctl`) and `List` (i.e., contains "main logic") come from https://github.com/containerd/nerdctl/issues/1855#issuecomment-1397034834. If this PR makes sense to you, to make the naming consistent, I can create a one-shot PR to rename all functions to be `XXXCommandHandler` according to https://github.com/containerd/nerdctl/issues/1855#issuecomment-1400837364.

Notes:

- The function comment of `List` comes from [command reference](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-blue_square-nerdctl-images). Not 100% sure about the comment of `nameAndRefFilter`, please let me know if it's inaccurate, thanks!

### DIFF
--- a/pkg/cmd/image/list.go
+++ b/pkg/cmd/image/list.go
@@ -44,16 +44,38 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func List(ctx context.Context, client *containerd.Client, options types.ImageListOptions) error {
-	var imageStore = client.ImageService()
-	imageList, err := imageStore.List(ctx, options.NameAndRefFilter...)
+// ListCommandHandler `List` and print images matching filters in `options`.
+func ListCommandHandler(ctx context.Context, client *containerd.Client, options types.ImageListOptions) error {
+	imageList, err := List(ctx, client, options.Filters, options.NameAndRefFilter)
 	if err != nil {
 		return err
 	}
-	if len(options.Filters) > 0 {
-		f, err := imgutil.ParseFilters(options.Filters)
+	return printImages(ctx, client, imageList, options)
+}
+
+// List queries containerd client to get image list and only returns those matching given filters.
+//
+// Supported filters:
+// - before=<image>[:<tag>]: Images created before given image (exclusive)
+// - since=<image>[:<tag>]: Images created after given image (exclusive)
+// - label=<key>[=<value>]: Matches images based on the presence of a label alone or a label and a value
+// - dangling=true: Filter images by dangling
+// - reference=<image>[:<tag>]: Filter images by reference (Matches both docker compatible wildcard pattern and regexp
+//
+// nameAndRefFilter has the format of `name==(<image>[:<tag>])|ID`,
+// and they will be used when getting images from containerd,
+// while the remaining filters are only applied after getting images from containerd,
+// which means that having nameAndRefFilter may speed up the process if there are a lot of images in containerd.
+func List(ctx context.Context, client *containerd.Client, filters, nameAndRefFilter []string) ([]images.Image, error) {
+	var imageStore = client.ImageService()
+	imageList, err := imageStore.List(ctx, nameAndRefFilter...)
+	if err != nil {
+		return nil, err
+	}
+	if len(filters) > 0 {
+		f, err := imgutil.ParseFilters(filters)
 		if err != nil {
-			return err
+			return nil, err
 		}
 
 		if f.Dangling != nil {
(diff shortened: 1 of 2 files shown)


pkg/cmd/image/list.go
```diff
@@ -45,22 +45,42 @@ import (
...
 	client, ctx, cancel, err := clientutil.NewClient(ctx, options.GOptions.Namespace, options.GOptions.Address)
 	if err != nil {
 		return err
 	}
 	defer cancel()
+	imageList, err := List(ctx, client, options.Filters, options.NameAndRefFilter)
+	if err != nil {
+		return err
+	}
+	return printImages(ctx, client, imageList, options)
```
djdongjin, line 58: Given that line 49-53 (client creation) will be moved to `cmd` side, I wonder if we should just export `printImages` (i.e. -> `PrintImages`) and let `cmd.imagesAction` first call `List` below  and then call `PrintImages` to print.

I'm also okay to have the `ListCommandHandler` (naming wise, maybe just `ListHandler`, or `ListAndPrint`?) as a wrapper of the two logic (list and print).

WDYT @Zheaoli @AkihiroSuda

davidhsingyuchen: > and then call PrintImages to print.

If we take this route, maybe we also want to make it consistent across all the commands (i.e., if there's anything to print, `cmd/nerdctl/X_Y.go` should call `PrintXXX` exported by `pkd/cmd/X/Y.go` to do so).

davidhsingyuchen:
Rebased on `main` to resolve merge conflicts and updated PR description accordingly.

pkg/cmd/image/list.go
```diff
@@ -44,16 +44,37 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func List(ctx context.Context, client *containerd.Client, options types.ImageListOptions) error {
-	var imageStore = client.ImageService()
-	imageList, err := imageStore.List(ctx, options.NameAndRefFilter...)
+func ListCommandHandler(ctx context.Context, client *containerd.Client, options types.ImageListOptions) error {
```
djdongjin, line 47: Can you add a comment here? Something like 

```
// ListCommandHandler `List` and print images matching filters in `options`.
```

djdongjin (approved):

pkg/cmd/image/list.go
davidhsingyuchen: > let cmd.imagesAction first call List below and then call PrintImages to print.

Also, if there's no single entrypoint for `cmd.imagesAction` to call, should we pass the whole `types.ImageListOptions` to both `List` and `PrintImages`? Some fields in `types.ImageListOptions` are only relevant to `List`, and some of them are only relevant to `PrintImages`.

AkihiroSuda (approved):
Thanks

PR #1924, opened 2023-01-25
[Refactor] refactor image save flagging process 

suyanhanx:
part of #1680 

Checklist:

* [x]  Create a file in `pkg/api/types/${cmd}_types.go`, and define the CommandOption for this command
* [x]  Create some file in `pkg/cmd/${cmd}`, and move the command entry point in real into this package

Signed-off-by: suyanhanx <suyanhanx@gmail.com>

### DIFF
--- a/pkg/cmd/image/save.go
+++ b/pkg/cmd/image/save.go
@@ -1,0 +1,90 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package image
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/images/archive"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/idutil/imagewalker"
+	"github.com/containerd/nerdctl/pkg/platformutil"
+	"github.com/containerd/nerdctl/pkg/strutil"
+	"github.com/mattn/go-isatty"
+)
+
+// Save will save one or more images to a tar archive (streamed to STDOUT by default).
+func Save(ctx context.Context, client *containerd.Client, images []string, options types.ImageSaveOptions) error {
+	if options.Output != "" {
+		f, err := os.OpenFile(options.Output, os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		options.Stdout = f
+	} else {
+		if isatty.IsTerminal(os.Stdout.Fd()) {
+			return fmt.Errorf("cowardly refusing to save to a terminal. Use the -o flag or redirect")
+		}
+	}
+
+	return SaveImages(ctx, client, images, options)
+}
+
+// SaveImages exports `images` to a `io.Writer` (e.g., a file writer, or os.Stdout) specified by `opt.Stdout`.
+func SaveImages(ctx context.Context, client *containerd.Client, images []string, options types.ImageSaveOptions, exportOpts ...archive.ExportOpt) error {
+	images = strutil.DedupeStrSlice(images)
+
+	platMC, err := platformutil.NewMatchComparer(options.AllPlatforms, options.Platform)
+	if err != nil {
+		return err
+	}
+
+	exportOpts = append(exportOpts, archive.WithPlatform(platMC))
+	imageStore := client.ImageService()
+
+	savedImages := make(map[string]struct{})
+	walker := &imagewalker.ImageWalker{
+		Client: client,
+		OnFound: func(ctx context.Context, found imagewalker.Found) error {
+			if found.UniqueImages > 1 {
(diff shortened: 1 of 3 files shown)


pkg/cmd/image/save.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/images/archive"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/idutil/imagewalker"
+	"github.com/containerd/nerdctl/pkg/platformutil"
+	"github.com/containerd/nerdctl/pkg/strutil"
+	"github.com/mattn/go-isatty"
+)
+
+func Save(ctx context.Context, client *containerd.Client, images []string, options types.ImageSaveOptions) error {
```
djdongjin, line 34: Can you add some comment here?

pkg/cmd/image/save.go
```diff
@@ -0,0 +1,90 @@
...
+	writer := options.Stdout
+	if options.Output != "" {
+		f, err := os.OpenFile(options.Output, os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		writer = f
```
djdongjin, line 42: should we update the `options.Stdout` in place instead of passing a `wirter` to `saveImages`? Otherwise, `saveImages` has two `io.writer` inputs (`writer` and `opt.Stdout`) and sometimes they're the same and sometimes not.

pkg/cmd/image/save.go
```diff
@@ -0,0 +1,90 @@
...
+	} else {
+		if isatty.IsTerminal(os.Stdout.Fd()) {
+			return fmt.Errorf("cowardly refusing to save to a terminal. Use the -o flag or redirect")
+		}
+	}
+
+	return saveImages(ctx, client, images, writer, options)
+}
+
+func saveImages(ctx context.Context, client *containerd.Client, images []string, writer io.Writer, opt types.ImageSaveOptions, exportOpts ...archive.ExportOpt) error {
```
djdongjin, line 52: Can we also expose this func? It makes sense to me that caller has own `io.writer` and call this func directly (i.e., they don't need us create file writer for them, as in `Save`)

pkg/cmd/image/save.go
suyanhanx: Sure.

pkg/cmd/image/save.go
suyanhanx: Makes sense. Currently, there is only one reference to it, so I kept it not exposed.

pkg/cmd/image/save.go
suyanhanx: Make sense. But their types are not always the same.

pkg/cmd/image/save.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		defer f.Close()
+		options.Stdout = f
+	} else {
+		if isatty.IsTerminal(os.Stdout.Fd()) {
+			return fmt.Errorf("cowardly refusing to save to a terminal. Use the -o flag or redirect")
+		}
+	}
+
+	return SaveImages(ctx, client, images, options.Stdout, options)
```
djdongjin, line 49: sorry I mean we can remove the `writer` arg in `SaveImages` right? Since `options` already has the `writer` from `options.Stdout`.

I think we should avoid having overlap between args within `options` and other args in a func (e.g., here `writer` v.s. `opt.Stdout`).

pkg/cmd/image/save.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func SaveImages(ctx context.Context, client *containerd.Client, images []string, writer io.Writer, opt types.ImageSaveOptions, exportOpts ...archive.ExportOpt) error {
```
djdongjin, line 52: Since it's exported now, can you add a comment here? Something like 

```
// SaveImages exports `images` to a `io.Writer` (e.g., a file writer, or os.Stdout) specified by `opt.Stdout`.
```

pkg/cmd/image/save.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		options.Stdout = f
+	} else {
+		if isatty.IsTerminal(os.Stdout.Fd()) {
+			return fmt.Errorf("cowardly refusing to save to a terminal. Use the -o flag or redirect")
+		}
+	}
```
djdongjin, line 47: (below is more for discussion and see other reviewers' thought).

I think we should remove `Output string` from `ImageSaveOptions`. Functionally, it represents the same thing as `Stdout io.Writer` and having both might cause inconsistency and confusing, for example:

- Even `Output != ""`, we init `Stdout = cmd.OutOrStdout()` when creating `ImageSaveOptions`, and then change it after the file writer created (i.e., here).
- Should a caller specify `Output`, or `Stdout io.Writer`?

If we remove `Output` and keep `Stdout` only:

1. `pkg/cmd` side logic doesn't need to know where the writer comes from. It just writes content to it.
2. On `cmd` side, we init `opt.Stdout = cmd.OutOrStdout()` if `output == ""` or create a file writer and use it as `opt.Stdout` (i.e., this pard of code).

suyanhanx: I think your thought is a pretty good idea.

pkg/cmd/image/save.go
suyanhanx: Yep. I'll remove it.

pkg/cmd/image/save.go
suyanhanx: Sure.

pkg/cmd/image/save.go
suyanhanx: /cc @AkihiroSuda @Zheaoli

Zheaoli: I think it's better to make another PR to simplify the logic. I prefer keep the same code here, (one PR for one thing)

Zheaoli (approved):
LGTM, Thanks

djdongjin (approved):

PR #1938, opened 2023-01-28
feat: add network in use check to `compose down`

### LINKED ISSUE #1425: `nerdctl compose down` with orphan container should keep the same behavior with docker

Retrospection:
Signed-off-by: YuRuosheng <371373446@qq.com>

Close #1425


Zheaoli (approved):
LGTM

djdongjin:
Can you add an integration test for this scenario?

Retrospection:
> Can you add an integration test for this scenario?

I may try~

cmd/nerdctl/compose_down_linux_test.go
```diff
@@ -24,6 +24,44 @@ import (
...
+	compFull := testutil.NewComposeDir(t, dockerComposeYAMLFull)
+	defer compFull.CleanUp()
+
+	projectName := fmt.Sprintf("nerdctl-compose-test-%d", time.Now().Unix())
+	t.Logf("projectName=%q", projectName)
+
+	// orphanContainer := fmt.Sprintf("%s_orphan_1", projectName)
+
+	base.ComposeCmd("-p", projectName, "-f", compFull.YAMLFullPath(), "up", "-d").AssertOK()
+	// defer base.ComposeCmd("-p", projectName, "-f", compFull.YAMLFullPath(), "down", "-v").Run()
```
fahedouch, line 59: should be deleted ?

pkg/cmd/compose/compose.go
```diff
@@ -58,6 +58,21 @@ func New(client *containerd.Client, globalOptions types.GlobalCommandOptions, op
 		return false, nil
 	}
 
+	options.NetworkInUse = func(ctx context.Context, netName string) (bool, error) {
+		usedMap, err := netutil.UsedNetworks(ctx, client)
+		if err != nil {
+			return false, err
+		}
+		for _, v := range usedMap {
+			for _, net := range v {
```
fahedouch, line 67: `v` here is a container slice not a network slice, right ?

pkg/composer/down.go
```diff
@@ -97,6 +97,15 @@ func (c *Composer) downNetwork(ctx context.Context, shortName string) error {
 	if err != nil {
 		return err
 	} else if netExists {
+		netUsed, err := c.NetworkInUse(ctx, fullName)
+		if err != nil {
+			return err
+		}
+		if netUsed {
+			logrus.Errorf("network %s is in use", fullName)
+			return nil
```
fahedouch, line 106: why not `return fmt.Errorf("network %s is in use", fullName)` ?

fahedouch (changes requested):

cmd/nerdctl/compose_down_linux_test.go
```diff
@@ -24,6 +24,44 @@ import (
...
+
+	base.ComposeCmd("-p", projectName, "-f", compOrphan.YAMLFullPath(), "down", "-v").AssertCombinedOutContains("is in use")
+	// base.ComposeCmd("-p", projectName, "-f", compFull.YAMLFullPath(), "ps").AssertOutNotContains(orphanContainer)
```
fahedouch, line 62: would you please explain your intention behind this integration test ? Thanks

AkihiroSuda:
> Signed-off-by: Retros

Please use a real name, as explained in https://github.com/cncf/foundation/issues/383#issuecomment-1178254458

cmd/nerdctl/compose_down_linux_test.go
Retrospection: yep，I pushed this commit just want to try to see if I can pass the CI test~ 

Will refactor this

cmd/nerdctl/compose_down_linux_test.go
Retrospection: I make this  integration test to ensure when run `compose down` command while the network is in use, the stderr will output error message "network xxxxx is in use". This behavior proves that the network in use inspection is working.

pkg/cmd/compose/compose.go
Zheaoli: Yes，I think we should use the key as the network name

Retrospection: > `v` here is a container slice not a network slice, right ?

nope, `useMap` is a map which key is namespace and value is network names.

suyanhanx:
Please also update your DCO sign in the PR description.

cmd/nerdctl/compose_down_linux_test.go
```diff
@@ -24,6 +24,43 @@ import (
...
+	compOrphan := testutil.NewComposeDir(t, dockerComposeYAMLOrphan)
+	defer compOrphan.CleanUp()
+	compFull := testutil.NewComposeDir(t, dockerComposeYAMLFull)
+	defer compFull.CleanUp()
+
+	projectName := fmt.Sprintf("nerdctl-compose-test-%d", time.Now().Unix())
+	t.Logf("projectName=%q", projectName)
+
+	base.ComposeCmd("-p", projectName, "-f", compFull.YAMLFullPath(), "up", "-d").AssertOK()
+	base.ComposeCmd("-p", projectName, "-f", compOrphan.YAMLFullPath(), "down", "-v").AssertCombinedOutContains("is in use")
```
suyanhanx, line 61: Need to clean up the orphan container.

PR #1943, opened 2023-01-28
compose: show detailed errors again

### LINKED ISSUE #1942: [Regression in v0.23.0] Compose does not print detailed error
```yaml
services:
  foo:
    image: hello-world
    runtime: invalid
```

v0.22.2
```
$ nerdctl compose up
INFO[0000] Ensuring image hello-world                   
INFO[0000] Re-creating container tmp_foo_1              
FATA[0000] failed to create shim task: OCI runtime create failed: unable to retrieve OCI runtime error (open /run/containerd/io.containerd.runtime.v2.task/default/88a3ee42b9cc54ee3a275b4a21eec2e610d8fd0470353bf3f89aa987acdf50c6/log.json: no such file or directory): exec: "invalid": executable file not found in $PATH: unknown 
FATA[0000] error while creating container tmp_foo_1: exit status 1
```

v0.23.0
```yaml
$ nerdctl compose up
INFO[0000] Ensuring image hello-world                   
INFO[0000] Re-creating container tmp_foo_1              
FATA[0000] error while creating container tmp_foo_1: exit status 1
```

Still not fixed in the main branch.

Seems a regression in 3296c1430679ffede4035744395ff4234296f27c

AkihiroSuda:
The detailed error was not printed since 3296c14 (v0.23.0).

Fix #1942

### DIFF
--- a/pkg/composer/up_service.go
+++ b/pkg/composer/up_service.go
@@ -140,8 +140,10 @@ func (c *Composer) upServiceContainer(ctx context.Context, service *serviceparse
 	defer os.RemoveAll(tempDir)
 	cidFilename := filepath.Join(tempDir, "cid")
 
-	if container.Detached && !service.Unparsed.StdinOpen && !service.Unparsed.Tty {
+	var runFlagD bool
+	if !service.Unparsed.StdinOpen && !service.Unparsed.Tty {
 		container.RunArgs = append([]string{"-d"}, container.RunArgs...)
+		runFlagD = true
 	}
 
 	//add metadata labels to container https://github.com/compose-spec/compose-spec/blob/master/spec.md#labels
@@ -164,11 +166,11 @@ func (c *Composer) upServiceContainer(ctx context.Context, service *serviceparse
 	if service.Unparsed.StdinOpen {
 		cmd.Stdin = os.Stdin
 	}
-
-	if !container.Detached {
+	if !runFlagD {
 		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
 	}
+	// Always propagate stderr to print detailed error messages (https://github.com/containerd/nerdctl/issues/1942)
+	cmd.Stderr = os.Stderr
 
 	err = cmd.Run()
 	if err != nil {
--- a/cmd/nerdctl/compose_up_test.go
+++ b/cmd/nerdctl/compose_up_test.go
@@ -1,0 +1,52 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"runtime"
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+	"gotest.tools/v3/icmd"
+)
+
+// https://github.com/containerd/nerdctl/issues/1942
+func TestComposeUpDetailedError(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pkg/composer/up_service.go
```diff
@@ -165,10 +165,15 @@ func (c *Composer) upServiceContainer(ctx context.Context, service *serviceparse
...
 	}
 
+	// Containers are always detached, even for `nerdctl compose up --detach=false`.
+	// (In that case, containers are detached and then the log viewer is "attached")
+	//
+	// See pkg/composer/serviceparser/serviceparser.go
 	if !container.Detached {
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+		panic("container.Detached can't be false for compose mode")
```
djdongjin, line 173: Should it return an error or panic directly?

AkihiroSuda: Intentionally panicking, because unlike other errors this error must not happen with any user operation.

djdongjin (approved):

pkg/composer/up_service.go
```diff
@@ -165,10 +165,15 @@ func (c *Composer) upServiceContainer(ctx context.Context, service *serviceparse
 		cmd.Stdin = os.Stdin
 	}
 
+	// Containers are always detached, even for `nerdctl compose up --detach=false`.
```
fahedouch, line 168: This message is confusing. containers created with `nerdctl  run` may be detached or not but compose should always read from container std streams no matter the underlying container is detached or not . In other worder `detach` option should have no effect on the compose level but should keep the standard effect on underlying `nerdcrl run` cmd

AkihiroSuda: - `nerdctl compose up --detach=true`: launches `nerdctl run -d`
- `nerdctl compose up --detach=false`: launches `nerdctl run -d` (same as above), and then "attach" `nerdctl logs`

fahedouch: Detach should be true to run nerdctl with `-d`  (see https://github.com/containerd/nerdctl/blob/4d9133f76f36efac6115aee2806c67a9b2ef2e8c/pkg/composer/up_service.go#L144)

fahedouch: `nerdctl compose up --detach=false` : why `nerdctl run -d` then attach the log viewer . Why not `nerdctl run` and connect cmd std streams to standard std streams ?

AkihiroSuda: To allow `nerdctl compose logs` after running `nerdctl compose up --detach=false`

fahedouch: OK I forgot that we enforced to write to log file even without `-d` . But I am not agreed with the fact that the underlying `nerdctl run` cmd is always running with the `-d` flag.

AkihiroSuda: You're right, the `container.Detached` is always true, but `nerdctl run` isn't always launched with `-d`.

I guess we should just remove the `container.Detached` struct field.

AkihiroSuda: Updated

manugupt1 (approved):

pkg/composer/up_service.go
fahedouch: that means `container.Detached` have no effect. Users can no longer drop stdout by setting `container.Detached` into the yaml definition ( I believe  it is not compatible with docker too)

I think we should keep `container.Detached` but not always true and then forward  cmd std streams to os std.

that means:

- `container.detached = true` :  `nerdctl run -d ` will log nothing => os std streams will receive nothing
- `container.detached = false` :  `nerdctl run` may log data => os std streams will receive data

```
	if container.Detached && !service.Unparsed.StdinOpen && !service.Unparsed.Tty {
		container.RunArgs = append([]string{"-d"}, container.RunArgs...)
	}

	if service.Unparsed.StdinOpen != service.Unparsed.Tty {
		return "", fmt.Errorf("currently StdinOpen(-i) and Tty(-t) should be same")
	}
	if service.Unparsed.StdinOpen {
		cmd.Stdin = os.Stdin
	}
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
```

AkihiroSuda: `Detached` was never settable via the YAML AFAICS
https://github.com/compose-spec/compose-spec/blob/master/spec.md

fahedouch: ok but strange! LGTM in this case

Issue #1946, opened 2023-01-28
[Regression in v1.2.0] `nerdctl logs` exits silently for `nerdctl run` w/o `-d`


AkihiroSuda:
```bash
nerdctl run --name=foo alpine echo foo
nerdctl logs foo
```

v1.1.0 fails as expected (because `-d` is not specified and the logging driver is not enabled)
```
FATA[0000] open /home/suda/.local/share/nerdctl/1935db59/containers/default/ed96e51c4c5f9b3d64cae624cba6b33ff3b9db3ed3b0224cc4bb02d1cbfc1a5c/log-config.json: no such file or directory
```

The main branch  exits silently, with exit code 0.

Regression in c67b102cb9d94a7df2657e410f28872f64555edd

fahedouch:
Because of this https://github.com/containerd/nerdctl/commit/c67b102cb9d94a7df2657e410f28872f64555edd#diff-464b1e1a904b2f79ca04e2202190403c19f87a135b6a9e1ef2e65fa52772a221R85
 
Condition should be replaced by  `if logURI != ""`

AkihiroSuda:
> Because of this [c67b102#diff-464b1e1a904b2f79ca04e2202190403c19f87a135b6a9e1ef2e65fa52772a221R85](https://github.com/containerd/nerdctl/commit/c67b102cb9d94a7df2657e410f28872f64555edd#diff-464b1e1a904b2f79ca04e2202190403c19f87a135b6a9e1ef2e65fa52772a221R85)
> 
> Condition should be replaced by `if logURI != ""`

Thanks, would you like to open a PR?

fahedouch:
> > Because of this [c67b102#diff-464b1e1a904b2f79ca04e2202190403c19f87a135b6a9e1ef2e65fa52772a221R85](https://github.com/containerd/nerdctl/commit/c67b102cb9d94a7df2657e410f28872f64555edd#diff-464b1e1a904b2f79ca04e2202190403c19f87a135b6a9e1ef2e65fa52772a221R85)
> > Condition should be replaced by `if logURI != ""`
> 
> Thanks, would you like to open a PR?

Sure

fahedouch:
We have an other issue here 
`nerdctl run --name=foo alpine echo foo` is not printing `foo` to stdout. Looks like `nerdctl logs` is not triggered behind the scene. Is it normal ?

AkihiroSuda:
[...]

PR #1947, opened 2023-01-28
fix nerdctl logs for run without flagD

fahedouch:


fahedouch:
test fails because of [this](https://github.com/containerd/nerdctl/issues/1946#issuecomment-1407468705). when `nerdctl nerdctl runs without `-d` , `nerdctl logs` should be triggered to print logs to stdout  in parallel with writing to log files

apostasie:
@fahedouch wondering if this is still a thing?

PR #1952, opened 2023-01-30
go.mod: github.com/containerd/cgroups/v3 v3.0.1, github.com/docker/docker v24.0.0

AkihiroSuda:
https://github.com/containerd/cgroups/releases/tag/v3.0.0


djdongjin:
There is also https://github.com/containerd/typeurl/releases/tag/v2.0.0

Should we update it as well?

https://github.com/containerd/nerdctl/blob/ca4a76377411295f8b77c5e3b16472e7bf318a34/go.mod#L20

AkihiroSuda:
Looks like `nerdctl stats` hangs up with this

AkihiroSuda:
> There is also https://github.com/containerd/typeurl/releases/tag/v2.0.0
> 
> Should we update it as well?
> 
> https://github.com/containerd/nerdctl/blob/ca4a76377411295f8b77c5e3b16472e7bf318a34/go.mod#L20

Not sure how it is relevant to this PR, but I'd prefer to wait for v2.1.0 (https://github.com/containerd/typeurl/pull/40)

djdongjin:
> Not sure how it is relevant to this PR, but I'd prefer to wait for v2.1.0 ([containerd/typeurl#40](https://github.com/containerd/typeurl/pull/40))

SGTM, it's not relevant to this PR, just for awareness since dependabot seems skip checking those dependencies with a specific commit. :)

cmd/nerdctl/container_stats_linux.go
```diff
@@ -23,8 +23,8 @@ import (
 	"strings"
 	"time"
 
-	v1 "github.com/containerd/cgroups/stats/v1"
-	v2 "github.com/containerd/cgroups/v2/stats"
```
fuweid, line 27: The line48 return error so that the `stats` command repeats to receive the metrics.
The repro steps is 

```
nerdctl rm testing
nerdctl run --name testing -d alpine sleep 5 && nerdctl stats --no-stream
```

Both versions are using the same name. 
* https://github.com/containerd/cgroups/blob/v3.0.1/cgroup1/stats/metrics.pb.go#L1705
* https://github.com/containerd/cgroups/blob/v1.1.0/stats/v1/metrics.pb.go#L705
* 
And the typeurl.UnmarshalAny prefers to use the type registerd by  `gogoproto.RegisterType`. So it will convert the data into v1 or v2. I think it is bug in cgroup v3. we should use diffierent package. 

https://github.com/containerd/cgroups/blob/49edf542aa03131efa6545a292f1deec4be5290e/cgroup1/stats/metrics.proto#L3

And it might cause regression if there are existing running shims using old version cgroup package.

https://github.com/containerd/containerd/blob/release/1.6/go.mod#L11

fuweid (approved):
LGTM

PR #1965, opened 2023-02-02
Refactor (un)pause in nerdctl and compose

djdongjin:
Part of #1680.

Refactor `nerdctl unpause`, `nerdctl compose (un)pause`.

### DIFF
--- a/pkg/cmd/container/unpause.go
+++ b/pkg/cmd/container/unpause.go
@@ -28,15 +28,15 @@ import (
 	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
 )
 
-// Pause pauses all containers specified by `reqs`.
-func Pause(ctx context.Context, client *containerd.Client, reqs []string, options types.ContainerPauseOptions) error {
+// Unpause unpauses all containers specified by `reqs`.
+func Unpause(ctx context.Context, client *containerd.Client, reqs []string, options types.ContainerUnpauseOptions) error {
 	walker := &containerwalker.ContainerWalker{
 		Client: client,
 		OnFound: func(ctx context.Context, found containerwalker.Found) error {
 			if found.MatchCount > 1 {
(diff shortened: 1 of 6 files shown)


pkg/cmd/container/unpause.go
```diff
@@ -0,0 +1,54 @@
...
+			return err
+		},
+	}
+	for _, req := range reqs {
+		n, err := walker.Walk(ctx, req)
+		if err != nil {
+			return err
+		} else if n == 0 {
+			return fmt.Errorf("no such container %s", req)
+		}
```
ginglis13, line 51: would this fail fast and is that expected? i.e., an alternative would be notifying the error but continuing to walk reqs and attempt to unpause those, perhaps returning multierror

djdongjin: yeah I also prefer to proceed to the next and print all errors at the end. 

I'm planning to keep the implementation unchanged in this refactor PR and fix in a follow-up PR, since it's not only in (un)pause, we have similar logic to update in `kill`/`start`/`kill`/`update` and some image subcommands.

djdongjin: And docker also did this printing errors at the end: https://github.com/docker/cli/blob/e92dd87c3209361f29b692ab4b8f0f9248779297/cli/command/container/kill.go#L50-L59

djdongjin: :) decided to update (un)pause to continue to walk and return all errors (combined) at the end if there is any. Didn't update compose (un)pause since it requires more changes and the current behavior matches docker compose:

https://github.com/docker/compose/blob/1640f155e94468e76ccb0b4f16c9592983f22aec/pkg/compose/pause.go#L47-L59

Zheaoli (approved):
LGTM, Thanks

AkihiroSuda:
@Zheaoli Please make sure to set the milestone next time, thanks

PR #1966, opened 2023-02-02
refactor: container list flagging process

davidhsingyuchen:


pkg/api/types/container_types.go
```diff
@@ -149,3 +149,24 @@ type ContainerExecOptions struct {
 	// Username or UID (format: <name|uid>[:<group|gid>])
 	User string
 }
+
+type ContainerListOptions struct {
```
djdongjin, line 153: Can you add a brief comment here?

pkg/cmd/container/list.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/containerd/containerd/pkg/progress"
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/containerd/nerdctl/pkg/formatter"
+	"github.com/containerd/nerdctl/pkg/imgutil"
+	"github.com/containerd/nerdctl/pkg/labels"
+	"github.com/containerd/nerdctl/pkg/labels/k8slabels"
+	"github.com/sirupsen/logrus"
+)
+
+func List(ctx context.Context, client *containerd.Client, options types.ContainerListOptions) error {
```
djdongjin, line 44: can you add a brief comment here?

davidhsingyuchen: > // List prints containers according to `options`.

Since there're things other than `Filters` that control what containers should be printed (e.g., `All`, `LastN`, etc.), I didn't go into explaining the details. Please let me know if this works for you, thanks!

djdongjin: SGTM.

AkihiroSuda (approved):
Thanks

PR #1970, opened 2023-02-03
[Refactor] Refactor the (container) create command flagging process

Laitr0n:
Checklist:

- [x] Create a file in pkg/api/types/container_types.go, and define the CommandOption for this command


cmd/nerdctl/container_create.go
```diff
@@ -50,36 +52,530 @@ func newCreateCommand() *cobra.Command {
 	return createCommand
 }
 
-func createAction(cmd *cobra.Command, args []string) error {
+func processContainerCreateOptions(cmd *cobra.Command) (types.ContainerCreateOptions, error) {
```
djdongjin, line 55: Suggest to use named return here (eg "opt types.ContainerCreatOptions, err error"), so you can simply "return" (instead of "return types.containercreateoptions{}, err"). You can take a look at "processContainerInspectOptions" for reference.

cmd/nerdctl/container_create.go
```diff
@@ -50,36 +52,529 @@ func newCreateCommand() *cobra.Command {
...
+	}
+	// #endregion
+	// #region for ipfs flags
+	ipfsAddressStr, err := cmd.Flags().GetString("ipfs-address")
+	if err != nil {
+		return
+	}
+	// #endregion
+
+	return types.ContainerCreateOptions{
```
djdongjin, line 457: nit: you can assign the value from `cmd` directly to corresponding field in `opt`, thus avoiding the whole repeated assignment again in line 457-548. For example:

```golang
	opt.IPFSAddress, err = cmd.Flags().GetString("ipfs-address")
	if err != nil {
		return
	}
```

I know writing this is tedious so no change is requested here :) I'm okay with this format. Actually, if you plan to make the change, I'd suggest to make a copy of the current format just in case other reviewers prefer this (more consistent with other `processXXXOptions`, though verbose)

Laitr0n: SGTM. 
I didn't notice that this consistent style has changed, and I will change it.

pkg/containerutil/containerutil.go
```diff
@@ -438,3 +441,36 @@ func Unpause(ctx context.Context, client *containerd.Client, id string) error {
...
+		return fmt.Errorf("container ID file found, make sure the other container isn't running or delete %s", path)
+	} else if errors.Is(err, os.ErrNotExist) {
+		f, err := os.Create(path)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		if err != nil {
+			return fmt.Errorf("failed to create the container ID file: %s", err)
+		}
```
djdongjin, line 457: The 2nd err check is redundant

pkg/api/types/container_types.go
```diff
@@ -40,6 +40,259 @@ type ContainerKillOptions struct {
...
+type ContainerCreateOptions struct {
+	Stdout io.Writer
+	Stderr io.Writer
+	// GOptions is the global options
+	GOptions GlobalCommandOptions
+
+	// NerdctlCmd is the command name of nerdctl
+	NerdctlCmd string
+	// NerdctlArgs is the arguments of nerdctl
+	NerdctlArgs []string
```
djdongjin, line 53: for myselfl: check how these are used on pkg side

cmd/nerdctl/container_create.go
```diff
@@ -50,36 +51,433 @@ func newCreateCommand() *cobra.Command {
 	return createCommand
 }
 
-func createAction(cmd *cobra.Command, args []string) error {
-	globalOptions, err := processRootCmdFlags(cmd)
+func processContainerCreateOptions(cmd *cobra.Command) (opt types.ContainerCreateOptions, err error) {
+	opt = types.ContainerCreateOptions{}
```
djdongjin, line 55: I believe with named return (opt), you don't need to explict init again.

cmd/nerdctl/container_create.go
```diff
@@ -50,36 +51,433 @@ func newCreateCommand() *cobra.Command {
...
+	// #region for init process flags
+	opt.InitProcessFlag, err = cmd.Flags().GetBool("init")
+	if err != nil {
+		return
+	}
+	opt.InitBinaryChanged = cmd.Flags().Changed("init-binary")
+	opt.InitBinary, err = cmd.Flags().GetString("init-binary")
+	if err != nil {
+		return
+	}
```
djdongjin, line 120: can we use a string pointer (InitBinary *string)? so we know `cmd.Flags().Changed("init-binary")` is true if the pointer is not nil.

djdongjin: same for other flags where `Changed` needs to be checked (opt.Network, opt.Net, opt.Publish, opt.DNS)

cmd/nerdctl/container_run.go
```diff
@@ -415,31 +408,27 @@ func createContainer(ctx context.Context, cmd *cobra.Command, client *containerd
...
 		}
 	}
-	dataStore, err := clientutil.DataStore(globalOptions.DataRoot, globalOptions.Address)
+	dataStore, err := clientutil.DataStore(options.GOptions.DataRoot, options.GOptions.Address)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	stateDir, err := getContainerStateDirPath(cmd, globalOptions, dataStore, id)
+	stateDir, err := containerutil.GetContainerStateDirPath(dataStore, id, options.GOptions.Namespace)
```
djdongjin, line 431: nit: `GetContainerStateDirPath` -> `ContainerStateDirPath`. IIRC, the convention is to not have `Get` in getter funcs.

cmd/nerdctl/container_run.go
```diff
@@ -749,64 +663,41 @@ func createContainer(ctx context.Context, cmd *cobra.Command, client *containerd
...
 
-		options, err := processPullCommandFlagsInRun(cmd)
+		imagePullOptions, err := processPullCommandFlagsInRun(options)
 		if err != nil {
 			return nil, nil, nil, err
 		}
 
-		options.GOptions = globalOptions
-		ensured, err = image.EnsureImage(ctx, client, rawRef, ocispecPlatforms, pull, nil, false, options)
+		imagePullOptions.GOptions = options.GOptions
```
djdongjin, line 699: can this be moved to `processPullCommandFlagsInRun`

djdongjin (commented):
thanks for the work, left a couple of comments :)

cmd/nerdctl/container_create.go
suyanhanx: I think it's okay to have a default value, as it doesn't affect the logic judgment here.

Laitr0n:
> need rebase

Yeah. I will finish it.

cmd/nerdctl/container_create.go
```diff
@@ -51,40 +53,388 @@ func newCreateCommand() *cobra.Command {
...
+	netManager, err := containerutil.NewNetworkingOptionsManager(createOpt.GOptions, netFlags)
+	if err != nil {
+		return err
+	}
+
+	imageVerifyOpt, err := processImageVerifyOptions(cmd)
+	if err != nil {
+		return err
+	}
+	imagePullOpt, err := processPullCommandFlagsInRun(imageVerifyOpt, createOpt)
```
djdongjin, line 432: note for myself: revisit this part later

cmd/nerdctl/container_run.go
```diff
@@ -334,12 +310,21 @@ func runAction(cmd *cobra.Command, args []string) (err error) {
...
 
-	c, gc, err := createContainer(ctx, cmd, client, netManager, globalOptions, args, platform, flagI, flagT, flagD)
+	imageVerifyOpt, err := processImageVerifyOptions(cmd)
+	if err != nil {
+		return err
+	}
+	imagePullOpt, err := processPullCommandFlagsInRun(imageVerifyOpt, createOpt)
+	if err != nil {
+		return err
+	}
```
djdongjin, line 325: note for myself: revisit this part

pkg/containerutil/containerutil.go
```diff
@@ -463,10 +466,96 @@ func Unpause(ctx context.Context, client *containerd.Client, id string) error {
...
+			return err
+		}
+		return nil
+	} else {
+		return err
+	}
+}
+
+// CheckRestartCapabilities checks the capabilities of the restart plugin.
+func checkRestartCapabilities(ctx context.Context, client *containerd.Client, restartFlag string) (bool, error) {
```
djdongjin, line 498: Is this func used in places other than `restart_options.go`? If not, it can be moved there to collocate related logic.

pkg/containerutil/containerutil.go
```diff
@@ -463,10 +466,96 @@ func Unpause(ctx context.Context, client *containerd.Client, id string) error {
...
+	if len(capabilities) == 0 {
+		capabilities = []string{"always"}
+	}
+	if !strutil.InStringSlice(capabilities, policySlice[0]) {
+		return false, fmt.Errorf("unsupported restart policy %q, supported policies are: %q", policySlice[0], capabilities)
+	}
+	return true, nil
+}
+
+func SetOOMScoreAdj(opts []oci.SpecOpts, oomScoreAdjChanged bool, oomScoreAdj int) ([]oci.SpecOpts, error) {
```
djdongjin, line 522: can you add a comment for exported funcs?

pkg/cmd/container/run.go
```diff
@@ -0,0 +1,297 @@
...
+	"github.com/containerd/nerdctl/pkg/platformutil"
+	"github.com/containerd/nerdctl/pkg/referenceutil"
+	"github.com/containerd/nerdctl/pkg/strutil"
+	dockercliopts "github.com/docker/cli/opts"
+	dockeropts "github.com/docker/docker/opts"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+)
+
+func Create(ctx context.Context, client *containerd.Client, args []string, netManager containerutil.NetworkOptionsManager, imagePullOptions types.ImagePullOptions, options types.ContainerCreateOptions) (containerd.Container, func(), error) {
```
djdongjin, line 48: can you add a comment for exported funcs?

pkg/containerutil/mount_options.go
```diff
@@ -40,11 +41,15 @@ import (
...
 	"github.com/opencontainers/image-spec/identity"
 	"github.com/opencontainers/runtime-spec/specs-go"
-
 	"github.com/sirupsen/logrus"
-	"github.com/spf13/cobra"
 )
 
+// getVolumeStore returns a volume store
+// that corresponds to a directory like `/var/lib/nerdctl/1935db59/volumes/default`
+func getVolumeStore(globalOptions types.GlobalCommandOptions) (volumestore.VolumeStore, error) {
```
djdongjin, line 49: can we remove this `getVolumeStore` and use `volume.Store` directly?

pkg/containerutil/labels.go
```diff
@@ -0,0 +1,44 @@
...
+package containerutil
+
+import (
+	"github.com/containerd/containerd"
+	"github.com/containerd/nerdctl/pkg/strutil"
+	dopts "github.com/docker/cli/opts"
+)
+
+// ParseKvStringsMapFromLabel parse label options KV entries and convert to Map
+func ParseKvStringsMapFromLabel(label, labelFile []string) (map[string]string, error) {
```
djdongjin, line 26: is this func used externally? If not, can we change to internal func?

pkg/containerutil/gc.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"os"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/nerdctl/pkg/namestore"
+	"github.com/sirupsen/logrus"
+)
+
+// GenerateGcFunc returns a function that can be used to garbage collect a container
+func GenerateGcFunc(ctx context.Context, container containerd.Container, ns, id, name, dataStore string, containerErr error, containerNameStore namestore.NameStore, netManager NetworkOptionsManager, internalLabels InternalLabels) func() {
```
djdongjin, line 29: where is this func used?

djdongjin:
I haven't looked into if the newly created files make sense (i.e. if they're created in the correct pkg/location etc). Will take a look later.

pkg/containerutil/restart_options.go
djdongjin, line 1: Will these restart options be used by other commands/functions other than `run`? If not, maybe we can just put at `./pkg/cmd/container/run_restart.go`.

djdongjin:
@AkihiroSuda sorry I haven't gotten a chance to take a look at the latest change. Will take a look today.

pkg/cmd/container/run_runtime.go
```diff
@@ -25,48 +25,42 @@ import (
...
-	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
-	"github.com/spf13/cobra"
 )
 
-func generateRuntimeCOpts(cmd *cobra.Command, globalOptions types.GlobalCommandOptions) ([]containerd.NewContainerOpts, error) {
-	runtime := plugin.RuntimeRuncV2
+func generateRuntimeCOpts(cgroupManager, runtime string) ([]containerd.NewContainerOpts, error) {
+	runTime := plugin.RuntimeRuncV2
```
djdongjin, line 33: Can we change the `runtime` arg to `runtimeStr`, and `runTime` to `runtime`?

It's more consistent with existing format and having a `Str` suffix is more clear than an uppercase `T`

pkg/cmd/container/run_cgroup_linux.go
```diff
@@ -40,73 +39,72 @@ type customMemoryOptions struct {
...
-		return nil, err
-	}
-	memStr, err := cmd.Flags().GetString("memory")
-	if err != nil {
-		return nil, err
-	}
-	memSwap, err := cmd.Flags().GetString("memory-swap")
-	if err != nil {
-		return nil, err
+func generateCgroupPath(id string, cgroupManager, cgroupParent string) (string, error) {
```
djdongjin, line 42: nit: can we keep the argument order unchanged and remove the redundant `string` arg type (`cgroupManager, cgroupParent, id string). Also if you move this func to after `GenerateOpts` and before `ParseDevice`, the diff will be much smaller.

djdongjin (approved):
Overall LGTM, I left some minor suggestions. Thanks!

Since it's a large PR it might be good to have another review/approval.

AkihiroSuda (approved):
Thanks

PR #1973, opened 2023-02-03
handle container deps removal with postsop hooks

fahedouch:
**#Before** 
the container deps ( `etchosts files`, `names`, `container state`) removal is deferred by [RemoveContainer](https://github.com/containerd/nerdctl/blob/main/pkg/cmd/container/remove.go#L118-LL152) which means container deletion happens before deps deletion.

**#After** 
Using  runtime-spec postStop Hook, we ensure that container deps deletion is successfully completed before the container deletion returns

### DIFF
--- a/cmd/nerdctl/container_restart_linux_test.go
+++ b/cmd/nerdctl/container_restart_linux_test.go
@@ -51,11 +51,11 @@ func TestRestartPIDContainer(t *testing.T) {
 	base := testutil.NewBase(t)
 
 	baseContainerName := testutil.Identifier(t)
-	base.Cmd("run", "-d", "--name", baseContainerName, testutil.AlpineImage, "sleep", "infinity").Run()
+	base.Cmd("run", "-d", "--name", baseContainerName, testutil.AlpineImage, "sleep", "infinity").AssertOK()
 	defer base.Cmd("rm", "-f", baseContainerName).Run()
 
-	sharedContainerName := testutil.Identifier(t)
-	base.Cmd("run", "-d", "--name", sharedContainerName, fmt.Sprintf("--pid=container:%s", baseContainerName), testutil.AlpineImage, "sleep", "infinity").Run()
+	sharedContainerName := fmt.Sprintf("%s-shared", baseContainerName)
+	base.Cmd("run", "-d", "--name", sharedContainerName, fmt.Sprintf("--pid=container:%s", baseContainerName), testutil.AlpineImage, "sleep", "infinity").AssertOK()
 	defer base.Cmd("rm", "-f", sharedContainerName).Run()
 
 	base.Cmd("restart", baseContainerName).AssertOK()
--- a/cmd/nerdctl/container_run.go
+++ b/cmd/nerdctl/container_run.go
@@ -358,7 +358,6 @@ func runAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	logURI := lab[labels.LogURI]
-
 	task, err := taskutil.NewTask(ctx, client, c, false, flagI, flagT, flagD, con, logURI)
 	if err != nil {
 		return err
--- a/pkg/cmd/container/remove.go
+++ b/pkg/cmd/container/remove.go
@@ -122,10 +122,10 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions 
 		if retErr != nil {
 			return
 		}
-
 		if err := os.RemoveAll(stateDir); err != nil {
 			logrus.WithError(retErr).Warnf("failed to remove container state dir %s", stateDir)
 		}
+		// enforce release name here in case the poststop hook name release fails
 		if name != "" {
 			if err := namst.Release(name, id); err != nil {
 				logrus.WithError(retErr).Warnf("failed to release container name %s", name)
@@ -135,6 +135,8 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions 
 			logrus.WithError(retErr).Warnf("failed to remove hosts file for container %q", id)
 		}
 	}()
+
+	// volume removal is not handled by the poststop hook lifecycle because it depends on removeAnonVolumes option
 	if anonVolumesJSON, ok := l[labels.AnonymousVolumes]; ok && removeAnonVolumes {
 		var anonVolumes []string
 		if err := json.Unmarshal([]byte(anonVolumesJSON), &anonVolumes); err != nil {
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -32,6 +32,7 @@ import (
 	"github.com/containerd/nerdctl/pkg/bypass4netnsutil"
 	"github.com/containerd/nerdctl/pkg/dnsutil/hostsstore"
 	"github.com/containerd/nerdctl/pkg/labels"
+	"github.com/containerd/nerdctl/pkg/namestore"
 	"github.com/containerd/nerdctl/pkg/netutil"
 	"github.com/containerd/nerdctl/pkg/netutil/nettype"
 	"github.com/containerd/nerdctl/pkg/rootlessutil"
@@ -445,6 +446,7 @@ func onCreateRuntime(opts *handlerOpts) error {
 
 func onPostStop(opts *handlerOpts) error {
 	ctx := context.Background()
+	ns := opts.state.Annotations[labels.Namespace]
 	if opts.cni != nil {
 		var err error
 		b4nnEnabled, err := bypass4netnsutil.IsBypass4netnsEnabled(opts.state.Annotations)
@@ -497,10 +499,17 @@ func onPostStop(opts *handlerOpts) error {
 		if err != nil {
 			return err
 		}
-		ns := opts.state.Annotations[labels.Namespace]
 		if err := hs.Release(ns, opts.state.ID); err != nil {
 			return err
 		}
+	}
+	namst, err := namestore.New(opts.dataStore, ns)
+	if err != nil {
+		return err
+	}
+	name := opts.state.Annotations[labels.Name]
+	if err := namst.Release(name, opts.state.ID); err != nil {
+		return fmt.Errorf("failed to release container name %s: %w", name, err)
 	}
 	return nil
 }
--- a/cmd/nerdctl/container_run_network_linux_test.go
+++ b/cmd/nerdctl/container_run_network_linux_test.go
@@ -404,6 +404,7 @@ func TestRunWithInvalidPortThenCleanUp(t *testing.T) {
 	t.Parallel()
 	base := testutil.NewBase(t)
 	containerName := testutil.Identifier(t)
+	defer base.Cmd("rm", "-f", containerName).Run()
 	base.Cmd("run", "--rm", "--name", containerName, "-p", "22200-22299:22200-22299", testutil.CommonImage).AssertFail()
 	base.Cmd("run", "--rm", "--name", containerName, "-p", "22200-22299:22200-22299", testutil.CommonImage).AssertCombinedOutContains(errdefs.ErrInvalidArgument.Error())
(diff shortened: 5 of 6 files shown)


fahedouch:
I still have work to here:

Once the runtime deps ( `host files`, `names`, `container state`...) removal is handled by postStop oci hook ( [see](https://github.com/opencontainers/runtime-spec/blob/main/config.md#summary) ), with the same logic the runtime deps ( `host files`, `names`, `container state`...)  creation should be handled the oci hook too, which is `createRuntime` hook. Acutally, it is our cmd that handles doing this creation which not a good design in my opinion; runtime deps creation/removal should be close the runtime level, then the runtime event hooks  (`createRuntime` / `postStop`) seems to be the best place to handle these actions.

WDYT @AkihiroSuda

djdongjin:
@fahedouch SGTM! as long as runtime/oci provides such capabilities we should utilize them other than handling in nerdctl. Just a few questions:

1. Is there any behavior changes? (e.g. new error-returnning logic/scenario).
2. How many container deps we can change to utilize the ocihook (you mentioned `etchosts files, names, container state`)? Mostly, any specific dep that is still need to be managed by nerdctl?
3. you mentioned `container name management`. Does that mean if we change to use oci hook, we can elimate the use of `NameStore `?

Thanks

manugupt1:
I have a basic question. Does taskStart and task.kill call ocihooks?
Can a container have more than 1 task? What happens then for let's say /etc/hosts file?

fahedouch:
@djdongjin 
> @fahedouch SGTM! as long as runtime/oci provides such capabilities we should utilize them other than handling in nerdctl. Just a few questions (this is new to me :)):
> 
> 1. Is there any behavior changes? (e.g. new error-returnning logic/scenario).

I believe that returned error error is handled by `runc` through OCI hook specification, same thing as hoststore error management 

> 2. How many container deps we can change to utilize the ocihook (you mentioned `etchosts files, names, container state`)? Mostly, any specific dep that is still need to be managed by nerdctl?

not all deps can be handled by the ocihook, such as [anon volume remove](https://github.com/containerd/nerdctl/blob/main/pkg/cmd/container/remove.go#L138-L152) as it depends on cli options. So the idea is to start moving theses deps under oci hook scope: 
- etchosts files
- names
- container states
And complete the rest laster ( different PRs)

> 3. you mentioned `container name management`. Does that mean if we change to use oci hook, we can elimate the use of `NameStore `?
`NameStore` interaction/management move under oci-hook => oci-hook will trigger NameStore interface

> 4. Does this impact which versions of containerd/runc that nerdctl supports?

No. `containerd/runc` out of technical scope. `runtime-spec` provide a decoupled architecture so we can implement the hook logic without impacting with runc itself. there is a good abstraction. 
> 


@manugupt1 
> I have a basic question. Does taskStart and task.kill call ocihooks?

`task.Start` => trigger `createRuntime` hook  (that need a double check)
`task.Delete` => trigger  `poststop` 
 
[full list of hooks](https://github.com/opencontainers/runtime-spec/blob/main/config.md#summary)

> Can a container have more than 1 task? 

[see](https://github.com/containerd/containerd/blob/main/runtime/runtime.go#L73)


> What happens then for let's say /etc/hosts file?

no changes for `/etc/hosts` because they are already handled by oci-hook

fahedouch:
I have some blocker(s) to make oci hooks managing the container deps. I can't figure out what is creating `hostsDirBasename` ( `/var/lib/nerdctl/<ADDRHASH>/etchosts`). 
Theoretically is done by the [onCreateRuntime] (https://github.com/containerd/nerdctl/blob/main/pkg/ocihook/ocihook.go#L378) `hook`, but  by  disabling the `onCreateRuntime` hook the `hostsDirBasename` is always create.  It seems like an other thing is creating/controlling the `hostsDirBasename` 🧐

fahedouch:
> I have some blocker(s) to make oci hooks managing the container deps. I can't figure out what is creating `hostsDirBasename` ( `/var/lib/nerdctl/<ADDRHASH>/etchosts`). Theoretically is done by the [onCreateRuntime] (https://github.com/containerd/nerdctl/blob/main/pkg/ocihook/ocihook.go#L378) `hook`, but by disabling the `onCreateRuntime` hook the `hostsDirBasename` is always create. It seems like an other thing is creating/controlling the `hostsDirBasename` 🧐

found my answer, I was targeting the wrong path :p , what I am looking for is the creation of `hostsDir` and not `hostsDirBasename` which is done by the `container_run_network.go`. I have to move creation login under the `onCreateRuntime` hook

fahedouch:
Unfortunately, I could not totally acheive my goal using runtime-spec oci hooks, for the simple reason that all hooks are triggered after the runtime environment has been created (according to the spec configuration in config.json):

So these deps cannot be handled by runtime-spec oci hooks:

- `etc/hosts files` is mount-bind file handled by the config.json within runtime creation  =>  out of `oci-hooks` scope

- `container states` are created by container and not containerd tasks => out of `oci-hooks` scope

But container `names` is now handled by the runtime-spec oci hooks.

manugupt1:
Hi @fahedouch; I am still building my understanding.
Were you able to verify this part:
1. task.New => trigger createRuntime hook (that need a double check)

I am also curious what happens when nerdctl run -d <...> sleep infinity and exec is called together in it. I am guessing it works because the tests are passing.

fahedouch:
> Hi @fahedouch; I am still building my understanding. Were you able to verify this part:
> 
> 1. task.New => trigger createRuntime hook (that need a double check)
> 
> I am also curious what happens when nerdctl run -d <...> sleep infinity and exec is called together in it. I am guessing it works because the tests are passing.

Hi @manugupt1 , you can verify this by checking the `/var/lib/nerdctl/1935db59/` content when  `task.New()` is execute. You can achieve this using some breakpoints

`task.New()` => `containerd` create a task => `runc`  create => trigger `createRuntime` hook

pkg/ocihook/ocihook.go
```diff
@@ -53,6 +54,10 @@ const (
 	NetworkNamespace = labels.Prefix + "network-namespace"
 )
 
+var (
+	containerNameStore namestore.NameStore
```
AkihiroSuda, line 58: why global?

fahedouch: fixed

cmd/nerdctl/container_restart_linux_test.go
```diff
@@ -51,11 +51,11 @@ func TestRestartPIDContainer(t *testing.T) {
...
 
 	baseContainerName := testutil.Identifier(t)
-	base.Cmd("run", "-d", "--name", baseContainerName, testutil.AlpineImage, "sleep", "infinity").Run()
+	base.Cmd("run", "-d", "--name", baseContainerName, testutil.AlpineImage, "sleep", "infinity").AssertOK()
 	defer base.Cmd("rm", "-f", baseContainerName).Run()
 
-	sharedContainerName := testutil.Identifier(t)
-	base.Cmd("run", "-d", "--name", sharedContainerName, fmt.Sprintf("--pid=container:%s", baseContainerName), testutil.AlpineImage, "sleep", "infinity").Run()
+	sharedContainerName := fmt.Sprintf("%s-shared", baseContainerName)
+	base.Cmd("run", "-d", "--name", sharedContainerName, fmt.Sprintf("--pid=container:%s", baseContainerName), testutil.AlpineImage, "sleep", "infinity").AssertOK()
```
manugupt1, line 58: nit: since we are changing Run --> AssertOk; can we do the same for defer statements in this test as well.

pkg/ocihook/ocihook.go
```diff
@@ -365,6 +366,16 @@ func getMACAddressOpts(opts *handlerOpts) ([]gocni.NamespaceOpts, error) {
 func onCreateRuntime(opts *handlerOpts) error {
 	loadAppArmor()
 
+	if opts.state.Annotations[labels.Name] != "" {
```
manugupt1, line 369: Q: Is it possible to have labels.Name as an empty string; should we throw an error if this annotation is empty string?

manugupt1:
Thank you @fahedouch for this PR; I learnt quite a bit.

pkg/ocihook/ocihook.go
suyanhanx: https://github.com/containerd/nerdctl/blob/1642bef99f0931b79fe75fd54c1566e9e061b178/pkg/labels/labels.go#L23-L30

It doesn't seem possible for it to be an empty string.

cmd/nerdctl/container_restart_linux_test.go
fahedouch: defer should not cause test to fail

pkg/ocihook/ocihook.go
fahedouch: it cannot be empty. So removed to avoid misunderstanding

cmd/nerdctl/container_restart_linux_test.go
manugupt1: The reason for doing this in defer is to prevent Leaks and avoid flakiness; when a wrong command is written by mistake.

manugupt1 (approved):
thanks!

suyanhanx (approved):

fahedouch:
cc @AkihiroSuda  PTAL!

AkihiroSuda:
This change is relatively huge, so let me release v1.2.1 before merging this
(waiting for https://github.com/moby/buildkit/pull/3651 to be released)

cmd/nerdctl/container_run.go
```diff
@@ -733,16 +721,6 @@ func createContainer(ctx context.Context, cmd *cobra.Command, client *containerd
...
-			if name != "" {
-				var errE error
-				if containerNameStore, errE = namestore.New(dataStore, globalOptions.Namespace); errE != nil {
-					isErr = true
-				}
-				if errE = containerNameStore.Release(name, id); errE != nil {
-					isErr = true
-				}
-
-			}
```
AkihiroSuda, line 745: The name has to be reserved as soon as the container is createad, so this code shouldn't be moved

pkg/cmd/container/remove.go
```diff
@@ -105,36 +104,26 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions
...
 		}
-
 		if err := os.RemoveAll(stateDir); err != nil {
 			logrus.WithError(retErr).Warnf("failed to remove container state dir %s", stateDir)
 		}
-		if name != "" {
-			if err := namst.Release(name, id); err != nil {
-				logrus.WithError(retErr).Warnf("failed to release container name %s", name)
-			}
-		}
```
AkihiroSuda, line 133: This code should be kept back too, in case the poststop hook fails

AkihiroSuda (changes requested):
Please see comments

cmd/nerdctl/container_run.go
fahedouch: ping @AkihiroSuda

AkihiroSuda: Sorry, my comment is for L690 `if err := containerNameStore.Acquire(name, id); err != nil {`

fahedouch: Fixed

fahedouch:
@AkihiroSuda PTAL !

AkihiroSuda:
CI failing

fahedouch:
because of double acquire, [this](fahedouch:handle-container-deps-removal-with-postsop-hooks) should be removed too

cmd/nerdctl/container_run.go
fahedouch: > The name has to be reserved as soon as the container is created, so this code shouldn't be moved

this code shouldn't be removed too, as failed creation in this level need a cleanup too

fahedouch:
@AkihiroSuda  PTAL again plz

AkihiroSuda (approved):
Thanks

PR #1975, opened 2023-02-04
Test for volume create command

suyanhanx:

### DIFF
--- a/cmd/nerdctl/volume_create_test.go
+++ b/cmd/nerdctl/volume_create_test.go
@@ -1,0 +1,58 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/volume_create_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+	"gotest.tools/v3/assert"
+)
+
+// Test TestVolumeCreate for creating volume with given name.
+func TestVolumeCreate(t *testing.T) {
```
djdongjin, line 27: `TestVolumeCreate` and `TestVolumeCreateWithLabels` seems duplicated with tests in `volume inspect`

https://github.com/containerd/nerdctl/blob/b80c9caa28c849c5191e834dd072c11ff80ee9b1/cmd/nerdctl/volume_inspect_test.go#L29

suyanhanx: Yes. I saw it. But it feels like this should be moved from the inspect side to the create side ?

fahedouch: As integration test, it is ok to keep both

cmd/nerdctl/volume_create_test.go
```diff
@@ -0,0 +1,58 @@
...
+	base := testutil.NewBase(t)
+	testVolume := testutil.Identifier(t)
+
+	base.Cmd("volume", "create", testVolume).AssertOK()
+	defer base.Cmd("volume", "rm", testVolume).Run()
```
fahedouch, line 32: `defer base.Cmd("volume", "rm", "-f", testVolume).Run()`

cmd/nerdctl/volume_create_test.go
djdongjin: SGTM

cmd/nerdctl/volume_create_test.go
suyanhanx: Make sense.

fahedouch (approved):
Thanks

PR #1976, opened 2023-02-05
Cache go artifacts that are not tested within docker

manugupt1:
Caching go modules should slightly improve speed if not by a lot.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,19 +7,22 @@ on:
       - 'release/**'
   pull_request:
 
+env:
+  GO_VERSION: 1.19.x
+
 jobs:
   project:
     name: Project Checks
     runs-on: ubuntu-22.04
     timeout-minutes: 20
     steps:
-      - uses: actions/setup-go@v3
-        with:
-          go-version: 1.19.x
       - uses: actions/checkout@v3.3.0
         with:
           path: src/github.com/containerd/nerdctl
           fetch-depth: 100
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
       - uses: containerd/project-checks@v1.1.0
         with:
           working-directory: src/github.com/containerd/nerdctl
@@ -38,7 +41,9 @@ jobs:
           fetch-depth: 1
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.19.x
+          go-version: ${{ env.GO_VERSION }}
+          check-latest: true
+          cache: true
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3.3.1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -51,12 +51,14 @@ jobs:
...
-        with:
-          go-version: 1.19.x
       - uses: actions/checkout@v3.3.0
         with:
           fetch-depth: 1
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
+          check-latest: true
+          cahche: true
```
AkihiroSuda, line 61: typo

.github/workflows/test.yml
```diff
@@ -51,12 +51,14 @@ jobs:
...
-        with:
-          go-version: 1.19.x
       - uses: actions/checkout@v3.3.0
         with:
           fetch-depth: 1
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
+          check-latest: true
+          cache: true
```
ktock, line 61: nit: Why not enable this in other tests? (e.g. `lint`, `project`)?

ktock (approved):

.github/workflows/test.yml
manugupt1: I did not realize that both the checks did go build; i.e the main reason.

golangci is already cached: https://github.com/containerd/nerdctl/actions/caches
I think having in one place is sufficient: See: https://github.com/containerd/nerdctl/actions/runs/4097163654/jobs/7065337105
Likely because a workflow is tied to a single runner.
https://github.com/containerd/nerdctl/actions/runs/4097163654/jobs/7065337009

I can make the change if you prefer.

.github/workflows/test.yml
```diff
@@ -13,13 +13,16 @@ jobs:
...
       - uses: actions/checkout@v3.3.0
         with:
           path: src/github.com/containerd/nerdctl
           fetch-depth: 100
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
+          check-latest: true
+          cache: true
+          cache-dependency-path: src/github.com/containerd/nerdctl
```
fahedouch, line 25: would you please set `cache-dependency-path` for other jobs

.github/workflows/test.yml
```diff
@@ -13,13 +13,16 @@ jobs:
...
-      - uses: actions/setup-go@v3
-        with:
-          go-version: 1.19.x
       - uses: actions/checkout@v3.3.0
         with:
           path: src/github.com/containerd/nerdctl
           fetch-depth: 100
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
```
fahedouch, line 22: can we expose the go version for the entire workflow, to have a single point change:
```
      - uses: actions/setup-go@v3
        with:
          check-latest: ${{ env.CHECK_LATEST }}
          cache-dependency-path: ${{ env.CHECK_DEPENDENCY_PATH }}
```

manugupt1: We do not need it as src/github.com/containerd/nerdctl is mapped at checkout actions in project checks only.

.github/workflows/test.yml
manugupt1: We do not need that as We do not need it as src/github.com/containerd/nerdctl is mapped at checkout actions in project checks only.

fahedouch (approved):
Thanks

PR #1978, opened 2023-02-07
feat: resolve special ip host-gateway

ningziwen:
https://github.com/containerd/nerdctl/issues/1979

Tests and docs will be added when the issue got concensus.

Signed-off-by: Ziwen Ning <ningziwe@amazon.com>

### DIFF
--- a/cmd/nerdctl/container_run.go
+++ b/cmd/nerdctl/container_run.go
@@ -57,7 +57,8 @@ import (
 	"github.com/containerd/nerdctl/pkg/referenceutil"
 	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/containerd/nerdctl/pkg/taskutil"
-	dopts "github.com/docker/cli/opts"
+	dockercliopts "github.com/docker/cli/opts"
+	dockeropts "github.com/docker/docker/opts"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -707,9 +708,19 @@ func createContainer(ctx context.Context, cmd *cobra.Command, client *containerd
 		return nil, nil, err
 	}
 	extraHosts = strutil.DedupeStrSlice(extraHosts)
-	for _, host := range extraHosts {
-		if _, err := dopts.ValidateExtraHost(host); err != nil {
+	for i, host := range extraHosts {
+		if _, err := dockercliopts.ValidateExtraHost(host); err != nil {
 			return nil, nil, err
+		}
+		parts := strings.SplitN(host, ":", 2)
+		// If the IP Address is a string called "host-gateway", replace this value with the IP address stored
+		// in the daemon level HostGateway IP config variable.
+		if parts[1] == dockeropts.HostGatewayName {
+			if globalOptions.HostGatewayIP == "" {
+				return nil, nil, fmt.Errorf("unable to derive the IP value for host-gateway")
+			}
+			parts[1] = globalOptions.HostGatewayIP
+			extraHosts[i] = fmt.Sprintf(`%s:%s`, parts[0], parts[1])
 		}
 	}
 	internalLabels.extraHosts = extraHosts
@@ -995,7 +1006,7 @@ func readKVStringsMapfFromLabel(cmd *cobra.Command) (map[string]string, error) {
 		return nil, err
 	}
 	labelsFilePath = strutil.DedupeStrSlice(labelsFilePath)
-	labels, err := dopts.ReadKVStrings(labelsFilePath, labelsMap)
+	labels, err := dockercliopts.ReadKVStrings(labelsFilePath, labelsMap)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/nerdctl/container_run_linux_test.go
+++ b/cmd/nerdctl/container_run_linux_test.go
@@ -19,8 +19,11 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -32,7 +35,6 @@ import (
 	"github.com/containerd/nerdctl/pkg/rootlessutil"
 	"github.com/containerd/nerdctl/pkg/strutil"
 	"github.com/containerd/nerdctl/pkg/testutil"
-
 	"gotest.tools/v3/assert"
 )
 
@@ -150,6 +152,38 @@ func TestRunAddHost(t *testing.T) {
 		return nil
 	})
 	base.Cmd("run", "--rm", "--add-host", "10.0.0.1:testing.example.com", testutil.AlpineImage, "cat", "/etc/hosts").AssertFail()
+
+	response := "This is the expected response for --add-host special IP test."
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		io.WriteString(w, response)
+	})
+	const hostPort = 8081
+	s := http.Server{Addr: fmt.Sprintf(":%d", hostPort), Handler: nil, ReadTimeout: 30 * time.Second}
+	go s.ListenAndServe()
+	defer s.Shutdown(context.Background())
+	base.Cmd("run", "--rm", "--add-host", "test:host-gateway", testutil.NginxAlpineImage, "curl", fmt.Sprintf("test:%d", hostPort)).AssertOutExactly(response)
+}
+
+func TestRunAddHostWithCustomHostGatewayIP(t *testing.T) {
+	// Not parallelizable (https://github.com/containerd/nerdctl/issues/1127)
+	base := testutil.NewBase(t)
+	testutil.DockerIncompatible(t)
+	base.Cmd("run", "--rm", "--host-gateway-ip", "192.168.5.2", "--add-host", "test:host-gateway", testutil.AlpineImage, "cat", "/etc/hosts").AssertOutWithFunc(func(stdout string) error {
+		var found bool
+		sc := bufio.NewScanner(bytes.NewBufferString(stdout))
+		for sc.Scan() {
+			//removing spaces and tabs separating items
+			line := strings.ReplaceAll(sc.Text(), " ", "")
+			line = strings.ReplaceAll(line, "\t", "")
+			if strings.Contains(line, "192.168.5.2test") {
+				found = true
+			}
+		}
+		if !found {
+			return errors.New("host was not added")
+		}
+		return nil
+	})
 }
 
 func TestRunUlimit(t *testing.T) {
--- a/cmd/nerdctl/flagutil.go
+++ b/cmd/nerdctl/flagutil.go
@@ -70,6 +70,10 @@ func processRootCmdFlags(cmd *cobra.Command) (types.GlobalCommandOptions, error)
 	if err != nil {
 		return types.GlobalCommandOptions{}, err
 	}
+	hostGatewayIP, err := cmd.Flags().GetString("host-gateway-ip")
+	if err != nil {
+		return types.GlobalCommandOptions{}, err
+	}
 	return types.GlobalCommandOptions{
 		Debug:            debug,
 		DebugFull:        debugFull,
@@ -83,5 +87,6 @@ func processRootCmdFlags(cmd *cobra.Command) (types.GlobalCommandOptions, error)
 		InsecureRegistry: insecureRegistry,
 		HostsDir:         hostsDir,
 		Experimental:     experimental,
+		HostGatewayIP:    hostGatewayIP,
 	}, nil
 }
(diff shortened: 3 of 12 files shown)


pkg/config/config.go
```diff
@@ -38,6 +38,7 @@ type Config struct {
 	InsecureRegistry bool     `toml:"insecure_registry"`
 	HostsDir         []string `toml:"hosts_dir"`
 	Experimental     bool     `toml:"experimental"`
+	HostGatewayIP    string   `toml:"host_gateway_ip"`
```
AkihiroSuda, line 41:
```suggestion
	HostGateway    *net.IP   `toml:"host_gateway_ip"`
```

AkihiroSuda: (Can remain as a `string` for ease of implementation)

cmd/nerdctl/main.go
```diff
@@ -171,6 +171,7 @@ func initRootCmdFlags(rootCmd *cobra.Command, tomlPath string) (*pflag.FlagSet,
 	rootCmd.PersistentFlags().StringSlice("hosts-dir", cfg.HostsDir, "A directory that contains <HOST:PORT>/hosts.toml (containerd style) or <HOST:PORT>/{ca.cert, cert.pem, key.pem} (docker style)")
 	// Experimental enable experimental feature, see in https://github.com/containerd/nerdctl/blob/main/docs/experimental.md
 	AddPersistentBoolFlag(rootCmd, "experimental", nil, nil, cfg.Experimental, "NERDCTL_EXPERIMENTAL", "Control experimental: https://github.com/containerd/nerdctl/blob/main/docs/experimental.md")
+	rootCmd.PersistentFlags().String("host-gateway-ip", cfg.HostGatewayIP, "IP address that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP address of the default bridge")
```
AkihiroSuda, line 174: Any reason not to just use the host IP?

AkihiroSuda: The default bridge IP isn't useful for rootless because the slirp is launched with `--disable-host-loopback` for the security reason
https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L123

pkg/config/config.go
ningziwen: I see for any other configs, the variable name is exactly same as the field name in toml. Is it good to make it special as "HostGateway" instead of "HostGatewayIP"?

AkihiroSuda: Sorry, just forgot to modify the toml tag in my comment above

ningziwen: `host-gateway-ip` is the key string in Docker config. https://github.com/moby/moby/blob/master/cmd/dockerd/config.go#L39

Should we keep Docker compatibiilty as much as possible or the config can be exempted?

AkihiroSuda: Ah, `host-gateway-ip` is the right choice then, sorry

cmd/nerdctl/main.go
ningziwen: Thanks for the suggestion. I just tried to use host ip as default, as the [commit](https://github.com/containerd/nerdctl/pull/1978/commits/08487b034c19b906b13bda9912aa1b553a613300) I pushed.

I tested the built nerdctl in Lima. The returned host ip can't be used to access host (curl failed). 

The static [Lima guest IP (192.168.5.15)](https://github.com/lima-vm/lima/blob/25b137d4c54a5a251fb7cca4adf1892097e1a6f2/docs/network.md#guest-ip-192168515) can be used to access host (curl succeeded). However it seems like a loopback IP which was skipped by the loopback check.

ningziwen: I tried in ubuntu linux host and it works. I think the host IP is VM has special setup. I will follow the route of making host IP the default as it should work for normal cases. VM users will need to use the config to set the their special IP.

ningziwen: I finally used env variable the pass the host IP to nsenter. Let me know if you have different thoughts.

pkg/testutil/testutil_linux.go
```diff
@@ -25,6 +25,7 @@ func mirrorOf(s string) string {
 
 var (
 	AlpineImage                 = mirrorOf("alpine:3.13")
+	AmazonLinux2Image           = "public.ecr.aws/amazonlinux/amazonlinux:2"
```
djdongjin, line 28: why is a new image needed here?

pkg/defaults/defaults_linux.go
```diff
@@ -130,3 +131,18 @@ func HostsDirs() []string {
 		filepath.Join(xch, "docker/certs.d"),
 	}
 }
+
+func HostGatewayIP() string {
```
djdongjin, line 135: can you add a comment here? since it's a new export non-test func.

cmd/nerdctl/container_run.go
```diff
@@ -707,10 +708,24 @@ func createContainer(ctx context.Context, cmd *cobra.Command, client *containerd
...
 			return nil, nil, err
 		}
+		parts := strings.SplitN(host, ":", 2)
+		// If the IP Address is a string called "host-gateway", replace this value with the IP address stored
+		// in the daemon level HostGateway IP config variable.
+		if parts[1] == dockerops.HostGatewayName {
+			gateway, err := cmd.Flags().GetString("host-gateway-ip")
+			if err != nil {
+				return nil, nil, err
+			}
```
djdongjin, line 722: `host-gateway-ip` has been parsed and passed via `globalOptions` right? If so you can use `globalOptions.HostGatewayIP`

cmd/nerdctl/flagutil.go
```diff
@@ -70,6 +70,10 @@ func processRootCmdFlags(cmd *cobra.Command) (types.GlobalCommandOptions, error)
 	if err != nil {
 		return types.GlobalCommandOptions{}, err
 	}
+	hostGatewayIP, err := cmd.Flags().GetString("host-gateway-ip")
+	if err != nil {
+		return types.GlobalCommandOptions{}, err
+	}
```
djdongjin, line 76: just to confirm, `HostGatewayIP` will never be empty right? Since the doc mentions it defauts to the IP address of the host.

pkg/testutil/testutil_linux.go
ningziwen: I need to use curl in tests but alpine doesn't have it and AL2 has it. Let me know if you have suggestions for better image choice.

djdongjin: I see. I'm okay with that. The [`nginx` image](https://github.com/containerd/nerdctl/blob/8a3331729b0d43fd51cec8aaaa5f45238ae05749/pkg/testutil/testutil_linux.go#L28) should have `curl`, can you try that?

cmd/nerdctl/flagutil.go
ningziwen: From nerdctl external it can't be empty. 

But it can be empty string "" when passing to here. When there was any issue getting host ip, it would be empty here. In container_run.go, it will return error when it is still "".

```
			if globalOptions.HostGatewayIP == "" {
				return nil, nil, fmt.Errorf("unable to derive the IP value for host-gateway")
			}
```

pkg/testutil/testutil_linux.go
ningziwen: That works. Changed it to.

cmd/nerdctl/container_run_linux_test.go
```diff
@@ -150,6 +152,38 @@ func TestRunAddHost(t *testing.T) {
...
+	s := http.Server{Addr: fmt.Sprintf(":%d", hostPort), Handler: nil, ReadTimeout: 30 * time.Second}
+	go s.ListenAndServe()
+	defer s.Shutdown(context.Background())
+	base.Cmd("run", "--rm", "--add-host", "test:host-gateway", testutil.NginxAlpineImage, "curl", fmt.Sprintf("test:%d", hostPort)).AssertOutExactly(response)
+}
+
+func TestRunAddHostWithCustomHostGatewayIP(t *testing.T) {
+	// Not parallelizable (https://github.com/containerd/nerdctl/issues/1127)
+	base := testutil.NewBase(t)
+	testutil.DockerIncompatible(t)
```
djdongjin, line 170: why is it docker incompatible?

cmd/nerdctl/container_run_linux_test.go
```diff
@@ -150,6 +152,38 @@ func TestRunAddHost(t *testing.T) {
...
+			line := strings.ReplaceAll(sc.Text(), " ", "")
+			line = strings.ReplaceAll(line, "\t", "")
+			if strings.Contains(line, "192.168.5.2test") {
+				found = true
+			}
+		}
+		if !found {
+			return errors.New("host was not added")
+		}
+		return nil
```
djdongjin, line 185: Does a `.AssertOutContains("192.168.5.2test")` work?

cmd/nerdctl/container_run_linux_test.go
ningziwen:
```
docker run --host-gateway-ip "192.168.5.2" --add-host=name:host-gateway -it run amazonlinux bash
unknown flag: --host-gateway-ip
See 'docker run --help'.
```

Docker has host-gateway-ip in [config](https://github.com/moby/moby/blob/master/cmd/dockerd/config.go#L39) but not in global flag. From my understanding, there is a convention in nerdctl that transferring all the config to global flag first before using them. So the global flag is docker incompatible.

djdongjin (approved):
LGTM, thanks

cmd/nerdctl/main.go
```diff
@@ -171,6 +171,7 @@ func initRootCmdFlags(rootCmd *cobra.Command, tomlPath string) (*pflag.FlagSet,
 	rootCmd.PersistentFlags().StringSlice("hosts-dir", cfg.HostsDir, "A directory that contains <HOST:PORT>/hosts.toml (containerd style) or <HOST:PORT>/{ca.cert, cert.pem, key.pem} (docker style)")
 	// Experimental enable experimental feature, see in https://github.com/containerd/nerdctl/blob/main/docs/experimental.md
 	AddPersistentBoolFlag(rootCmd, "experimental", nil, nil, cfg.Experimental, "NERDCTL_EXPERIMENTAL", "Control experimental: https://github.com/containerd/nerdctl/blob/main/docs/experimental.md")
+	AddPersistentStringFlag(rootCmd, "host-gateway-ip", nil, nil, nil, aliasToBeInherited, cfg.HostGatewayIP, "HOST_GATEWAY_IP", "IP address that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP address of the default bridge")
```
AkihiroSuda, line 174: Does Docker use `HOST_GATEWAY_IP`?

Otherwise:
```suggestion
	AddPersistentStringFlag(rootCmd, "host-gateway-ip", nil, nil, nil, aliasToBeInherited, cfg.HostGatewayIP, "NERDCTL_HOST_GATEWAY_IP", "IP address that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP address of the default bridge")
```

AkihiroSuda: > Defaults to the IP address of the default bridge

Doesn't seem consistent with the actual implementation

pkg/defaults/defaults_linux.go
```diff
@@ -130,3 +131,19 @@ func HostsDirs() []string {
...
+
+// HostGatewayIP returns the non-loop-back host ip if available and returns empty string if running into error.
+func HostGatewayIP() string {
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return ""
+	}
+	for _, addr := range addrs {
+		if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil || ipnet.IP.To16() != nil {
```
fahedouch, line 143: we do not support `ipv6` yet. You should return empty string for `ipv6`

pkg/defaults/defaults_linux.go
```diff
@@ -130,3 +131,19 @@ func HostsDirs() []string {
 		filepath.Join(xch, "docker/certs.d"),
 	}
 }
+
+// HostGatewayIP returns the non-loop-back host ip if available and returns empty string if running into error.
+func HostGatewayIP() string {
```
fahedouch, line 136: the `HostGatewayIP` should be the container network interface IP and not from other network interfaces

ningziwen: I think by defaulit we are returning the host's real ip now instead of any interface IP. The host ip is accessible from containers.

fahedouch: docker are setting the host bridge net interface IP. not sure if the host real IP will fit here, beacuse the host net interface IP that's attached to the container need to be the gateway to route traffic from the container to host.

fahedouch: SGTM the real host IP

fahedouch (approved):
Thanks

AkihiroSuda (approved):
Thanks

Issue #1675, opened 2022-12-21
`nerdctl push` interop issue with Harbor registry


l2dy:
### Description

Hi, I'm trying `nerdctl push` on a private Harbor registry with Bearer authentication and a custom token service, but the command fails with `401 Unauthorized`. After further investigation, it seems to be an interoperability issue between nerdctl and Harbor registries that have disabled basic authentication.

When nerdctl requests `GET /v2/<name>/blobs/<digest>`, the registry returns a `WWW-Authenticate` header that asks the client to request for a Bearer token with the scope of `repository:<name>:pull` from the specified `realm`. nerdctl would faithfully do so and retry the request with the right credentials, so far so good.

But when nerdctl reuses the same token for `POST /v2/<name>/blobs/uploads/` requests, the scope required is `repository:<name>:pull,push`, so this request is denied and the registry returns `401 Unauthorized`. The problem is that when a request contains a `Authorization` header but failed to authenticate, Harbor uses `Basic realm="harbor"` as the authentication challenge[[ref]](https://github.com/goharbor/harbor/blob/6113469a567623054a58d0f5506b90a55adca7ff/src/server/middleware/v2auth/auth.go#L95-L98), instead of the token service configured.

The registry I'm accessing exclusively relies on Bearer tokens for authentication and has been configured to deny all requests with basic auth. Being mislead by the new `WWW-Authenticate` header, all subsequent requests made by nerdctl would fail.

To solve this interop issue, nerdctl could do one of the following:

1. Make a "preflight" request to the `GET /v2/` API Version Check endpoint to retrieve a token without scope constraint.
2. Retry the request without an `Authorization` header. In this case, Harbor would return the correct `WWW-Authenticate` header that points to the token service.

On Harbor's side, there is https://github.com/goharbor/harbor/issues/17930, but no decision has been made yet.

### Steps to reproduce the issue

1. `nerdctl login ...`
2. `nerdctl push <image>`

### Describe the results you received and expected

Received: `unexpected status from POST request to .../blobs/uploads/: 401 Unauthorized`

Expected: Command succeed without errors.

### What version of nerdctl are you using?

```
Client:
 Version:	v1.0.0
 OS/Arch:	linux/amd64
 Git commit:	c00780a1f5b905b09812722459c54936c9e070e6
 buildctl:
  Version:	v0.10.5
  GitCommit:	bc26045116045516ff2427201abd299043eaf8f7

Server:
 containerd:
  Version:	v1.6.8
[...]

PR #1987, opened 2023-02-08
Fix compatibility issue with Harbor token service

l2dy:
When a request contains `Authorization` header but fails to authenticate, Harbor registry returns `Basic realm="harbor"` as the authentication challenge [1], instead of the token service configured.

nerdctl caches pull-scoped tokens for later use, so when a *push* request is sent with the cached pull-scoped token, Harbor denies the request and instructs the client to attempt "Basic" authentication, which consistently fails if "Basic" authentication is disabled or reserved for special accounts only.

`nerdctl push` to a Harbor registry could fall into this "Basic" authentication loop and eventually fail with 401 Unauthorized.

This patch detects such pattern of repeated requests and retries the request without `Authorization` header to get a proper "Bearer" challenge from the registry.

[1]: https://github.com/goharbor/harbor/blob/6113469a567623054a58d0f5506b90a55adca7ff/src/server/middleware/v2auth/auth.go#L95-L98

Fixes: https://github.com/containerd/nerdctl/issues/1675

### DIFF
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -225,8 +225,12 @@ func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost) error {
 				req.Header.Add(k, vv)
 			}
 		}
-		if err := rh.Authorizer.Authorize(ctx, req); err != nil {
-			return fmt.Errorf("failed to call rh.Authorizer.Authorize: %w", err)
+		// If stuck in a loop of the same unmet `WWW-Authenticate` challenge,
+		// send a request without `Authorization` header.
+		if !infiniteAuthorizationLoop(ress) {
+			if err := rh.Authorizer.Authorize(ctx, req); err != nil {
+				return fmt.Errorf("failed to call rh.Authorizer.Authorize: %w", err)
+			}
 		}
 		res, err := ctxhttp.Do(ctx, rh.Client, req)
 		if err != nil {
@@ -320,4 +324,29 @@ func convertToHostname(serverAddress string) (string, error) {
 	}
 
 	return serverAddress, nil
+}
+
+func infiniteAuthorizationLoop(responses []*http.Response) bool {
+	n := len(responses)
+	if n == 1 || (n > 1 && !sameUnmetAuthorizationChallenge(responses[n-2], responses[n-1])) {
+		return false
+	}
+
+	return true
+}
+
+func sameUnmetAuthorizationChallenge(r1, r2 *http.Response) bool {
+	if r1.StatusCode != 401 || r2.StatusCode != 401 {
+		return false
+	}
+	if r1.Request.Method != r2.Request.Method {
+		return false
+	}
+	if *r1.Request.URL != *r2.Request.URL {
+		return false
+	}
+	if r1.Header.Get("WWW-Authenticate") != r2.Header.Get("WWW-Authenticate") {
+		return false
+	}
+	return true
 }


l2dy:
@fahedouch Should this fix be implemented in nerdctl or containerd's `remotes/docker/authorizer.go`? I will proceed testing this change after your decision.

fahedouch (changes requested):

fahedouch:
> @fahedouch Should this fix be implemented in nerdctl or containerd's `remotes/docker/authorizer.go`? I will proceed testing this change after your decision.

login is on nerdctl side. SGTM to keep it on nerdctl

pkg/cmd/login/login.go
```diff
@@ -225,8 +225,12 @@ func tryLoginWithRegHost(ctx context.Context, rh docker.RegistryHost) error {
...
 			}
 		}
-		if err := rh.Authorizer.Authorize(ctx, req); err != nil {
-			return fmt.Errorf("failed to call rh.Authorizer.Authorize: %w", err)
+		// If stuck in a loop of the same unmet `WWW-Authenticate` challenge,
+		// send a request without `Authorization` header.
+		if !infiniteAuthorizationLoop(ress) {
+			if err := rh.Authorizer.Authorize(ctx, req); err != nil {
+				return fmt.Errorf("failed to call rh.Authorizer.Authorize: %w", err)
+			}
```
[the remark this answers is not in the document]
l2dy: > if **NOT** `infiniteAuthorizationLoop` => `Authorize`, So how to authorize when we have an `infiniteAuthorizationLoop`? .

Proper authorization should happen in the next iteration, if the authentication method that the server requests (in its response) in this iteration has changed.

A side effect is that if the server returns 401 and the same `WWW-Authenticate` header for a temporary failure, retry only works once. Later retries will not contain an `Authorization` Header at all.

I'm assuming that this package is only used in the command line, where state is not persisted across executions, so this side effect is negligible.

> as discussed int t[he related issue ticket](https://github.com/containerd/nerdctl/issues/1675), an `Authorization` Header within an authorization request is the cause of the infiniteLoop when targeting harbor registry, So we have to remove this header to break the infiniteLoop and get a valid `token`. I believe that no need for an authorization header when making an Authorization request. So we can, by default, ensure that Authorization request is proceeded without `Authorization` Header, by removing the `Authorization` Header from `req` before `rh.Authorizer.Authorize(ctx, req)`. So, then drop the infiniteLoop check mechanisms

`rh.Authorizer.Authorize(ctx, req)` sets the `Authorization` Header, so clearing it before the statement is no-op.

pkg/cmd/login/login.go
```diff
@@ -321,3 +325,28 @@ func convertToHostname(serverAddress string) (string, error) {
...
 }
+
+func infiniteAuthorizationLoop(responses []*http.Response) bool {
+	n := len(responses)
+	if n == 1 || (n > 1 && !sameUnmetAuthorizationChallenge(responses[n-2], responses[n-1])) {
+		return false
+	}
+
+	return true
+}
```
AkihiroSuda, line 336: Hard to understand this function

fahedouch, line 336: When `n = 0` you return `true`. Is it normal ?
If `n = 0` return false => no need for next iteration to make a propre Authorization

l2dy: It's not, thanks for spotting this. I copied the logic from https://github.com/containerd/containerd/blob/f0ce2f6d369a35d9b0bf3e6a31404ee3dbfdb9bb/remotes/docker/authorizer.go#L332-L344, but did not notice the difference.

l2dy:
> > @fahedouch Should this fix be implemented in nerdctl or containerd's `remotes/docker/authorizer.go`? I will proceed testing this change after your decision.
> 
> login is on nerdctl side. SGTM to keep it on nerdctl

I missed that login is not where this issue appears, it probably has to be done in authorizer.go, which is also used in [resolver.go](https://github.com/containerd/containerd/blob/ba243ffaaf47f279e18aca980f90f9b152b257d4/remotes/docker/resolver.go#L595-L612) where the retry happens. Given that it's not CLI code anymore, I will need to make the logic more robust.

Sorry for submitting an early-stage unverified PR, I was eager to get some feedback.

fahedouch:
> > > @fahedouch Should this fix be implemented in nerdctl or containerd's `remotes/docker/authorizer.go`? I will proceed testing this change after your decision.
> > 
> > 
> > login is on nerdctl side. SGTM to keep it on nerdctl
> 
> I missed that login is not where this issue appears, it probably has to be done in authorizer.go, which is also used in [resolver.go](https://github.com/containerd/containerd/blob/ba243ffaaf47f279e18aca980f90f9b152b257d4/remotes/docker/resolver.go#L595-L612) where the retry happens. Given that it's not CLI code anymore, I will need to make the logic more robust.
> 
> Sorry for submitting an early-stage unverified PR, I was eager to get some feedback.

login flow and corner case should be kept in the nerdctl side, the harbor auth corner case abstract for containerd.

Issue #1997, opened 2023-02-10
Testutil images cannot be directly used to validate `nerdctl images` output


djdongjin:
### Description

Some integration tests need to check if an image (mostly in [testutil](https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/pkg/testutil/testutil_linux.go#L27-L43) exist in `nerdctl images` output to validate some opeartions (e.g. `nerdctl rmi`).

However, integration tests use different ways to do this and some have bugs:

1. [Use `images --names`](https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/cmd/nerdctl/image_list_test.go#L36-L37): this works since the output combines repo and tag.
2. [Split `testutil.xxxImage`](https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/cmd/nerdctl/image_load_linux_test.go#L47) and only check repo: this works but it's too manual.
3. [`Cmd("images").AssertOutContains(testutil.xxxImage)`](https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/cmd/nerdctl/image_remove_linux_test.go#L40): this is bug because the output has repo and tag separated but not the image var.

For the 3rd one, if check `AssertOutNo` then it always return true because of the format; if check `AssertOutContains` then it bug is surfaced:

```shell
=== RUN   TestRemovePausedImage
    image_remove_linux_test.go:65: assertion failed: 
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test images
        ExitCode: 0
        Stdout:   REPOSITORY                          TAG         IMAGE ID        CREATED                   PLATFORM       SIZE       BLOB SIZE
        ghcr.io/stargz-containers/alpine    3.13-org    ec14c7992a97    Less than a second ago    linux/amd64    5.6 MiB    2.7 MiB
        
        Stderr:   
        
        Failures:
        Expected stdout to contain "ghcr.io/stargz-containers/alpine:3.13-org"
--- FAIL: TestRemovePausedImage (1.68s)
```

### Describe the results you received and expected

Make images check in tests no bug and consistent.

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #1998, opened 2023-02-10
Fix `rmi` bug when checking containers with deleted image

### LINKED ISSUE #1988: `create` container and `rmi` its image causes `rmi` failure
### Description

As titled, after `create` and `rmi`(1) its image, any follwoup `rmi`(2) command will fail. The exact unexpected behavior can be 1 of the 2 and I'm not sure which one we should fix:

1. `rmi`(1) shouldn't be able to remove an image if a container is created/running using that image.
2. `rmi`(2) should skip the container (i.e. when getting used images) if its image is already removed.

`snd` = `sudo nerdctl`

```shell
# a container is created using `alpine`
$ snd ps -a
CONTAINER ID    IMAGE                              COMMAND      CREATED          STATUS     PORTS    NAMES
d8cf1159f0ed    docker.io/library/alpine:latest    "/bin/sh"    5 minutes ago    Created             alpine-d8cf1
$ snd images -a
REPOSITORY    TAG       IMAGE ID        CREATED          PLATFORM       SIZE        BLOB SIZE
alpine        latest    f271e74b17ce    6 minutes ago    linux/amd64    7.0 MiB     3.2 MiB
ubuntu        latest    9a0bdde4188b    6 minutes ago    linux/amd64    80.8 MiB    28.2 MiB

# rmi alpine succeeds and container still in `Created` state
$ snd rmi -f alpine
Untagged: docker.io/library/alpine:latest@sha256:f271e74b17ced29b915d351685fd4644785c6d1559dd1f2d4189a5e851ef753a
Deleted: sha256:8e012198eea15b2554b07014081c85fec4967a1b9cc4b65bd9a4bce3ae1c0c88
$ snd images -a
REPOSITORY    TAG       IMAGE ID        CREATED          PLATFORM       SIZE        BLOB SIZE
ubuntu        latest    9a0bdde4188b    6 minutes ago    linux/amd64    80.8 MiB    28.2 MiB
$ snd ps -a
CONTAINER ID    IMAGE                              COMMAND      CREATED          STATUS     PORTS    NAMES
d8cf1159f0ed    docker.io/library/alpine:latest    "/bin/sh"    6 minutes ago    Created             alpine-d8cf1

# any follow-up rmi will fail now
$ snd rmi -f ubuntu
FATA[0000] failed to get image docker.io/library/alpine:latest for container: image "docker.io/library/alpine:latest": not found
```

Depending on which unexpected behaviors we should fix

for (1): the 1st `rmi` should fail to remove alpine.
for (2): the 2nd `rmi` should skip the `alpine not exist` when getting `used images`.

main

None

### Host information

_No response_

djdongjin:
Fix #1988 (~and part of #1997~ fixed in a separate PR).

Previous: when `rmi` checks `runningImages` by containers, return directly if there is error.
Now: skip this container (since no image returned from it).

### DIFF
--- a/cmd/nerdctl/image_encrypt_linux_test.go
+++ b/cmd/nerdctl/image_encrypt_linux_test.go
@@ -68,7 +68,11 @@ func newJWEKeyPair(t testing.TB) *jweKeyPair {
 func rmiAll(base *testutil.Base) {
 	base.T.Logf("Pruning images")
 	imageIDs := base.Cmd("images", "--no-trunc", "-a", "-q").OutLines()
-	base.Cmd(append([]string{"rmi", "-f"}, imageIDs...)...).AssertOK()
+	// remove empty output line at the end
+	imageIDs = imageIDs[:len(imageIDs)-1]
+	// use `Run` on purpose (same below) because `rmi all` may fail on individual
+	// image id that has an expected running container (e.g. a registry)
+	base.Cmd(append([]string{"rmi", "-f"}, imageIDs...)...).Run()
 
 	base.T.Logf("Pruning build caches")
 	if _, err := buildkitutil.GetBuildkitHost(testutil.Namespace); err == nil {
@@ -97,7 +101,7 @@ func rmiAll(base *testutil.Base) {
 
 		base.T.Logf("Pruning all images (again?)")
 		imageIDs = base.Cmd("images", "--no-trunc", "-a", "-q").OutLines()
-		base.Cmd(append([]string{"rmi", "-f"}, imageIDs...)...).AssertOK()
+		base.Cmd(append([]string{"rmi", "-f"}, imageIDs...)...).Run()
 	}
 }
 
--- a/cmd/nerdctl/image_remove_linux_test.go
+++ b/cmd/nerdctl/image_remove_linux_test.go
@@ -31,7 +31,7 @@ func TestRemoveImage(t *testing.T) {
 	base.Cmd("rmi", "-f", tID).AssertOK()
 
 	base.Cmd("run", "--name", tID, testutil.CommonImage).AssertOK()
-	defer base.Cmd("rm", "-f", tID).Run()
+	defer base.Cmd("rm", "-f", tID).AssertOK()
 
 	base.Cmd("rmi", testutil.CommonImage).AssertFail()
 	defer base.Cmd("rmi", "-f", testutil.CommonImage).Run()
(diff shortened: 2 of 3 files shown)


djdongjin:
Seems `--names` is not compatible with docker :)

```shell
=== RUN   TestRemovePausedImage
    image_remove_linux_test.go:65: assertion failed: 
        Command:  /usr/bin/docker images --names
        ExitCode: 125
        Error:    exit status 125
        Stdout:   
        Stderr:   unknown flag: --names
        See 'docker images --help'.
```

djdongjin:
okay seems docker and nerdctl have different behavior :)

If passing image id, then docker prints errors and doesn't delete images with running/paused containers. If passing image names, docker just untag images:

```shell
# d = docker
$ d ps -a
CONTAINER ID   IMAGE     COMMAND      CREATED          STATUS                   PORTS     NAMES
f2cacb0af911   ubuntu    "sleep 1h"   4 seconds ago    Up 3 seconds                       musing_hellman
c19d9113d612   alpine    "sleep 1h"   28 seconds ago   Up 28 seconds (Paused)             compassionate_hugle
$ d rmi -f $(d images -q)
Error response from daemon: conflict: unable to delete 58db3edaf2be (cannot be forced) - image is being used by running container f2cacb0af911
Error response from daemon: conflict: unable to delete 042a816809aa (cannot be forced) - image is being used by running container c19d9113d612
$ d images
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
ubuntu       latest    58db3edaf2be   2 weeks ago   77.8MB
alpine       latest    042a816809aa   4 weeks ago   7.05MB
$ d rmi -f ubuntu
Untagged: ubuntu:latest
Untagged: ubuntu@sha256:9a0bdde4188b896a372804be2384015e90e3f84906b750c1a53539b585fbbe7f
$ d images
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
<none>       <none>    58db3edaf2be   2 weeks ago   77.8MB
alpine       latest    042a816809aa   4 weeks ago   7.05MB
```

manugupt1:
`--names` is a nerdctl extension.
https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-blue_square-nerdctl-images

cmd/nerdctl/image_remove_linux_test.go
```diff
@@ -27,27 +27,66 @@ func TestRemoveImage(t *testing.T) {
...
+	base.Cmd("create", "--name", tID, testutil.AlpineImage, "sleep", "infinity").AssertOK()
+	defer base.Cmd("rm", "-f", tID).AssertOK()
+
+	// `rmi -f` can remove image associated with created container (`-f` only logs err, not return err)
+	base.Cmd("rmi", testutil.AlpineImage).AssertFail()
+	base.Cmd("rmi", "-f", testutil.AlpineImage).AssertOK()
+	base.Cmd("images").AssertNoOut(testutil.ImageRepo(testutil.AlpineImage))
+
+	// a created container with removed image doesn't impact other `rmi` command
+	base.Cmd("rmi", "-f", testutil.NginxAlpineImage).AssertOK()
```
manugupt1, line 90: Can this work without force rmi with a retagged image name?

cmd/nerdctl/image_remove_linux_test.go
```diff
@@ -27,27 +27,66 @@ func TestRemoveImage(t *testing.T) {
...
 	base.Cmd("run", "--name", tID, testutil.CommonImage).AssertOK()
-	defer base.Cmd("rm", "-f", tID).Run()
+	defer base.Cmd("rm", "-f", tID).AssertOK()
 
 	base.Cmd("rmi", testutil.CommonImage).AssertFail()
-	defer base.Cmd("rmi", "-f", testutil.CommonImage).Run()
 	base.Cmd("rmi", "-f", testutil.CommonImage).AssertOK()
-
-	base.Cmd("images").AssertNoOut(testutil.CommonImage)
+	base.Cmd("images").AssertNoOut(testutil.ImageRepo(testutil.CommonImage))
```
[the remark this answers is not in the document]
djdongjin: Do you mean we should avoid `prune/rmi -f` common images (e.g. `testutil.CommonImage`) and only `prune/rmi -f` a custom name (either from `tag` or `build`)? Both here and other integration tests

pkg/cmd/image/remove.go
```diff
@@ -89,10 +92,27 @@ func Remove(ctx context.Context, client *containerd.Client, args []string, optio
...
+			errs = append(errs, err.Error())
+		}
+	}
+
+	if len(errs) > 0 {
+		msg := fmt.Sprintf("%d errors:\n%s", len(errs), strings.Join(errs, "\n"))
+		if !options.Force || fatalErr {
+			return errors.New(msg)
+		}
+		logrus.Error(msg)
```
djdongjin, line 115: This is to exactly match docker's [error handling behavior](https://github.com/docker/cli/blob/dfb36eaef8729914749500b148649e05ceeddd03/cli/command/image/remove.go#L61-L88) (for each err, if not `IsErrNotFound`, then it's fatal; if there is any fatal error, return a joined error, else log error and return nil).

If using `WalkAll`, an edge case will be inconsistent with docker: if all errors are `IsErrNotFound` (e.g. `docker|nerdctl rmi -f afsd afdafaf aff`), docker return nil and nerdctl return a joined err. TBH I don't feel it makes much sense to distinguish the specific edge case in docker...

cmd/nerdctl/image_encrypt_linux_test.go
```diff
@@ -68,7 +68,7 @@ func newJWEKeyPair(t testing.TB) *jweKeyPair {
 func rmiAll(base *testutil.Base) {
 	base.T.Logf("Pruning images")
 	imageIDs := base.Cmd("images", "--no-trunc", "-a", "-q").OutLines()
-	base.Cmd(append([]string{"rmi", "-f"}, imageIDs...)...).AssertOK()
+	base.Cmd(append([]string{"rmi", "-f"}, imageIDs...)...).Run()
```
djdongjin, line 71: This change is because when calling `rmi -f allIDs`, tests cannot ensure no running containers. (e.g., in `TestImageEncryptJWE` there is always a registry container running, causing this error:

```
conflict: unable to delete sha256:5bb9b919833aa955dfe1d1121cc038330b025ec6506ce47066c9192927e3dc3d (cannot be forced) - image is being used by running container c1b56cc721c5d1717f842ca7cc3277122e09cb964fcabca22b4bbeea4d57880a
```

pkg/testutil/testutil.go
```diff
@@ -487,7 +487,7 @@ func (c *Cmd) OutLines() []string {
 	c.Base.T.Helper()
 	out := c.Out()
 	// FIXME: improve memory efficiency
-	return strings.Split(out, "\n")
+	return strings.Split(strings.TrimSpace(out), "\n")
```
djdongjin, line 490: This change is because otherwise it produces an empty string, which is useless and when used in follow up cmds (e.g. `rmi`) causes errors like this:

```
filters: parse error: [name==]: expected value or quoted: invalid argument: invalid argument"
```

cmd/nerdctl/image_remove_linux_test.go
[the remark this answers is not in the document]
manugupt1: Yep (also other integ tests; but not in this PR).
I think we should start with this PR or when we write new tests should be fine.

cmd/nerdctl/image_encrypt_linux_test.go
djdongjin: added

cmd/nerdctl/image_remove_linux_test.go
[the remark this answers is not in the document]
djdongjin: tested locally. using `tag` makes it incompatible with docker.

If there is a container (assume exited) from an image, `docker rmi image` will succeed if `image` is a tag (e.g. `foobar`, so `foobar` is untagged) but fail if `image` is the real image name (e.g., `alpine`)

cmd/nerdctl/image_remove_linux_test.go
djdongjin: replied in comment above.

AkihiroSuda (approved):
Thanks

PR #1999, opened 2023-02-10
Fix images output check bug

### LINKED ISSUE #1997: Testutil images cannot be directly used to validate `nerdctl images` output

djdongjin:
Fix #1997


manugupt1 (approved):
Thanks!

pkg/testutil/testutil.go
```diff
@@ -450,17 +450,6 @@ func (c *Cmd) AssertOutStreamsExactly(stdout, stderr string) {
...
-	c.Base.T.Helper()
-	fn := func(stdout string) error {
-		if strings.Contains(stdout, s) {
-			return fmt.Errorf("expected not to contain %q, got %q", s, stdout)
-		}
-		return nil
-	}
-	c.AssertOutWithFunc(fn)
-}
-
```
AkihiroSuda, line 463: Is this change relevant to the bug?
Otherwise please consider splitting the commit

pkg/testutil/testutil_linux.go
```diff
@@ -16,7 +16,9 @@
 
 package testutil
 
-import "fmt"
+import (
+	"fmt"
+)
```
AkihiroSuda, line 21: Not necessary

pkg/testutil/testutil.go
```diff
@@ -684,3 +684,9 @@ func Identifier(t testing.TB) string {
 	}
 	return s
 }
+
+// ImageRepo is a test helper that simply splits a full image name (i.e., not id/digest)
+// by `:` and returns the 1st part.
+func ImageRepo(s string) string {
+	return strings.Split(s, ":")[0]
```
AkihiroSuda, line 691: Doesn't seem to work when the image name contains a registry port number, e.g., `example.com:8080/example`

manugupt1: I think we can use imgutil.ParseRepoTag and return the first part.

AkihiroSuda (approved):
Thanks

PR #2009, opened 2023-02-14
Retag commonimage before loading to ensure it does not fail.

manugupt1:
Removing CommonImage can fail if testimage can be used by some other test. This leads to flakiness. Retagging to ensure that it does not fail. Also add t.Parallel() to ensure this test can run concurrently.

Failed test case can be found here: [TestLoadStdinFromPipe](https://github.com/containerd/nerdctl/actions/runs/4171221656/jobs/7220936048)


AkihiroSuda (approved):
Thanks

PR #2021, opened 2023-02-17
Migrate from k8s.gcr.io to registry.k8s.io

aniruddha2000:
This PR migrate from `k8s.gcr.io` to `registry.k8s.io`

Part Of https://github.com/kubernetes/k8s.io/issues/4780

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -375,7 +375,7 @@ After update, the following will be written to your Docker config file located a
     "us.gcr.io": "gcloud",
     "eu.gcr.io": "gcloud",
     "asia.gcr.io": "gcloud",
-    "staging-k8s.gcr.io": "gcloud",
+    "staging-registry.k8s.io": "gcloud",
     "marketplace.gcr.io": "gcloud"
   }
 }


docs/registry.md
```diff
@@ -375,7 +375,7 @@ After update, the following will be written to your Docker config file located a
     "us.gcr.io": "gcloud",
     "eu.gcr.io": "gcloud",
     "asia.gcr.io": "gcloud",
-    "staging-k8s.gcr.io": "gcloud",
+    "staging-registry.k8s.io": "gcloud",
```
AkihiroSuda, line 378:
```console
$ nslookup staging-registry.k8s.io
;; connection timed out; no servers could be reached
```

AkihiroSuda:
gcloud (418.0.0) seems still using `staging-k8s.gcr.io`

mrbobbytables:
Just an FYI - staging-k8s.gcr.io used to be a push only endpoint, however that is no longer the case and largely acts as a placeholder to swap out for other addresses. It should be safe to remove, but also fine to leave as is. It won't really be used for anything.

PR #2025, opened 2023-02-17
update Go (1.20)

AkihiroSuda:


fahedouch (approved):
So cool to onboard v1.20 😀

PR #2027, opened 2023-02-17
follow-up: upgrade to go1.20

fahedouch:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containerd/nerdctl
 
-go 1.19
```
AkihiroSuda, line 3: This is intended to be the minimum supported version

fahedouch: Why ? is there sthg blocking on containerd side ?

fahedouch: Sorry I miss understood your comment! ok but is it the right place to put  the minimum supported version ? the `go.mod` ?

fahedouch: SGTM `A list of minimum versions of other modules required by the current module.` from official [golang doc](https://go.dev/doc/modules/gomod-ref)

fahedouch: Thanks for your feedback

PR #1985, opened 2023-02-08
[Refactor] refactor system prune flagging process

suyanhanx:
Signed-off-by: Han Xu <suyanhanx@gmail.com>

PR #2030, opened 2023-02-18
Make confirm logic in `volume prune` be consistent with other Prune

suyanhanx:
Currently, in the logic of `volume prune`, `Stdin` is only used for confirmation of `force` flag. Therefore, this PR moves the confirmation logic from `pkg` to `cmd` side, and removes `Stdin` from the options at the same time.

This is also the case for other prune commands.

reference: https://github.com/containerd/nerdctl/pull/1985#discussion_r1110022415

### DIFF
--- a/cmd/nerdctl/system_prune.go
+++ b/cmd/nerdctl/system_prune.go
@@ -121,7 +121,6 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
 			GOptions: globalOptions,
 			Force:    true,
 			Stdout:   cmd.OutOrStdout(),
-			Stdin:    cmd.InOrStdin(),
 		}); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


cmd/nerdctl/system_prune.go
```diff
@@ -121,7 +121,6 @@ func systemPruneAction(cmd *cobra.Command, args []string) error {
 			GOptions: globalOptions,
 			Force:    true,
 			Stdout:   cmd.OutOrStdout(),
-			Stdin:    cmd.InOrStdin(),
```
AkihiroSuda, line 124: ?

AkihiroSuda:
Could you explain this PR?

suyanhanx:
> Could you explain this PR?

Sorry for not explaining this PR. 

Description updated.

AkihiroSuda (approved):
Thanks

Issue #1940, opened 2023-01-28
[Refactor] simplify the stdout rewrite logic in image.Save


Zheaoli:
(below is more for discussion and see other reviewers' thought).

I think we should remove `Output string` from `ImageSaveOptions`. Functionally, it represents the same thing as `Stdout io.Writer` and having both might cause inconsistency and confusing, for example:

- Even `Output != ""`, we init `Stdout = cmd.OutOrStdout()` when creating `ImageSaveOptions`, and then change it after the file writer created (i.e., here).
- Should a caller specify `Output`, or `Stdout io.Writer`?

If we remove `Output` and keep `Stdout` only:

1. `pkg/cmd` side logic doesn't need to know where the writer comes from. It just writes content to it.
2. On `cmd` side, we init `opt.Stdout = cmd.OutOrStdout()` if `output == ""` or create a file writer and use it as `opt.Stdout` (i.e., this pard of code).

_Originally posted by @djdongjin in https://github.com/containerd/nerdctl/pull/1924#discussion_r1087419176_

Zheaoli:
@djdongjin @suyanhanx

suyanhanx:
Thanks for issue.

PR #2037, opened 2023-02-20
Refactor: func `Save` in package `pkg/cmd/image` shouldn't create file from `options.Output`

rkonfj:
> `pkg` side should assume a `StdOut` is given (i.e., not create a file from `options.Output`)

Signed-off-by: Yakun Jing <rkonfj@gmail.com>

### DIFF
--- a/cmd/nerdctl/image_save.go
+++ b/cmd/nerdctl/image_save.go
@@ -18,10 +18,12 @@ package main
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/containerd/nerdctl/pkg/api/types"
 	"github.com/containerd/nerdctl/pkg/clientutil"
 	"github.com/containerd/nerdctl/pkg/cmd/image"
+	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 )
 
@@ -54,10 +56,6 @@ func processImageSaveOptions(cmd *cobra.Command) (types.ImageSaveOptions, error)
 		return types.ImageSaveOptions{}, err
 	}
 
-	output, err := cmd.Flags().GetString("output")
-	if err != nil {
-		return types.ImageSaveOptions{}, err
-	}
 	allPlatforms, err := cmd.Flags().GetBool("all-platforms")
 	if err != nil {
 		return types.ImageSaveOptions{}, err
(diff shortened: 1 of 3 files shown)


rkonfj:
What I'm a bit unsure about is whether we need to move the following code into the `main` package of command?
https://github.com/containerd/nerdctl/blob/bba23d57744dbdd14c7674efab3e743600a3ba1f/pkg/cmd/image/save.go#L35-L44

djdongjin:
> What I'm a bit unsure about is whether we need to move the following code into the main package of command?

Yes this should be part of the `cmd` side (i.e., cmd/nerdctl/image_save.go). `pkg` side should assume a `StdOut` is given (i.e., not create a file from `options.Output`)

cmd/nerdctl/image_save.go
```diff
@@ -69,29 +67,56 @@ func processImageSaveOptions(cmd *cobra.Command) (types.ImageSaveOptions, error)
...
 		AllPlatforms: allPlatforms,
-		Output:       output,
 		Platform:     platform,
 	}, err
 }
 
-func saveAction(cmd *cobra.Command, args []string) error {
-	if len(args) == 0 {
-		return fmt.Errorf("requires at least 1 argument")
+func determineSaveToFile(cmd *cobra.Command) (*os.File, error) {
```
suyanhanx, line 75: Maybe `determineSaveToFile` -> `determineFileToSave` ?

cmd/nerdctl/image_save.go
```diff
@@ -69,29 +67,56 @@ func processImageSaveOptions(cmd *cobra.Command) (types.ImageSaveOptions, error)
...
 		AllPlatforms: allPlatforms,
-		Output:       output,
 		Platform:     platform,
 	}, err
 }
 
-func saveAction(cmd *cobra.Command, args []string) error {
-	if len(args) == 0 {
-		return fmt.Errorf("requires at least 1 argument")
+func determineFileToSave(cmd *cobra.Command) (*os.File, error) {
```
djdongjin, line 75: maybe just put the logic within `saveAction`, which should be cleaner. Most of the logic are just err check.

djdongjin (approved):

suyanhanx (approved):

suyanhanx:
close #1940 .

AkihiroSuda (approved):
Thanks

PR #2041, opened 2023-02-21
refactor(container-list): separate main logic out of printing logic

davidhsingyuchen:
PR moves the "exclude non-running containers if `All` is `true`" logic out of `printContainers`. The naming pattern in #1923 (i.e., `XXXCommandHandler()` for use by `cmd/nerdctl` & `XXX()` for the main logic func) is followed.


Zheaoli (approved):
LGTM

Issue #2040, opened 2023-02-21
Snapshotters: use containerd's AppendInfoHandlerWrapper


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/v1.2.0/pkg/imgutil/snapshotter.go should be replaced with:
- https://github.com/containerd/containerd/pull/8036

danishprakash:
I can see this is tagged expert, does that mean it's out of bounds for external contributors? I'd like to look into this.

AkihiroSuda:
> I can see this is tagged expert, does that mean it's out of bounds for external contributors? I'd like to look into this.

Thanks, contribution is welcome from anybody, but needs to be familiar with stargz, nydus, overlaybd, etc.

PR #2063, opened 2023-03-01
remote snapshotters: migrate to containerd-based label management

ktock:
Fixes: https://github.com/containerd/nerdctl/issues/2040

This PR migrates https://github.com/containerd/nerdctl/blob/v1.2.0/pkg/imgutil/snapshotter.go to `github.com/containerd/containerd/pkg/snapshotters`-based remote snapshotter labels management introduced in https://github.com/containerd/containerd/pull/8036 .


pkg/imgutil/snapshotter.go
```diff
@@ -90,45 +93,3 @@ func (dsn *defaultSnapshotterOpts) apply(config *pull.Config, _ref string) {
...
-// stargzSnapshotterOpts for stargz snapshotter
-type stargzSnapshotterOpts struct {
-	remoteSnapshotter
-}
-
-func (ssn *stargzSnapshotterOpts) apply(config *pull.Config, ref string) {
-	// TODO: support "skip-content-verify"
-	config.RemoteOpts = append(
-		config.RemoteOpts,
-		containerd.WithImageHandlerWrapper(source.AppendDefaultLabelsHandlerWrapper(ref, prefetchSize)),
```
djdongjin, line 103: Hi @ktock, I have a question about this change. Take stargz as an example (other snapshotter may not have extra lables/annotations), compared to `ctdsnapshotters.AppendInfoHandlerWrapper(ref)`, [`source.AppendDefaultLabelsHandlerWrapper(ref, prefetchSize)`](https://github.com/containerd/stargz-snapshotter/blob/f93d3dcfbff83ef82f97f99c99553bf55ff24d4d/fs/source/source.go#L146) has some [extra labels](https://github.com/containerd/stargz-snapshotter/blob/f93d3dcfbff83ef82f97f99c99553bf55ff24d4d/fs/source/source.go#L180-L183). I guess for `config.TargetPrefetchSizeLabel`, stargz may just take default value if it doesn't exist; but I'm not sure about `targetURLsLabel`.

In general, I'm curious how does this will impact a remote snapshotter? Also what if a remote snapshotter needs extra labels (e.g., `TargetPrefetchSizeLabel/targetURLsLabel` in stargz)?

Thanks.

ktock: Thank you for the comments.

> extra labels

Yes, we need to determine which extra labels need to be handled in nerdctl because containerd's labels cover only basic functionalities of lazy pulling (e.g. lack IPFS support).

> targetURLsLabel

This is needed for IPFS support. This label propagates `urls` field in OCI descriptor. https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/specs-go/v1/descriptor.go#L33
I'm currently working on neccessary changes on stargz-snapshotter's side: https://github.com/containerd/stargz-snapshotter/pull/1131

> In general, I'm curious how does this will impact a remote snapshotter?

No impact if the snapshotter already understand these containerd labels. If the snapshotter doesn't understand these labels and if they need to use some of the information passed via these labels, they need to add support for them. (But most of snapshotter should already support them because they have been used by containerd CRI plugin for Kuberenetes integration.)

> Also what if a remote snapshotter needs extra labels (e.g., TargetPrefetchSizeLabel/targetURLsLabel in stargz)?

In that case, we need to add the label handling logic to this pkg (`pkg/imgutil/snapshotter.go`).

cc @imeoer @BigVan PTAL this change and please let us know if you want extra labels for nydus and overlaybd snapshotters. IIRC nydus already support these labels.

imeoer: Thanks @ktock for reminding us, this change LGTM, no additional labels are needed by nydus for the time being.

djdongjin: Another thing I notice is that stargz and the containerd-replacement have different label strings even for labels of the same purpose:

https://github.com/containerd/stargz-snapshotter/blob/f93d3dcfbff83ef82f97f99c99553bf55ff24d4d/fs/source/source.go#L64-L72

v.s. 

https://github.com/containerd/containerd/blob/759a6350bc16523c64639e0f75ef9dfc71a4832e/pkg/snapshotters/annotations.go#L34-L44

Is the plan for stargz to remove its own label key and use containerd's? (e.g. replacing [source.AppendDefaultLabelsHandlerWrapper(ref, prefetchSize)](https://github.com/containerd/stargz-snapshotter/blob/f93d3dcfbff83ef82f97f99c99553bf55ff24d4d/fs/source/source.go#L146) with `ctdsnapshotters.AppendInfoHandlerWrapper(ref)` in nerdctl not only removes the mentioned labels in my previous comment, but also change the key strings of common labels)

BigVan: Thanks for the notice. @ktock 
Currently, overlaybd doesn't need more extra labels.

AkihiroSuda (approved):
Thanks

PR #2067, opened 2023-03-02
Update naming of filter containers function && `container list` handler

suyanhanx:
Update the naming of the containers filter function and `container list` handler to match the current naming pattern.

### DIFF
--- a/cmd/nerdctl/container_list.go
+++ b/cmd/nerdctl/container_list.go
@@ -124,5 +124,5 @@ func psAction(cmd *cobra.Command, args []string) error {
 	}
 	defer cancel()
 
-	return container.ListCommandHandler(ctx, client, options)
+	return container.List(ctx, client, options)
 }
--- a/pkg/cmd/container/list.go
+++ b/pkg/cmd/container/list.go
@@ -40,22 +40,22 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// ListCommandHandler prints containers according to `options`.
-func ListCommandHandler(ctx context.Context, client *containerd.Client, options types.ContainerListOptions) error {
-	containers, err := List(ctx, client, options.Filters, options.LastN, options.All)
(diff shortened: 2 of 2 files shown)


Zheaoli (approved):
LGTM

Zheaoli:
This PR is a little bit different with #2041. I think we should keep one style in command style. I prefer #2067

suyanhanx:
I forgot to review it. Otherwise, I would have commented in that PR.

davidhsingyuchen:
@Zheaoli Thanks for merging #2041.

> This PR is a little bit different with #2041. I think we should keep one style in command style. I prefer #2067

I agree that ideally there should be only command style, but #2067 (i.e., this PR) unexports the "main logic" function (see https://github.com/containerd/nerdctl/issues/1855#issuecomment-1400837364 for more details), which means that other projects cannot reuse that function.

I went with `ListCommandHandler` in #2041 because #1923 had been merged, and I'm willing to do a one-shot PR to change all the commands to use that style, but if that style is not preferred, could you help shed some light on if I can just go ahead and change `filterContainers` to be `FilterContainers`?

As a side note, I feel that `filterContainers` sounds that the input of it is already a list of containers, and what it does is only filters out some of them, while it's not the case, so the semantics of it could be a bit off. Don't have a strong opinion on this though.

suyanhanx:
> but #2067 (i.e., this PR) unexports the "main logic" function (see [#1855 (comment)](https://github.com/containerd/nerdctl/issues/1855#issuecomment-1400837364) for more details), which means that other projects cannot reuse that function.

This is a minor issue. I think we can change it when needed.

> As a side note, I feel that `filterContainers` sounds that the input of it is already a list of containers, and what it does is only filters out some of them, while it's not the case, so the semantics of it could be a bit off. Don't have a strong opinion on this though.

Not without reason. If you come up with other good names, you can also propose them.

Issue #2052, opened 2023-02-25
Support State.Error field in inspect command


ningziwen:
### What is the problem you're trying to solve

`nerdctl inspect` cannot return State.Error field today.

```
$ nerdctl run -d public.ecr.aws/amazonlinux/amazonlinux:2
2fd4c9d197320279055bfca5a109433935e762fc91bcbe371815a6cf7f39fe03
$ nerdctl inspect --format {{.State.Status}}:{{.State.Error}} 2fd4c9d197320279055bfca5a109433935e762fc91bcbe371815a6cf7f39fe03
WARN[0000] failed to inspect NetNS                       error="failed to Statfs \"/proc/4058/ns/net\": no such file or directory" id=2fd4c9d197320279055bfca5a109433935e762fc91bcbe371815a6cf7f39fe03
FATA[0000] 1 errors: [template parsing error: template: :1:26: executing "" at <.State.Error>: map has no entry for key "Error"]
```

### Describe the solution you'd like

State.Error exists in the inspect response, and it is an existing TODO in comment. https://github.com/containerd/nerdctl/blob/e09d7f8a419ee2d1152386ab654afc55f6e2201b/pkg/inspecttypes/dockercompat/dockercompat.go#L170

### Additional context

https://github.com/runfinch/finch/issues/242

fahedouch:
container `exit error` is rendered when the container process exit : 
https://github.com/containerd/containerd/blob/v1.6.19/process.go#L44

we need to store it in container metadata so we can retrieve it later

I will take this ticket :)

fahedouch:
Closed by https://github.com/containerd/nerdctl/pull/2073

PR #2073, opened 2023-03-06
Support State.Error field in inspect command

fahedouch:
[see](https://github.com/containerd/nerdctl/issues/2052)


pkg/labels/labels.go
```diff
@@ -90,6 +89,9 @@ const (
 	// PIDContainer is the `nerdctl run --pid` for restarting
 	PIDContainer = Prefix + "pid-container"
 
+	// RunError encapsulates a container run error.
```
AkihiroSuda, line 92: Please add how this is consumed

pkg/labels/labels.go
```diff
@@ -90,6 +89,9 @@ const (
...
+	RunError = Prefix + "run-error"
```
AkihiroSuda, line 93: Wonder if this should be `start-error` or just `error` or `errors`

fahedouch: `start-error` LGTM. `error` has a large scope.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@ningziwen PTAL

ningziwen:
Thanks! @fahedouch 

I think this change only populates the error from `container create` and `container run`.

If we want to make this field Docker compatible, the scope of the ErrorMsg also includes container start, stop or remove, based on the [comment](https://github.com/moby/moby/blob/2cf63891427d68661bb8e3c58969058c62dbc85d/container/state.go#L32). 

The [line](https://github.com/moby/moby/blob/2cf63891427d68661bb8e3c58969058c62dbc85d/daemon/start.go#L125) to populate errorMsg in docker start

I didn't find the line explicitly in docker stop/remove but found [here](https://github.com/moby/moby/blob/2cf63891427d68661bb8e3c58969058c62dbc85d/daemon/monitor.go#L61) is probably used to populate when container exits.

Maybe we can decide if we want to populate errorMsg in the above scenarios. Based on the decision, we can change the commit message to reflect the change scope or extend the PR.

pkg/labels/labels.go
fahedouch: switch for `error` to handle `stop` and `remove` verb(s)

fahedouch:
> If we want to make this field Docker compatible, the scope of the ErrorMsg also includes container start, stop or remove

`start` is [supported](https://github.com/containerd/nerdctl/pull/2073/files#diff-f21a0764536449a426504492e373a01950701927368d6c6968d895fb316a9a9cR200-R204) 

`stop` and `remove` added.

@ningziwen PTAL

ningziwen (approved):

fahedouch:
@AkihiroSuda PTAL , I added some changes

cmd/nerdctl/container_inspect_linux_test.go
```diff
@@ -154,3 +155,54 @@ func TestContainerInspectContainsLabel(t *testing.T) {
...
+	testContainer := testutil.Identifier(t)
+	base := testutil.NewBase(t)
+
+	type testCase struct {
+		name, containerName, cmd string
+		want                     dockercompat.ContainerState
+	}
+	errStatus := ""
+	if base.Target == testutil.Docker {
+		errStatus = "created"
```
AkihiroSuda, line 170: Why differs for Docker?

fahedouch: nerdctl : run error means task is nil so the `Status` is empty because `Status` comes from Task.
docker : run error gives => `Status` = `created`.

AkihiroSuda: Please add that in the code comment lines

fahedouch: @AkihiroSuda Fixed

AkihiroSuda (approved):
Thanks

PR #2109, opened 2023-03-18
Upgrade cosign to 2.0.0 in tests

ningziwen:
Upgrade cosign to 2.0.0 in tests.

Need to add --yes to skip prompt because the [prompt](https://github.com/sigstore/cosign/pull/1909) was added in 1.9.0.

### DIFF
--- a/pkg/signutil/cosignutil.go
+++ b/pkg/signutil/cosignutil.go
@@ -46,6 +46,7 @@ func SignCosign(rawRef string, keyRef string) error {
 		cosignCmd.Env = append(cosignCmd.Env, "COSIGN_EXPERIMENTAL=true")
 	}
 
+	cosignCmd.Args = append(cosignCmd.Args, "--yes")
 	cosignCmd.Args = append(cosignCmd.Args, rawRef)
 
 	logrus.Debugf("running %s %v", cosignExecutable, cosignCmd.Args)
--- a/Dockerfile
+++ b/Dockerfile
@@ -282,7 +282,7 @@ WORKDIR /go/src/github.com/containerd/nerdctl
 VOLUME /tmp
 ENV CGO_ENABLED=0
 # copy cosign binary for integration test
-COPY --from=gcr.io/projectsigstore/cosign:v1.3.1@sha256:3cd9b3a866579dc2e0cf2fdea547f4c9a27139276cc373165c26842bc594b8bd /ko-app/cosign /usr/local/bin/cosign
+COPY --from=gcr.io/projectsigstore/cosign:v2.0.0@sha256:728944a9542a7235b4358c4ab2bcea855840e9d4b9594febca5c2207f5da7f38 /ko-app/cosign /usr/local/bin/cosign
 # enable offline ipfs for integration test
 COPY ./Dockerfile.d/test-integration-etc_containerd-stargz-grpc_config.toml /etc/containerd-stargz-grpc/config.toml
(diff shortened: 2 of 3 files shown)


pkg/signutil/cosignutil.go
```diff
@@ -46,6 +46,7 @@ func SignCosign(rawRef string, keyRef string) error {
 		cosignCmd.Env = append(cosignCmd.Env, "COSIGN_EXPERIMENTAL=true")
 	}
 
+	cosignCmd.Args = append(cosignCmd.Args, "--yes")
```
fahedouch, line 49: Hi @ningziwen , Is it relevant to expose the prompt feature as nerdctl options ? And make it silent within integration tests .

ningziwen: This is not to expose --yes, but to add --yes to make it silent by default. So it can be silent within integration tests and in the downstream software that is comsuming nerdctl.

The alternative option is to expose --cosign-yes in nerdctl. But I feel as Cosign is treated as the dependency of nerdctl feature, nerdctl can be kind of opininated to say yes to it, and make users have the expection of saying yes to this. (we can also add some documentation if needed)

fahedouch: LGTM but please introduce some documentation

fahedouch: FYI prompt  can be disablied by exporting `
COSIGN_PASSWORD` 
[source](https://github.com/containerd/nerdctl/blob/main/docs/cosign.md )
```
# Export your COSIGN_PASSWORD to prevent CLI prompting
$ export COSIGN_PASSWORD=$COSIGN_PASSWORD
```

ningziwen: I think COSIGN_PASSWORD only disables the prompt of inputting cosign key passphrase. The prompt in this PR is only for private statement, which can't be disabled by COSIGN_PASSWORD.

```
	PrivacyStatement = `
        Note that there may be personally identifiable information associated with this signed artifact.
        This may include the email address associated with the account with which you authenticate.
        This information will be used for signing this artifact and will be stored in public transparency logs and cannot be removed later.`
	PrivacyStatementConfirmation = "        By typing 'y', you attest that you grant (or have permission to grant) and agree to have this information stored permanently in transparency logs."
```

`--yes` is also specifically for privacy statement.

ningziwen: Added documentation.

fahedouch (commented):
LGTM. Just small last thing; would you please append `Upgrade cosign to 2.0.0 in integration tests` to your commit text for traceability purposes

pkg/signutil/cosignutil.go
manugupt1: From a privacy point of view, I think it is better to make it explicit for nerdctl users imo. We should make our users opt-in.

ningziwen:
> LGTM. Just small last thing; would you please append `Upgrade cosign to 2.0.0 in tests` to your commit text for traceability purposes

@fahedouch Fixed

pkg/signutil/cosignutil.go
fahedouch: agree regarding `privacy point of view` purpose, but I think the idea here is to unblock the upgrade to `2.0.0`, so let's accept it as default value, and make new PR to make it explicit for nerdctl users

fahedouch (approved):

fahedouch:
Thanks

PR #2112, opened 2023-03-20
fix(docs): update installation steps for win and linux

profnandaa:
This PR updates the installation instructions for Windows and Linux.

We didn't have Windows installation steps and the Linux one was using `brew` (needs another extra installation);
sort of an anti-pattern from how we install the rest of the ecosystem tools like `containerd`, etc.


profnandaa:
`ctr run` works okay but getting this when I run `nerdctl run`, let me fix that and update the README too:

```
nerdctl run --rm -it mcr.microsoft.com/windows/nanoserver:ltsc2022 cmd
time="2023-03-20T04:32:35Z" level=fatal msg="failed to create default network: needs CNI plugin \"nat\" to be installed in CNI_PATH (\"C:\\\\Program Files\\\\containerd\\\\cni\\\\bin\"), see https://github.com/containernetworking/plugins/releases: exec: \"C:\\\\Program Files\\\\containerd\\\\cni\\\\bin\\\\nat\": file does not exist"
```

README.md
```diff
@@ -134,16 +134,20 @@ In addition to containerd, the following components should be installed:
 
 These dependencies are included in `nerdctl-full-<VERSION>-<OS>-<ARCH>.tar.gz`, but not included in `nerdctl-<VERSION>-<OS>-<ARCH>.tar.gz`.
 
-### Brew
```
AkihiroSuda, line 137: Why remove?

README.md
```diff
@@ -154,14 +158,30 @@ $ limactl start
...
+```
+If you already added `$Env:ProgramFiles\containerd` in the `PATH`, you should have `nerdctl` accessible from anywhere now:
 
-- Linux containers: Known to work on WSL2
-- Windows containers: experimental support for Windows (see below for features that are currently known to work)
+```powershell
+nerdctl --version
+```
+
+> For Linux Containers on Windows (LCOW), you can use WSL2 and follow similar steps for [Linux](#linux).
```
AkihiroSuda, line 180: To me, LCOW means running Linux containers on a Windows executable of `containerd.exe`.
nerdctl doesn't support this LCOW.

README.md
profnandaa: I just renamed to `Linux` the other items are based on the OS types instead, i.e.  Windows, FreeBSD, MacOS.

README.md
profnandaa: Ooh, yes, this mixes up stuff. I'll revert back to the previous NOTE instead.

profnandaa:
@AkihiroSuda -- thanks for the review, going to make the changes.
I finally have `nerdctl` running but the way I've set up the CNI-plugin is hacky hence I get this warning:
```
level=warning msg="default network named \"nat\" does not have an internal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl"
```
What is the best way of setting up the CNI plugin for windows.

I was trying to adapt this [bash script here](https://github.com/containerd/containerd/blob/main/script/setup/install-cni-windows) (which I think should be done in PowerShell instead, working on that) - https://gist.github.com/profnandaa/e9ab7bd7e069ce2728840b041ce4c300

aznashwan:
@profnandaa don't worry about that warning message, even if you didn't explicitly ask for network settings for your container, it's due to the fact that [`nat` is the network name `nerdctl` defaults to on Windows](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil_windows.go#L28).

This special treatment of the default network is the result of https://github.com/containerd/nerdctl/pull/1538, after which `nerdctl` can tell whether a network was initially created by itself as the default network by [setting a special label on it](https://github.com/containerd/nerdctl/blob/main/pkg/netutil/netutil.go#L358).

> I was trying to adapt this [bash script here](https://github.com/containerd/containerd/blob/main/script/setup/install-cni-windows)

That's perfect since it's basically the "reference configuration" which is used in both the [containerd CI](https://github.com/containerd/containerd/blob/main/.github/workflows/windows-periodic.yml#L201), as well as being basically 1:1 lifted to be used in the [nerdctl CI too](https://github.com/containerd/nerdctl/blob/main/hack/configure-windows-ci.ps1#L32).

Although there's basically no reason to worry about that warning message, and I wouldn't necessarily bother mentioning it in the docs, but you could dodge it by opening the network config file shown in `nerdctl network list` and either:
* renaming that network to something other than `nat` (NOTE: the `type` should still stay `nat`!)
* keep the `nat` name but add the following line to the top-level of the config:
```js
{
  ...
   // NOTE: this ID field is all you'll really need, and you can invent your own ID if needed.
  "nerdctlID": "d919a100ce6b45524d415d52d088d5817587c6dd8c3691b03b8063c44d043523",
  "nerdctlLabels": {
    "nerdctl/default-network": "true"
  },
  ...
}
```

README.md
profnandaa: @jsturtevant -- I've replaced that with the standard installation process on Linux. I don't see a reason to install `brew` first so that you can install `nerdctl` when most likely they have already installed `containerd`, etc. using the standard process of getting the binaries from Github with `curl`.

PR #2117, opened 2023-03-21
rootlessutil: always claim non-rootless on non-Linux.

### LINKED ISSUE #2115: Properly handle Windows WCOW containers in `pkg/rootlessutil`.
### Description

TL;DR: on Windows, `nerdctl` wrongly thinks it's always running as a rootless parent (read background below), so we should try to reach consensus on whether Windows containers should be considered rootless or not.
Q1: Would you consider Windows containers managed by HCS to be "rootless" or not?
Q2: Should it depend on the isolation mode of the Windows container? (i.e. Hyper-V vs. Process containers?)

### Background.

Containerd and nerdctl can run in containers in [rootless mode](https://rootlesscontaine.rs), where both the containers **as well as the runtime** can be run under non-root-users, thus providing an additional layer of isolation.

On Linux, Nerdctl leverages [RootlessKit](https://github.com/rootless-containers/rootlesskit) to run containers in rootless mode, and performs several checks arounds the codebase using `rootless.IsRootless` which leads to some special codepaths such as:
* setting up [`slirp4dns` for rootless children](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_network.go#L323-L328).
* setting AppArmor [policies for rootless children](https://github.com/containerd/nerdctl/blob/main/pkg/infoutil/infoutil_linux.go#L57).
* added setup for [port mappings for rootless children](https://github.com/containerd/nerdctl/blob/main/pkg/ocihook/ocihook.go#L288-L329).

[`rootlessutil.IsRootless()`](https://github.com/containerd/nerdctl/blob/main/pkg/rootlessutil/rootlessutil.go#L29) should return `true` if the process is either a rootless parent or child, whose subchecks consist of:
* `IsRootlessParent()` just checks whether [`os.Getuid() != 0`](https://github.com/containerd/nerdctl/blob/main/pkg/rootlessutil/parent.go#L33).
* `IsRootlessChild()` checks that `!IsRootlessParent()` as well as [whether the container is running within a Linux user namespace](https://github.com/containerd/nerdctl/blob/main/pkg/rootlessutil/child.go#L26).

### Current Windows behavior.

Currently, because `rootlessutil.IsRootlessParent()` checks that [`os.Getuid() != 0`](https://github.com/containerd/nerdctl/blob/main/pkg/rootlessutil/parent.go#L33), and [`os.Getuid()` always returns `-1` on Windows](https://pkg.go.dev/os#Getuid), all `IsRootlessParent()` calls will return `true` on Windows, and thus **all `rootlessutil.IsRootless()` calls will also return `true`**.

Fortunately, this does not really affect any important logic, since most of the special logic is only required if `IsRootlessChild()` is true.

### Discussion.

The discussion boils down to the following questions:
* should Windows Process Containers be considered rootless?
    * probably not, since even if the container is run as `ContainerUser` or some other non-admin user, it'll still be managed by HCS, which is a privileged service and thus breaks the "rootless" definition
* should Windows Hyper-V Containers be considered rootless?
    * probably so when considering that it has its own guest OS in the micro-VM it's running in, and so from the PoV of the containerd host, the container and its runtime "look rootless" despite being root-ful within the micro-VM
    * also, probably not, since a case could be made that HCS/Hyper-V themselves are privileged system services. so anything they run cannot be counted as "rootless"

Depending on the consensus, the actual changes to the codebase I'll be making will be very minimal, but it's important that a consensus is reached so things can be planned better moving forward.

### Misc.

FWIW, I've also **tried hardcoding `false` in all `IsRootless*()` calls on Windows and things seemed to mostly work**, with the only notable pain point I could find were some syslog-related tests which were ignored if `IsRootless*() == true`. (the tests set up their own syslog server, and everything was technically fine, just [that the tests run `echo`s](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_log_driver_syslog_test.go#L97) which will contain added double quotes `""` around the message due to the Windows `testutil.CommonImage` having `cmd.exe` as the default entrypoint)

I think it's also be worth noting that `rootlessutil.IsRootless*()` only checks whether the **current** `nerdctl` parent process is rootless, and creates rootless containers based on that.
On Windows, however, the question of rootlessness might depend on the `--isolation` flag, but should not theoretically depend on the `nerdctl` process at all (only precondition will be the user running `nerdctl` having access to containerd's named pipe to make calls to it)

_No response_

Expected `nerdctl` to have explicit knowledge/logic on whether it's running in rootless mode on Windows or not. (currently, `rootlessutil.IsRootlessParent()` is always `true` due to a minor oversight with `os.Getuid()`, see above)

Latest `main`.

None

### Host information

Will happen on any Windows host.

aznashwan:
Make `pkg/rootlesskit` always deny rottlessness and error out if
attempted to be called on non-Linux platforms.

Considering RootlessKit only works on Linux, and no supported
containerd runtime on Windows can be considered rootless,
any calls to the rootlessutil package should error out
on any non-Linux platform.

Fixes #2115.


pkg/rootlessutil/child.go
AkihiroSuda, line 1: Maybe we can just move `pkg/rootlessutil/*.go` to `pkg/rootlessutil/*_linux.go` and add `pkg/rootlessutil/*_others.go` for non-Linux

pkg/ocihook/rootless_other.go
AkihiroSuda, line 1: The files should be `_linux.go`(Linux)  and `_others.go` (Windows, FreeBSD)

aznashwan: I realize RootlessKit only works on Linux but as it stands the code in `rootlessutils` is shared on FreeBSD as well, and I wanted to avoid accidentally removing any logic, which is why I initially chose the `_windows.go` vs. `_others.go` naming convention.

We may need to have at least a FreeBSD implementation of `IsRootlessParent()` so that nerdctl can properly realize it's being executed as a non-root user and keep correctly failing on FreeBSD when anyone tries.

I'll switch it to your original suggestion (and probably spam pushes to re-run the FreeBSD CI) soon.

jsturtevant:
thanks @aznashwan, lgtm for windows pending comments on file naming

AkihiroSuda (approved):
Thanks

PR #2128, opened 2023-03-23
feat: support '--detach-keys' for 'nerdctl (run|start)'

davidhsingyuchen:
## Control Flow

1. Detect if the detach keys are present in stdin.
1. After detecting the detach keys, cancel the current I/O operations. This means that the underlying I/O FIFOs should be left intact, and containerd should keep writing to/reading from those FIFOs, but nerdctl should stop writing to/reading from them. It seems that `cio.IO.Cancel()` should be used for this, but there's an issue: containerd/containerd#8326.
1. After starting the task, we need to call `IO.Wait`, and after it returns, we need to see if the container exits or it's just the user detaching from the container by checking the state of the container.

## Notes

- It may be better to support `nerdctl exec --detach-keys` in a follow-up PR so that we can minimize the changes needed in this PR as the control flow of it is a bit different from that of `run|start`.

## TODOs

Before this PR can be merged, I need to:

- [x] Revert the `temporarily use the containerd feature branch` commit and use the released containerd version.

### DIFF
--- a/cmd/nerdctl/container_start_linux_test.go
+++ b/cmd/nerdctl/container_start_linux_test.go
@@ -1,0 +1,69 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+	"gotest.tools/v3/assert"
+)
+
+func TestStartDetachKeys(t *testing.T) {
+	t.Parallel()
+
+	if testutil.GetTarget() == testutil.Docker {
+		t.Skip("When detaching from a container, for a session started with 'docker attach'" +
+			", it prints 'read escape sequence', but for one started with 'docker (run|start)', it prints nothing." +
+			" However, the flag is called '--detach-keys' in all cases" +
+			", so nerdctl prints 'read detach keys' for all cases" +
+			", and that's why this test is skipped for Docker.")
+	}
+
+	base := testutil.NewBase(t)
+	containerName := testutil.Identifier(t)
+
+	defer base.Cmd("container", "rm", "-f", containerName).AssertOK()
+	opts := []func(*testutil.Cmd){
+		// If NewDelayOnceReader is not used,
+		// the container state will be Created instead of Exited.
+		// Maybe `unbuffer` exits too early in that case?
+		testutil.WithStdin(testutil.NewDelayOnceReader(strings.NewReader("exit\n"))),
+	}
+	// unbuffer(1) emulates tty, which is required by `nerdctl run -t`.
+	// unbuffer(1) can be installed with `apt-get install expect`.
+	//
+	// "-p" is needed because we need unbuffer to read from stdin, and from [1]:
+	// "Normally, unbuffer does not read from stdin. This simplifies use of unbuffer in some situations.
+	//  To use unbuffer in a pipeline, use the -p flag."
+	//
+	// [1] https://linux.die.net/man/1/unbuffer
+	base.CmdWithHelper([]string{"unbuffer", "-p"}, "run", "-it", "--name", containerName, testutil.CommonImage).
+		CmdOption(opts...).AssertOK()
+	container := base.InspectContainer(containerName)
+	assert.Equal(base.T, container.State.Running, false)
+
+	opts = []func(*testutil.Cmd){
+		testutil.WithStdin(testutil.NewDelayOnceReader(bytes.NewReader([]byte{1, 2}))), // https://www.physics.udel.edu/~watson/scen103/ascii.html
+	}
+	base.CmdWithHelper([]string{"unbuffer", "-p"}, "start", "-a", "--detach-keys=ctrl-a,ctrl-b", containerName).
+		CmdOption(opts...).AssertOutContains("read detach keys")
+	container = base.InspectContainer(containerName)
+	assert.Equal(base.T, container.State.Running, true)
+}
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/moby/sys/mount v0.3.3
 	github.com/moby/sys/signal v0.7.0
+	github.com/moby/term v0.5.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.0-rc3
 	github.com/opencontainers/runtime-spec v1.1.0-rc.2
@@ -63,6 +64,7 @@ require (
 require (
 	github.com/AdaLogics/go-fuzz-headers v0.0.0-20230106234847-43070de90fa1 // indirect
 	github.com/AdamKorcz/go-118-fuzz-build v0.0.0-20221215162035-5330a85ea652 // indirect
+	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/cilium/ebpf v0.9.1 // indirect
 	github.com/containerd/cgroups v1.1.0 // indirect
 	github.com/containerd/fifo v1.1.0 // indirect
--- a/pkg/consoleutil/detach.go
+++ b/pkg/consoleutil/detach.go
@@ -1,0 +1,63 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package consoleutil
+
+import (
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/moby/term"
+	"github.com/sirupsen/logrus"
+)
+
+const DefaultDetachKeys = "ctrl-p,ctrl-q"
+
+type detachableStdin struct {
+	stdin  io.Reader
+	closer func()
+}
+
+// NewDetachableStdin returns an io.Reader that
+// uses a TTY proxy reader to read from stdin and detect when the specified detach keys are read,
+// in which case closer will be called.
+func NewDetachableStdin(stdin io.Reader, keys string, closer func()) (io.Reader, error) {
+	if len(keys) == 0 {
+		keys = DefaultDetachKeys
+	}
+	b, err := term.ToBytes(keys)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert the detach keys to bytes: %w", err)
+	}
+	return &detachableStdin{
+		stdin:  term.NewEscapeProxy(stdin, b),
+		closer: closer,
+	}, nil
+}
+
+func (ds *detachableStdin) Read(p []byte) (int, error) {
+	n, err := ds.stdin.Read(p)
+	var eerr term.EscapeError
+	if errors.As(err, &eerr) {
+		logrus.Info("read detach keys")
+		if ds.closer != nil {
+			ds.closer()
+		}
+		return n, io.EOF
+	}
+	return n, err
+}
--- a/pkg/taskutil/taskutil.go
+++ b/pkg/taskutil/taskutil.go
@@ -30,6 +30,7 @@ import (
 	"github.com/containerd/console"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
+	"github.com/containerd/nerdctl/pkg/consoleutil"
 	"github.com/containerd/nerdctl/pkg/infoutil"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/term"
@@ -37,12 +38,34 @@ import (
 
 // NewTask is from https://github.com/containerd/containerd/blob/v1.4.3/cmd/ctr/commands/tasks/tasks_unix.go#L70-L108
 func NewTask(ctx context.Context, client *containerd.Client, container containerd.Container,
-	flagA, flagI, flagT, flagD bool, con console.Console, logURI string) (containerd.Task, error) {
+	flagA, flagI, flagT, flagD bool, con console.Console, logURI, detachKeys string, detachC chan<- struct{}) (containerd.Task, error) {
+	var t containerd.Task
+	closer := func() {
+		if detachC != nil {
+			detachC <- struct{}{}
+		}
+		// t will be set by container.NewTask at the end of this function.
+		//
+		// We cannot use container.Task(ctx, cio.Load) to get the IO here
+		// because the `cancel` field of the returned `*cio` is nil. [1]
+		//
+		// [1] https://github.com/containerd/containerd/blob/8f756bc8c26465bd93e78d9cd42082b66f276e10/cio/io.go#L358-L359
+		io := t.IO()
+		if io == nil {
+			logrus.Errorf("got a nil io")
+			return
+		}
+		io.Cancel()
+	}
 	var ioCreator cio.Creator
 	if flagA {
 		logrus.Debug("attaching output instead of using the log-uri")
 		if flagT {
-			ioCreator = cio.NewCreator(cio.WithStreams(con, con, nil), cio.WithTerminal)
+			in, err := consoleutil.NewDetachableStdin(con, detachKeys, closer)
+			if err != nil {
+				return nil, err
+			}
+			ioCreator = cio.NewCreator(cio.WithStreams(in, con, nil), cio.WithTerminal)
 		} else {
 			ioCreator = cio.NewCreator(cio.WithStdio)
 		}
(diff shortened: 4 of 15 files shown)


davidhsingyuchen:
@AkihiroSuda PTAL, thanks!

pkg/taskutil/taskutil.go
```diff
@@ -38,11 +39,23 @@ import (
 // NewTask is from https://github.com/containerd/containerd/blob/v1.4.3/cmd/ctr/commands/tasks/tasks_unix.go#L70-L108
 func NewTask(ctx context.Context, client *containerd.Client, container containerd.Container,
 	flagA, flagI, flagT, flagD bool, con console.Console, logURI string) (containerd.Task, error) {
+	closer := func() {
+		if t, err := container.Task(ctx, nil); err != nil {
```
fahedouch, line 43: you need `t, err := container.Task(ctx, cio.Load)` to load task with `IO`

pkg/consoleutil/detach.go
```diff
@@ -0,0 +1,72 @@
...
+var defaultDetachKeys = "ctrl-p,ctrl-q"
+
+type detachableStdin struct {
+	stdin  io.Reader
+	closer func()
+	mu     sync.Mutex
+	closed bool
+}
+
+// NewDetachableStdin returns a new TTY proxy reader
```
fahedouch, line 38: NewDetachableStdin returns a new `detachableStdin` which uses a TTY proxy reader to read specified detach keys from stdin, in which case closer will be called

fahedouch:
> After detecting the detach keys, cancel the current I/O operations. This means that the underlying I/O FIFOs should be left intact, and containerd should keep writing to/reading from those FIFOs, but nerdctl should stop writing to/reading from them. It seems that cio.IO.Cancel() should work, but I'm not 100% sure due to the nil IO() thing below.

 `Cancel` will aborts all current io operations ( from nerdctl and containerd)

davidhsingyuchen:
> > After detecting the detach keys, cancel the current I/O operations. This means that the underlying I/O FIFOs should be left intact, and containerd should keep writing to/reading from those FIFOs, but nerdctl should stop writing to/reading from them. It seems that cio.IO.Cancel() should work, but I'm not 100% sure due to the nil IO() thing below.
> 
> `Cancel` will aborts all current io operations ( from nerdctl and containerd). but nerdctl is fully controlling the io of the task (nerdctl is creating task with specefic io) , so I am wondering why not making `task.CloseIO()` ( this will abort io and clean fifos) , I also think that we no longer use the exsiting fifos after detach

After we detach from a container, we should be able to reattach to it again. Docker:

```sh
➜  ~ docker attach test
/ #
/ #
/ # read escape sequence
➜  ~ docker attach test
/ #
/ #
/ # read escape sequence
```

pkg/consoleutil/detach.go
davidhsingyuchen: Hmm, I'm not sure if we want to mention the underlying unexported struct here, maybe the users of this function shouldn't need to know that implementation detail?

pkg/taskutil/taskutil.go
davidhsingyuchen: Thanks a ton! Only realized how obvious it is just now 😅

pkg/consoleutil/detach.go
```diff
@@ -0,0 +1,72 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"sync"
+	"syscall"
+
+	"github.com/moby/term"
+)
+
+var defaultDetachKeys = "ctrl-p,ctrl-q"
```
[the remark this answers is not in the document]
davidhsingyuchen: replaced with `const`

go.mod
```diff
@@ -58,6 +58,8 @@ require (
 	gotest.tools/v3 v3.4.0
 )
 
+require github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
```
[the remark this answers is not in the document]
davidhsingyuchen: fixed

pkg/consoleutil/detach.go
fahedouch: LGTM, but just one thing was confusing me is : `NewDetachableStdin returns a new TTY proxy reader`. In fact the `NewDetachableStdin` is not returning a TTY proxy reader but it is wrapping a a TTY proxy reader

cmd/nerdctl/container_run_test.go
```diff
@@ -467,3 +468,21 @@ func TestRunWithTtyAndDetached(t *testing.T) {
 	withTtyContainer := base.InspectContainer(withTtyContainerName)
 	assert.Equal(base.T, 0, withTtyContainer.State.ExitCode)
 }
+
+func TestRunWithDetachKeys(t *testing.T) {
+	t.Parallel()
+
+	base := testutil.NewBase(t)
+	if testutil.GetTarget() == testutil.Nerdctl {
+		testutil.RequireDaemonVersion(base, ">= 1.6.0-0")
```
fahedouch, line 477: why this check ?

cmd/nerdctl/container_run_test.go
```diff
@@ -467,3 +468,21 @@ func TestRunWithTtyAndDetached(t *testing.T) {
...
+	}
+
+	containerName := testutil.Identifier(t)
+	opts := []func(*testutil.Cmd){
+		testutil.WithStdin(bytes.NewReader([]byte{17, 16})), // https://www.physics.udel.edu/~watson/scen103/ascii.html
+	}
+	base.Cmd("run", "-it", "--detach-keys=ctrl-q,ctrl-p", "--name", containerName, testutil.CommonImage).CmdOption(opts...).AssertOK()
+	defer base.Cmd("container", "rm", "-f", containerName).AssertOK()
```
fahedouch, line 485: please put defer before this [line](https://github.com/containerd/nerdctl/pull/2128/files#diff-2e856777b9ddfcd376cce19411cd5372a666ea2c5a6c3f4adeda04c983a14f5eR484)

cmd/nerdctl/container_run_test.go
```diff
@@ -467,3 +468,21 @@ func TestRunWithTtyAndDetached(t *testing.T) {
...
+	base := testutil.NewBase(t)
+	if testutil.GetTarget() == testutil.Nerdctl {
+		testutil.RequireDaemonVersion(base, ">= 1.6.0-0")
+	}
+
+	containerName := testutil.Identifier(t)
+	opts := []func(*testutil.Cmd){
+		testutil.WithStdin(bytes.NewReader([]byte{17, 16})), // https://www.physics.udel.edu/~watson/scen103/ascii.html
+	}
+	base.Cmd("run", "-it", "--detach-keys=ctrl-q,ctrl-p", "--name", containerName, testutil.CommonImage).CmdOption(opts...).AssertOK()
```
fahedouch, line 484: `testutil.CommonImage`  exits directly we need a long running image to ensure that `detach-keys` takes effects, such as `testutil.NginxAlpineImage`

cmd/nerdctl/container_run_test.go
davidhsingyuchen: Removed. I copied from [`TestRunStdin`](https://github.com/containerd/nerdctl/blob/225a70bdc3b93cdb00efac7db1ceb50c098a8a16/cmd/nerdctl/container_run_test.go#L223-L225), but it turned out this check is only needed when `-t` is not specified with `-i` ([source](https://github.com/containerd/nerdctl/blob/225a70bdc3b93cdb00efac7db1ceb50c098a8a16/pkg/taskutil/taskutil.go#L97-L99)). Thanks for pointing it out!

cmd/nerdctl/container_run_test.go
davidhsingyuchen: fixed

davidhsingyuchen:
@fahedouch Thanks a lot for the quick review! In the new revision, besides addressing the comments, I also made it print `read detach keys` when detaching from the container so that it's clearer to the users what's happening, and it's also aligned to UX of docker. Also updated e2e tests accordingly.

davidhsingyuchen:
If I `replace` `containerd/containerd` in nerdctl's `go.mod` with the version from https://github.com/containerd/containerd/pull/8334, it seems to be working.

```sh
davidhyc@lima-dev:/Users/davidhyc/dev/containerd/nerdctl$ ./_output/nerdctl run -it --name detach busybox
/ #
/ # INFO[0001] read detach keys
davidhyc@lima-dev:/Users/davidhyc/dev/containerd/nerdctl$ nerdctl ps -a
CONTAINER ID    IMAGE                               COMMAND    CREATED          STATUS    PORTS    NAMES
fd3b77abf607    docker.io/library/busybox:latest    "sh"       4 seconds ago    Up                 detach
```

pkg/consoleutil/detach.go
```diff
@@ -0,0 +1,75 @@
...
+		// which means that the loop becomes an infinite loop,
+		// and it will only "exits" when IO.Wait() of the current task returns,
+		// as the whole nerdctl process will exit after that.
+		//
+		// The implication is that "read detach keys" will be printed once for every iteration of the loop,
+		// while we only want it to be printed a single time,
+		// so we just return a non-nil error here to make things simpler.
+		//
+		// [1] https://github.com/golang/go/blob/c0fd7f79fe445ad49e11bf42c8c785cb71b3bedf/src/io/io.go#L426-L452
+		// [2] https://github.com/containerd/containerd/blob/8f756bc8c26465bd93e78d9cd42082b66f276e10/cio/io_unix.go#L68
```
fahedouch, line 71: is it possible to keep referencing the issue ticket opened in containerd repo. Thanks

pkg/taskutil/taskutil.go
```diff
@@ -30,19 +30,38 @@ import (
...
 // NewTask is from https://github.com/containerd/containerd/blob/v1.4.3/cmd/ctr/commands/tasks/tasks_unix.go#L70-L108
 func NewTask(ctx context.Context, client *containerd.Client, container containerd.Container,
-	flagA, flagI, flagT, flagD bool, con console.Console, logURI string) (containerd.Task, error) {
+	flagA, flagI, flagT, flagD bool, con console.Console, logURI, detachKeys string) (containerd.Task, error) {
+	var t containerd.Task
+	closer := func() {
+		// t will be set by container.NewTask at the end of this function.
+		//
+		// We cannot use container.Task(ctx, cio.Load) to get the IO here
+		// because the `cancel` field of the returned `*cio` is nil. [1]
```
fahedouch, line 47: it says that the io to be load for a task for `deletion`. So I think they assume the the task io is already canceled. But It is not intuitive at all we may load and then cancel  :/ [see](https://github.com/containerd/containerd/blob/8f756bc8c26465bd93e78d9cd42082b66f276e10/cio/io.go#L354-L355)

fahedouch:
the integration is failing too , it is possible that it is related to the cancellation of io or `t.Paralle()`  is impacting the test flow..

pkg/consoleutil/detach.go
davidhsingyuchen: Could you share which issue you're referring to? This is not really an "issue" though, so I'm a bit confused.

As a side note, after a second thought, maybe we can return an `io.EOF` instead of the original `err` so that the semantics is more accurate to code that doesn't know `term.EscapeError`, wdyt?

pkg/taskutil/taskutil.go
davidhsingyuchen: That's true. Are you suggesting to open an issue at containerd side to make `cio.Load` also usable for cancel?

davidhsingyuchen:
> the integration is failing too , it is possible that it is related to the cancellation of io or `t.Paralle()` is impacting the test flow..

Thanks for the review. The tests were failing because `io.Wait()` returns also when the container exits, so the `ExitCoder` was not used in the previous revision. Updated that part of code accordingly.

Besides that, in the new revision:

- Use `unbuffer -p` for the detach e2e tests.
- Added an e2e test for `nerdctl start`.
- Added a temp commit (`temporarily use the containerd feature branch`) to make the e2e tests pass on CI, which can hopefully facilitate reviewing this PR.

pkg/taskutil/taskutil.go
fahedouch: Yes. I can take care of opening the issue if LGTY

pkg/consoleutil/detach.go
fahedouch: I am pointing to the issue related to this [PR](https://github.com/containerd/containerd/pull/8334).

> As a side note, after a second thought, maybe we can return an io.EOF instead of the original err so that the semantics is more accurate to code that doesn't know term.EscapeError, wdyt?

totaly agree. This is what we need. So even if containerd closed fifos after mergeing this [PR](https://github.com/containerd/containerd/pull/8334). We need to return `io.EOF` to indicated `io` end of life in the  nerdctl side

go.mod
```diff
@@ -2,6 +2,11 @@ module github.com/containerd/nerdctl
 
 go 1.19
 
+// 82ec62b86582 refers to the feature branch of https://github.com/containerd/containerd/pull/8334.
+// It's used to make the e2e tests pass.
+// TODO: Before merging this PR, we need to remove this `replace` directive and use the released containerd version.
+replace github.com/containerd/containerd => github.com/davidhsingyuchen/containerd v0.0.0-20230331013921-82ec62b86582
```
fahedouch, line 8: Why ?

pkg/taskutil/taskutil.go
```diff
@@ -30,19 +30,42 @@ import (
...
 	"golang.org/x/term"
 )
 
 // NewTask is from https://github.com/containerd/containerd/blob/v1.4.3/cmd/ctr/commands/tasks/tasks_unix.go#L70-L108
 func NewTask(ctx context.Context, client *containerd.Client, container containerd.Container,
-	flagA, flagI, flagT, flagD bool, con console.Console, logURI string) (containerd.Task, error) {
+	flagA, flagI, flagT, flagD bool, con console.Console, logURI, detachKeys string, detachC chan<- struct{}) (containerd.Task, error) {
+	var t containerd.Task
+	closer := func() {
+		if detachC != nil {
```
fahedouch, line 44: always `True`

cmd/nerdctl/container_start_linux_test.go
```diff
@@ -0,0 +1,69 @@
...
+			", and that's why this test is skipped for Docker.")
+	}
+
+	base := testutil.NewBase(t)
+	containerName := testutil.Identifier(t)
+
+	defer base.Cmd("container", "rm", "-f", containerName).AssertOK()
+	opts := []func(*testutil.Cmd){
+		// If NewDelayOnceReader is not used,
+		// the container state will be Created instead of Exited.
```
fahedouch, line 45: > the container state will be Created instead of Exited.
you mean `Running` rather then `Exited` ?

pkg/taskutil/taskutil.go
davidhsingyuchen: Sure, please go ahead, thanks! I can then link the issue as a code comment here.

cmd/nerdctl/container_start_linux_test.go
davidhsingyuchen: I meant `Exited`. It should be `Exited` because we start the container with `sh`, and then we pass `exit\n` to it.

go.mod
davidhsingyuchen: Without this temp change, https://github.com/containerd/containerd/issues/8326 makes `IO.Wait()` hang forever even if `IO.Cancel()` has been invoked.

pkg/consoleutil/detach.go
davidhsingyuchen: > I am pointing to the issue related to this https://github.com/containerd/containerd/pull/8334.

I see, sorry for the confusion. That issue makes `IO.Wait()` hang forever even if `IO.Cancel()` has been invoked, while the comment here is referring to the fact that even after containerd/containerd#8334 is merged, there will still be some time difference between:

1. `IO.Cancel()` is invoked because the detach keys are read.
2. `IO.Wait()` returns and then the whole nerdctl process exits so that we would stop printing `read detach keys`.

> totaly agree. This is what we need. So even if containerd closed fifos after mergeing this https://github.com/containerd/containerd/pull/8334. We need to return io.EOF to indicated io end of life in the nerdctl side

Done. Also removed the comments as I feel that the semantics of `io.EOF` does fit into this context and makes the comments a bit unnecessary.

pkg/consoleutil/detach.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const DefaultDetachKeys = "ctrl-p,ctrl-q"
+
+type detachableStdin struct {
+	stdin  io.Reader
+	closer func()
+}
+
+// NewDetachableStdin returns a io.Reader that
```
fahedouch, line 35: typo `a io.Reader` => `an io.Reader`

fahedouch:
LGTM, thanks

pkg/consoleutil/detach.go
davidhsingyuchen: fixed

fahedouch:
@davidhsingyuchen 

> It seems that cio.IO.Cancel() should be used for this, but there's an issue: https://github.com/containerd/containerd/issues/8326.

please [see](https://github.com/containerd/containerd/issues/8326#issuecomment-1537480670)

AkihiroSuda:
Needs rebase

davidhsingyuchen:
@AkihiroSuda Thank you for the reminder. Could you help take a look at https://github.com/containerd/containerd/issues/8326 and/or https://github.com/containerd/containerd/pull/8334? I'll rebase the feature branch after the blocker is resolved, thanks!

davidhsingyuchen:
@fahedouch PTAL, thanks! The failed jobs seem to be flaky as all the jobs pass at least once:

- https://github.com/containerd/nerdctl/actions/runs/5194348122
- https://github.com/containerd/nerdctl/actions/runs/5194508608

fahedouch (approved):

go.mod
```diff
@@ -57,12 +58,14 @@ require (
 	golang.org/x/term v0.8.0
 	golang.org/x/text v0.9.0
 	gopkg.in/yaml.v3 v3.0.1
+	gotest.tools v2.2.0+incompatible
 	gotest.tools/v3 v3.4.0
```
AkihiroSuda, line 62: Can we avoid mixing up v2 and v3?

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/moby/sys/mount v0.3.3
 	github.com/moby/sys/signal v0.7.0
+	github.com/moby/term v0.0.0-20221205130635-1aeaba878587
```
AkihiroSuda, line 40: Can we use https://github.com/moby/term/releases/tag/v0.5.0

davidhsingyuchen:
@AkihiroSuda Thanks for the reminder. Both are fixed.

AkihiroSuda (approved):
Thanks

PR #2131, opened 2023-03-25
Refactor image sign and verify logic

djdongjin:
This PR extracts image sign/verify options into their own option strcut and move the switch logic into a signutil helper which reduces duplications and make further modification simplier. This also makes sign/verify related logic more centralized in a single package `signutil`.

Some examples:

1. To move one tool out of experimental:

before: change logic in push, pull, and compose.
after: only change logic in `signutil.go`.

2. To add a new verify flag (suppose it's used in `n` cmds):

before: add the flag to (1) `n` cobra cmds, (2) `n` cmd process funcs, (3) `n` cmd option structs.
after: add the flag to (1) `n` cobra cmds, (2). only 1 cmd process logic (the newly added `processImageVerifyOptions`).

3. To add a new tool (not consider flags which is similar to 2):

before: update the switch logic in all places (pull, push, compose).
after: update the switch logic only in `signutil.go`.


AkihiroSuda (approved):
Thanks

PR #2145, opened 2023-03-31
Add Windows tests for the container commands

dardelean:
Added more unit tests for 'nerdctl container' command on Windows. Also added the 'HyperVContainer' helper to check with hcsshim if the container is a hyperv container.

Restart and stop tests are skipped since they fail atm on WS 2019.

### DIFF
--- a/cmd/nerdctl/container_create_windows_test.go
+++ b/cmd/nerdctl/container_create_windows_test.go
@@ -1,0 +1,57 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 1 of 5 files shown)


cmd/nerdctl/container_create_windows_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/containerd/nerdctl/pkg/testutil"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+func TestCreateProcessContainer(t *testing.T) {
```
manugupt1, line 27: Why do we need a separate set of tests? Can we not reuse the same tests and mark windows incompatible tests and windows only tests in testutil like docker

aznashwan: The [`--isolation` flag](https://github.com/containerd/nerdctl/pull/2145/files/a176d8cca3cc9cfae0f10f3abc5a872dc75c6ec6#diff-aa8815c52e48ac75edf93e1ba0ef7a71e3d16bec6b485505a9d614b0ae2863c2R60) mandates the need for at least some dedicated tests on Windows.

I agree that `TestCreateProcessContainer` should explicitly set `--isolation process` though, so maybe the actual testing logic in this and `TestCreateHyperVContainer` could be abstracted into a testing helper which takes as an argument the isolation type it should use? @dardelean

manugupt1: Can you see `testutil.DockerIncompatible(t)` and implement something as `testutil.DockerHyperVCompatible(t)` and run those specific tests when this flag is set. You would probably need testutl_linux.go that basically always skips the test running on a linux machine.

dardelean:
Docker test seems to be flaky, and it's a linux test it's failing on. Can this be merged?

AkihiroSuda (approved):
Thanks

PR #2157, opened 2023-04-05
add image to compose output

gerbercloud:


cmd/nerdctl/compose_ps_linux_test.go
```diff
@@ -95,6 +95,9 @@ volumes:
 	// check all services are up (can be marshalled and unmarshalled)
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "--format", "json").
 		AssertOutWithFunc(assertHandler("all", 2, `"Service":"wordpress"`, `"Service":"db"`))
+	// check Image field
+	base.ComposeCmd("-f", comp.YAMLFullPath(), "ps", "--format", "json").
+		AssertOutWithFunc(assertHandler("all", 2, fmt.Sprintf(`"Image":"%s"`, testutil.WordpressImage), fmt.Sprintf(`"Image":"%s"`, testutil.MariaDBImage)))
```
djdongjin, line 100: This can be merged into line 97 right? Since it accepts variable-length fields to be checked.

gerbercloud: Correct, changed it.

djdongjin (approved):
LGTM, thanks. 

Can you squash the commits? You can merged the 4th one to the 3rd one since they're doing the same task (add tests). Then having these 3 commits probably should be fine since they're doing different tasks, unless suggested by other reviewers :)

gerbercloud:
> Can you squash the commits?

Done.

gerbercloud:
> > > Can you squash the commits?
> > 
> > 
> > Done.
> 
> Doesn't seem squashed

I squashed the last two commits so it's back to three commits for three different tasks. Should I squash everything into one commit or is it okay like this?

fahedouch (approved):
Thanks

fahedouch:
> > > > Can you squash the commits?
> > > 
> > > 
> > > Done.
> > 
> > 
> > Doesn't seem squashed
> 
> I squashed the last two commits so it's back to three commits for three different tasks. Should I squash everything into one commit or is it okay like this?


each commit addresses a different task so that's fine. Thanks

cmd/nerdctl/compose_images_linux_test.go
```diff
@@ -75,3 +76,80 @@ volumes:
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "images").AssertOutContains(dbImageName)
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "images").AssertOutContains(wordpressImageName)
 }
+
+func TestComposeImagesJson(t *testing.T) {
+	// `--format` is only supported in compose v2.
+	// Currently, CI is using compose v1.
```
fahedouch, line 82: please update as above:
`--format is only supported in docker compose v2.`
`Currently, CI is using docker compose v1.`  

because we also support compose into nerdctl too

would you please update these [lines](https://github.com/containerd/nerdctl/pull/2157/files#diff-cf5e5f2df14711c5dd73f6508b0b5ce392bd61c21d06a05dd2f4b953d88f0d31R30) too

gerbercloud: Thanks. I updated the comments.

PR #2160, opened 2023-04-06
Remove colons in VOLUME mappings on Windows.

aznashwan:
Attempting to run Windows images with VOLUME declarations which are absolute paths would lead to `securejoin.SecureJoin()` concatenating two absolute paths when determining the host mount dir and then failing to `os.Lstat()` the resulting path, as it contained an illegal second `:`.


aznashwan:
This should alleviate (but **not** fix) https://github.com/containerd/nerdctl/issues/759, as it enables nerdctl to at least [reach the point where it sends the OCI spec over](https://github.com/containerd/nerdctl/pull/924#discussion_r871002561).

cmd/nerdctl/container_run_mount_windows.go
```diff
@@ -0,0 +1,40 @@
...
+	"strings"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+)
+
+// Joins the given host path with the given guest path while guaranteeing the provided
+// path is scoped within the host path. (i.e. joining with `../..` or symlinks cannot
+// lead to a higher level path than the provided hostPath, and are simply ignored)
+// Any colons in the guest path will be removed completely in order to have a valid
+// host path while maining the guest path's drive letter.
```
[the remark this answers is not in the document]
aznashwan: I am honestly unsure how Docker does it, at it looks like `VOLUME`s are not working properly for WCOW, as the mounted volume in the container seems empty.

```powershell
# Docker CE on Windows 10 in WCOW container mode.
# Image has a `VOLUME C:\test_dir`:
# https://github.com/containerd/containerd/blob/main/integration/images/volume-copy-up/Dockerfile_windows#L29-L37
PS C:\Users\Nashu> docker run  -ti ghcr.io/containerd/volume-copy-up:2.1 cmd /c "dir C:\test_dir"
 Volume in drive C has no label. 
 Volume Serial Number is 7AFF-76D2            
                                              
 Directory of C:\test_dir                     
                                              
# NOTE: should have contained a `test_file` with some `test_contents` in it too.
04/07/2023  06:05 AM    <DIR>          .      
04/07/2023  06:05 AM    <DIR>          ..     
               0 File(s)              0 bytes 
               2 Dir(s)  142,581,915,648 bytes free 
```

What I can say though is that this [`tempDir` is created here](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_mount.go#L155) and will have [image layers mounted under their individual IDs in it](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_mount.go#L185).

In this sense, there should be no [collisions when we create/mount the `VOLUME`](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_mount.go#L272-L275) based on its mountpoint within that `tempDir` with whatever might be in the image layers.

@gabriel-samfira did, rightfully point out the possibility of collisions when having things like `VOLUME ["C:/test_dir", "/C/test_dir"]`, which are technically valid and are worth considering.

At this stage, I'm simply tempted to `securejoin.SecureJoin(tempDir, anonVolName)` as [`anonVolName` is an ID generated right above the join](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_mount.go#L263), what do y'all think?

cmd/nerdctl/container_run_mount.go
```diff
@@ -228,7 +227,7 @@ func generateMountOpts(ctx context.Context, cmd *cobra.Command, client *containe
 			ociMounts[i] = x.Mount
 			mounted[filepath.Clean(x.Mount.Destination)] = struct{}{}
 
-			target, err := securejoin.SecureJoin(tempDir, x.Mount.Destination)
+			target, err := joinHostPath(tempDir, x.Mount.Destination)
```
gabriel-samfira, line 230: This issue is similar to: https://github.com/containerd/containerd/pull/8362

For non ```C:\``` volumes, we need to skip trying to copy existing contents. For ```C:``` volumes, ```containerd``` already strips the drive letter here:

https://github.com/containerd/containerd/blob/474a257b16f0aa79421038736126080696605e9d/pkg/cri/opts/container.go#L100-L101

I think it should be enough to mimic that.

gabriel-samfira: Note: a folder on the host for a potential non ```C:``` volume should already exist. We just need to skip the attempt to populate it.

aznashwan: > For non C:\ volumes, we need to skip trying to copy existing contents. For C: volumes, containerd strips the drive letter

Correct, but it only trims "C:/" specifically, and AFAICT [Docker should allow for any drive other than C too](https://docs.docker.com/engine/reference/builder/#volume).

Considering the names of these host dirs seems to be irrelevant, I suggest we just use `anonVolName`. (which is actually a random ID generated on the spot)

I'll also log the mapping for some debug-ability...

gabriel-samfira: I was thinking of something like this. This should mimic what is/will (be) happening in containerd.

```suggestion
			if len(x.Mount.Destination) >= 2 && string(x.Mount.Destination[1]) == ":" && !strings.EqualFold(string(x.Mount.Destination[0]), "c") {
				continue
			}
			destination := x.Mount.Destination
			if len(destination) >= 2 && strings.EqualFold((destination[:2]), "C:") {
				destination = destination[2:]
			}
			target, err := joinHostPath(tempDir, destination)
```

TBBle (changes requested):
This code won't work correctly, AFAICT.

The path being created (`target`) is the path to the _source_ of the copying in the temporary local mount of the source image's View, so it cannot be an arbitrary path. https://github.com/containerd/nerdctl/pull/2160/files#r1160795974 is the right path-transformation, but we can't `continue` as (unlike the containerd-cri equivalent to this code) we still need to do the things after the `copyExistingContents` call to mount the new volume into the system.

When the path to be copied is `C:\some\path`, then its location is `<tempDir>\some\path`; when the path to be copied is `D:`, then there's actually no possible valid path, as Windows container images cannot contain data for any drive but `C:`.

So perhaps something like what is here now, but the Windows split of the `joinHostPath` strips C: if present, and if given `D:`..`Z:`, returns an empty string which then skips _only_ the `copyExistingContents` step (_the only use of `joinHostPath`'s return value here_).

I also suspect this code won't work with current containerd main branch (and not with released containerd) as it has a [Windows codepath](https://github.com/containerd/nerdctl/blob/8f2940e74dd36fda70ebdebed48f94446f876951/cmd/nerdctl/container_run_mount.go#L180-L191) that mounts the copying source image at a subdirectory of `<temp_dir>`, so it's impossible to actually generate a correct path right now. This block could be removed, as the `else` path is correct (`mount.All(mounts, tempDir)`) for current containerd main (along with revendoring to current continerd main, I expect to pick up the new mount implementation, and possibly requiring )

cmd/nerdctl/container_run_mount_windows.go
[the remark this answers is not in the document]
TBBle: It's true that Docker does not implement populating the temporary volume on Windows, see https://github.com/moby/moby/blob/dd3b71d17c614f837c4bba18baed9fa2cb31f1a4/daemon/create_windows.go#L42-L70. However, containerd-cri does; that implementatoin is currently faulty for non-C: drives, see https://github.com/containerd/containerd/issues/8171 and proposed fix at https://github.com/containerd/containerd/pull/8362.

> What I can say though is that this [`tempDir` is created here](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_mount.go#L155) and will have [image layers mounted under their individual IDs in it](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_mount.go#L185).

This code is separately faulty, I've gone into more detail in my review comment, but this loop produces mounts at paths that the rest of the code does not know about. It's also no longer necessary with recent improvements to the containerd code it's calling, as there'll only be one mount returned from `s.View` on Windows. See https://github.com/containerd/containerd/commit/474a257b16f0aa79421038736126080696605e9d#diff-2f84625f0f53cdff6461e071ffd95fba797e037a6195d2879ed9fd8bafdb7b5dL89-R97

aznashwan:
After some post-KubeCon delay, I have resorted to adding an explicit non-C-drive check.

I've also added a simple test with containerd's [volume-copy-up image](https://github.com/containerd/containerd/blob/main/integration/images/volume-copy-up/Dockerfile_windows) FWIW, which is currently expected to fail with this: https://github.com/containerd/nerdctl/pull/924#discussion_r871002561.

I'll need to investigate further where/if that "fstype" needs setting.

TBBle (changes requested):
Per https://github.com/containerd/nerdctl/pull/2160#pullrequestreview-1376686249, we're still failing when using non-C: volume mounts, i.e. `VOLUME ["D:"]` will now error out with an "unsupported" complaint, which appears to blame "Windows containers", but is just a parsing failure in nerdctl.

It also doesn't fix the issue in the `runtime.GOOS == "windows"` mounting code path which means that the content copying in both use-cases won't work anyway as the data we need is mounted not at `tempDir`, but a subdirectory of `tempDir`. I don't understand how the given test-case can pass because of this, which makes me suspect `VOLUME` is not actually being applied.

I'd also get rid of the new `logrus.Debugf` calls before merging, they seem overly-verbose.

It'd be nice to have tests of the `-v` code-path this code changes too, but perhaps #924 is covering that? (Per the above-linked comment, I'm not totally clear how `-v` doesn't clobber existing volume contents if pointed at a directory that contains data already; I suspect we lack a test for that flow on any platform.)

Issue #1854, opened 2023-01-14
support nerdctl export


saurik:
### What is the problem you're trying to solve

I am trying to use nerdctl as a drop-in replacement for docker, but it seems there is no equivalent yet for "docker export":

```
time="2023-01-14T12:58:38Z" level=fatal msg="unknown subcommand \"export\" for \"nerdctl\"\n\nDid you mean this?\n\tport\n"
```

### Describe the solution you'd like

Ideally, nerdctl should implement docker export. Barring that, it might be nice to have a place that lists all of the ways in which nerdctl is not actually Docker-compatible, as right now the project is misleading :(.

AkihiroSuda:
> Barring that, it might be nice to have a place that lists all of the ways in which nerdctl is not actually Docker-compatible, as right now the project is misleading :(.

Documented here https://github.com/containerd/nerdctl/tree/v1.1.0#unimplemented-docker-commands

saurik:
@AkihiroSuda Huh. I really thought I'd searched for "export" on that README and didn't see anything, but apparently I didn't. Sorry, and thanks.

turan18:
I can take a crack at this, if thats fine

djdongjin:
> I can take a crack at this, if thats fine

please go ahead :)

reneleonhardt:
@saurik Could you try the https://github.com/containerd/nerdctl/releases/tag/v2.1.4 export feature?

PR #2161, opened 2023-04-06
[WIP] Support `container export` 

turan18:
Adding support for container export.  Still WIP.

Resolves: #1854 

TODO:

* [ ] Map container uid/gid to host uid/gid (currently no uid/gid is set)
* [ ] Add unit test
* [ ] Make the tarballer more extensible to potentially support compression, chown options.. etc., although, `docker export` [currently does not utilize these options by default](https://github.com/moby/moby/blob/b52765c8c997bc9d009b73f63b1c790aeb7641da/daemon/export.go#L46).


pkg/tarutil/tarutil.go
```diff
@@ -55,3 +69,118 @@ func FindTarBinary() (string, bool, error) {
...
+
+type Tarballer struct {
+	Buffer    *bufio.Writer
+	TarWriter *tar.Writer
+	seenFiles map[uint64]string
+}
+
+// TODO: Add tar options for compression, whiteout files, chown ..etc
+
+func NewTarballer(writer io.Writer) *Tarballer {
```
AkihiroSuda, line 81: Probably you can just use https://pkg.go.dev/github.com/containerd/containerd/archive#WriteDiff .
(input `A` = container rootfs, input `B` = an empty dir)

Pymann:
I really would appreciate any updates...

Issue #2171, opened 2023-04-12
[v1.3.0 regression] Nerdctl push with Cosign not working with real registries


ningziwen:
### Description

Nerdctl push with Cosign seems not working with real registries (like Docker Hub or ECR) for me with the latest nerdctl, although the local registry may still work because it is part of nerdctl CI.

With the latest nerdctl:
```
sudo -E COSIGN_PASSWORD=123 ./_output/nerdctl push --sign cosign --cosign-key cosign.key charlienzw/sign:tagname --debug

...
DEBU[0000] running /home/ubuntu/go/bin/cosign [/home/ubuntu/go/bin/cosign sign --key cosign.key --yes docker.io/charlienzw/sign:tagname@sha256:4e83453afed1b4fa1a3500525091dbfca6ce1e66903fd4c01ff015dbcb1ba33e] 
INFO[0000] cosign: Error: signing [docker.io/charlienzw/sign:tagname@sha256:4e83453afed1b4fa1a3500525091dbfca6ce1e66903fd4c01ff015dbcb1ba33e]: accessing image: entity not found in registry 
INFO[0000] cosign: main.go:74: error during command execution: signing [docker.io/charlienzw/sign:tagname@sha256:4e83453afed1b4fa1a3500525091dbfca6ce1e66903fd4c01ff015dbcb1ba33e]: accessing image: entity not found in registry 
```

Reverted this [commit](https://github.com/containerd/nerdctl/pull/2136) locally and it works and the related debug log is 
```
DEBU[0000] running /home/ubuntu/go/bin/cosign [/home/ubuntu/go/bin/cosign sign --key cosign.key charlienzw/sign:tagname]
```

Both of the experiments are using Cosign 2.0.0.

Looks like the reason is that the latest nerdctl adds the digest at the end of image before passing to Cosign, and the digest added is from local, which is different with the one in Dockerhub/ECR (because container registry compresses the image which changes the digest displayed in registry, and local registry may not have this concern).

### Steps to reproduce the issue

Build nerdctl from main or use 1.3.0, then follow https://github.com/containerd/nerdctl/blob/main/docs/cosign.md to push a image with test key to Docker Hub, with Cosign 2.0 installed.

### Describe the results you received and expected

I expect the latest version of nerdctl also succeeds for pushing with Cosign.

### What version of nerdctl are you using?

1.3.0 or main branch can reproduce.

Reverting the [commit](https://github.com/containerd/nerdctl/pull/2136) can work.

[...]

PR #2172, opened 2023-04-13
revert the change of using digest instead of a tag while signing

ningziwen:
https://github.com/containerd/nerdctl/issues/2171


pkg/cmd/image/push.go
```diff
@@ -150,18 +149,7 @@ func Push(ctx context.Context, client *containerd.Client, rawRef string, options
...
-	}
-	refSpec, err := reference.Parse(ref)
-	if err != nil {
-		return err
-	}
-	signRef := fmt.Sprintf("%s@%s", refSpec.String(), img.Target.Digest.String())
-	if err = signutil.Sign(signRef,
-		options.GOptions.Experimental,
-		options.SignOptions); err != nil {
+	if err = signutil.Sign(rawRef, options.GOptions.Experimental, options.SignOptions); err != nil {
```
AkihiroSuda, line 152: Probably no need to revert this, but use pushRef instead of ref

ningziwen: That works for me.

AkihiroSuda (approved):
Thanks

fahedouch (approved):

PR #2176, opened 2023-04-13
Resolve Go cache warning in release workflow

austinvazquez:
actions/setup-go@v4.0.0 release enabled Go cache by default. This resulted in a warning in the release workflow for restore cache failure. Resolve the warning by having checkout before setup-go step.

A recent [release workflow run](https://github.com/containerd/nerdctl/actions/runs/4618402058) with Go cache restore failure warning.

Signed-off-by: Austin Vazquez <macedonv@amazon.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,7 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: ${{ env.GO_VERSION }}
+          cache-dependency-path: src/github.com/containerd/nerdctl
       - uses: containerd/project-checks@v1.1.0
         with:
           working-directory: src/github.com/containerd/nerdctl
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-22.04
     timeout-minutes: 40
     steps:
+    - uses: actions/checkout@v3.5.1
     - uses: actions/setup-go@v4
       with:
         go-version: 1.20.x
-    - uses: actions/checkout@v3.5.1
     - name: "Compile binaries"
       run: make artifacts
     - name: "SHA256SUMS"


austinvazquez:
Oh interesting, I found another occurrence during Project checks. See [here](https://github.com/containerd/nerdctl/actions/runs/4687450931).

Let me look into this more.

austinvazquez:
> Oh interesting, I found another occurrence during Project checks. See [here](https://github.com/containerd/nerdctl/actions/runs/4687450931).
> 
> Let me look into this more.

Resolved this by specifying cache-dependency-path for setup-go step. This is needed when cache is enabled and checkout is to a subdirectory.

.github/workflows/test.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: ${{ env.GO_VERSION }}
+          cache-dependency-path: src/github.com/containerd/nerdctl
```
fahedouch, line 26: IMHO this is useless if we remove this line https://github.com/containerd/nerdctl/pull/2176/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R21

fahedouch: I think we can add checkout depth : 1 for this job
```
with:
          fetch-depth: 1
```

fahedouch (commented):
[](https://github.com/containerd/nerdctl/pull/2176/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R21)

.github/workflows/test.yml
fahedouch: Sorry looks like one commit is fetched by default 

 > Only a single commit is fetched by default, for the ref/SHA that triggered the workflow

From  https://github.com/actions/checkout#checkout-v3

fahedouch (approved):
Thanks

AkihiroSuda (approved):
Thanks

PR #2203, opened 2023-04-26
bump nydus snapshotter v0.8.0

imeoer:
The PR includes 2 commits:

**vendor: bump nydus-snapshotter v0.8.0**

**convert: enable oci option for nydus conversion type by default**

The newer nydus image converter package (>=0.7.3) provides an option
`OCI: true` for letting the user explicitly specify whether they want
it or not convert Docker media types to OCI media types.

We should enable `OCI: true` in nerdctl for enhanced features of nydus.

Fixed the broken CI in https://github.com/containerd/nerdctl/pull/2200


imeoer:
@AkihiroSuda It fixes https://github.com/containerd/nerdctl/pull/2200, PTAL, thanks!

AkihiroSuda (approved):
Thanks

Issue #2240, opened 2023-05-17
Support --volumes-from in nerdctl run


ningziwen:
### What is the problem you're trying to solve

Mount volumes from existing containers.

### Describe the solution you'd like

```nerdctl run --volumes-from <container> IMAGE```

Feature parity with Docker: https://docs.docker.com/storage/volumes/

ningziwen:
Close as merged.

PR #2241, opened 2023-05-17
feat: add --volumes-from support in run command

ningziwen:
https://github.com/containerd/nerdctl/issues/2240

### DIFF
--- a/cmd/nerdctl/container_create.go
+++ b/cmd/nerdctl/container_create.go
@@ -282,6 +282,10 @@ func processContainerCreateOptions(cmd *cobra.Command) (opt types.ContainerCreat
 	if err != nil {
 		return
 	}
+	opt.VolumesFrom, err = cmd.Flags().GetStringArray("volumes-from")
+	if err != nil {
+		return
+	}
 	// #endregion
 
 	// #region for rootfs flags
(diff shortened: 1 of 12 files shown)


ningziwen:
Running into an issue that the spec of task is expected the but mount is not in destination container.

In containerd, are the mounts in one running container deterministic by the OCI spec of current task?

I used the commands to create the source container with a mount and another destination container to import the mount from the source container. I printed the task.Spec() before the task starts. https://github.com/containerd/nerdctl/pull/2241/files#diff-79f01a47e3272b4712676344123e9b763054e51f44d200a986193a7307c87204R364

```
sudo ./_output/nerdctl run -v /home/ubuntu/nerdctl:/dbdata --name dbstore ubuntu ls /dbdata 

sudo ./_output/nerdctl run --rm --volumes-from dbstore ubuntu ls /dbdata
```

However, the first one can display the correct directory but the second one returns:
```
ls: cannot access '/dbdata': No such file or directory
```

The mounts field in the printed spec of both run command are exactly same as this
```
mounts:[{"Type":"bind","Source":"/home/ubuntu/nerdctl","Destination":"/dbdata","Mode":"","RW":false,"Propagation":""}]
```

AkihiroSuda (approved):
Thanks

PR #2278, opened 2023-06-06
Adding ability to inject secrets into compose build command

andrewbelu:
Injection of secrets is done via the '--secret' argument to nerdctl

Build compose spec shows secrets: https://github.com/compose-spec/compose-spec/blob/master/schema/compose-spec.json#L110
Docker compose shows only usage of Source and Target as aliases for secret: https://github.com/docker/compose/blob/v2/pkg/compose/build.go#L423


pkg/composer/serviceparser/build.go
```diff
@@ -82,6 +83,29 @@ func parseBuildConfig(c *types.BuildConfig, project *types.Project, imageName st
...
+		projectSecret, ok := project.Secrets[fileRef.Source]
+		if !ok {
+			return nil, fmt.Errorf("build: secret %s is undefined", fileRef.Source)
+		}
+		var src string
+		if filepath.IsAbs(projectSecret.File) {
+			logrus.Warnf("build.secrets should be relative path, got %q", projectSecret.File)
+			src = projectSecret.File
+		} else {
+			src = filepath.Join(ctxDir, projectSecret.File)
```
AkihiroSuda, line 100:
```suggestion
			src, err = securejoin.SecureJoin(ctxDir, projectSecret.File)
```

to block  ../../../… attacks

AkihiroSuda (approved):
Thanks

AkihiroSuda (approved):
Thanks

Issue #2295, opened 2023-06-13
Do not throw error when original image was missing when committing a running container


cardyok:
### What is the problem you're trying to solve

I'm not sure whether this is appropriate or not.

When image of a running container is deleted, should this container still be available for commit?

I think in current containerd design this is doable in two ways.

1. reconstruct the content, create diff, ingest into content store and recreate layers for the newly committed image (history of original image will be lost)
2. if image is missing, ask user to provide credentials for original image repo and refetch the missing layers.

there are also similar cases when image is not deleted while contents of some layers are missing (ctr image check image is incomplete), these cases can also be covered.

### Describe the solution you'd like

Some feedback about this feature.

AkihiroSuda:
> reconstruct the content, create diff, ingest into content store and recreate layers for the newly committed image (history of original image will be lost)

SGTM

PR #2297, opened 2023-06-13
reconstruct the content for newly committed image if compressed layers are missing

cardyok:
implements #2295

### DIFF
--- a/pkg/imgutil/commit/commit.go
+++ b/pkg/imgutil/commit/commit.go
@@ -94,6 +94,10 @@ func Commit(ctx context.Context, client *containerd.Client, container containerd
 	if err != nil {
 		return emptyDigest, err
 	}
+	baseManifest, _, err := imgutil.ReadManifest(ctx, baseImg)
+	if err != nil {
+		return emptyDigest, err
+	}
 
 	task, err := container.Task(ctx, cio.Load)
 	if err != nil {
@@ -128,6 +132,7 @@ func Commit(ctx context.Context, client *containerd.Client, container containerd
 		differ = client.DiffService()
 		snName = info.Snapshotter
 		sn     = client.SnapshotService(snName)
+		cs     = baseImg.ContentStore()
 	)
 
 	// Don't gc me and clean the dirty data after 1 hour!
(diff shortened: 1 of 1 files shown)


pkg/imgutil/commit/commit.go
```diff
@@ -100,6 +100,7 @@ func Commit(ctx context.Context, client *containerd.Client, container containerd
 		return emptyDigest, err
 	}
 
+	pauseSwitch := false
```
AkihiroSuda, line 103: var paused bool

AkihiroSuda:
Is this still WIP?

cardyok:
> Is this still WIP?

yes, this pr includes 2 commits, current commit resumes container once upper layer finish committing to reduce the unavailable time for container
second commit includes reconstructing missing layers. Code has been finished and I am still locally testing, I think can be finished this week.

cardyok:
kindly ping @AkihiroSuda

pkg/imgutil/commit/commit.go
```diff
@@ -173,6 +191,34 @@ func Commit(ctx context.Context, client *containerd.Client, container containerd
...
+	}
+	for i, layer := range manifest.Layers {
+		_, err := cs.Info(ctx, layer.Digest)
+		if err == nil {
+			continue
+		}
+		if !errdefs.IsNotFound(err) {
+			return fmt.Errorf("get layer %v failed: %w", layer.Digest.String(), err)
+		}
+		snapshotID := identity.ChainID(config.RootFS.DiffIDs[:i+1]).String()
```
ktock, line 207: Can we get the snapshot ID from the metadata store (e.g. `containers.Container`) rather than manually computing it here?

pkg/imgutil/commit/commit.go
```diff
@@ -173,6 +191,34 @@ func Commit(ctx context.Context, client *containerd.Client, container containerd
...
+		logrus.Infof("creating layer chainid:%v config diffid:%v\n", snapshotID, config.RootFS.DiffIDs[i+1])
+		diffLayerDesc, diffID, err := createDiff(ctx, snapshotID, sn, cs, comparer)
+		if err != nil {
+			return fmt.Errorf("failed to compress layer for %v: %w", layer.Digest.String(), err)
+		}
+		if diffID != config.RootFS.DiffIDs[i] {
+			logrus.Infof("newly created Diff id %v does not match with diff id in config %v: %v", diffID, config.RootFS.DiffIDs[i], err)
```
ktock, line 214: This should be a warning. Also, it should be warned that the contents of the image has been changed by this diff computation.

ktock:
Could you add tests?

pkg/imgutil/commit/commit.go
cardyok: This func does not change content of the original image, it reconstructs the content for new image if original content is missing. So if nothing i changed for original image consumers.

pkg/imgutil/commit/commit.go
cardyok: yea that works too. My concern is that since we are reconstructing layers (digest recorded in manifest is not found), we better rely on manifest/config which is image relevant instead of snapshotter plugin..

cardyok:
> Could you add tests?

Sure thing! thx

PR #2303, opened 2023-06-14
Prevent timing flakyness for `--since/--until` in TestLogs.

aznashwan:
This patch updates `TestLogs` to prevent potential test timing issues while testing the behaviour of the `--since/--until` flags.

Although extremely rare, I have [encountered this timing issue in a workflow run](https://github.com/aznashwan/nerdctl/actions/runs/5255660956/jobs/9495877023#step:9:12594).

### DIFF
--- a/cmd/nerdctl/container_logs_test.go
+++ b/cmd/nerdctl/container_logs_test.go
@@ -36,9 +36,15 @@ bar`
 	base.Cmd("run", "-d", "--name", containerName, testutil.CommonImage,
 		"sh", "-euxc", "echo foo; echo bar").AssertOK()
 
+	//test since / until flag
 	time.Sleep(3 * time.Second)
+	base.Cmd("logs", "--since", "1s", containerName).AssertNoOut(expected)
+	base.Cmd("logs", "--since", "10s", containerName).AssertOutContains(expected)
+	base.Cmd("logs", "--until", "10s", containerName).AssertNoOut(expected)
+	base.Cmd("logs", "--until", "1s", containerName).AssertOutContains(expected)
+
+	// Ensure follow flag works as expected:
 	base.Cmd("logs", "-f", containerName).AssertOutContains("bar")
-	// Run logs twice, make sure that the logs are not removed
 	base.Cmd("logs", "-f", containerName).AssertOutContains("foo")
 
 	//test timestamps flag
@@ -53,12 +59,6 @@ bar`
 		}
 		return nil
 	})
-
-	//test since / until flag
-	base.Cmd("logs", "--since", "1s", containerName).AssertNoOut(expected)
-	base.Cmd("logs", "--since", "5s", containerName).AssertOutContains(expected)
-	base.Cmd("logs", "--until", "5s", containerName).AssertNoOut(expected)
-	base.Cmd("logs", "--until", "1s", containerName).AssertOutContains(expected)
 
 	base.Cmd("rm", "-f", containerName).AssertOK()
 }


cmd/nerdctl/container_logs_test.go
```diff
@@ -36,9 +36,15 @@ bar`
 	base.Cmd("run", "-d", "--name", containerName, testutil.CommonImage,
 		"sh", "-euxc", "echo foo; echo bar").AssertOK()
 
+	//test since / until flag
 	time.Sleep(3 * time.Second)
+	base.Cmd("logs", "--since", "1s", containerName).AssertNoOut(expected)
+	base.Cmd("logs", "--since", "10s", containerName).AssertOutContains(expected)
+	base.Cmd("logs", "--until", "10s", containerName).AssertNoOut(expected)
```
fahedouch, line 43: I believe that `"--until", "10s"` should contains `foo`  if  `"--until", "1s"` contains `foo`

aznashwan: I was slightly confused on how these arguments work when passing a duration too, but since `--until` basically means "up to that point in time", and 10s ago the container didn't exist, let alone output anything, [this check is correct](https://github.com/containerd/nerdctl/actions/runs/5268906176/jobs/9526269301?pr=2303#step:6:496).

The [failure I mentioned](https://github.com/aznashwan/nerdctl/actions/runs/5255660956/jobs/9495877023#step:9:12594) happened because the old `--until 5s ` `AssertNoOut` check would sometimes execute after 5 seconds had passed because of all the other `nerdctl logs` commands being tested up to that point, so just moving the logic here should have been enough; I just increased the tolerances to `10s` instead of `5` to be safe.

fahedouch: Sorry but I am confused here.  I was looking at [docker example](https://docs.docker.com/engine/reference/commandline/logs/#until)  then from my understanding `--until 10s` means from the beginning  of logs **up to** `10s`  that means it should contains `expected` value. 
I checked nerdctl code, it seems we are not correctly [calculating](https://github.com/containerd/nerdctl/blob/main/pkg/logging/jsonfile/jsonfile.go#L94) the until timestamp value ( actually it is `now() - amount_of_duration`), as describe in this [example](https://docs.docker.com/engine/reference/commandline/logs/#until) the `up to` should be relative to log starting point and not relative to current time

aznashwan: > I was looking at [docker example](https://docs.docker.com/engine/reference/commandline/logs/#until)

Their official example is admittedly even more confusing than our test, here's the gist of it:
*  `--until $TIMESTAMP` will show all logs **before** said point in time.
* `--since $TIMESTAMP` will show all logs **after** said point in time.
* `10s` is simple shorthand notation for `$CURRENT_TIMESTAMP - 10s`, nothing more. You can even use negative `-10s` to get a timestamp in the future.

> from my understanding --until 10s means from the beginning of logs up to 10s that means it should contains expected value

I could not find any mention of the "beginning of the logs" part in any of Docker's documentation, it [only uses the phrase "a specific point in time"](https://github.com/docker/cli/blob/master/docs/reference/commandline/logs.md?plain=1#L63).

Looking at their code, both `--until` and `--since` flags [always get processed by `timeutil.GetTimestamp()`](https://github.com/moby/moby/blob/master/client/container_logs.go#L46C1-L55), which [always subtracts the duration](https://github.com/moby/moby/blob/master/api/types/time/timestamp.go#L24C1-L26) like `10s` from the `reference` time you give it. (and the `reference` is always `time.Now()` from the previous link)

The above is [exactly what nerdctl does](https://github.com/containerd/nerdctl/blob/main/pkg/logging/jsonfile/jsonfile.go#L79-L94), so `logs --until 10s` would print logs **before** `10s` ago, which was before the container was event created, which means there should correctly be zero output, as [the original implementer](https://github.com/containerd/nerdctl/commit/828c0210f) intended.

For completeness, I have re-enacted Docker's example on Docker Server+Client `20.10.21` with this script and got the expected results:
```bash
#!/bin/bash

set -eu

ID=$(docker run --name logs-until-test -d busybox sh -c "while true; do $(echo date); sleep 1; done")

echo "### Started container at: $(date). Sleeping 5s."
sleep 5

echo -e '\n### Logs --until 3s ago':
docker logs -f --until=3s $ID

echo -e '\n### Logs --until 15s ago (before container was created = no output):'
docker logs -f --until 15s $ID

# Funnily enough, a negative value will be converted into a future timestamp,
# and `-f` will make it follow until that future point is reached.
echo -e '\n### Logs --until -10s ago (negative seconds = 10s in the future):'
docker logs -f --until -10s $ID

# NOTE: might be one second gap between executions, so 1-2 entries may be missing:
echo -e '\n### Logs --since 10s ago until stopped':
docker logs -f --since 10s $ID
```

fahedouch: I agree, but I still question the logic behind this implementation, it's not at all intuitive this `said point in time`.

fahedouch (approved):
Thanks

PR #2306, opened 2023-06-15
refactor: container stats command

ningziwen:
https://github.com/containerd/nerdctl/issues/1680

### DIFF
--- a/cmd/nerdctl/container_stats.go
+++ b/cmd/nerdctl/container_stats.go
@@ -17,32 +17,10 @@
 package main
 
 import (
-	"bytes"
-	"context"
-	"errors"
-	"fmt"
-	"strings"
-	"sync"
-	"text/tabwriter"
-	"text/template"
-	"time"
-
 	"github.com/containerd/containerd"
-	eventstypes "github.com/containerd/containerd/api/events"
-	"github.com/containerd/containerd/errdefs"
-	"github.com/containerd/containerd/events"
 	"github.com/containerd/nerdctl/pkg/api/types"
 	"github.com/containerd/nerdctl/pkg/clientutil"
-	"github.com/containerd/nerdctl/pkg/containerinspector"
-	"github.com/containerd/nerdctl/pkg/eventutil"
-	"github.com/containerd/nerdctl/pkg/formatter"
-	"github.com/containerd/nerdctl/pkg/idutil/containerwalker"
-	"github.com/containerd/nerdctl/pkg/infoutil"
-	"github.com/containerd/nerdctl/pkg/labels"
-	"github.com/containerd/nerdctl/pkg/rootlessutil"
-	"github.com/containerd/nerdctl/pkg/statsutil"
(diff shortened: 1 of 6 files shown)


cmd/nerdctl/container_stats.go
```diff
@@ -68,417 +46,50 @@ func addStatsFlags(cmd *cobra.Command) {
...
+		return types.ContainerStatsOptions{}, err
 	}
 
 	noStream, err := cmd.Flags().GetBool("no-stream")
 	if err != nil {
-		return err
+		return types.ContainerStatsOptions{}, err
 	}
 
 	format, err := cmd.Flags().GetString("format")
```
AkihiroSuda, line 65: Lint is failing

ningziwen:
@AkihiroSuda Fixed!

cmd/nerdctl/container_stats.go
ningziwen: Fixed

pkg/cmd/container/stats.go
```diff
@@ -0,0 +1,432 @@
...
+			}
+
+			netNS, err := containerinspector.InspectNetNS(ctx, int(task.Pid()))
+			if err != nil {
+				u <- err
+				continue
+			}
+
+			// when (firstSet == true), we only set container stats without rendering stat entry
+			statsEntry, err := setContainerStatsAndRenderStatsEntry(previousStats, firstSet, anydata, int(task.Pid()), netNS.Interfaces)
```
fahedouch, line 389:
```suggestion
			statsEntry, err := setContainerStatsThenRenderStatsEntry(previousStats, firstSet, anydata, int(task.Pid()), netNS.Interfaces)
```

fahedouch:
I left small comment, Globally LGTM , Thanks

pkg/cmd/container/stats.go
ningziwen: "And" seems more common to me than "Then" in my impression. Any specific reasons of preferring "Then"?

(This is out of the scope of the current refactoring goal but I'm good to quickly include it if it makes sense)

fahedouch: what happens under vu: the first iteration we define only `ContainerStats`, then the second iteration we define a new `ContainerStats` and render the statistics by making differences. My idea was to reflect this strategy in the title of the function but we can keep it because you don't feel the need @AkihiroSuda @ningziwen

ningziwen: To me, for the meanings, "and" and "then" should be interchangable in Go naming. 

Even In daily usage, they seem very similar in most scenarios.
"I opened the door, and entered the room"
"I opened the door, then entered the room"

I'm leaning to keeping the more common one "and" but don't have strong preference.

fahedouch (approved):
Thanks

AkihiroSuda (approved):
Thanks

PR #734, opened 2022-01-26
Cirrus: skip BuildKit tests

AkihiroSuda:
#622 is too much flaky

Issue #622, opened 2021-12-15
Rootless CI: `ERROR: failed to do request: Head "https://ghcr.io/v2/stargz-containers/alpine/manifests/3.13-org": dial tcp: lookup ghcr.io on 10.0.2.3:53: read udp 10.0.2.100:50602->10.0.2.3:53: i/o timeout`


AkihiroSuda:
```
ERROR: failed to do request: Head "https://ghcr.io/v2/stargz-containers/alpine/manifests/3.13-org": dial tcp: lookup ghcr.io on 10.0.2.3:53: read udp 10.0.2.100:50602->10.0.2.3:53: i/o timeout
```

Intermittently happens with rootless CI (slirp4netns).
Observed on both GHA and Cirrus.

Likely to be an issue of libslirp.

AkihiroSuda:
#615 added a workaround to use 8.8.8.8 , but still facing a similar error

https://cirrus-ci.com/task/6332163160276992?logs=rootless_test#L433

```
--- FAIL: TestImageVerifyWithCosign (10.78s)
=== RUN   TestImageVerifyWithCosignShouldFailWhenKeyIsNotCorrect
    push_linux_test.go:81: hostIP="172.17.0.2", listenIP="0.0.0.0", listenPort=5000
    pull_linux_test.go:102: localhost IP="127.0.0.1"
    pull_linux_test.go:105: testImageRef="127.0.0.1:5000/nerdctl-testimageverifywithcosignshouldfailwhenkeyisnotcorrect"
    pull_linux_test.go:115: assertion failed: expression is false: res.ExitCode == exitCode: #1 [internal] load build definition from Dockerfile
        #1 transferring dockerfile: 127B done
        #1 DONE 0.0s
        #2 [internal] load .dockerignore
        #2 transferring context: 2B done
        #2 DONE 0.0s
        #3 [internal] load metadata for ghcr.io/stargz-containers/alpine:3.13-org
        #3 ERROR: failed to do request: Head "https://ghcr.io/v2/stargz-containers/alpine/manifests/3.13-org": dial tcp: lookup ghcr.io on 8.8.8.8:53: read udp 10.0.2.100:35508->8.8.8.8:53: i/o timeout
        ------
         > [internal] load metadata for ghcr.io/stargz-containers/alpine:3.13-org:
        ------
        Dockerfile:1
        --------------------
           1 | >>> FROM ghcr.io/stargz-containers/alpine:3.13-org
[...]

PR #2318, opened 2023-06-22
Skip the BuildKit tests in the rootless integration 

yankay:
https://github.com/containerd/nerdctl/issues/622 is too much flaky
The CI is often broken by the issue.

So like https://github.com/containerd/nerdctl/pull/734, skip the test in the rootless integration .

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -149,7 +149,7 @@ jobs:
       - name: "Prepare (network driver=slirp4netns, port driver=builtin)"
         run: DOCKER_BUILDKIT=1 docker build -t ${TEST_TARGET} --target ${TEST_TARGET} --build-arg UBUNTU_VERSION=${UBUNTU_VERSION} --build-arg CONTAINERD_VERSION=${CONTAINERD_VERSION} .
       - name: "Test (network driver=slirp4netns, port driver=builtin)"
-        run: docker run -t --rm --privileged ${TEST_TARGET}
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -149,7 +149,7 @@ jobs:
       - name: "Prepare (network driver=slirp4netns, port driver=builtin)"
         run: DOCKER_BUILDKIT=1 docker build -t ${TEST_TARGET} --target ${TEST_TARGET} --build-arg UBUNTU_VERSION=${UBUNTU_VERSION} --build-arg CONTAINERD_VERSION=${CONTAINERD_VERSION} .
       - name: "Test (network driver=slirp4netns, port driver=builtin)"
-        run: docker run -t --rm --privileged ${TEST_TARGET}
+        run: docker run -t --rm --privileged -e WORKAROUND_CIRRUS=1 ${TEST_TARGET}
```
AkihiroSuda, line 152: This is not Cirrus, so I guess we have to rename the var to something like WORKAROUND_ISSUE_622

AkihiroSuda: I’m also wondering if we can try other workarounds, such as using 8.8.8.8? (From my memory we might have already tried that though)

yankay: Thanks @AkihiroSuda 

It has been change to the WORKAROUND_ISSUE_622 :-)

AkihiroSuda (approved):
Thanks

yankay:
Thanks @AkihiroSuda for the PR review

PR #2332, opened 2023-06-26
Optimize for itoa operations

austinvazquez:
Use `strconv` package for itoa operations instead of `fmt.Sprintf`. Note this change is non-functional and is not required.

### DIFF
--- a/cmd/nerdctl/container_run_log_driver_syslog_test.go
+++ b/cmd/nerdctl/container_run_log_driver_syslog_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


austinvazquez:
I did some manual performance testing to validate my assumptions using the following benchmarks.

Tests:
```
var numToConvert uint = 100000000

func BenchmarkFmt(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = fmt.Sprintf("%d", numToConvert)
	}
}

func BenchmarkStrconv(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = strconv.FormatUint(uint64(numToConvert), 10)
	}
}
```

Results:
```
BenchmarkFmt-8           7690650               154.8 ns/op
BenchmarkStrconv-8      21786693                53.33 ns/op
```

AkihiroSuda (approved):

PR #2333, opened 2023-06-27
refactor: extract the printing logic out of pkg for container list

ningziwen:
> And in the community, people may need to add some features of their own, like https://github.com/containerd/nerdctl/issues/1631, so it's time to refactor the command package!

Based on the [issue](https://github.com/containerd/nerdctl/issues/1680), one of the reason of refactoring nerdctl is to make it easy for developers to import subcommand package. The current refactoring pattern of container list mixed printing logic in List() in pkg, which makes it not flexible to be imported to use. The printing logic is very specific to nerdctl cmd layer, so the idea is to moving the printing logic out of pkg, and let List() return a generic container list.

This is not the only solution to unblock the use case. The alternative is to simply make filterContainers and other functions public.

### DIFF
--- a/pkg/cmd/container/list.go
+++ b/pkg/cmd/container/list.go
@@ -17,15 +17,11 @@
 package container
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"sort"
 	"strings"
-	"text/tabwriter"
-	"text/template"
 	"time"
 
 	"github.com/containerd/containerd"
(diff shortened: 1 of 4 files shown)


pkg/cmd/container/list.go
```diff
@@ -90,147 +86,69 @@ func filterContainers(ctx context.Context, client *containerd.Client, filters []
...
-type containerPrintable struct {
+type ListItem struct {
 	Command   string
-	CreatedAt string
+	CreatedAt time.Time
 	ID        string
 	Image     string
 	Platform  string // nerdctl extension
-	Names     string
+	Names     []string
```
AkihiroSuda, line 95: ?

AkihiroSuda (approved):
Thanks

pkg/cmd/container/list.go
ningziwen: The general principle I'm following for this PR, is to make the pkg layer's data type reflect the intrinsic nature, generic without impacting by the printing preference.

One example to demonstrate this principle is I changed `CreatedAt` from `string` to `time.Time.

For "Names", the naming is a bit confusing because each container can only have one name.

But I looked into the Docker API [spec](https://docs.docker.com/engine/api/v1.43/#tag/Container/operation/ContainerList) and found Names is Array in `/containers/json` response. That made me think it is not just confusing naming but there may be some design consideration behind the scene to keep the flexibility of supporting multiple names in the future.

So I tend to think []string can reflect more intrinsic nature. However, the evidences are not that strong so I don't have strong preference. Happy to ignore the naming and make it completely `string` if you prefer it.

Issue #2269, opened 2023-06-01
`nerdctl inspect --format` behaviour is inconsistent  with docker


wzshiming:
### Description

https://github.com/containerd/nerdctl/blob/fa24191dc73710df0ef2b2600088415f0ad9133b/cmd/nerdctl/inspect.go#L55-L58

The usage message is the same as docker, but the actual behavior is not the same

`nerdctl inspect --format=json`

that docker and podman are the same used `{{ json . }}`

`docker inspect --format='{{ json . }}'`

`podman inspect --format='{{ json . }}'`

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

Compatible with `{{ json . }}`

### What version of nerdctl are you using?

latest

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

AkihiroSuda:
What behavior is inconsistent?

wzshiming:
```
docker create --name=nginx nginx:alpine
docker inspect nginx

[{
   ...
   "State": {...},
   ...
}]
```

```
nerdctl create --name=nginx nginx:alpine
nerdctl inspect nginx

[{
   ...
   "State": null,
   ...
}]
```

The state field is null, which causes the `--format={{ json .State.Running }}` is not work.

wzshiming:
```
nerdctl create --name=nginx nginx:alpine
nerdctl inspect nginx --format='{{ json . }}'

WARN[0000] failed to inspect Task                        error="no running task found: task ... not found: not found" 
{...}
```

Maybe we can remove this warn?

yldoge:
Hey @AkihiroSuda, I'm new here. Could you assign it to me?

yldoge:
Hey guys do we want that the `Status` field returned by `nerdctl inspect` to be `created` whenever the container doesn't have a task, which is consistent with docker?

In test for `inspect`, it says:
```
// nerdctl: run error produces a nil Task, so the Status is empty because Status comes from Task.
// docker : run error gives => `Status=created` as  in docker there is no a separation between container and Task.
```
So it expects an empty `Status` when task is `nil`
But in `nerdctl ps` command code i found this:
```
task, err := c.Task(ctx, nil)
	if err != nil {
		// NOTE: NotFound doesn't mean that container hasn't started.
		// In docker/CRI-containerd plugin, the task will be deleted
		// when it exits. So, the status will be "created" for this
		// case.
		if errdefs.IsNotFound(err) {
[...]

PR #2341, opened 2023-06-30
#2269/Remove warning msg & Fix empty Status field when inspect container that is not running

yldoge:
https://github.com/containerd/nerdctl/issues/2269
Removed the warning message and filled the `Status` field with `created` when `nerdctl inspect` a container that doesn't have a task.


AkihiroSuda:
Please squash and sign the commits

PR #2342, opened 2023-06-30
refactor: allow custom handlers for processing push/pull status updates

mharwani:
Contributing to the refactoring issue #1680 

Currently, the way image push/pull commands are refactored, it is difficult to obtain status updates programmatically. Nerdctl has its own printing logic, but users may want to handle status updates directly via code when importing those subcommands. These changes allow users to define custom handlers for processing status updates during push/pull.


pkg/api/types/image_types.go
```diff
@@ -165,14 +169,15 @@ type ImagePushOptions struct {
...
 	Quiet bool
+	// If non-nil, the Push job will send upload statuses to the handler instead of Stdout
+	ProgressHandler jobs.StatusHandler
 	// AllowNondistributableArtifacts allow pushing non-distributable artifacts
 	AllowNondistributableArtifacts bool
 }
 
 // ImagePullOptions specifies options for `nerdctl (image) pull`.
 type ImagePullOptions struct {
 	Stdout        io.Writer
```
ningziwen, line 180: How about just removing `Stdout` and pass `progressHandler` as `jobs.PrintProgress(cmd.OutOrStderr())` in CLI layer?

mharwani: Sounds good to me. The push command also prints the image reference to stdout if `--quiet` flag is set, so I have changed the push/pull method signature to return image reference along with error.

ningziwen (approved):
LGTM for the refactoring pattern.

ningziwen (commented):
The integration tests are failing and please squash the commits to one.

AkihiroSuda:
Needs rebase

apostasie:
Hey @mharwani 

Do you think you could rebase this?

PR #2346, opened 2023-06-30
Remove warning msg when inspect containers without task

yldoge:
#2269 
Remove warning message when `inspect` containers without task.
Sorry I messed a previous one with a merge main.


fahedouch (approved):
Thanks

Issue #2223, opened 2023-05-10
compose: lacking support for `replicas = 0`


furkanmustafa:
### Description

`nerdctl`, when evaluating compose.yaml, does a sanity check on replica count, and enforces that it has to be greater than zero. `docker compose` itself, allows scaling down to zero, similar to all other container/workload scaling options out there.

PS. There is a flaky behavior as well, where this is not enforced for some commands, some of my services got up, but enforced for some others (not when trying to run this specific service). (could not narrow it down yet. will fill in another bug report for that)

### Steps to reproduce the issue

```
% ls
compose.yaml

% cat compose.yaml 
services:
  test:
    image: busybox
    deploy: { replicas: 0 }
    command: [ "echo", "hello!" ]

% lima nerdctl compose run --rm test
FATA[0000] invalid replicas: 0 
```

### Describe the results you received and expected

Expected (`docker compose` behavior);

```
$ docker compose run --rm test
[+] Running 1/0
 ⠿ Network test_default  Created                                                                                                                                                          0.0s
hello!
```

`nerdctl compose` behavior;

```
% lima nerdctl compose run --rm test
FATA[0000] invalid replicas: 0 
```

### What version of nerdctl are you using?

```
% lima nerdctl version              
Client:
 Version:       v1.3.1
 OS/Arch:       linux/arm64
 Git commit:    b224b280ff3086516763c7335fc0e0997aca617a
 buildctl:
  Version:      v0.11.5
  GitCommit:    252ae63bcf2a9b62777add4838df5a257b86e991

Server:
 containerd:
  Version:      v1.7.0
  GitCommit:    1fbd70374134b891f97ce19c70b6e50c7b9f4e0d
 runc:
  Version:      1.1.5
  GitCommit:    v1.1.5-0-gf19387a6
```

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Lima

### Host information

<details>
  <summary>`lima nerdctl info`</summary>

```
Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v1.7.0
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Log: fluentd journald json-file syslog
  Storage: native overlayfs fuse-overlayfs stargz
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
  rootless
 Kernel Version: 5.19.0-38-generic
[...]

PR #2361, opened 2023-07-06
Fix compose lacking support for replicas = 0

yankay:
Allow the compose service replica set to 0.

fix https://github.com/containerd/nerdctl/issues/2223


ningziwen:
Suggest adding a test case under https://github.com/containerd/nerdctl/blob/main/pkg/composer/serviceparser/serviceparser_test.go to cover the scenario.

ningziwen (approved):

AkihiroSuda (approved):
Thanks

PR #2209, opened 2023-04-28
Fix: incorrect json format output with `stat` command

yuchanns:
Fix #2196

PR #2364, opened 2023-07-07
Fix incorrect usage of fmt.Fprintf()

mook-as:
In quite a few places, we take the pattern of

```go
  b := <template.Execute() output>
  fmt.Printf(..., b.String() + "\n")
```

This means we end up passing arbitrary strings as a printf-style formatting string, leading to issues whenever there is a literal `%` in the output.  As in #2209, convert to `fmt.Fprintln()` as appropriate so that we emit the rendered template as-is.

This was noticed with `nerdctl ps --format json` (where the `nerdctl/log-uri` label has URL-escaped `%2F`), but given that this was the second time I'm seeing this I figured I might as well just `git grep -P 'fmt.Fprintf\([^,]*,\s+[^"]'` and fix all instances instead.

### DIFF
--- a/cmd/nerdctl/container_list.go
+++ b/cmd/nerdctl/container_list.go
@@ -194,7 +194,7 @@ func formatAndPrintContainerInfo(containers []container.ListItem, options Format
 			if err := tmpl.Execute(&b, c); err != nil {
 				return err
 			}
-			if _, err := fmt.Fprintf(w, b.String()+"\n"); err != nil {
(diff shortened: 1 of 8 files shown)


fahedouch:
Thanks, would you please clean your commit by including only your signature `Signed-off-by: Kay Yan <kay.yan@daocloud.io>`

yankay:
The code LGTM :-)

AkihiroSuda (approved):
Thanks

PR #2366, opened 2023-07-07
Add metadata to volume create return

sbuckfelder:
Adds the `*native.Volume` to the volume create return, currently this is discarded.  This is beneficial as if you use nerdctl as a library it allows you to more easily inspect the newly created volume.

**NOTE**: I was not able to find any unit tests or integ tests for this, where should I add testing?

### DIFF
--- a/cmd/nerdctl/volume_create.go
+++ b/cmd/nerdctl/volume_create.go
@@ -57,5 +57,7 @@ func volumeCreateAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


sbuckfelder:
> You can add ut to `cmd/nerdctl/volume_create_test.go`

I was looking through this file, the tests here are done at the cli cmd level, which shouldn't have any changes.  I think the good news is that this does confirm that the change does not introduce anything different in current behavior from a cli standpoint.

ginglis13 (approved):

AkihiroSuda (approved):
Thanks

Issue #2327, opened 2023-06-23
flaky test: `content digest sha256:xxx: not found` in CI 


yankay:
## Description

There are often some test errors in the CI,

It seems the image is not pulled.

Log:
```
2023-06-23T10:00:28.6954646Z === RUN   TestRunCustomRootfs
2023-06-23T10:00:28.6956107Z     container_run_linux_test.go:57: assertion failed: res.ExitCode is not exitCode: time="2023-06-23T10:00:28Z" level=fatal msg="failed to get reader: content digest sha256:ca3cd42a7c9525f6ce3d64c1a70982613a8235f0cc057ec9244052921853ef15: not found"
2023-06-23T10:00:28.6956612Z         
2023-06-23T10:00:28.6956905Z --- FAIL: TestRunCustomRootfs (0.30s)
2023-06-23T10:00:28.6957247Z FAIL github.com/containerd/nerdctl/cmd/nerdctl.TestRunCustomRootfs (0.30s)
2023-06-23T10:03:31.0098479Z === RUN   TestImageConvertNydus
2023-06-23T10:03:31.0099449Z     image_convert_linux_test.go:41: assertion failed: res.ExitCode is not exitCode: time="2023-06-23T10:03:31Z" level=fatal msg="content digest sha256:ca3cd42a7c9525f6ce3d64c1a70982613a8235f0cc057ec9244052921853ef15: not found"
2023-06-23T10:03:31.0099954Z         
2023-06-23T10:03:31.0100273Z --- FAIL: TestImageConvertNydus (0.37s)
2023-06-23T10:03:31.0100643Z FAIL github.com/containerd/nerdctl/cmd/nerdctl.TestImageConvertNydus (0.37s)
2023-06-23T10:03:31.4190709Z === RUN   TestImageConvertZstdChunked
2023-06-23T10:03:31.4197555Z     image_convert_test.go:52: assertion failed: res.ExitCode is not exitCode: time="2023-06-23T10:03:31Z" level=fatal msg="content digest sha256:ca3cd42a7c9525f6ce3d64c1a70982613a8235f0cc057ec9244052921853ef15: not found"
2023-06-23T10:03:31.4203446Z         
2023-06-23T10:03:31.4209424Z --- FAIL: TestImageConvertZstdChunked (0.41s)
2023-06-23T10:03:31.4215481Z FAIL github.com/containerd/nerdctl/cmd/nerdctl.TestImageConvertZstdChunked (0.41s)
[...]

PR #2367, opened 2023-07-10
Deflake TestRunCustomRootfs by workaround

yankay:
Attempt to fix https://github.com/containerd/nerdctl/issues/2327 by workaround

The failure is because the nerdctl can not pull all the layers, 
and the `ctr `can do it.

### DIFF
--- a/cmd/nerdctl/container_run_linux_test.go
+++ b/cmd/nerdctl/container_run_linux_test.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


fahedouch:
Thanks, but I don't think that is a good idea to add a dependency to `ctr` . IMHO we have to investigate why `nerdctl` do not attempt to fetch correctly layers inside the ci

yankay:
Thanks @fahedouch 

After investigating the issue, the root cause is here:
https://github.com/containerd/containerd/blob/main/pkg/unpack/unpacker.go#L300C14-L300C14
The layer is skipped because the snapshot can be found. 

But I have no idea how to fix it :-)

fahedouch:
> Thanks @fahedouch
> 
> After investigating the issue, the root cause is here: https://github.com/containerd/containerd/blob/main/pkg/unpack/unpacker.go#L300C14-L300C14 The layer is skipped because the snapshot can be found.
> 
> But I have no idea how to fix it now :-)

So it is no longer a cli issue.

apostasie:
A workaround has been put in place in #3179 that seems to address the issue on the CI.

As for the underlying issue, 2 cents is on the fact we do not ensure the image is extracted (will look into that lead later).

apostasie:
@AkihiroSuda I guess this can be closed?

AkihiroSuda:
> @AkihiroSuda I guess this can be closed?

Sure

PR #2384, opened 2023-07-20
Refactored container wait and make it public

KevinLiAWS:
Wait container is useful for external callers.
The current container wait private method only return the error without the container status code, instead, it print the code.
This PR refactored the code to return both the status code and error, and make it public.
From docker API, the status code is int64, so converted the original uint32 to int64, and also use -1 to indicate non container wait error status.

### DIFF
--- a/pkg/cmd/container/wait.go
+++ b/pkg/cmd/container/wait.go
@@ -19,7 +19,6 @@ package container
 import (
 	"context"
 	"fmt"
-	"io"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/nerdctl/pkg/api/types"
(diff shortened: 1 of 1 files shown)


pkg/cmd/container/wait.go
```diff
@@ -49,29 +48,27 @@ func Wait(ctx context.Context, client *containerd.Client, reqs []string, options
...
+		fmt.Fprintln(w, code)
+		if waitErr != nil {
 			allErr = multierror.Append(allErr, waitErr)
 		}
 	}
 	return allErr
 }
 
-func waitContainer(ctx context.Context, w io.Writer, container containerd.Container) error {
+func WaitContainer(ctx context.Context, container containerd.Container) (code int64, err error) {
```
AkihiroSuda, line 60: This should be `pkg/containerutil.Wait()`
https://github.com/containerd/nerdctl/blob/main/pkg/containerutil/containerutil.go

ningziwen: If we want to keep the pattern consistent with other recent refactoring, instead of making methods public, we can also move stdout to cmd from pkg. 
https://github.com/containerd/nerdctl/pull/2333
https://github.com/containerd/nerdctl/pull/2342/files

AkihiroSuda:
Needs rebase

apostasie:
Hey @KevinLiAWS 

Do you think you could rebase this?

PR #2404, opened 2023-07-31
fix: registry missing layer when concurrently push image

zhenxuanxin:
this PR fix the following issues:
1. MANIFEST_BLOB_UNKNOWN occur in registry when push image that most layer are same;
2. image repository not create in registry when push image that all layer are same.


AkihiroSuda (approved):
Thanks

PR #2405, opened 2023-08-01
Add lease and remove special case

gabriel-samfira:
This change adds a lease to the view creation. Without a lease, there is a high probability that the newly created view will be removed by the containerd garbage collector shortly after we create the view and before we get a chance to use it.

Depending on when the GC runs during the call to `m.Mount()`, we may get a set of different errors (at least on Windows) ranging from `The system cannot find the path specified` to `The parameter is incorrect` to `Access is denied.`

Adding a lease prevents containerd from removing the view mid-flight. 

This change also adds a drive-by fix. The special case for Windows when mounting the view, was removed. 

Fixes https://github.com/containerd/containerd/issues/8891

### DIFF
--- a/pkg/cmd/container/run_mount.go
+++ b/pkg/cmd/container/run_mount.go
@@ -25,10 +25,12 @@ import (
 	"runtime"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/containers"
 	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/leases"
 	"github.com/containerd/containerd/mount"
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/pkg/userns"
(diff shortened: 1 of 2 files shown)


gabriel-samfira:
CC: @AkihiroSuda

gabriel-samfira:
The failure seems to be in `TestStart`. That seems to be a bug unrelated to the snapshotter issue.

gabriel-samfira:
The `TestStart` fail seems to only happen on `ltsc2019`

AkihiroSuda:
Thank you, could you try running CI with containerd v1.7.3?
https://github.com/containerd/nerdctl/blob/b33a58f288bc42351404a016e694190b897cd252/.cirrus.yml#L36

gabriel-samfira:
CI seems happy on 1.7.3

pkg/cmd/container/run_mount.go
```diff
@@ -159,6 +161,14 @@ func generateMountOpts(ctx context.Context, client *containerd.Client, ensuredIm
 		// When the Unmount fails, RemoveAll will incorrectly delete data from the mounted dir
 		defer os.Remove(tempDir)
 
+		// Add a lease of 1 hour to the view so that it is not garbage collected
```
gabriel-samfira, line 164: @AkihiroSuda I added a 1 hour lease. Let me know if you prefer a different value.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Not sure if relevant, but this one is now consistently failing
- https://github.com/containerd/nerdctl/issues/2412

gabriel-samfira:
I don't think it was triggered by this change. It seems to be a networking issue, which I have seen on ltsc2019, but does not happen on ltsc2022. Does the CI also run on ltsc2022? Are you seeing it there as well?

AkihiroSuda:
Currently the Cirrus CI only runs on  "windows-2019-core-for-containers"
https://github.com/cirruslabs/vm-images/blob/master/googlecompute/windows_images.json

Maybe we should migrate to GitHub Actions?

PR #2413, opened 2023-08-03
fix: Support copy out of volumes in stopped containers

vsiravar:
### What does this PR do?
- When a container is stopped, copying out of mounted volumes does not work. This change supports copying in and out of mounted volumes in a stopped/created container.
- If a volume is marked read-only, added check to throw error if copying to the read-only volume
- If the container rootfs is read-only, added check to throw error if copying to the read-only rootfs

### Testing
- Added integration test to ensure copying in/out of volumes has the same behaviour in a stopped container as a running container

### DIFF
--- a/cmd/nerdctl/container_cp_linux_test.go
+++ b/cmd/nerdctl/container_cp_linux_test.go
@@ -51,8 +51,8 @@ func TestCopyToContainer(t *testing.T) {
 
 	assertCat := func(catPath string, testContainer string, stopped bool) {
 		if stopped {
-			base.Cmd("start", testStoppedContainer).AssertOK()
-			defer base.Cmd("stop", testStoppedContainer).AssertOK()
+			base.Cmd("start", testContainer).AssertOK()
+			defer base.Cmd("stop", testContainer).AssertOK()
 		}
 		t.Logf("catPath=%q", catPath)
 		base.Cmd("exec", testContainer, "cat", catPath).AssertOutExactly(string(srcFileContent))
@@ -107,6 +107,75 @@ func TestCopyToContainer(t *testing.T) {
 			base.Cmd("stop", testStoppedContainer).AssertOK()
 			base.Cmd("cp", srcPath, testStoppedContainer+":"+destPath).AssertOK()
 			assertCat(catPath, testStoppedContainer, true)
+		})
+		t.Run("DEST_PATH is in a volume", func(t *testing.T) {
+			// Create a volume
+			vol := "somevol"
+			base.Cmd("volume", "create", vol).AssertOK()
+			defer base.Cmd("volume", "rm", vol).Run()
+			con := fmt.Sprintf("%s-with-volume", testContainer)
+			mountDir := "/some_dir"
+			base.Cmd("run", "-d", "--name", con, "-v", fmt.Sprintf("%s:%s", vol, mountDir), testutil.CommonImage, "sleep", "1h").AssertOK()
+			defer base.Cmd("rm", "-f", con).Run()
+			catPath := filepath.Join(mountDir, filepath.Base(srcFile))
+			// Running container test
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertOK()
+			assertCat(catPath, con, false)
+
+			// Skip for rootless
+			if rootlessutil.IsRootless() {
+				t.Skip("Test skipped in rootless mode for testStoppedContainer")
+			}
+			// Stopped container test
+			// Delete previously copied file
+			base.Cmd("exec", con, "rm", catPath).AssertOK()
+			base.Cmd("stop", con).AssertOK()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertOK()
+			assertCat(catPath, con, true)
+		})
+		t.Run("Destination path is a read-only", func(t *testing.T) {
+			vol := "somevol"
+			base.Cmd("volume", "create", vol).AssertOK()
+			defer base.Cmd("volume", "rm", vol).Run()
+			con := fmt.Sprintf("%s-with-read-only-volume", testContainer)
+			mountDir := "/some_dir"
+			// Create container with read-only volume mounted
+			base.Cmd("run", "-d", "--name", con, "-v", fmt.Sprintf("%s:%s:ro", vol, mountDir), testutil.CommonImage, "sleep", "1h").AssertOK()
+			defer base.Cmd("rm", "-f", con).Run()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertFail()
+
+			// Skip for rootless
+			if rootlessutil.IsRootless() {
+				t.Skip("Test skipped in rootless mode for testStoppedContainer")
+			}
+
+			// Stopped container test
+			// Delete previously copied file
+			base.Cmd("stop", con).AssertOK()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertFail()
+		})
+		t.Run("Destination path is a read-only and default tmpfs mount point", func(t *testing.T) {
+			vol := "somevol"
+			base.Cmd("volume", "create", vol).AssertOK()
+			defer base.Cmd("volume", "rm", vol).Run()
+			con := fmt.Sprintf("%s-with-read-only-volume", testContainer)
+
+			// /tmp is from rootfs of alpine
+			mountDir := "/tmp"
+			// Create container with read-only mounted volume mounted at /tmp
+			base.Cmd("run", "-d", "--name", con, "-v", fmt.Sprintf("%s:%s:ro", vol, mountDir), testutil.CommonImage, "sleep", "1h").AssertOK()
+			defer base.Cmd("rm", "-f", con).Run()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertFail()
+
+			// Skip for rootless
+			if rootlessutil.IsRootless() {
+				t.Skip("Test skipped in rootless mode for testStoppedContainer")
+			}
+
+			// Stopped container test
+			// Delete previously copied file
+			base.Cmd("stop", con).AssertOK()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertFail()
 		})
 	})
 	t.Run("SRC_PATH specifies a directory", func(t *testing.T) {
@@ -176,7 +245,6 @@ func TestCopyFromContainer(t *testing.T) {
 	base := testutil.NewBase(t)
 	testContainer := testutil.Identifier(t)
 	testStoppedContainer := "stopped-container-" + testutil.Identifier(t)
-
 	base.Cmd("run", "-d", "--name", testContainer, testutil.CommonImage, "sleep", "1h").AssertOK()
 	defer base.Cmd("rm", "-f", testContainer).Run()
 
@@ -253,6 +321,45 @@ func TestCopyFromContainer(t *testing.T) {
 				t.Skip("Test skipped in rootless mode for testStoppedContainer")
 			}
 			base.Cmd("cp", testStoppedContainer+":"+srcPath, destPath).AssertOK()
+			assertCat(catPath)
+		})
+		t.Run("SRC_PATH is in a volume", func(t *testing.T) {
+			// Setup
+			// Create a volume
+			vol := "somevol"
+			base.Cmd("volume", "create", vol).AssertOK()
+			defer base.Cmd("volume", "rm", "-f", vol).Run()
+
+			// Create container for test
+			con := fmt.Sprintf("%s-with-volume", testContainer)
+
+			mountDir := "/some_dir"
+			base.Cmd("run", "-d", "--name", con, "-v", fmt.Sprintf("%s:%s", vol, mountDir), testutil.CommonImage, "sleep", "1h").AssertOK()
+			defer base.Cmd("rm", "-f", con).Run()
+
+			// Create a file to mounted volume
+			mountedVolFile := filepath.Join(mountDir, "test-file")
+			mkSrcScript = fmt.Sprintf("echo -n %q >%q && chown %d %q", srcFileContent, mountedVolFile, srcUID, mountedVolFile)
+			base.Cmd("exec", con, "sh", "-euc", mkSrcScript).AssertOK()
+
+			// Create destination directory on host for copy
+			destPath := filepath.Join(td, "dest-dir")
+			err := os.Mkdir(destPath, 0o700)
+			assert.NilError(t, err)
+
+			catPath := filepath.Join(destPath, filepath.Base(mountedVolFile))
+
+			// Running container test
+			base.Cmd("cp", con+":"+mountedVolFile, destPath).AssertOK()
+			assertCat(catPath)
+
+			// Skip for rootless
+			if rootlessutil.IsRootless() {
+				t.Skip("Test skipped in rootless mode for testStoppedContainer")
+			}
+			// Stopped container test
+			base.Cmd("stop", con).AssertOK()
+			base.Cmd("cp", con+":"+mountedVolFile, destPath).AssertOK()
 			assertCat(catPath)
 		})
 	})
--- a/pkg/containerutil/cp_linux.go
+++ b/pkg/containerutil/cp_linux.go
@@ -18,10 +18,12 @@ package containerutil
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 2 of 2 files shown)


pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,44 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+	}
+	if spec.Root.Readonly {
+		return "", "", fmt.Errorf("container rootfs: %s is marked read-only", spec.Root.Path)
+	}
+
+	for _, mount := range spec.Mounts {
+		for _, option := range mount.Options {
+			if isParentDir(filePath, mount.Destination) {
+				if option == "ro" {
+					return "", "", fmt.Errorf("mount point %s is marked read-only", filePath)
```
ktock, line 293: This seems to correctly checked only when ro is the first option? Is this an intended behaviour?

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,44 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+				}
+				return mount.Source, mount.Destination, nil
+			}
+		}
+	}
+	return "", "", nil
+}
+
+func isParentDir(filePath, potentialParent string) bool {
```
ktock, line 302: Paths should be cleaned with filepath.Clean.

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,44 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
 	}
 	return tempDir, cleanup, nil
 }
+
+func getContainerMountInfo(ctx context.Context, con containerd.Container, containerPath string) (string, string, error) {
+	filePath := filepath.Clean(containerPath)
```
ktock, line 280: Package name shouldn't be overwritten.

pkg/containerutil/cp_linux.go
vsiravar: Good catch! Updated.

pkg/containerutil/cp_linux.go
vsiravar: Can you expand a little bit, unsure what you mean.

ktock: nvm, I confused `filePath` with `filepath`.

cmd/nerdctl/container_cp_linux_test.go
```diff
@@ -108,6 +107,31 @@ func TestCopyToContainer(t *testing.T) {
...
+			if rootlessutil.IsRootless() {
+				t.Skip("Test skipped in rootless mode for testStoppedContainer")
+			}
+			// Stopped container test
+			// Delete previously copied file
+			base.Cmd("exec", con, "rm", catPath).AssertOK()
+			base.Cmd("stop", con).AssertOK()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertOK()
+			assertCat(catPath, con, true)
+		})
```
ktock, line 134: can we have tests for read-only volume?

.github/workflows/release.yml
```diff
@@ -1,6 +1,7 @@
 # See https://github.com/containerd/nerdctl/blob/main/MAINTAINERS_GUIDE.md for how to make a release.
 name: Release
 on:
+  workflow_dispatch:
```
AkihiroSuda, line 4: ?

vsiravar: My bad, I was experimenting in my fork with GitHub actions. Did not realize it was this branch. Updated.

cmd/nerdctl/container_cp_linux_test.go
vsiravar: Updated with a test.

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,46 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+	spec, err := con.Spec(ctx)
+	if err != nil {
+		return "", "", err
+	}
+	if spec.Root.Readonly {
+		return "", "", fmt.Errorf("container rootfs: %s is marked read-only", spec.Root.Path)
+	}
+
+	for _, mount := range spec.Mounts {
+		if isParentDir(filePath, mount.Destination) {
```
ktock, line 290: We should manage the priority of mounts with the same destination. For example if a user do `-v /dir:/tmp:ro` we should avoid copying files into `/tmp` which is readonly. But it looks like the first mount is prioritized here so it can cause unexpected behaviour in the situation like the above?

vsiravar: `isParentDir` checks is user is trying to copy to subdirectory of a mounted volume. Eg `isParentDir("/tmp/some_nested_dir", "/tmp")` will return `true `and if `/tmp` is a `ro`, it'll throw an error. 

- If a user creates multiple volumes with the same mount point `sudo nerdctl run -v myvol:/tmp:ro -v myvol2:/tmp alpine`, it's anyway ambiguous to whether /tmp should be read-only or not irrespective of the order. Although nerdctl allows mounting multiple  volumes to the same mount point(/tmp in this example), docker throws an error with message `Duplicate mount point: /tmp`. Should this perhaps be fixed in the run command to not allow same mount point?

ktock: > Should this perhaps be fixed in the run command to not allow same mount point?

I think yes it should. Even with this limitation, I believe user still can specify a single `-v /dir:/tmp:ro` which cause duplication on /tmp for the default tmpfs mount and the user-specified mount so the duplication management is still needed here.

vsiravar: From the [runtime spec](https://github.com/opencontainers/runtime-spec/blob/main/specs-go/config.go#L18) `spec.Mounts` contains additional mounts on top of root so the default tmps mount won't show up in this list.

> single -v /dir:/tmp:ro which cause duplication on /tmp for the default tmpfs mount

I added a [test](https://github.com/containerd/nerdctl/pull/2413/files#diff-fc1f66854a3bb061fe38564a668b1b07c4be8fe9f404e8248b80ab2edad6e22fR156) for this scenario to confirm correct behaviour of not allowing to write to /tmp. Maybe I missed something here, can you help define a test case to describe the scenario for duplication and allowing write to a read-only volume?

pkg/containerutil/cp_linux.go
```diff
@@ -54,13 +55,27 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
 		if !errdefs.IsNotFound(err) || rootlessutil.IsRootless() {
 			return err
```
fahedouch, line 56: possible to add a message to mention that copying to a container without root is not yet supported ?

pkg/containerutil/cp_linux.go
```diff
@@ -54,13 +55,27 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
 		}
-		root, cleanup, err = mountSnapshotForContainer(ctx, client, container, snapshotter)
-		if cleanup != nil {
-			defer cleanup()
+		if !container2host {
+			path = dst
+		} else {
+			path = src
```
fahedouch, line 61: positive if clause if simpler to red
```
		if container2host {
			path = src
		} else {
			path = dst
                 }
```

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,46 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+			return mount.Source, mount.Destination, nil
+		}
+	}
+	return "", "", nil
+}
+
+func isParentDir(filePath, potentialParent string) bool {
+	filePath = filepath.Clean(filePath)
+	potentialParent = filepath.Clean(potentialParent)
+	if filePath == potentialParent {
```
fahedouch, line 305: `/etc/var/test` is not a parent of `/etc/var/test`
```suggestion
	if path.Dir(filePath) == potentialParent
```

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,46 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+		return true
+	}
+	if filePath == "/" {
+		return false
+	}
+	return isParentDir(path.Dir(filePath), potentialParent)
```
fahedouch, line 311:
```suggestion
	return isParentDir(filePath, potentialParent)
```

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,46 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+}
+
+// When the path is in volume remove directory that volume is mounted on from the path
+func handleVolumePaths(container2host bool, dst string, src string, destination string) (string, string) {
```
fahedouch, line 315: `dst` and `destination` are really confusing, please choose more meaningful names

pkg/containerutil/cp_linux.go
```diff
@@ -54,13 +55,27 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
-		root, cleanup, err = mountSnapshotForContainer(ctx, client, container, snapshotter)
-		if cleanup != nil {
-			defer cleanup()
+		if !container2host {
+			path = dst
+		} else {
+			path = src
 		}
+		// Check if path is in a volume
+		root, destination, err = getContainerMountInfo(ctx, container, path)
```
fahedouch, line 64: this check concerns only `host  to container` copy , please put in the above if clause

pkg/containerutil/cp_linux.go
```diff
@@ -54,13 +55,27 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
 		if err != nil {
 			return err
 		}
+		dst, src = handleVolumePaths(container2host, dst, src, destination)
+		// path is not in a volume and is inside container.
```
fahedouch, line 69:
```suggestion
		// path is not in a volume and is inside container and container rootfs is not read-only
```

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,46 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+				}
+			}
+			return mount.Source, mount.Destination, nil
+		}
+	}
+	return "", "", nil
+}
+
+func isParentDir(filePath, potentialParent string) bool {
+	filePath = filepath.Clean(filePath)
```
fahedouch, line 303:
```suggestion
	if filePath == "/"  ||  filePath = "" ||  potentialParent = "" {
		return false
	}
	filePath = filepath.Clean(filePath)
```
we need a base case here;  `if filePath == "/"  ||  filePath = "" ||  potentialParent = ""` no need to make any check

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,46 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+					return "", "", fmt.Errorf("mount point %s is marked read-only", filePath)
+				}
+			}
+			return mount.Source, mount.Destination, nil
+		}
+	}
+	return "", "", nil
+}
+
+func isParentDir(filePath, potentialParent string) bool {
```
fahedouch, line 302:
```suggestion
	if filePath == "/"  ||  filePath == "" ||  potentialParent == "" {
		return false
	}
	filePath = filepath.Clean(filePath)
```
we need a base case here;  
`if filePath == "/"` : filepath is root, it does not have parent
`filePath = "" ||  potentialParent = ""`  this func can be used as utility for other context, so this check is need

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +275,46 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+	filePath = filepath.Clean(filePath)
+	potentialParent = filepath.Clean(potentialParent)
+	if filePath == potentialParent {
+		return true
+	}
+	if filePath == "/" {
+		return false
+	}
+	return isParentDir(path.Dir(filePath), potentialParent)
+}
```
fahedouch, line 312: I have better suggestion for `isParentDir`   func :

```
  func isParentDir(filePath, potentialParent string) bool {
         // base case
         if filePath == "/"  ||  filePath == "" ||  potentialParent == "" {
		return false
	}

        filePath = filepath.Clean(filePath)
	potentialParent = filepath.Clean(potentialParent)
	if filePath == potentialParent {
            return true
         }
         return isParentDir(path.Dir(filePath), potentialParent)
}

File: pkg/containerutil/cp_linux.go (hunk 9 above)
Comment (vsiravar) [reply pkg/containerutil/cp_linux.go:305]: We want isParentDir("/etc/var/test", "/etc/var/test") to return `true` since `/etc/var/test` is in the volume so is `/etc/var/test/test1`. I will rename the function to reflect that.

**Comment by vsiravar** [comment]
@fahedouch Thanks for the review. I have addressed your feedback in my PR. PTAL. Thanks!

File: pkg/containerutil/cp_linux.go [hunk 17]
```diff
@@ -43,24 +44,42 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
 		return err
 	}
 	logrus.Debugf("Detected tar binary %q (GNU=%v)", tarBinary, isGNUTar)
-	var srcFull, dstFull, root string
+	var srcFull, dstFull, root, mountDestination, path string
 	var cleanup func()
 	task, err := container.Task(ctx, nil)
 	if err != nil {
-		// if the task is simply not found, we should try to mount the snapshot. any other type of error from Task() is fatal here.
 		// Rootless does not support copying into/out of stopped/created containers as we need to nsenter into the user namespace of the
```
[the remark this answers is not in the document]
vsiravar: Do you have any ideas to fix this for rootless while preserving the UID/GID mapping. I can implement it in a different PR once brainstorm the idea.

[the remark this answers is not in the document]
fahedouch: you  can access the user ns with `nsenter` to gain caps and then do the copy. Please [see](https://github.com/containerd/nerdctl/blob/main/pkg/containerutil/cp_linux.go#L182-L189)

fahedouch (commented):
LGTM overall I left some new feedback :)

cmd/nerdctl/container_cp_linux_test.go
```diff
@@ -108,6 +108,75 @@ func TestCopyToContainer(t *testing.T) {
...
+
+			// /tmp is from rootfs of alpine
+			mountDir := "/tmp"
+			// Create container with read-only mounted volume mounted at /tmp
+			base.Cmd("run", "-d", "--name", con, "-v", fmt.Sprintf("%s:%s:ro", vol, mountDir), testutil.CommonImage, "sleep", "1h").AssertOK()
+			defer base.Cmd("rm", "-f", con).Run()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertFail()
+
+			// Skip for rootless
+			if rootlessutil.IsRootless() {
```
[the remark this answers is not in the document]
vsiravar: I am a bit confused with this, do we not want to run `TestCopyToContainer` in rootless for running containers?

cmd/nerdctl/container_cp_linux_test.go
```diff
@@ -108,6 +108,75 @@ func TestCopyToContainer(t *testing.T) {
...
+				t.Skip("Test skipped in rootless mode for testStoppedContainer")
+			}
+
+			// Stopped container test
+			// Delete previously copied file
+			base.Cmd("stop", con).AssertOK()
+			base.Cmd("cp", srcPath, con+":"+mountDir).AssertFail()
+		})
+		t.Run("Destination path is a read-only and shared", func(t *testing.T) {
```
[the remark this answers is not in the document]
vsiravar: Updated

cmd/nerdctl/container_cp_linux_test.go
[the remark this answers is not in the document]
fahedouch: no, the idea is to drop the whole `TestCopyToContainer`  for rootless mode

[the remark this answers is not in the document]
vsiravar: Hmm, but copy to container works for rootless mode when container is running. Matrix below for cp support for rootless and rootful for different container states. I prefer to test copy in rootless mode when container status = Running. I might have missed something here, do let me know if this seems wrong. 
| Mode    | Status = running | Status = created  | Status =  stopped | Status = paused |
| ----------- | ----------- | ------------------- | ------------------ | ---------------- |
|  Rootful      |  true | true | true | true |
| Rootless   | true | false | false | false |

fahedouch (approved):
Thanks

pkg/containerutil/cp_linux.go
```diff
@@ -43,24 +45,42 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
 	}
 	logrus.Debugf("Detected tar binary %q (GNU=%v)", tarBinary, isGNUTar)
-	var srcFull, dstFull, root string
+	var srcFull, dstFull, root, mountDestination, containerPath string
 	var cleanup func()
 	task, err := container.Task(ctx, nil)
 	if err != nil {
-		// if the task is simply not found, we should try to mount the snapshot. any other type of error from Task() is fatal here.
-		// Rootless does not support copying into/out of stopped/created containers as we need to nsenter into the user namespace of the
+		// FIXME: Rootless does not support copying into/out of stopped/created containers as we need to nsenter into the user namespace of the
```
ktock, line 52: Can we add this to the doc?

pkg/containerutil/cp_linux.go
```diff
@@ -247,3 +282,50 @@ func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, c
...
+	spec, err := con.Spec(ctx)
+	if err != nil {
+		return "", "", err
+	}
+	// read-only applies only while copying into container from host
+	if !container2host && spec.Root.Readonly {
+		return "", "", fmt.Errorf("container rootfs: %s is marked read-only", spec.Root.Path)
+	}
+
+	for _, mount := range spec.Mounts {
```
ktock, line 297: It seems that the `nerdctl cp` behaviour is different between running and stopping containers when they have duplicated volumes:

When a file is copied to the runing container, it looks file the file is copied to the second volue:

```console
# nerdctl run -it --name voltest -v test1:/mnt/ -v test2:/mnt/ ubuntu:22.04 /bin/bash
...
# echo hi > /tmp/hi
# nerdctl cp /tmp/hi voltest:/mnt/
# nerdctl run --rm -it -v test2:/mnt/ ubuntu:22.04 cat /mnt/hi
hi
```

After stopping `voltest`:

```console
# echo aaa > /tmp/hi2
# nerdctl cp /tmp/hi2 voltest:/mnt/
# nerdctl run --rm -it -v test1:/mnt/ ubuntu:22.04 ls -al /mnt/
total 12
drwxr-xr-x 2 root root 4096 Aug 26 06:25 .
drwxr-xr-x 1 root root 4096 Aug 26 06:27 ..
-rw-r--r-- 1 root root    4 Aug 26 06:25 mnt
# nerdctl run --rm -it -v test1:/mnt/ ubuntu:22.04 cat /mnt/mnt
aaa
```

Now it looks like the file `hi2` is copied to the volume `test1` and the name also changed to `mnt` (not `hi2`).

vsiravar: @ktock This is a valid point, however having duplicate mount points does not make a lot of sense IMO. Maybe this fix should be in `run_mount.go` to not allow duplicate mount points and keep the behaviour consistent with docker? Test in docker. 
```
 docker run -it --name voltest -v test1:/mnt/ -v test2:/mnt/ ubuntu:22.04 /bin/bash
Unable to find image 'ubuntu:22.04' locally
22.04: Pulling from library/ubuntu
db76c1f8aa17: Pull complete 
Digest: sha256:ec050c32e4a6085b423d36ecd025c0d3ff00c38ab93a3d71a460ff1c44fa6d77
Status: Downloaded newer image for ubuntu:22.04
docker: Error response from daemon: Duplicate mount point: /mnt.
See 'docker run --help'.
```

ktock: > not allow duplicate mount points and keep the behaviour consistent with docker

SGTM

AkihiroSuda:
@ktock Can we merge this and revisit other issues in follow-up PRs?

ktock:
SGTM :+1:

Issue #2420, opened 2023-08-05
`nerdctl --help` is messed up with `nerdctl attach` caveats


AkihiroSuda:
```console
$ nerdctl --help
nerdctl is a command line interface for containerd

Config file ($NERDCTL_TOML): /home/suda/.config/nerdctl/nerdctl.toml

Usage: nerdctl [flags]

Management commands:
  apparmor   Manage AppArmor profiles
  builder    Manage builds
  container  Manage containers
  image      Manage images
  ipfs       Distributing images on IPFS
  namespace  Manage containerd namespaces
  network    Manage networks
  system     Manage containerd
  volume     Manage volumes

Commands:
  attach      Attach stdin, stdout, and stderr to a running container. For example:

1. 'nerdctl run -it --name test busybox' to start a container with a pty
2. 'ctrl-p ctrl-q' to detach from the container
3. 'nerdctl attach test' to attach to the container

Caveats:

- Currently only one attach session is allowed. When the second session tries to attach, curre
ntly no error will be returned from nerdctl.
  However, since behind the scenes, there's only one FIFO for stdin, stdout, and stderr respec
tively,
  if there are multiple sessions, all the sessions will be reading from and writing to the sam
e 3 FIFOs, which will result in mixed input and partial output.
- Until dual logging (issue #1946) is implemented,
  a container that is spun up by either 'nerdctl run -d' or 'nerdctl start' (without '--attach
') cannot be attached to.
  build       Build an image from a Dockerfile. Needs buildkitd to be running.
...
```

https://github.com/containerd/nerdctl/commit/f0140a5c1b13b89bd046cab4a5dc60d161b765a6

PR #2422, opened 2023-08-07
Fix nerdctl --help is messed up with nerdctl attach caveats

yankay:
Create a long help to 'attach' commands.

fix https://github.com/containerd/nerdctl/issues/2420


AkihiroSuda (approved):
Thanks

PR #2428, opened 2023-08-10
Update to Go 1.21

austinvazquez:
[Go 1.21](https://go.dev/blog/go1.21) is released!

This PR updates the Go version in CI and includes a linter (golangci-lint) version update for the new Go version support.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containerd/nerdctl
 
-go 1.19
+go 1.20
```
austinvazquez, line 3: @AkihiroSuda, curious on your thoughts if we should upgrade the min Go version or give folks a little more time to migrate to supported Go versions. i.e. Go 1.20 and 1.21 now

AkihiroSuda: We still have to support 1.20

fahedouch (approved):
Thanks

Issue #2432, opened 2023-08-11
anonymous volume create behaviour is not compatible with Docker


yankay:
### Description

Anonymous volume create behaviour is not compatible with Docker

### Steps to reproduce the issue

1. nerdctl volume create

### Describe the results you received and expected

The results received
```
FATA[0000] "nerdctl volume create" requires exactly 1 argument(s).
See 'nerdctl volume create --help'.

Usage:  nerdctl volume create [flags] VOLUME

Create a volume
```

Expected:
```
root@kay200:~# docker volume create
968922466b3e48f7004ee1b9aaad6892fc0ab27a9d618811079e2d274b93686c
```

### What version of nerdctl are you using?

Client:
 Version:       v1.5.0-11-g14ec1cc
 OS/Arch:       linux/amd64
 Git commit:    14ec1cc63de3a84ab248fd6765df19403c14f96a
 buildctl:
  Version:      v0.12.1
  GitCommit:    bb857a0d49f45aa0ce9cd554b78d4075553e20f9

Server:
 containerd:
  Version:      v1.7.1
  GitCommit:    1677a17964311325ed1c31e2c0a3589ce6d5c30d
 runc:
  Version:      1.1.7
  GitCommit:    v1.1.7-0-g860f061b

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

Issue #2421, opened 2023-08-05
test-integration-docker-compatibility: support Docker/Moby 23


AkihiroSuda:
`test-integration-docker-compatibility` was working for https://github.com/actions/runner-images/blob/ubuntu22/20230724.1/images/linux/Ubuntu2204-Readme.md (Docker-Moby Server 20.10.25+azure-2) but now failing for https://github.com/actions/runner-images/blob/ubuntu22/20230728.3/images/linux/Ubuntu2204-Readme.md (Docker-Moby Server 23.0.6+azure-2)

https://github.com/containerd/nerdctl/actions/runs/5767158960/job/15636390969?pr=2417

```
=== NAME  TestBuildWithBuildArg/InvalidBuildArgCausesError
    testing.go:1471: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestBuildWithBuildArg (4.91s)
    --- PASS: TestBuildWithBuildArg/ArgValueOverridesDefault (0.75s)
    --- PASS: TestBuildWithBuildArg/EmptyArgValueOverridesDefault (0.76s)
    --- PASS: TestBuildWithBuildArg/UnsetArgKeyPreservesDefault (0.71s)
    --- PASS: TestBuildWithBuildArg/EnvValueOverridesDefault (0.73s)
    --- PASS: TestBuildWithBuildArg/EmptyEnvValueOverridesDefault (0.72s)
    --- FAIL: TestBuildWithBuildArg/InvalidBuildArgCausesError (0.28s)
```

```
=== RUN   TestSystemPrune
    system_prune_linux_test.go:58: assertion failed: error is not nil: expected not to contain "nerdctl-testsystemprune", got "DRIVER    VOLUME NAME\nlocal     nerdctl-testsystemprune\n": DRIVER    VOLUME NAME
        local     nerdctl-testsystemprune
        
--- FAIL: TestSystemPrune (1.71s)
```

```
=== RUN   TestVolumePrune
    volume_prune_linux_test.go:37: assertion failed: 
        Command:  /usr/bin/docker volume prune -f
        ExitCode: 0
        Stdout:   Total reclaimed space: 0B
        
        Stderr:   
        
        Failures:
        Expected stdout to contain "nerdctl-testvolumeprune-2"
--- FAIL: TestVolumePrune (1.25s)
```

```
=== RUN   TestComposeUpDetailedError
    compose_up_test.go:54: assertion failed: 
[...]

PR #2433, opened 2023-08-11
Support create anonymous volume 

yankay:
Add a feature to create an anonymous volume like docker.

When investigating the https://github.com/containerd/nerdctl/issues/2421,  the `docker system prune` behavior is updated by the https://github.com/moby/moby/pull/44216.
So the  create anonymous volume should be support first :-)

to Fix: https://github.com/containerd/nerdctl/issues/2432


AkihiroSuda (approved):
Thanks

PR #2434, opened 2023-08-11
Remove dependency on github.com/hashicorp/go-multierror

AkihiroSuda:


fahedouch (approved):
Thanks

djdongjin:
One behavior difference between `multierror.Append` and `errors.Join` is that the former will append to an existing multierror if it is (so the new one has a length of N + 1), while the latter will create a new `joinError` which includes previous error and the new one (so the new one has a length of 2)

https://go.dev/play/p/ryeimUucDyb

I feel the expected use pattern of `errors.Join` is to have a `errs [] error` to save all the errors along the way and a `errors.Join(errs...)` at the end to produce one `joinError`.

https://github.com/golang/go/issues/60209

https://github.com/golang/go/issues/53435#issuecomment-1189495745

AkihiroSuda:
@djdongjin Thanks for finding this, opened:
- https://github.com/containerd/nerdctl/pull/2443

PR #2447, opened 2023-08-23
makefile fixes

metux:
Several makefile fixes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ VERSION_TRIMMED := $(VERSION:v%=%)
 REVISION ?= $(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
 
 GO_BUILD_LDFLAGS ?= -s -w
+GO_BUILD_FLAGS ?=
 export GO_BUILD=GO111MODULE=on CGO_ENABLED=0 GOOS=$(GOOS) $(GO) build -ldflags "$(GO_BUILD_LDFLAGS) -X $(PACKAGE)/pkg/version.Version=$(VERSION) -X $(PACKAGE)/pkg/version.Revision=$(REVISION)"
 
 ifdef VERBOSE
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -61,6 +67,7 @@ install:
 	install -D -m 755 $(CURDIR)/_output/nerdctl $(DESTDIR)$(BINDIR)/nerdctl
 	install -D -m 755 $(CURDIR)/extras/rootless/containerd-rootless.sh $(DESTDIR)$(BINDIR)/containerd-rootless.sh
 	install -D -m 755 $(CURDIR)/extras/rootless/containerd-rootless-setuptool.sh $(DESTDIR)$(BINDIR)/containerd-rootless-setuptool.sh
+	install -D -m 644 -t $(DESTDIR)$(DOCDIR)/nerdctl docs/*.md
```
fahedouch, line 70:
```suggestion
	install -D -m 644  $(CURDIR)/docs/*.md  $(DESTDIR)$(DOCDIR)/nerdctl
```

Makefile
```diff
@@ -19,19 +19,25 @@
...
+BINDIR  ?= $(PREFIX)/bin
+DATADIR ?= $(PREFIX)/share
+DOCDIR  ?= $(DATADIR)/doc
 
 VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
 VERSION_TRIMMED := $(VERSION:v%=%)
 REVISION ?= $(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
 
 GO_BUILD_LDFLAGS ?= -s -w
+GO_BUILD_FLAGS ?=
```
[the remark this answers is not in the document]
metux: It's meant to be passed by the caller - just here for reference.

Makefile
metux: Nice though, but doesn't work that way: the command won't create the "nerdctl" dir ('-D' creates only the leading path, but explicitly leaving off the last component). That's why I'm using '-t', which does create that dir.

metux:
Looks like the node where "test / test-integration-rootless (20.04, v1.7.3, test-integration-rootless)" is running got a problem - IPFS test gets orphan containers, thus can't remove the image.

Issue #1945, opened 2023-01-28
`nerdctl run -v` fails with bind mount if the host folder doesn't exist


djdongjin:
### Description

```shell
# folder not exist
➜  compenv ls /tmp/nerdctl/non-exist-folder
ls: cannot access /tmp/nerdctl/non-exist-folder: No such file or directory
# nerdctl run fails
➜  compenv sudo nerdctl run -v=/tmp/nerdctl/non-exist-folder:/var/lib/random-folder alpine
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/tmp/nerdctl/non-exist-folder" to rootfs at "/var/lib/random-folder": stat /tmp/nerdctl/non-exist-folder: no such file or directory: unknown
# create the folder
➜  compenv mkdir -p /tmp/nerdctl/non-exist-folder
# nerdctl run succeeds
➜  compenv sudo nerdctl run -v=/tmp/nerdctl/non-exist-folder:/var/lib/random-folder alpine
```

### Describe the results you received and expected

nerdctl creates the folder in host if not exist, and nerdctl run succeeds with expected bind mount.

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

AkihiroSuda:
Fun fact: While `-v /tmp/nerdctl/non-exist-folder:/var/lib/random-folder` must create the src dir, `--mount type=bind,src=/tmp/nerdctl/non-exist-folder,target=/var/lib/random-folder` must NOT create the src dir

PR #2461, opened 2023-08-29
fix: fix nerdctl run -v when host folder does not exist for bind mounts

vsiravar:
Fixes: https://github.com/containerd/nerdctl/issues/1945 and improves [compatibility](https://docs.docker.com/storage/bind-mounts/#differences-between--v-and---mount-behavior) with docker.


AkihiroSuda (approved):
Thanks

PR #2467, opened 2023-09-02
feat: support push with soci

ningziwen:
## High level approach

The current approach integrates both `soci create` and `soci push` to `nerdctl push` when soci is specified in `nerdctl push`. The change is based on the assumption that we want to integrate whole soci workflow to nerdctl seamlessly without letting users run any soci commands by themselves.

One alternative is to integrate `soci create` to `nerdctl build` and `soci push` to `nerdctl push`. `nerdctl build` makes sense to be independent with `nerdctl push` because the image can ran or exported locally without pushing later, but soci index is for lazy pull, which doesn't make sense without pushing to a registry later. The soci getting-started [guide](https://github.com/awslabs/soci-snapshotter/blob/main/docs/getting-started.md) also puts `soci create` after pushing image to registry. So didn't choose this approach.

## SOCI flags

When calling `soci` commands in `nerdctl push`. There are many flags in `soci create` and `soci push` to be set properly. There are several possible ways of setting them:

- If Nerdctl has the same flag and it is reasonable to pass through, pass throught the flags from Nerdctl input to soci.
- If the flag is reasonable to expose to Nerdctl users but not available in Nerdctl, add new flags to Nerdctl prefixed with `soci`. (Similar to the [style](https://github.com/containerd/nerdctl/blob/593b94f977d802db43801baccc148e7bed499e95/docs/command-reference.md?plain=1#L365) in Cosign)
- Hardcode to specific values in soci commands.
- Keep the flags as default in soci commands if we don't see obvious benefit of exposing them to Nerdctl users. It is a two-way-door to expose them later.

|  SOCI flag             |  SOCI command      | Recommended Value |
|--------------------------|-----------------|-------------------|
| **--address**                | global          | Pass through from Nerdctl global flag --address                  |
| **--namespace**              | global          | Pass through from Nerdctl global flag --namespace        |
| --timeout                | global          |  Keep default as no timeout          |
| --debug                  | global          |    Keep default as false, which is consistent with the pattern of other dependencies such as Cosign               |
| --content-store          | global          |   Keep default. The content store is anyways managed by SOCI. [Discussion](https://github.com/awslabs/soci-snapshotter/issues/423#issuecomment-1720368676)                |
| **--all-platforms**          | create and push | Pass through from Nerdctl push flag --all-platforms                  |
| **--platform**               | create and push | Pass through from Nerdctl push flag --platform                  |
| **--span-size**              | create          |  Add new Nerdctl push flag --soci-span-size to expose it                 |
| **--min-layer-size**         | create          | Add new Nerdctl push flag --soci-min-layer-size to expose it                  |
| **--skip-verify**       | push            |        Pass through from Nerdctl push flag --insecure-registry           |
| **--plain-http**       | push            |         Pass through from Nerdctl push flag --insecure-registry (from this [discussion](https://github.com/containerd/containerd/issues/4826#issuecomment-742280578), both are needed)          |
| --user      | push            |    Keep default as empty. Soci could and should share nerdctl's credentials               |
| --refresh     | push            |   Keep default as empty. Soci could and should share nerdctl's credentials                |
| **--hosts-dir**    | push            |   Pass through from Nerdctl config "hosts_dir" (TBD. Seems unsupported. To be confirmed. https://github.com/awslabs/soci-snapshotter/discussions/839)             |
| --tlscacert    | push            |  Keep default. The config has too low level to be exposed in Nerdctl                      |
| --tlscert      | push            |    Keep default. The config has too low level to be exposed in Nerdctl                    |
| --tlskey       | push            |    Keep default. The config has too low level to be exposed in Nerdctl                     |
| --http-dump      | push            |  Keep default. The config has too low level to be exposed in Nerdctl                        |
| --http-trace       | push            |   Keep default. The config has too low level to be exposed in Nerdctl                     |
| --label          | push            | Keep default. SOCI index should not need special labels and any labels in Nerdctl should not apply to index                  |
| --snapshotter     | push            |  Keep default. The user input snapshotter in Nerdctl is soci. SOCI index should use the default one instead of using SOCI snapshotter again                 |
| --existing-index         | push            |  Keep default as "warn". Can expose later if needed.                |
| --max-concurrent-uploads | push            | Keep default as "10". Can expose later if needed.                   |
| --quiet                  | push            |  Keep default as false. No matter whether `quiet` is true in soci, Nerdctl quiet should already suppress anything |

### DIFF
--- a/cmd/nerdctl/image_push_linux_test.go
+++ b/cmd/nerdctl/image_push_linux_test.go
@@ -174,3 +174,19 @@ func TestPushNonDistributableArtifacts(t *testing.T) {
 	}
 	assert.Equal(t, resp.StatusCode, http.StatusOK, "non-distributable blob should be available")
 }
+
+func TestPushSoci(t *testing.T) {
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	requiresSoci(base)
+	reg := testregistry.NewPlainHTTP(base, 5000)
+	defer reg.Cleanup()
+
+	base.Cmd("pull", testutil.UbuntuImage).AssertOK()
+	testImageRef := fmt.Sprintf("%s:%d/%s:%s",
+		reg.IP.String(), reg.ListenPort, testutil.Identifier(t), strings.Split(testutil.UbuntuImage, ":")[1])
+	t.Logf("testImageRef=%q", testImageRef)
+	base.Cmd("tag", testutil.UbuntuImage, testImageRef).AssertOK()
+
+	base.Cmd("--snapshotter=soci", "--insecure-registry", "push", "--soci-span-size=2097152", "--soci-min-layer-size=20971520", testImageRef).AssertOK()
+}
--- a/pkg/snapshotterutil/sociutil.go
+++ b/pkg/snapshotterutil/sociutil.go
@@ -1,0 +1,166 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package snapshotterutil
+
+import (
+	"bufio"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/containerd/nerdctl/pkg/api/types"
+	"github.com/sirupsen/logrus"
+)
+
+// CreateSoci creates a SOCI index(`rawRef`)
+func CreateSoci(rawRef string, gOpts types.GlobalCommandOptions, allPlatform bool, platforms []string, sOpts types.SociOptions) error {
+	sociExecutable, err := exec.LookPath("soci")
+	if err != nil {
+		logrus.WithError(err).Error("soci executable not found in path $PATH")
+		logrus.Info("you might consider installing soci from: https://github.com/awslabs/soci-snapshotter/blob/main/docs/install.md")
+		return err
+	}
+
+	sociCmd := exec.Command(sociExecutable)
+	sociCmd.Env = os.Environ()
+
+	// #region for global flags.
+	if gOpts.Address != "" {
+		sociCmd.Args = append(sociCmd.Args, "--address", gOpts.Address)
+	}
+	if gOpts.Namespace != "" {
+		sociCmd.Args = append(sociCmd.Args, "--namespace", gOpts.Namespace)
+	}
+	// #endregion
+
+	// Global flags have to be put before subcommand before soci upgrades to urfave v3.
+	// https://github.com/urfave/cli/issues/1113
+	sociCmd.Args = append(sociCmd.Args, "create")
+
+	if allPlatform {
+		sociCmd.Args = append(sociCmd.Args, "--all-platforms", strconv.FormatBool(allPlatform))
+	}
+	if len(platforms) > 0 {
+		sociCmd.Args = append(sociCmd.Args, "--platform")
+		sociCmd.Args = append(sociCmd.Args, strings.Join(platforms, ","))
+	}
+	if sOpts.SpanSize != -1 {
+		sociCmd.Args = append(sociCmd.Args, "--span-size", strconv.FormatInt(sOpts.SpanSize, 10))
+	}
+	if sOpts.MinLayerSize != -1 {
+		sociCmd.Args = append(sociCmd.Args, "--min-layer-size", strconv.FormatInt(sOpts.MinLayerSize, 10))
+	}
+	// --timeout, --debug, --content-store
+	sociCmd.Args = append(sociCmd.Args, rawRef)
+
+	logrus.Debugf("running %s %v", sociExecutable, sociCmd.Args)
+
+	err = processSociIO(sociCmd)
+	if err != nil {
+		return err
+	}
+
+	return sociCmd.Wait()
+}
+
+// PushSoci pushes a SOCI index(`rawRef`)
+// `hostsDirs` are used to resolve image `rawRef`
+func PushSoci(rawRef string, gOpts types.GlobalCommandOptions, allPlatform bool, platforms []string) error {
+	logrus.Debugf("pushing SOCI index: %s", rawRef)
+
+	sociExecutable, err := exec.LookPath("soci")
+	if err != nil {
+		logrus.WithError(err).Error("soci executable not found in path $PATH")
+		logrus.Info("you might consider installing soci from: https://github.com/awslabs/soci-snapshotter/blob/main/docs/install.md")
+		return err
+	}
+
+	sociCmd := exec.Command(sociExecutable)
+	sociCmd.Env = os.Environ()
+
+	// #region for global flags.
+	if gOpts.Address != "" {
+		sociCmd.Args = append(sociCmd.Args, "--address", gOpts.Address)
+	}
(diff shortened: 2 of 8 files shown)


pkg/snapshotterutil/sociutil.go
```diff
@@ -0,0 +1,125 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/containerd/nerdctl/pkg/imgutil"
+	"github.com/sirupsen/logrus"
+)
+
+// CreateSoci creates a SOCI index(`rawRef`)
+func CreateSoci(rawRef string) error {
+	sociExecutable, err := exec.LookPath("soci")
```
Kern--, line 32: Is shelling out to the SOCI CLI a design goal or is it just because the CLI has non-trivial logic?

I would think using SOCI as a library here would be better long term.

pkg/snapshotterutil/sociutil.go
```diff
@@ -0,0 +1,125 @@
...
+// PushSoci pushes a SOCI index(`rawRef`)
+// `hostsDirs` are used to resolve image `rawRef`
+func PushSoci(ctx context.Context, rawRef string, hostsDirs []string) error {
+	digest, err := imgutil.ResolveDigest(ctx, rawRef, false, hostsDirs)
+	if err != nil {
+		logrus.WithError(err).Errorf("unable to resolve digest for an image %s: %v", rawRef, err)
+		return err
+	}
+	ref := rawRef
+	if !strings.Contains(ref, "@") {
```
Kern--, line 63: Is there a use-case where this is true? The current call site always contains a specific digest reference, right?

pkg/snapshotterutil/sociutil.go
ningziwen: If I understand correctly, the pros of using SOCI library include:

- Does not require SOCI CLI to be installed in Nerdctl env
- Cleaner code to call library instead of calling commands

However, there is already a common practice of calling commands of dependencies in Nerdctl, such as Buildkit, Cosign, Notation, and `soci-snapshotter-grpc` is anyways needed to be pre-configured as part of SOCI provisioning. So the two pros don't seem significant.

Here are the reasons of leaning to using SOCI CLI instead of library for now.

- SOCI CLI has well-defined documentation and the backward compatibility is respected by default. The library code doesn't have documentation and it is not clear about how much backward compatibility it will have. It may also be difficult for SOCI maintainers to promise and define that.
- SOCI CLI layer has some business logic which can call multiple methods of library code, which is tedious to duplicate in Nerdctl. 

For both above, some refactoring in SOCI similar to [what Nerdctl is doing](https://github.com/containerd/nerdctl/issues/1680) could help (with that, the library code can include all the business logic and can be clearly defined and promised backward compatibility) but may not deserve.

ningziwen:
@AkihiroSuda @djdongjin May I get basic concensus of the high level approach (See PR description) before polishing the implementation and adding tests?

AkihiroSuda:
> @AkihiroSuda @djdongjin May I get basic concensus of the high level approach (See PR description) before polishing the implementation and adding tests?

👍  on exec-ing the `soci` binary.
We have been doing the same for cosign, notation, IPFS, buildg, etc., to avoid inflating `go.mod` dependencies.

pkg/snapshotterutil/sociutil.go
ningziwen: You are right. Fixed.

pkg/snapshotterutil/sociutil.go
```diff
@@ -0,0 +1,108 @@
...
+// CreateSoci creates a SOCI index(`rawRef`)
+func CreateSoci(rawRef string) error {
+	sociExecutable, err := exec.LookPath("soci")
+	if err != nil {
+		logrus.WithError(err).Error("soci executable not found in path $PATH")
+		logrus.Info("you might consider installing soci from: https://github.com/awslabs/soci-snapshotter/blob/main/docs/install.md")
+		return err
+	}
+
+	sociCmd := exec.Command(sociExecutable, []string{"create"}...)
```
Kern--, line 36: You should also pass `--namespace`, `--address` (containerd address), and maybe `--timeout`.

cmd/nerdctl/image_push_linux_test.go
```diff
@@ -174,3 +176,69 @@ func TestPushNonDistributableArtifacts(t *testing.T) {
 	}
 	assert.Equal(t, resp.StatusCode, http.StatusOK, "non-distributable blob should be available")
 }
+
+func TestPushSoci(t *testing.T) {
```
ginglis13, line 180: test LGTM. maybe could deduplicate some of https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_soci_linux_test.go

pkg/snapshotterutil/sociutil.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if len(platforms) > 0 {
+		sociCmd.Args = append(sociCmd.Args, "--platform")
+		sociCmd.Args = append(sociCmd.Args, platforms...)
+	}
+	if gOpts.Namespace != "" {
+		sociCmd.Args = append(sociCmd.Args, "--namespace", gOpts.Namespace)
+	}
+	if sOpts.SpanSize != -1 {
+		sociCmd.Args = append(sociCmd.Args, "--span-size", strconv.FormatInt(sOpts.SpanSize, 10))
```
ginglis13, line 65: Looks like values are in bytes in soci cli. See https://github.com/awslabs/soci-snapshotter/blob/f556b45e8e241ff243776e15352a6b7bd38f2161/soci/soci_index.go#L57-L58
```go
defaultSpanSize     = int64(1 << 22) // 4MiB
defaultMinLayerSize = 10 << 20       // 10MiB
```

Maybe room for improved UX to specify value with `mib`, `m`, `kib`,  `k` suffix and do the conversion?

pkg/snapshotterutil/sociutil.go
```diff
@@ -0,0 +1,161 @@
...
+	stdout, err := sociCmd.StdoutPipe()
+	if err != nil {
+		logrus.Warn("soci: " + err.Error())
+	}
+	stderr, err := sociCmd.StderrPipe()
+	if err != nil {
+		logrus.Warn("soci: " + err.Error())
+	}
+	if err := sociCmd.Start(); err != nil {
+		// only return err if it's critical (soci start failed.)
```
ginglis13, line 140: nit:
```suggestion
		// only return err if it's critical (soci command failed to start.)
```

pkg/snapshotterutil/sociutil.go
ningziwen: Yeah. Maybe this can contribute to soci CLI later.

pkg/testutil/testutil_linux.go
```diff
@@ -60,6 +60,7 @@ var (
 const (
 	FedoraESGZImage = "ghcr.io/stargz-containers/fedora:30-esgz"            // eStargz
 	FfmpegSociImage = "public.ecr.aws/soci-workshop-examples/ffmpeg:latest" // SOCI
+	UbuntuImage     = "public.ecr.aws/docker/library/ubuntu:latest"         // Large enough for testing soci index creation
```
AkihiroSuda, line 63: Please pin the version

AkihiroSuda (approved):
Thanks, LGTM if CI is green

AkihiroSuda:
Failing
```
=== RUN   TestPushSoci
    testregistry_linux.go:49: hostIP="10.4.0.1", listenIP="0.0.0.0", listenPort=5000
    image_push_linux_test.go:188: testImageRef="10.4.0.1:5000/nerdctl-testpushsoci:23.10"
    image_push_linux_test.go:191: assertion failed: res.ExitCode is not exitCode: time="2023-09-18T10:26:25Z" level=fatal msg="invalid argument \"1<<21\" for \"--soci-span-size\" flag: strconv.ParseInt: parsing \"1<<21\": invalid syntax"
        
--- FAIL: TestPushSoci (1.50s)
FAIL cmd/nerdctl.TestPushSoci (re-run 2) (1.50s)
```

ginglis13 (approved):

AkihiroSuda (approved):

PR #2488, opened 2023-09-12
remove useless option from containerd-rootless-setuptool

fahedouch:
`containerd-rootless-setuptool.sh install -f` throws an error when passed with `-f`


AkihiroSuda (approved):
Thanks

PR #2514, opened 2023-09-18
nerdctl version: skip inspecting rootless daemon if the daemon is not running

### LINKED ISSUE #2455: nerdctl version requires rootless containerd
### Description

I don't know if this is intended, but `nerdctl version` requires rootless containerd to be running, is that the expected behavior?:

```bash
~ ❯ nerdctl --version
nerdctl version 1.5.0
~ ❯ nerdctl version
FATA[0000] rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): stat /run/user/1000/containerd-rootless: no such file or directory

```

1. Run `nerdctl version`, without rootless containerd

I would expect that `nerdctl version` shows the version of whatever it is I'm running

1.5.0

None

### Host information

Surprisingly, `nerctl info` also requires rootless containerd:

```bash
~   nerdctl info
FATA[0000] rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): stat /run/user/1000/containerd-rootless: no such file or directory
```
FWIW, I'm using Arch Linux, but even building from the tip of main, the error persists.

AkihiroSuda:
Fix #2455

### DIFF
--- a/cmd/nerdctl/version.go
+++ b/cmd/nerdctl/version.go
@@ -23,11 +23,12 @@ import (
 	"os"
 	"text/template"
 
-	"github.com/containerd/nerdctl/pkg/api/types"
 	"github.com/containerd/nerdctl/pkg/clientutil"
 	"github.com/containerd/nerdctl/pkg/formatter"
 	"github.com/containerd/nerdctl/pkg/infoutil"
 	"github.com/containerd/nerdctl/pkg/inspecttypes/dockercompat"
+	"github.com/containerd/nerdctl/pkg/rootlessutil"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
@@ -66,7 +67,17 @@ func versionAction(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	v, vErr := versionInfo(cmd, globalOptions)
+	address := globalOptions.Address
+	// rootless `nerdctl version` runs in the host namespaces, so the address is different
+	if rootlessutil.IsRootless() {
+		address, err = rootlessutil.RootlessContainredSockAddress()
+		if err != nil {
+			logrus.WithError(err).Warning("failed to inspect the rootless containerd socket address")
+			address = ""
+		}
+	}
+
+	v, vErr := versionInfo(cmd, globalOptions.Namespace, address)
 	if tmpl != nil {
 		var b bytes.Buffer
 		if err := tmpl.Execute(&b, v); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/version.go
```diff
@@ -66,7 +67,17 @@ func versionAction(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	v, vErr := versionInfo(cmd, globalOptions)
+	address := globalOptions.Address
+	// rootless `nerdctl version` runs in the host namespaces, so the address is different
```
fahedouch, line 71:
```suggestion
	// rootless `nerdctl version` runs in the host namespaces (parent), so the address is different
```

cmd/nerdctl/version.go
```diff
@@ -66,7 +67,17 @@ func versionAction(cmd *cobra.Command, args []string) error {
...
+	if rootlessutil.IsRootless() {
+		address, err = rootlessutil.RootlessContainredSockAddress()
+		if err != nil {
+			logrus.WithError(err).Warning("failed to inspect the rootless containerd socket address")
+			address = ""
```
fahedouch, line 76: `address = ""` is useless

cmd/nerdctl/version.go
```diff
@@ -102,13 +113,16 @@ func versionAction(cmd *cobra.Command, args []string) error {
 	return vErr
 }
 
-// versionInfo may return partial VersionInfo on error
-func versionInfo(cmd *cobra.Command, globalOptions types.GlobalCommandOptions) (dockercompat.VersionInfo, error) {
-
+// versionInfo may return partial VersionInfo on error.
+// Address can be empty to skip inspecting the server.
```
fahedouch, line 117: Nice to have; Address can be a ptr type to avoid allocating usless memory space when empty

cmd/nerdctl/version.go
AkihiroSuda: "host namespaces" should be already enough

cmd/nerdctl/version.go
AkihiroSuda: The overhead is negligible

cmd/nerdctl/version.go
AkihiroSuda: This is added because a future version of `RootlessContainredSockAddress` may potentially return a non-empty address with a non-nil err.

fahedouch: Why ?

fahedouch (approved):
LGTM

ktock (approved):

Issue #2520, opened 2023-09-21
`nerdctl network ls --filter ` is missing


yankay:
### What is the problem you're trying to solve

When using nerdctl to create kind cluster, there needs the `--filter` option.

```
root@kay201:~/oss/nerdctl# kind create cluster --config /root/kind.yml --image release-ci.daocloud.io/docker.m.daocloud.io/kindest/node:v${KUBE_VERSION}
Creating cluster "kind" ...
ERROR: failed to create cluster: failed to ensure docker network: command "docker network ls '--filter=name=^kind$' '--format={{.ID}}'" failed with error: exit status 1
Command Output: time="2023-09-21T15:20:01Z" level=fatal msg="unknown flag: --filter"
```

### Describe the solution you'd like

enhance the feature like docker.

PR #2522, opened 2023-09-21
Add network `--filter` option

yankay:
Implement compose cp command. docker doc: https://docs.docker.com/engine/reference/commandline/network_ls/

fixes : https://github.com/containerd/nerdctl/issues/2520


AkihiroSuda (approved):
Thanks

PR #2531, opened 2023-09-25
fix: NewNetworkingOptionsManager is modified to take a client

testwill:


pkg/containerutil/container_network_manager.go
```diff
@@ -260,7 +260,11 @@ func (m *containerNetworkManager) getNetworkingContainerForArgument(ctx context.
 	if err != nil {
 		return nil, err
 	}
-	defer cancel()
+	defer func(){
+		cancel()
+		client.Close()
+	}()
+
```
AkihiroSuda, line 267: NewNetworkingOptionsManager should be modified to take a client, then no need to create and close another client here

AkihiroSuda:
Please squash commits

AkihiroSuda (approved):
Thanks, I'm merging this, but in the next time please clean up the commit message.

PR #2541, opened 2023-09-28
CNI: prioritize `/opt/cni/bin` over `/usr/lib/cni`

AkihiroSuda:
When the both of `/opt/cni/bin` and `/usr/lib/cni` are present, the latter one is likely to be a distro package and likely to be older than `/opt/cni/bin`.


fahedouch (approved):
LGTM

Issue #2523, opened 2023-09-21
Replace logrus with `github.com/containerd/log`


AkihiroSuda:
Direct dependencies on `github.com/sirupsen/logrus` and `github.com/containerd/containerd/log` have to be replaced with https://github.com/containerd/log

TinaMor:
@AkihiroSuda, I'd like to take up this task since it's marked as a good first issue.

djdongjin:
@TinaMor please go ahead :)

PR #2543, opened 2023-10-03
[Chore] 2523: Use https://github.com/containerd/log for logging

TinaMor:
# PR Description

This PR replaces direct dependencies on github.com/sirupsen/logrus and github.com/containerd/containerd/log with https://github.com/containerd/log to streamline our logging dependencies.

We continue to use logrus.SetOutput() in [ocihook.go](https://github.com/TinaMor/nerdctl/blob/replace-logrus-package/pkg/ocihook/ocihook.go) because not all Logrus APIs are provided by the new library.

Github issues #2523

### DIFF
--- a/cmd/nerdctl/builder.go
+++ b/cmd/nerdctl/builder.go
@@ -22,9 +22,9 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/pkg/buildkitutil"
 	"github.com/containerd/nerdctl/pkg/defaults"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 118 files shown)


.gitignore
```diff
@@ -8,3 +8,7 @@ _output
 # vagrant
 /.vagrant
 Vagrantfile
+
+# IDEs and editors
+/.vs
+/.vscode
```
AkihiroSuda, line 14: This shouldn't be committed here

AkihiroSuda: ping @TinaMor 

Also, please squash commits

aznashwan:
> We continue to use logrus.SetOutput() in [ocihook.go](https://github.com/TinaMor/nerdctl/blob/replace-logrus-package/pkg/ocihook/ocihook.go) because not all Logrus APIs are provided by the new library.

@TinaMor while I did not personally test this myself, it looks like in the current implementation of the [`containerd/log.L.Logger` struct field delegates to `logrust.Entry`](https://github.com/containerd/log/blob/main/context.go#L71), which has a [public `Logger` field](https://pkg.go.dev/github.com/sirupsen/logrus#Entry) which we could directly call [`SetOutput()`](https://pkg.go.dev/github.com/sirupsen/logrus#Logger.SetOutput) on as follows

```go
logger := log.GetLogger(log.L.Context)
logger.Logger.SetOutput(io.MultiWriter(stderr, logFile))
```

Considering that the current implementation of `containerd/log` simply delegates to the now-maintenance-mode `logrus` anyway, I don't believe it's that big of a deal either way in its current state.

profnandaa (approved):
LGTM. Also means, should the project want to switch from `logrus` to something else with the same API, it will just be changing in one place.

yankay:
Thanks@TinaMor
/lgtm

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@aznashwan Do you want to open a follow-up PR?

TinaMor:
@AkihiroSuda  I'm looking into @aznashwan's suggestion. Will do a followup PR

PR #2586, opened 2023-10-20
docs: minor soci push fix

ollypom:
Removing the duplicate `push` in the soci docs. 

(Looks like my linter caught the table formatting too)


AkihiroSuda (approved):
Thanks

Issue #2521, opened 2023-09-21
`nerdctl network create --ipv6` is missing


yankay:
### What is the problem you're trying to solve

When using nerdctl to create kind cluster, there needs the ----ipv6 option.
```
Creating cluster "kind" ...
ERROR: failed to create cluster: failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1
Command Output: time="2023-09-21T17:14:26Z" level=fatal msg="unknown flag: --ipv6"
```

### Describe the solution you'd like

enhance the feature like docker.

yankay:
relate https://github.com/containerd/nerdctl/issues/1547

AkihiroSuda:
> relate #1547

Let me close as a duplicate

PR #2592, opened 2023-10-25
Support create network with `com.docker.network.bridge.enable_ip_masquerade` option

yankay:
Support create network with `com.docker.network.bridge.enable_ip_masquerade` option, like:

```
nerdctl network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true kind
```

The docker docs: https://docs.docker.com/engine/reference/commandline/network_create/#bridge-driver-options

related to: https://github.com/containerd/nerdctl/issues/2521


AkihiroSuda (approved):
Thanks

Issue #2593, opened 2023-10-25
`nerdctl ps --format {{.Names}}` behaviour is inconsistent with docker 


yankay:
### Description

The `docker ps --format`  is inconsistent with docker , the brackets is not appeared in the docker.

### Steps to reproduce the issue

1. nerdctl create --label io.x-k8s.kind.cluster=kind  busybox
2. nerdctl ps -a --filter label=io.x-k8s.kind.cluster=kind --format {{.Names}}
[busybox-0235b]

### Describe the results you received and expected

`busybox-0235b` not `[busybox-0235b]`

### What version of nerdctl are you using?

nerdctl version
Client:
 Version:       v1.6.2-29-g35966845
 OS/Arch:       linux/amd64
 Git commit:    359668452f1019d752487f629f5e2e356d830240
 buildctl:
  Version:      v0.12.1
  GitCommit:    bb857a0d49f45aa0ce9cd554b78d4075553e20f9

Server:
 containerd:
  Version:      v1.7.1
  GitCommit:    1677a17964311325ed1c31e2c0a3589ce6d5c30d
 runc:
  Version:      1.1.7
  GitCommit:    v1.1.7-0-g860f061b

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Others

PR #2594, opened 2023-10-25
Fix  `nerdctl ps --format {{.Names}} ` behaviour

yankay:
The `docker ps --format`  is inconsistent with docker , the brackets is not appeared in the docker.

fix: https://github.com/containerd/nerdctl/issues/2593


AkihiroSuda (approved):
Thanks

PR #2632, opened 2023-11-07
[build][iidfile] Set the same permissions as docker

amrmahdi:
Docker makes the file readable by other users, nerdctl makes only readable by the user created the file, which breaks compatibility.

See https://github.com/docker/buildx/blob/ff8bca206b389bc09e8e0a9afb5bf21f3d51686d/commands/build.go#L297C1-L298C1


splusq (approved):

fahedouch (approved):
Thanks

Issue #2637, opened 2023-11-14
Add --soci-index-digest to nerdctl pull


sondavidb:
### What is the problem you're trying to solve

`soci image rpull` allows a user to specify a specific SOCI index digest. nerdctl ignores this flag as it's not necessary, but in some edge cases it might be useful to specify this. For example if multiple indexes have been made, the referrers API will grab the most recently created one when that might not be the one needed. Another example might be wanting to run containers with SOCI but not having write access to a repository. Both of these cases are rare, but it would be nice to cover them regardless.

### Describe the solution you'd like

A new flag, `--soci-index-digest`, that would mimic the `--soci-index-digest` flag in `soci image rpull`.

sondavidb:
Implemented in #2638

PR #2638, opened 2023-11-14
Add SOCI index digest flag

sondavidb:
Addresses #2637.

`soci image rpull` and the SOCI library both allow for a particular SOCI index to be specified, so this change adds support for passing this in.

Since no existing structure was in place to pass flags to remote snapshotters, this change also introduces the structures needed to support such future changes. I did my best to ensure that, should any supported remote snapshotters need any additional flags, the changes needed to support such changes would be minimal, but please let me know if this does not scale properly.

I also added another test case to TestPullSoci with a garbage SOCI index to ensure it behaves the same as `soci image rpull`. I attempted to make the testing suite run on my machine, but couldn't get it to work, so apologies if that test case fails.

### DIFF
--- a/cmd/nerdctl/image_pull.go
+++ b/cmd/nerdctl/image_pull.go
@@ -20,6 +20,7 @@ import (
 	"github.com/containerd/nerdctl/pkg/api/types"
 	"github.com/containerd/nerdctl/pkg/clientutil"
 	"github.com/containerd/nerdctl/pkg/cmd/image"
+	"github.com/containerd/nerdctl/pkg/imgutil"
 	"github.com/spf13/cobra"
 )
 
@@ -56,6 +57,10 @@ func newPullCommand() *cobra.Command {
 	pullCommand.Flags().String("cosign-certificate-oidc-issuer-regexp", "", "A regular expression alternative to --certificate-oidc-issuer for --verify=cosign,. Accepts the Go regular expression syntax described at https://golang.org/s/re2syntax. Either --cosign-certificate-oidc-issuer or --cosign-certificate-oidc-issuer-regexp must be set for keyless flows")
 	// #endregion
 
+	// #region socipull flags
+	pullCommand.Flags().String("soci-index-digest", "", "Specify a particular index digest for SOCI. If left empty, SOCI will automatically use the index determined by the selection policy.")
+	// #endregion
+
 	pullCommand.Flags().BoolP("quiet", "q", false, "Suppress verbose output")
 
 	pullCommand.Flags().String("ipfs-address", "", "multiaddr of IPFS API (default uses $IPFS_PATH env variable if defined or local directory ~/.ipfs)")
@@ -81,6 +86,7 @@ func processPullCommandFlags(cmd *cobra.Command) (types.ImagePullOptions, error)
 	if err != nil {
 		return types.ImagePullOptions{}, err
 	}
+
 	quiet, err := cmd.Flags().GetBool("quiet")
 	if err != nil {
 		return types.ImagePullOptions{}, err
@@ -89,6 +95,12 @@ func processPullCommandFlags(cmd *cobra.Command) (types.ImagePullOptions, error)
 	if err != nil {
 		return types.ImagePullOptions{}, err
 	}
+
+	sociIndexDigest, err := cmd.Flags().GetString("soci-index-digest")
+	if err != nil {
+		return types.ImagePullOptions{}, err
+	}
+
 	verifyOptions, err := processImageVerifyOptions(cmd)
 	if err != nil {
 		return types.ImagePullOptions{}, err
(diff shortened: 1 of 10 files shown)


cmd/nerdctl/image_pull.go
```diff
@@ -56,6 +56,10 @@ func newPullCommand() *cobra.Command {
 	pullCommand.Flags().String("cosign-certificate-oidc-issuer-regexp", "", "A regular expression alternative to --certificate-oidc-issuer for --verify=cosign,. Accepts the Go regular expression syntax described at https://golang.org/s/re2syntax. Either --cosign-certificate-oidc-issuer or --cosign-certificate-oidc-issuer-regexp must be set for keyless flows")
 	// #endregion
 
+	// #region socipull flags
+	pullCommand.Flags().String("soci-index-digest", "", "Specify a particular index digest for SOCI. If left empty, SOCI will automatically use the latest index generated from the pulled registry.")
```
ktock, line 60: Please add this to the usage docs.

pkg/cmd/compose/compose.go
```diff
@@ -125,7 +125,7 @@ func New(client *containerd.Client, globalOptions types.GlobalCommandOptions, op
 				ipfsPath = dir
 			}
 			_, err = ipfs.EnsureImage(ctx, client, stdout, stderr, globalOptions.Snapshotter, scheme, ref,
-				pullMode, ocispecPlatforms, nil, quiet, ipfsPath)
+				pullMode, ocispecPlatforms, nil, quiet, ipfsPath, &imgutil.RemoteSnapshotterFlags{})
```
djdongjin, line 128: should we also put a `RemoteSnapshotterFlags` in `ImagePullOptions`, and make `SociIndexDigest` part of `RemoteSnapshotterFlags`?

pkg/imgutil/pull/pull.go
```diff
@@ -30,6 +30,10 @@ import (
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+type RemoteSnapshotterFlags struct {
+	SociIndexDigest string
+}
```
djdongjin, line 35: can we move this to `pkg/api/types`? I think this is not only related to `pull` (i.e., we can also use these as part of `run` command right`)?

sondavidb: Thanks. This makes a lot more sense and adds a couple of imports but makes the code much easier to follow.

> (i.e., we can also use these as part of run command right)

For this case specifically, I don't think so. Technically, since nerdctl attempts to pull images it doesn't have locally, it could be passed in for that purpose, and at that point is there a reason to not just pull first?

But other remote snapshotters can certainly leverage this in the future, so I think it's still a good place for it to reside.

sondavidb (commented):
Fixed an issue with the apply function that should hopefully make the tests pass

sondavidb:
Just rebased from main

pkg/imgutil/pull/pull.go
sondavidb: I could move it to types, but I still had to keep this in. By importing the types package everywhere I needed it, I ended up creating a circular dependency when importing it into `imgutil.go` due to `network_types.go` in types importing the `netutil` package, which somehow created a circular dependency.

sondavidb:
Looking through other tests very quickly, some of the tests seem flaky. However, I see that `TestImageEncryptJWE` is a new test failing that I haven't seen elsewhere, and I think that it is causing the rest of the tests to fail in [20.04 1.7.8](https://github.com/containerd/nerdctl/actions/runs/6871592567/job/18688688831?pr=2638#step:6:387). I'm trying to figure out if this is normal or if I shouldn't worry about it.

I also am not really sure how else to approach my idea here. Redeclaring a struct essentially 3 times seems poor but unsure if there's an easier solution after another run-through.

pkg/api/types/image_types.go
```diff
@@ -193,6 +198,8 @@ type ImagePullOptions struct {
 	Quiet bool
 	// multiaddr of IPFS API (default uses $IPFS_PATH env variable if defined or local directory ~/.ipfs)
 	IPFSAddress string
+	// Flags to pass into remote snapshotters
+	RFlags *RemoteSnapshotterFlags
```
djdongjin, line 202: I don't think we need a pointer here.

djdongjin: I think in general we don't need pointer for these flag types (they're not that large)

pkg/imgutil/pull/pull.go
djdongjin: If there is circular dependency, it's okay to put `RemoteSnapshotterFlags` in `imgutil` pkg directly. And in `pkg/api/types` you can just use as `imgutil.RemoteSnapshotterFlags`.

sondavidb: The issue is that pull.go cannot import imgutil without creating another circular dependency. Perhaps I can have it live in snapshotter.go somehow? I'm not sure exactly how that might look, but if I do that it might solve this issue since it shares the same package. What do you think?

sondavidb:
I ended up mutating the function headers in snapshotter.go so that you can just pass a RemoteSnapshotterFlags object into `apply()`. The new RemoteSnapshotterFlags now lives in the imgutil package in snapshotter.go This solves the issue of having to redeclare the struct multiple times. PTAL @djdongjin

I also reran the testing suite from main branch here and I'm still seeing test failures similar to #2327, but I'd like a sanity check on this.

ktock (approved):
LGTM

AkihiroSuda (approved):
Thanks

Issue #2639, opened 2023-11-14
`TestComposeExecTTY` has to be compatible with Docker


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/171d761effdce70ef1d65cb6d42b33c8bd37ca5c/cmd/nerdctl/compose_exec_linux_test.go#L165-L167

https://github.com/containerd/nerdctl/pull/2623#discussion_r1382352885

PR #2650, opened 2023-11-21
Fix incompatible in `TestComposeExecTTY`

yankay:
fix: https://github.com/containerd/nerdctl/issues/2639
using similar code in https://github.com/docker/compose/blob/v2.23.1/cmd/compose/exec.go#L77

### DIFF
--- a/cmd/nerdctl/compose_exec.go
+++ b/cmd/nerdctl/compose_exec.go
@@ -36,9 +36,7 @@ func newComposeExecCommand() *cobra.Command {
 	}
 	composeExecCommand.Flags().SetInterspersed(false)
 
-	composeExecCommand.Flags().BoolP("tty", "t", true, "Allocate a pseudo-TTY") // missing in Docker Compose v2?
 	composeExecCommand.Flags().BoolP("no-TTY", "T", false, "Disable pseudo-TTY allocation. By default nerdctl compose exec allocates a TTY.")
-	composeExecCommand.Flags().BoolP("interactive", "i", true, "Keep STDIN open even if not attached")
 	composeExecCommand.Flags().BoolP("detach", "d", false, "Detached mode: Run containers in the background")
 	composeExecCommand.Flags().StringP("workdir", "w", "", "Working directory inside the container")
 	// env needs to be StringArray, not StringSlice, to prevent "FOO=foo1,foo2" from being split to {"FOO=foo1", "foo2"}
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/compose_exec.go
```diff
@@ -46,17 +46,8 @@ func newComposeExecCommand() *cobra.Command {
...
-		if noTTY, _ := flags.GetBool("no-TTY"); noTTY {
-			if err := flags.Set("tty", "false"); err != nil {
-				return err
-			}
-		}
-		return nil
-	}
-
+	composeExecCommand.Flags().MarkHidden("interactive")
+	composeExecCommand.Flags().MarkHidden("tty")
```
AkihiroSuda, line 50: I guess we can just remove `--tty`

cmd/nerdctl/compose_exec.go
```diff
@@ -46,17 +46,8 @@ func newComposeExecCommand() *cobra.Command {
...
-		flags := cmd.Flags()
-		if noTTY, _ := flags.GetBool("no-TTY"); noTTY {
-			if err := flags.Set("tty", "false"); err != nil {
-				return err
-			}
-		}
-		return nil
-	}
-
+	composeExecCommand.Flags().MarkHidden("interactive")
```
AkihiroSuda, line 49: This doesn't seem to exist in Docker Compose, so we can just remove it

cmd/nerdctl/compose_exec_linux_test.go
```diff
@@ -165,7 +165,6 @@ services:
 func TestComposeExecTTY(t *testing.T) {
 	// `-i` in `compose run & exec` is only supported in compose v2.
 	// Currently CI is using compose v1.
```
AkihiroSuda, line 167: Docker Compose v2 doesn't seem to have `-i`

AkihiroSuda, line 167: " Currently CI is using compose v1" is no longer true

cmd/nerdctl/compose_exec.go
```diff
@@ -69,7 +60,7 @@ func composeExecAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	tty, err := cmd.Flags().GetBool("tty")
+	noTty, err := cmd.Flags().GetBool("no-TTY")
```
AkihiroSuda, line 63: Shouldn't we still check `--tty` too?

It is hidden in Docker Compose v2, but it still exists
https://github.com/docker/compose/blob/b1a26dac1d3dc201d0d026cc16cb9baadaacffd7/cmd/compose/exec.go#L77

yankay: HI @AkihiroSuda 

Thanks for the PR review.

When using `docker compose exec -t=true --no-TTY`, the docker do not check the `--tty`.

```
root@kay200:/tmp# docker compose -f a.yaml exec --tty=true --no-TTY  svc0 stty
stty: standard input: Not a tty
```

So it be removed to keep compatibility :-)

cmd/nerdctl/compose_exec_linux_test.go
yankay: Thanks , it has changed. :-)

cmd/nerdctl/compose_exec.go
AkihiroSuda: 👀 Could you add a code comment to explain that behavior?

yankay: Thanks 
There are comments on it :-)

AkihiroSuda (approved):
Thanks

PR #2661, opened 2023-11-27
Allow for binding a named pipe on Windows

### LINKED ISSUE #2199: nerdctl run does not allow for binding a named pipe
### Description

I'm trying to bind the named pipe for the containerd socket to a container on Windows using the following command:

```
nerdctl run -it --rm -v \\.\pipe\containerd-containerd:\\.\pipe\containerd-containerd mydockerhub/my-image my-container
```

This yields the following error:

```
time="2023-04-24T07:10:13Z" level=fatal msg="error while parsing \"\\\\\\\\.\\\\pipe\\\\containerd-containerd:\\\\\\\\.\\\\pipe\\\\containerd-containerd\": failed to get volume \"\\\\\\\\.\\\\pipe\\\\containerd-containerd\": malformed name \\\\.\\pipe\\containerd-containerd: identifier \"\\\\\\\\.\\\\pipe\\\\containerd-containerd\" must match ^[A-Za-z0-9]+(?:[._-](?:[A-Za-z0-9]+))*$: invalid argument"
```

When running the image on a machine with Docker using the Docker CLI, this kind of command works, only using `\\.\pipe\docker_engine` for the name of the pipe:

```
docker run -it --rm -v \\.\pipe\docker_engine:\\.\pipe\docker_engine mydockerhub/my-image my-container
```

Clearly the regular expression used in nerdctl to check the volume name does not account for a named pipe. Is this an oversight or is it intentional for some reason?

1. Run the command:
```
nerdctl run --rm -v \\.\pipe\containerd-containerd:\\.\pipe\containerd-containerd hello-world
```

The container should run successfully and be bound to the containerd named pipe.

1.3.1

None

### Host information

_No response_

TinaMor:
# PR Description

This PR fixes:
- #759 to allow users to mount volumes on Windows
- #2199 to allow users to bind a named pipe on Windows

Also, adds a check in [create.go#L298](https://github.com/containerd/nerdctl/blob/93b90990df9a5871354e7dcfde60ebd528290db6/pkg/cmd/container/create.go#L298) to log a warning only when there is an error

### DIFF
--- a/cmd/nerdctl/container_run_mount_linux_test.go
+++ b/cmd/nerdctl/container_run_mount_linux_test.go
@@ -85,8 +85,22 @@ func TestRunVolume(t *testing.T) {
 func TestRunAnonymousVolume(t *testing.T) {
 	t.Parallel()
 	base := testutil.NewBase(t)
-	base.Cmd("run", "--rm", "-v", "/foo", testutil.AlpineImage,
-		"mountpoint", "-q", "/foo").AssertOK()
+	base.Cmd("run", "--rm", "-v", "/foo", testutil.AlpineImage).AssertOK()
+	base.Cmd("run", "--rm", "-v", "TestVolume2:/foo", testutil.AlpineImage).AssertOK()
+	base.Cmd("run", "--rm", "-v", "TestVolume", testutil.AlpineImage).AssertOK()
+
+	// Destination must be an absolute path not named volume
+	base.Cmd("run", "--rm", "-v", "TestVolume2:TestVolumes", testutil.AlpineImage).AssertFail()
+}
+
+func TestRunVolumeRelativePath(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+	base.Cmd("run", "--rm", "-v", "./foo:/mnt/foo", testutil.AlpineImage).AssertOK()
+	base.Cmd("run", "--rm", "-v", "./foo", testutil.AlpineImage).AssertOK()
+
+	// Destination must be an absolute path not a relative path
+	base.Cmd("run", "--rm", "-v", "./foo:./foo", testutil.AlpineImage).AssertFail()
 }
 
 func TestRunAnonymousVolumeWithTypeMountFlag(t *testing.T) {
--- a/cmd/nerdctl/container_run_mount_windows_test.go
+++ b/cmd/nerdctl/container_run_mount_windows_test.go
@@ -1,0 +1,214 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/containerd/nerdctl/pkg/inspecttypes/dockercompat"
+	"github.com/containerd/nerdctl/pkg/testutil"
+	"gotest.tools/v3/assert"
+)
+
+func TestRunMountVolume(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+	tID := testutil.Identifier(t)
+	rwDir, err := os.MkdirTemp(t.TempDir(), "rw")
+	if err != nil {
+		t.Fatal(err)
+	}
+	roDir, err := os.MkdirTemp(t.TempDir(), "ro")
+	if err != nil {
+		t.Fatal(err)
+	}
+	rwVolName := tID + "-rw"
+	roVolName := tID + "-ro"
+	for _, v := range []string{rwVolName, roVolName} {
+		defer base.Cmd("volume", "rm", "-f", v).Run()
+		base.Cmd("volume", "create", v).AssertOK()
+	}
+
+	containerName := tID
+	defer base.Cmd("rm", "-f", containerName).AssertOK()
+	base.Cmd("run",
+		"-d",
+		"--name", containerName,
+		"-v", fmt.Sprintf("%s:C:/mnt1", rwDir),
+		"-v", fmt.Sprintf("%s:C:/mnt2:ro", roDir),
+		"-v", fmt.Sprintf("%s:C:/mnt3", rwVolName),
+		"-v", fmt.Sprintf("%s:C:/mnt4:ro", roVolName),
+		testutil.CommonImage,
+		"ping localhost -t",
+	).AssertOK()
+
+	base.Cmd("exec", containerName, "cmd", "/c", "echo -n str1 > C:/mnt1/file1").AssertOK()
+	base.Cmd("exec", containerName, "cmd", "/c", "echo -n str2 > C:/mnt2/file2").AssertFail()
+	base.Cmd("exec", containerName, "cmd", "/c", "echo -n str3 > C:/mnt3/file3").AssertOK()
+	base.Cmd("exec", containerName, "cmd", "/c", "echo -n str4 > C:/mnt4/file4").AssertFail()
+	base.Cmd("rm", "-f", containerName).AssertOK()
+
+	base.Cmd("run",
+		"--rm",
+		"-v", fmt.Sprintf("%s:C:/mnt1", rwDir),
+		"-v", fmt.Sprintf("%s:C:/mnt3", rwVolName),
+		testutil.CommonImage,
+		"cat", "C:/mnt1/file1", "C:/mnt3/file3",
+	).AssertOutContainsAll("str1", "str3")
+	base.Cmd("run",
+		"--rm",
+		"-v", fmt.Sprintf("%s:C:/mnt3/mnt1", rwDir),
+		"-v", fmt.Sprintf("%s:C:/mnt3", rwVolName),
+		testutil.CommonImage,
+		"cat", "C:/mnt3/mnt1/file1", "C:/mnt3/file3",
+	).AssertOutContainsAll("str1", "str3")
+}
+
+func TestRunMountVolumeInspect(t *testing.T) {
+	base := testutil.NewBase(t)
+	testContainer := testutil.Identifier(t)
+	testVolume := testutil.Identifier(t)
+
+	defer base.Cmd("volume", "rm", "-f", testVolume).Run()
+	base.Cmd("volume", "create", testVolume).AssertOK()
+	inspectVolume := base.InspectVolume(testVolume)
+	namedVolumeSource := inspectVolume.Mountpoint
+
+	base.Cmd(
+		"run", "-d", "--name", testContainer,
+		"-v", "C:/mnt1",
+		"-v", "C:/mnt2:C:/mnt2",
+		"-v", "\\\\.\\pipe\\containerd-containerd:\\\\.\\pipe\\containerd-containerd",
+		"-v", fmt.Sprintf("%s:C:/mnt3", testVolume),
+		testutil.CommonImage,
+	).AssertOK()
+
+	inspect := base.InspectContainer(testContainer)
+	// convert array to map to get by key of Destination
+	actual := make(map[string]dockercompat.MountPoint)
+	for i := range inspect.Mounts {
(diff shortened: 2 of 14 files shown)


cmd/nerdctl/container_run_mount_windows_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestRunNamedPipeVolume(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+	base.Cmd("run", "--rm", "-v", `\\.\pipe\containerd-containerd`, testutil.CommonImage).AssertOK()
+}
+
+func TestInvalidPath(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+	base.Cmd("run", "--rm", "-v", `InvalidPathC:\TestVolume:C:\Mount`, testutil.CommonImage).AssertFail()
```
aznashwan, line 125: Please consider adding more potentially invalid cases such as:

* no colons in source/target/neither
* multiple colons being present
* anything other formatting issue which may cross your mind

cmd/nerdctl/container_run_mount_windows_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
```
aznashwan, line 126: Depending on how you decide to treat `~` on Windows, please also add a test specifically to test its expansion into the home path. (or lack thereof if we decide `~` should be an error on Windows.

aznashwan (changes requested):
First review pass done, mostly LGTM but I suggested some added checks or minor changes to the newly-added helper functions.

pkg/mountutil/mountutil.go
```diff
@@ -49,90 +53,60 @@ type Processed struct {
...
-		var err error
-		var specOpts []oci.SpecOpts
-		options, specOpts, err = parseVolumeOptions(res.Type, src, rawOpts)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse volume options (%q, %q, %q): %w", res.Type, src, rawOpts, err)
-		}
-		res.Opts = append(res.Opts, specOpts...)
-	default:
+	if len(split) < 1 || len(split) > 3 {
 		return nil, fmt.Errorf("failed to parse %q", s)
```
aznashwan, line 70: Please give this a more descriptive error message like:

`fmt.Errorf("volume arg %q has was split into an incorrect number of elements (expected 1-3, got %d): %v", s, len(split), split)`

pkg/mountutil/mountutil.go
```diff
@@ -49,90 +53,60 @@ type Processed struct {
...
+
+	rawOpts := res.Mode
+	options, spectOpts, err := getVolumeOptions(src, res.Type, rawOpts)
+	if err != nil {
+		return nil, err
+	}
+	res.Opts = spectOpts
+
+	fstype := DefaultMountType
 	if runtime.GOOS != "freebsd" {
```
aznashwan, line 109: We should ideally strive to have ZERO platform-dependant code in the main `mountutil.go` file, so could you please move this logic into a  platform-specific function like `ensureBindOption()`?

pkg/mountutil/mountutil.go
TinaMor: This method is used to parse --mount and --tmpfs flags which have different formats. The [nerdctl command reference](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-blue_square-nerdctl-run) clearly documents the format of each input

pkg/mountutil/mountutil.go
TinaMor: While striving for ZERO platform-dependant code is ideal, in some cases it may not apply. The code works for both Windows and Linux and creating a file for non-freebsd code with a single function does not add any value. Should there be future requirements for a non-freebsd platform, then we could create the file then

pkg/mountutil/mountutil.go
```diff
@@ -161,5 +135,129 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore, createDir bool) (*
...
+	// assume src is a volume name
+	if err := identifiers.Validate(source); err == nil {
+		if isAnonymousVol {
+			return handleAnonymousVolumes(source, volStore)
+		}
+		return handleNamedVolumes(source, volStore)
+	}
+
+	// Handle named pipes
+	if isNamedPipe(source) {
```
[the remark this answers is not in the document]
TinaMor: The implementation of this method on non-windows platform returns false

pkg/mountutil/mountutil.go
```diff
@@ -49,90 +52,84 @@ type Processed struct {
...
 	)
 
-	s = strings.TrimLeft(s, ":")
-	split := strings.Split(s, ":")
+	split, err := splitVolumeSpec(s)
+	if err != nil {
+		return nil, fmt.Errorf("failed to split volume mount specification %q: %v", s, err)
+	}
+
+	if len(split) < 1 || len(split) > 3 {
```
profnandaa, line 75: nit: `<1` for `len` is `0` , I think `len(split) == 0` is more explicit.

pkg/mountutil/mountutil.go
```diff
@@ -161,5 +158,150 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore, createDir bool) (*
...
+	res.Source = source
+
+	// Handle named volumes
+	if isNamedVolume(source) {
+		return handleNamedVolumes(source, volStore)
+	}
+
+	// Handle named pipes
+	if isNamedPipe(source) {
+		res.Type = Npipe
```
profnandaa, line 178: nit: this is not necessarily for this PR, but I feel like these types could have been `enum`s?

pkg/mountutil/mountutil.go
```diff
@@ -161,5 +158,150 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore, createDir bool) (*
...
+		return res, nil
+	}
+
+	// Handle relative paths
+	if !filepath.IsAbs(source) {
+		log.L.Warnf("expected an absolute path, got a relative path %q (allowed for nerdctl, but disallowed for Docker, so unrecommended)", source)
```
profnandaa, line 184: suggestion: the warning is quite long, perhaps excluding the explanation in ()? But that's just preference.

pkg/mountutil/mountutil.go
```diff
@@ -161,5 +158,150 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore, createDir bool) (*
...
+		absPath, err := filepath.Abs(source)
+		if err != nil {
+			return res, fmt.Errorf("failed to get the absolute path of %q: %w", source, err)
+		}
+		res.Source = absPath
+	}
+
+	// Create dir if it does not exist
+	if err := createDirOnHostIfNotExist(source, createDir); err != nil {
```
profnandaa, line 193: nit: perhaps `createDirOnHost` would do? The _if not exist_ part can be implicit IMO.

pkg/mountutil/mountutil.go
```diff
@@ -161,5 +158,150 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore, createDir bool) (*
...
+
+func getDestination(split []string) (string, error) {
+	var dst string
+
+	if len(split) == 1 {
+		dst = split[0]
+
+		// For anonymous named volumes, the dst value is set to the created volume for the source
+		if isNamedVolume(dst) {
+			return dst, nil
```
profnandaa, line 248: nit: from this logic ,once `len(split)` is `1`, doesn't matter if `isNamedVolume(dst)`, it will always return `dst, nil` either at L248 or L259, unless L255 happens. Any way of re-arranging this logic; perhaps starting off with error catching L255, etc?

profnandaa (commented):
Dropped in comments just for sanity checks. I'm not fully conversant with the project, so didn't give any functional related feedback.

Good work!

pkg/mountutil/mountutil.go
```diff
@@ -161,5 +158,150 @@ func ProcessFlagV(s string, volStore volumestore.VolumeStore, createDir bool) (*
...
+			log.L.Warnf("mount source %q does not exist. Please make sure to create the directory on the host.", src)
+			return nil
+		}
+		return fmt.Errorf("failed to stat %q: %w", src, err)
+	}
+
+	if !os.IsNotExist(err) {
+		return fmt.Errorf("failed to stat %q: %w", src, err)
+	}
+	if err := os.MkdirAll(src, 0o755); err != nil {
```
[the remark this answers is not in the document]
TinaMor: Just moved this part of the code. But read and execute permission on the dir for users other than the owner seems ok.

@AkihiroSuda, should we change the permission?

pkg/mountutil/mountutil.go
TinaMor: Will create a separate PR for this.

cmd/nerdctl/container_run_mount_linux_test.go
```diff
@@ -85,15 +85,28 @@ func TestRunVolume(t *testing.T) {
...
+
+	// Destination must be an absolute path not a relative path
+	base.Cmd("run", "--rm", "-v", "./foo:./foo", testutil.AlpineImage).AssertFail()
 }
 
 func TestRunAnonymousVolumeWithTypeMountFlag(t *testing.T) {
 	t.Parallel()
 	base := testutil.NewBase(t)
-	base.Cmd("run", "--rm", "--mount", "type=volume,dst=/foo", testutil.AlpineImage,
-		"mountpoint", "-q", "/foo").AssertOK()
```
jsturtevant, line 96: this should still be included to verify that the folder point is a mount

cmd/nerdctl/container_run_mount_linux_test.go
```diff
@@ -113,8 +126,7 @@ VOLUME /foo
 	defer os.RemoveAll(buildCtx)
 
 	base.Cmd("build", "-t", imageName, buildCtx).AssertOK()
-	base.Cmd("run", "--rm", "-v", "/foo", testutil.AlpineImage,
-		"mountpoint", "-q", "/foo").AssertOK()
```
jsturtevant, line 117: same here

pkg/mountutil/mountutil_nonwindows.go
jsturtevant, line 1: should this file name be `mountutil_others.go`?

pkg/mountutil/mountutil_windows.go
```diff
@@ -68,3 +76,39 @@ func ProcessFlagTmpfs(s string) (*Processed, error) {
 func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, error) {
 	return nil, errdefs.ErrNotImplemented
 }
+
+func splitVolumeSpec(s string) ([]string, error) {
+	s = strings.TrimPrefix(s, "npipe:")
+	s = strings.TrimLeft(s, ":")
+	volSpec := strings.ReplaceAll(s, "/", pathSeparator)
+	pattern, err := regexp.Compile(`(\w*:\\)?(\\*[\w\_\-\s\.\,\~]+\\*)*`)
```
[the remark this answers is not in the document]
jsturtevant: It would be nice to see some tests around this

jsturtevant (commented):
Thanks for taking this on!  Looks good over all, a few minor comments. I think this also resolves #759 and would replace #924.

pkg/mountutil/mountutil_windows.go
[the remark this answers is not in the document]
TinaMor: @jsturtevant Refactored the regex borrowing from [docker windows_parser.go#L53](https://github.com/moby/moby/blob/master/volume/mounts/windows_parser.go#L53)

cmd/nerdctl/container_run_mount_linux_test.go
```diff
@@ -85,8 +85,22 @@ func TestRunVolume(t *testing.T) {
 func TestRunAnonymousVolume(t *testing.T) {
 	t.Parallel()
 	base := testutil.NewBase(t)
-	base.Cmd("run", "--rm", "-v", "/foo", testutil.AlpineImage,
-		"mountpoint", "-q", "/foo").AssertOK()
```
jsturtevant, line 89: can we include the mountpoint -q here?

pkg/mountutil/mountutil_windows.go
```diff
@@ -14,10 +16,19 @@
    limitations under the License.
 */
 
+/*
+   Portions from https://github.com/moby/moby/blob/master/volume/mounts/windows_parser.go#L53
```
jsturtevant, line 20: Oh I think this is the attribution, can you use a tagged/branch version instead of the `master` branch?

jsturtevant (commented):
couple minor comments but overall LGTM

AkihiroSuda:
Needs rebase

AkihiroSuda (approved):
Thanks

PR #2668, opened 2023-11-29
Do not enter rootless mode when socket is accessible

koct9i:
Containerd socket could be simply set group-accessible.

Signed-off-by: Konstantin Khlebnikov <koct9i@gmail.com>


AkihiroSuda:
Just accessing the socket is not enough, as it needs to access `/var/lib`, etc.

The right way to use rootful nerdctl via a non-root user is to use sudo.

Alternatively you may also use setuid binary
https://github.com/containerd/nerdctl/blob/v1.7.0/docs/faq.md#does-nerdctl-have-an-equivalent-of-sudo-usermod--ag-docker-user-

PR #2682, opened 2023-12-05
nerdctl run: reset the default value of limits.

jiusanzhou:
With the oci spec, we have the default rlimits RLIMIT_NOFILE=1024:1024, If we start a pod in kubelet, this value we be reset to nil by cri. We use the nerdctl to start containers in production, this difference will cause user confused.

When run a container if don't get any ulimit opts, reset the rlimits to nil. This will remove the default 1024 file limit.
```go
if len(ulimitOpts) == 0 {
	ulimitOpts = append(ulimitOpts, withRlimits(nil))
}
```

Warning: this PR will make `nerdctl run` behave differently than before.

### DIFF
--- a/pkg/cmd/container/run_linux.go
+++ b/pkg/cmd/container/run_linux.go
@@ -95,6 +95,13 @@ func setPlatformOptions(ctx context.Context, client *containerd.Client, id, uts 
 	if err != nil {
 		return nil, err
 	}
+
+	// If without any ulimitOpts, we need to reset the default value from spec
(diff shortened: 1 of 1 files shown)


yankay (approved):
Thanks @jiusanzhou 

I've invesgated, the behavior is also compatible Docker.

AkihiroSuda (approved):
Thanks, I'm merging this, but please consider adding tests to prevent regression in future

PR #2715, opened 2023-12-20
Remove "soci image rpull" from testing suite

sondavidb:
Since SOCI plans on removing this functionality from the CLI, we can preemptively remove our test's dependency on this component of the CLI by hardcoding the expected number of mounts per image. This is fine because the same image with the same SOCI config will always have the same amount of mounts. The only caveat is that one must know how many mounts to expect from a certain image.

Additionally, small refactoring was done to allow for more future test coverage (e.g. testing different images with SOCI), and to allow TestSociRun to be as flexible as TestSociPull. (Though the tests are so similar, I wonder if it may be better to just group the two.)

### DIFF
--- a/cmd/nerdctl/container_run_soci_linux_test.go
+++ b/cmd/nerdctl/container_run_soci_linux_test.go
@@ -17,7 +17,6 @@
 package main
 
 import (
-	"os"
 	"os/exec"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


ktock (approved):

cmd/nerdctl/container_run_soci_linux_test.go
```diff
@@ -27,60 +26,52 @@ import (
...
-	expectedMounts, err := exec.Command("mount").Output()
-	if err != nil {
-		t.Fatal(err)
-	}
+			rmiOutput := base.Cmd("rmi", testutil.FfmpegSociImage).Out()
+			base.T.Logf("rmi output: %s", rmiOutput)
 
-	remoteSnapshotsExpectedCount := strings.Count(string(expectedMounts), "fuse.rawBridge")
-	base.T.Logf("number of expected mounts: %v", remoteSnapshotsExpectedCount)
+			remoteSnapshotsExpectedCount := 11
```
djdongjin, line 67: this should be `tt.remoteSnapshotsExpectedCount` right?

(also since `tt` is used within `t.Run`, it will be safer to add a line `tt := tt` between line 41 and 42)

sondavidb: Oh, sorry. I initially did not make a struct for this and hardcoded in the values instead. Fixed this.

> (also since tt is used within t.Run, it will be safer to add a line tt := tt between line 41 and 42)

From my understanding of this Go semantics intricacy [here](https://github.com/golang/go/discussions/56010), it seems this is only a problem when making references to any of the loop variables outside of the context of this loop. Given this, I think it would be better to not have an extra line of code. But, if I'm wrong, or if you feel particularly strongly about this, I'm happy to add that line.

djdongjin: > this is only a problem when making references to any of the loop variables outside of the context of this loop

My understanding is it will be a problem  if any goroutines created within the for-loop body use the loop variables. For example, if we parallel these `t.Run` sub tests, it will be a problem. Here since you're not parallel these sub tests, it should be okay.

djdongjin (approved):
LGTM thanks

Issue #2716, opened 2023-12-22
The `nerdctl run` cannot run correctly when using the gpu-operator (`FATA[0000] exec: "nvidia-container-cli": executable file not found in $PATH`)


yankay:
### Description

In a cluster with the [gpu-operator](https://github.com/NVIDIA/gpu-operator) .
The cuda pod can start correctly , but the `nerdctl run` cannot run correctly

### Steps to reproduce the issue

1. Using the  [gpu-operator](https://github.com/NVIDIA/gpu-operator)  to install GPU drivers. https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/getting-started.html
2. nerdctl run --gpus=1 nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda11.7.1-ubuntu20.04
```
FATA[0000] exec: "nvidia-container-cli": executable file not found in $PATH
```

### Describe the results you received and expected

The nerdctl shim cannot start. And it can be fixed by the steps:
```
export PATH=$PATH:/usr/local/nvidia/toolkit
echo "/run/nvidia/driver/usr/lib/x86_64-linux-gnu" > /etc/ld.so.conf.d/nvidia.conf
ldconfig
```

And then it can be run correctly:
```
root@kay113-gpu:~# nerdctl run --gpus=1 nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda11.7.1-ubuntu20.04
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
```

### What version of nerdctl are you using?

nerdctl  version
```
WARN[0000] unable to determine buildctl version: exec: "buildctl": executable file not found in $PATH
Client:
 Version:       v1.7.1
 OS/Arch:       linux/amd64
 Git commit:    4692b3eaafc5dc3a3883d3b1cc0e70a35e3b5d32
 buildctl:
  Version:

Server:
 containerd:
  Version:      v1.7.11
  GitCommit:    64b8a811b07ba6288238eefc14d898ee0b5b99ba
 runc:
  Version:      1.1.10
  GitCommit:    v1.1.10-0-g18a0cb0f
```

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

AkihiroSuda:
I don't think this is a bug.
[...]

PR #2720, opened 2023-12-29
Add docs for trouble shooting "nvidia gpu operator"

yankay:
The nerdctl run cannot run correctly when using the gpu-operator with error message (FATA[0000] exec: "nvidia-container-cli": executable file not found in $PATH) 

Add docs so that the user can solve the problem.

Related to:  https://github.com/containerd/nerdctl/issues/2716#issuecomment-1868269556

### DIFF
--- a/docs/gpu.md
+++ b/docs/gpu.md
@@ -62,3 +62,24 @@ services:
           - capabilities: ["utility"]
             count: all
 ```
+
+## Trouble Shooting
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

panpan0000:
see the same issue, thanks for documents.

PR #2740, opened 2024-01-10
fix: return error if volume create failed

zjumoon01:
If processing volume create command options failed, return error.


AkihiroSuda (approved):
Thanks

yankay:
Thanks a lot 
/lgtm

yankay (approved):

Issue #2755, opened 2024-01-16
docs/gpu.md: `docker.io/nvidia/cuda:9.0-base: not found`


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/v1.7.2/docs/gpu.md#options-for-nerdctl-run---gpus

`nvidia/cuda:9.0-base` image no longer seems to exist:
```console
$ nerdctl run -it --rm --gpus all nvidia/cuda:9.0-base nvidia-smi
docker.io/nvidia/cuda:9.0-base: resolving      |--------------------------------------| 
elapsed: 1.1 s                  total:   0.0 B (0.0 B/s)                                         
INFO[0001] trying next host - response was http.StatusNotFound  host=registry-1.docker.io
FATA[0001] failed to resolve reference "docker.io/nvidia/cuda:9.0-base": docker.io/nvidia/cuda:9.0-base: not found
```

The plain ubuntu image still works though
```console
$ nerdctl run -it --rm --gpus all ubuntu nvidia-smi
Tue Jan 16 07:27:01 2024       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.146.02             Driver Version: 535.146.02   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Tesla T4                       Off | 00000000:00:1E.0 Off |                    0 |
| N/A   24C    P8               8W /  70W |      2MiB / 15360MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

cc @ktock

yankay:
The Nvidia CUDA image has been updated at https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda/tags.
[...]

PR #2756, opened 2024-01-16
docs: update gpu cuda image to 12.3.1-base-ubuntu20.04

yankay:
The `nvidia/cuda:9.0-base` image has been deleted. And The Nvidia CUDA image has been updated at https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda/tags.
So update it to `12.3.1-base-ubuntu20.0` like https://docs.docker.com/compose/gpu-support/#example-of-a-compose-file-for-running-a-service-with-access-to-1-gpu-device.

fixes: https://github.com/containerd/nerdctl/issues/2755

### DIFF
--- a/docs/gpu.md
+++ b/docs/gpu.md
@@ -20,7 +20,7 @@ You can specify number of GPUs to use via `--gpus` option.
 The following example exposes all available GPUs.
 
 ```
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

ktock (approved):
Thanks!

PR #2769, opened 2024-01-23
Update soci-snapshotter dependency to 8d2b306690f9484a21c29963c6c5631…

austinvazquez:
Alternative to #2761 until the next SOCI release.

The v0.5.0 github.com/awslabs/soci-snapshotter release added transitive dependency on k8s.io/cri-api.
https://github.com/awslabs/soci-snapshotter/commit/8d2b306690f9484a21c29963c6c56315580258fa was added to allow nerdctl to continue to use soci-snapshotter's handler implementation without taking k8s.io/cri-api dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/Microsoft/go-winio v0.6.1
 	github.com/Microsoft/hcsshim v0.11.4
-	github.com/awslabs/soci-snapshotter v0.4.1
+	github.com/awslabs/soci-snapshotter v0.0.0-20240123154419-8d2b306690f9
 	github.com/compose-spec/compose-go v1.20.2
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/Microsoft/go-winio v0.6.1
 	github.com/Microsoft/hcsshim v0.11.4
-	github.com/awslabs/soci-snapshotter v0.4.1
+	github.com/awslabs/soci-snapshotter v0.0.0-20240123154419-8d2b306690f9
```
AkihiroSuda, line 9: Thanks, can we have a new tag?

austinvazquez: @AkihiroSuda , sure I'll help get the ball started on that on the SOCI side. Can we close this PR and #2761 so dependabot can catch v0.5.1 when it is cut?

Issue #2644, opened 2023-11-16
Create SOCI Indexes when pushing multiple platforms


ollypom:
### What is the problem you're trying to solve

A SOCI index is tied to a container image manifest not a manifest list / image index. In https://github.com/containerd/nerdctl/pull/2467 we added support to generate a SOCI Index on push with `nerdctl push --snapshotter soci` however today this fails if you attempt to push multiple platforms at the same time. 

Current behavior today:

```bash
$ nerdctl image ls
REPOSITORY                                                 TAG       IMAGE ID        CREATED           PLATFORM       SIZE         BLOB SIZE
111222333444.dkr.ecr.eu-west-1.amazonaws.com/nginx-demo    latest    cd6043ce93a0    9 minutes ago     linux/arm64    670.5 MiB    235.1 MiB
111222333444.dkr.ecr.eu-west-1.amazonaws.com/nginx-demo    latest    cd6043ce93a0    9 minutes ago     linux/amd64    0.0 B        241.6 MiB
```

With the `--all-platforms` flag results in a confusing "true" not found error.

```bash
$ nerdctl push --snapshotter soci --all-platforms 111222333444.dkr.ecr.eu-west-1.amazonaws.com/nginx-demo
index-latest@sha256:cd6043ce93a09b7ab1086837945401dbd63ad78da23c50fd2128ce8e80aaf5a1: done           |++++++++++++++++++++++++++++++++++++++|
manifest-sha256:efbf016ba665281867463004baa20cd764f049fb520d1d1d2dcdeb5c01bfb386:     done           |++++++++++++++++++++++++++++++++++++++|
manifest-sha256:bdae03687df2d8f4ecbfef44a51228064c012cc20a29eaf315a1d5ee38fb6716:     done           |++++++++++++++++++++++++++++++++++++++|
config-sha256:d3e6c5969420b426f5a909b2520fe2699c385ef3011f99b3e297a73c7c30c03e:       done           |++++++++++++++++++++++++++++++++++++++|
config-sha256:8207b6dd7ee1e3b31d05b519da6affa53c28e031e636dc1f490476c734f3edaf:       done           |++++++++++++++++++++++++++++++++++++++|
elapsed: 1.4 s                                                                        total:  5.0 Ki (3.6 KiB/s)
INFO[0001] soci: soci: image "true": not found         
```

When specifying specific platforms the regex fails because of the comma.

```bash
[...]

PR #2813, opened 2024-02-09
Fix soci image push with multi-platforms flags

monirul:
Addresses Issue - https://github.com/containerd/nerdctl/issues/2644

This PR resolves command flag issues encountered with the `nerdctl image push --snapshotter soci` command.

The first issue related to the `--all-platforms` flag. Previously, when set, `nerdctl` would erroneously append a `true` value when invoking the `soci` binary (`--all-platforms true`). This led to an error because `soci` interprets `--all-platforms` as a standalone boolean flag. The fix ensures that the flag is used correctly, as shown below:
`soci create --all-platforms <IMAGE>`

The second issue involved the `--platform` flag. `nerdctl` was combining multiple platform values into a single, comma-separated string, contrary to `soci`'s expectation for separate, repeating flags (as it relies on `github.com/urfave/cli`). The PR corrects this by adjusting the flag usage to match `soci`'s requirements:
`soci create --platform amd64 --platform arm64 <IMAGE>`

### DIFF
--- a/pkg/snapshotterutil/sociutil.go
+++ b/pkg/snapshotterutil/sociutil.go
@@ -53,12 +53,16 @@ func CreateSoci(rawRef string, gOpts types.GlobalCommandOptions, allPlatform boo
 	sociCmd.Args = append(sociCmd.Args, "create")
 
 	if allPlatform {
-		sociCmd.Args = append(sociCmd.Args, "--all-platforms", strconv.FormatBool(allPlatform))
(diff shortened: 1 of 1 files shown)


TinaMor:
LGTM. Nice catch on the `--all-platforms true`

ktock (approved):

AkihiroSuda (approved):
Thanks

PR #2823, opened 2024-02-15
doc: should use master tag for binfmt

zwpaper:
the latest tag for binfmt was updated 2 years ago, users should use the master tag which is updated regularly.

### DIFF
--- a/docs/multi-platform.md
+++ b/docs/multi-platform.md
@@ -11,7 +11,7 @@ e.g., ARM on Intel, and vice versa.
 ```console
 $ sudo systemctl start containerd
 
-$ sudo nerdctl run --privileged --rm tonistiigi/binfmt --install all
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
> the latest tag for binfmt was updated 2 years ago, users should use the master tag which is updated regularly.

cc @tonistiigi 
Can we just have a new latest tag?

zwpaper:
hi @AkihiroSuda, from https://github.com/tonistiigi/binfmt/issues/165#issuecomment-2030398232, it looks like the latest is not updated on purpose for historical reasons, what should we do for this one? 

should we document it to use the master tag, or the versioned one? it may not be a good choice to use the legacy tag latest, WDYT?

AkihiroSuda:
Let's merge this, but we may revert this later

PR #2824, opened 2024-02-15
platform: support loongarch64

zwpaper:
binfmt has the loongarch64 support now: https://github.com/tonistiigi/binfmt/blob/51a64e64ef8e523f7ab4d5bbbd376d0ff76fcd3c/cmd/binfmt/config.go#L68


AkihiroSuda (approved):
Thanks

Issue #2810, opened 2024-02-09
Logic for checking isImageSharable creating performance delays due to tarball generation


Shubhranshu153:
### Discussed in https://github.com/containerd/nerdctl/discussions/2808

<div type='discussions-op-text'>

<sup>Originally posted by **Shubhranshu153** February  7, 2024</sup>
Running command:
nerdctl build . --platform <platform name> vs docker build . --platform <platform name> has a different behaviour
With nerdctl it falls back to type docker/oci which leads it to export a tar file and then loading the image, this has a significant performance impact.

Checking the logic of: 
```
executor == "containerd" && containerdUUID == uuid && containerdNamespace == namespace && workerSnapshotter == snapshotter && len(platform) == 0, nil
```

is it possible this is mapping too many conditional for is-shareable?  This is creating some unintended behavior and delays due to tar creation

Logic in docker for setting exporter type to  "image"
```
switch len(opt.Exports) {
	case 1:
		// valid
	case 0:
		if nodeDriver.IsMobyDriver() && !noDefaultLoad() {
			// backwards compat for docker driver only:
			// this ensures the build results in a docker image.
			opt.Exports = []client.ExportEntry{{Type: "image", Attrs: map[string]string{}}}
		}
	default:
		return nil, nil, errors.Errorf("multiple outputs currently unsupported")
	}
```

Thank You</div>

AkihiroSuda:
I don't think the conditions can be reduced ?

Shubhranshu153:
As per the comment it seems the platform option is the driving factor for this. 
```
// NOTE: It's possible that BuildKit doesn't download the base image of non-default platform (e.g. when the provided
    //       Dockerfile doesn't contain instructions require base images like RUN) even if --output type=image,unpack=true
    //       is passed to BuildKit. Thus, we need to use type=docker or type=oci when nerdctl builds non-default platform
    //       image using platform option.
```
isImageShareable it seems if the image build is shareable among runtimes. Not sure the purpose so here is what i am thinking:

1. executor== containerd, if other executor it is not shareable as the build image is specific to containerd. Hence needed
2. containerdUUID == uuid : not sure how this is getting used. if the first one is true then the containedUUID need not match?
3. workerSnapshotter == snapshotter : same question as above.
4. containerdNamespace == namespace: same question as above.
5. len(platform) ==0: This one i think we might want to check if the platform is actually different, if not we can use the type=image.

AkihiroSuda:
[...]

PR #2834, opened 2024-02-20
Enhance default platform comparision for image shareablity #2810

Shubhranshu153:
Partially Addresses: https://github.com/containerd/nerdctl/issues/2810

There are cases where --platform flag may be passed during nerdctl build. In such a situation if platform matches the runtime platform the image is shareable.

### DIFF
--- a/pkg/cmd/builder/build.go
+++ b/pkg/cmd/builder/build.go
@@ -42,6 +42,21 @@ import (
 	"github.com/containerd/platforms"
 )
 
+type PlatformParser interface {
+	Parse(platform string) (platforms.Platform, error)
+	DefaultSpec() platforms.Platform
+}
+
+type platformParser struct{}
+
+func (p platformParser) Parse(platform string) (platforms.Platform, error) {
+	return platforms.Parse(platform)
+}
+
+func (p platformParser) DefaultSpec() platforms.Platform {
+	return platforms.DefaultSpec()
+}
+
 func Build(ctx context.Context, client *containerd.Client, options types.BuilderBuildOptions) error {
 	buildctlBinary, buildctlArgs, needsLoading, metaFile, tags, cleanup, err := generateBuildctlArgs(ctx, client, options)
 	if err != nil {
@@ -419,6 +434,29 @@ func getDigestFromMetaFile(path string) (string, error) {
 	return digest, nil
 }
 
+func isMatchingRuntimePlatform(platform string, parser PlatformParser) bool {
+	p, err := parser.Parse(platform)
+	if err != nil {
+		return false
+	}
+	d := parser.DefaultSpec()
+
+	if p.OS == d.OS && p.Architecture == d.Architecture && (p.Variant == "" || p.Variant == d.Variant) {
+		return true
+	}
+
+	return false
+}
+
+func isBuildPlatformDefault(platform []string, parser PlatformParser) bool {
+	if len(platform) == 0 {
+		return true
+	} else if len(platform) == 1 {
+		return isMatchingRuntimePlatform(platform[0], parser)
+	}
+	return false
+}
+
 func isImageSharable(buildkitHost, namespace, uuid, snapshotter string, platform []string) (bool, error) {
 	labels, err := buildkitutil.GetWorkerLabels(buildkitHost)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/cmd/builder/build.go
```diff
@@ -419,6 +419,29 @@ func getDigestFromMetaFile(path string) (string, error) {
 	return digest, nil
 }
 
+func isMatchingRuntimePlatform(platform string) bool {
+	p, err := platforms.Parse(platform)
+	if err != nil {
+		return false
```
pendo324, line 425: Should we log this error?
```suggestion
                log.L.WithError(err).Errorf("failed to parse platform")
		return false
```

Shubhranshu153: i don't think its important to log here, as the Parse will be called elsewhere and be handled. Here it just important to figure out if the image is shareable, if any thing goes wrong it will attempt to create a non-shareable image.

ktock:
Thanks. can we add a test for ` len(platform) == 1` case?

Shubhranshu153:
@ktock 
was thinking of adding a unit test file for build.go called build_test.go and add unit test for is_shreable() func. I dont see any other place where this can go. Let me know your thoughts.

ktock:
Maybe we can add a test case here: https://github.com/containerd/nerdctl/blob/265d6b9cf526ce7d9ed8d34a0e3c3066901cc463/cmd/nerdctl/builder_build_test.go#L100

Shubhranshu153:
Not sure about the test failures, some of the test seem flaky.

pkg/cmd/builder/build.go
```diff
@@ -42,6 +42,21 @@ import (
 	"github.com/containerd/platforms"
 )
 
+type PlatformParser interface {
```
ktock, line 45: why do we need to introduce a new wrapper form platforms pkg?

ktock:
Test failed on CI: https://github.com/containerd/nerdctl/actions/runs/8026538828/job/21929197370?pr=2834#step:7:13

```
=== RUN   TestBuildIsShareableForCompatiblePlatform
    builder_build_test.go:86: assertion failed: 
        Command:  /usr/bin/docker build /tmp/nerdctl-build-test2104019243 -t nerdctl-testbuildisshareableforcompatibleplatform --platform linux/arm/v7 --progress plain
        ExitCode: 0
        Stdout:   
        Stderr:   #0 building with "default" instance using docker driver
        
        #1 [internal] load .dockerignore
        #1 transferring context: 2B done
        #1 DONE 0.0s
        
        #2 [internal] load build definition from Dockerfile
        #2 transferring dockerfile: 127B done
        #2 DONE 0.0s
        
        #3 [internal] load metadata for ghcr.io/stargz-containers/alpine:3.13-org
        #3 DONE 0.2s
        
        #4 [1/1] FROM ghcr.io/stargz-containers/alpine:3.13-org@sha256:ec14c7992a97fc11425907e908340c6c3d6ff602f5f13d899e6b7027c9b4133a
        #4 CACHED
        
        #5 exporting to image
        #5 exporting layers done
        #5 writing image sha256:6f5021027f826fc1f7353f081bd8ff6364f8db76b5ce718a9007e023faf033ab done
        #5 naming to docker.io/library/nerdctl-testbuildisshareableforcompatibleplatform done
        #5 DONE 0.0s
        
        
        Failures:
        Expected stderr to contain "tarball"
--- FAIL: TestBuildIsShareableForCompatiblePlatform (2.11s)

```

pkg/cmd/builder/build.go
Shubhranshu153: wanted to add a unit test via mocking the platform package, seemed like the preferred way to do a dependency injection to facilitate mocking.

Shubhranshu153:
Fixed the above docker failed test.

ktock (approved):
Thanks

go.mod
```diff
@@ -50,6 +50,7 @@ require (
 	github.com/rootless-containers/rootlesskit/v2 v2.0.1
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.8.4
```
AkihiroSuda, line 53: Can we avoid this dependency?

Shubhranshu153:
The lint error seems to be some sort of caching issue, is it possible to have no-cache option.
integ test not sure why it failed.

pkg/cmd/builder/build_test.go
```diff
@@ -0,0 +1,189 @@
...
+*/
+
+package builder
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/containerd/containerd/platforms"
+	"gotest.tools/v3/assert"
```
AkihiroSuda, line 24:
```
  Running [/home/runner/golangci-lint-1.55.2-linux-amd64/golangci-lint run --out-format=github-actions --verbose] in [] ...
  Error: File is not `gofmt`-ed with `-s` (gofmt)
```

AkihiroSuda (approved):
Thanks

pkg/testutil/testutil.go
```diff
@@ -430,6 +438,15 @@ func (c *Cmd) AssertOutNotContains(s string) {
 	})
 }
 
+func (c *Cmd) AssertErrNotContains(s string) {
```
apostasie, line 441: Hey @Shubhranshu153 

Looks like this is testing stdout (`c.AssertOutWithFunc`) and not stderr.
Can you clarify if the intention here was to test stderr?

PR #2854, opened 2024-03-04
Use json marshal/unmarshal to copy

dmcgowan:
Just call marshal/unmarshal rather depend on a CRI util package. This package is now internal in containerd 2.0.

### DIFF
--- a/cmd/nerdctl/container_update.go
+++ b/cmd/nerdctl/container_update.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"runtime"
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Thanks, but why not just expose DeepCopy ?

dmcgowan:
To avoid creating another util package for a tiny utility. I would also say the `DeepCopy` is poorly named for what it actually does, `CopyWithJSON` doesn't make the function sound nice to use. There are probably much more efficient ways to copy an object and I think it worth trying to prevent all those one-off utils written for CRI from being used everywhere.

AkihiroSuda (approved):

PR #2865, opened 2024-03-08
fix: cleanup volume dir if create volume failed

zjumoon01:
Volume path should be cleaned up if create volume failed.

Otherwise, the volume will be created failed again next time with a 'file exists' error on os.Mkdir.

### DIFF
--- a/pkg/mountutil/volumestore/volumestore.go
+++ b/pkg/mountutil/volumestore/volumestore.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/identifiers"
+	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/native"
 	"github.com/containerd/nerdctl/v2/pkg/lockutil"
 	"github.com/containerd/nerdctl/v2/pkg/strutil"
@@ -83,13 +84,24 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
 	}
 	volPath := filepath.Join(vs.dir, name)
 	volDataPath := filepath.Join(volPath, DataDirName)
-	fn := func() error {
+	volFilePath := filepath.Join(volPath, volumeJSONFileName)
+	fn := func() (err error) {
 		if err := os.Mkdir(volPath, 0700); err != nil {
 			return err
 		}
+		defer func() {
+			if err != nil {
+				os.Remove(volPath)
+			}
+		}()
 		if err := os.Mkdir(volDataPath, 0755); err != nil {
 			return err
 		}
+		defer func() {
+			if err != nil {
+				os.Remove(volDataPath)
+			}
+		}()
 
 		type volumeOpts struct {
 			Labels map[string]string `json:"labels"`
@@ -106,13 +118,24 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
 			return err
 		}
 
-		volFilePath := filepath.Join(volPath, volumeJSONFileName)
+		defer func() {
+			if err != nil {
+				if _, statErr := os.Stat(volFilePath); statErr != nil && !os.IsNotExist(statErr) {
+					log.L.Warnf("failed to stat volume file: %v", statErr)
+					return
+				} else if statErr == nil {
+					os.Remove(volFilePath)
+				}
+			}
+		}()
 		return os.WriteFile(volFilePath, labelsJSON, 0644)
 	}
 
 	if err := lockutil.WithDirLock(vs.dir, fn); err != nil {
 		return nil, err
 	}
+
+	// If other new actions that might fail are added below, we should move the cleanup function out of fn.
 
 	vol := &native.Volume{
 		Name:       name,


AkihiroSuda:
How to test this?

zjumoon01:
> How to test this?

This scene is quite rare. There is no space left on device, and a directory is just deleted. 

step1
#nerdctl volume create v2
FATA[0000] mkdir /var/lib/nerdctl/1935db59/volumes/default/v2/_data: no space left on device

step2
Cleanup my disk and release some space.
Create volume again.
#nerdctl volume create v2
FATA[0000] mkdir /var/lib/nerdctl/1935db59/volumes/default/v2: file exists

Look into /var/lib/nerdctl/1935db59/volumes/default/v2, it's an empty dir

#nerdctl volume ls
FATA[0000] volume "v2" not found: not found

AkihiroSuda:
This case is quite rare, so I expect the user to just run `nerdctl volume rm`

zjumoon01:
> This case is quite rare, so I expect the user to just run `nerdctl volume rm`

Volume is not created successfully. The creation process was only halfway complete.
volume rm also got 'not found' error.

#nerdctl volume rm v2
FATA[0000] volume "v2" not found: not found

In fact, get function is called before remove function.
Get function in volume store only check volume dataPath

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -83,13 +83,23 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
...
 	volPath := filepath.Join(vs.dir, name)
 	volDataPath := filepath.Join(volPath, DataDirName)
-	fn := func() error {
+	fn := func() (err error) {
 		if err := os.Mkdir(volPath, 0700); err != nil {
 			return err
 		}
+		defer func() {
+			if err != nil {
+				os.Remove(volPath)
```
ktock, line 92: > Volume path should be cleaned up if create volume failed.

cleanup doesn't seem to happen for errors outside of this fn?

zjumoon01: cleanup is called in this fn for the two errors, `json.MarshalIndent` and `os.WriteFile`
Creation is done after this fn  is completed. No other new errors outside of it. If some action added after `lockutil.WithDirLock(vs.dir, fn)`  in future, cleanup will be moved out of this fn.

ktock: > cleanup is called in this fn for the two errors, json.MarshalIndent and os.WriteFile

cleanup seems to fail if that dir isn't empty

>  in future, cleanup will be moved out of this fn.

should be documented

zjumoon01: > cleanup seems to fail if that dir isn't empty
dir is empty as expected if writing json file failed. Or use `os.RemoveAll` instead ? 

> should be documented
OK, should I add some comments？

AkihiroSuda: RemoveAll can be dangerous, so I don't prefer it

ktock: > dir is empty as expected if writing json file failed. 

https://pkg.go.dev/os#WriteFile
>  Since WriteFile requires multiple system calls to complete, a failure mid-operation can leave the file in a partially written state.

zjumoon01: How about these steps in cleanup ?
step 1.  check json file exists and remove it
step 2. remove volume data path dir
step 3. remove volume path dir

fahedouch: the path was created on multiple steps so I beleive that cleanup should follow the same pattern and be  cleaned on multiple steps

zjumoon01: > How about these steps in cleanup ?
> step 1. check json file exists and remove it

fixed
Use `stat` to check json file exist
Also add some comments.

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -83,13 +83,29 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
...
+			}
+		}()
 		if err := os.Mkdir(volDataPath, 0755); err != nil {
 			return err
 		}
+		defer func() {
+			if err != nil {
+				// ignore cleanup error
+				if _, serr := os.Stat(volFilePath); serr == nil || !os.IsNotExist(serr) {
```
ktock, line 103: If stat returns error for the file, it should print error and shouldn't continue cleanup.

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -83,13 +83,29 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
...
+				// ignore cleanup error
+				os.Remove(volPath)
+			}
+		}()
 		if err := os.Mkdir(volDataPath, 0755); err != nil {
 			return err
 		}
+		defer func() {
+			if err != nil {
+				// ignore cleanup error
```
ktock, line 102: error shouldn't be ignored. can we print log for this?

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -83,13 +83,29 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
...
 	volDataPath := filepath.Join(volPath, DataDirName)
-	fn := func() error {
+	volFilePath := filepath.Join(volPath, volumeJSONFileName)
+	fn := func() (err error) {
 		if err := os.Mkdir(volPath, 0700); err != nil {
 			return err
 		}
+		defer func() {
+			if err != nil {
+				// ignore cleanup error
```
fahedouch, line 93: IMO, this message is useless

pkg/mountutil/volumestore/volumestore.go
fahedouch: > it should print error and shouldn't continue cleanup.

`volFilePath` does not exist and `volDataPath` exists  => we should continue cleanup

just drop `stat` and log remove error should be fine

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -106,14 +122,15 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
...
 
-		volFilePath := filepath.Join(volPath, volumeJSONFileName)
 		return os.WriteFile(volFilePath, labelsJSON, 0644)
 	}
 
 	if err := lockutil.WithDirLock(vs.dir, fn); err != nil {
 		return nil, err
 	}
 
+	// If other new actions are added below, we should move cleanup function out of fn on failure.
```
fahedouch, line 132:
```suggestion
	// If other new actions that might fail are added below, we should move the cleanup function out of fn.
```

zjumoon01: fixed,  thanks    @fahedouch

pkg/mountutil/volumestore/volumestore.go
zjumoon01: > If stat returns error for the file, it should print error and shouldn't continue cleanup.

Fixed.

pkg/mountutil/volumestore/volumestore.go
zjumoon01: > error shouldn't be ignored. can we print log for this?

I don't print log for cleanup error as some other cleanup functions `defer os.Remove` in the code.
Of course, I can also add some logs for these remove errors

ktock (approved):

ktock (approved):
LGTM on green

zjumoon01:
> LGTM on green

@ktock  Could you please rerun the integration test?   The failure is not likely related with nerdctl/volume

djdongjin:
@zjumoon01 can you rebase to get the latest change in main? The cosign-related CI errors should be fixed now https://github.com/containerd/nerdctl/pull/2896.

zjumoon01:
> @zjumoon01 can you rebase to get the latest change in main? The cosign-related CI errors should be fixed now #2896.

I rebased, but CI failed again caused by 'run container error' in TestVolumeRemove. Do you have any ideas ?

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -83,13 +84,30 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
...
+			if err != nil {
+				os.Remove(volPath)
+			}
+		}()
 		if err := os.Mkdir(volDataPath, 0755); err != nil {
 			return err
 		}
+		defer func() {
+			if err != nil {
+				if _, statErr := os.Stat(volFilePath); statErr != nil && !os.IsNotExist(statErr) {
```
fahedouch, line 102: IMO it may be more readable with a defer func that remove `volFilePath`  just before `os.WriteFile(volFilePath, labelsJSON, 0644)`

zjumoon01: > IMO code should be more readable by doing thing step by step: with a seperate defer func that removes `volFilePath` just before `os.WriteFile(volFilePath, labelsJSON, 0644)`

Fixed.
cc @ktock

PR #2892, opened 2024-03-22
pkg/api/types: remove dependencies on nerdctl

sondavidb:
pkg/api/types had a dependency on netutil and imgutil packages. This would create issues with circular dependencies, e.g. in #2638 imgutil could not import types due to its dependency on netutil. This change moves NetworkCreateOptions and RemoteSnapshotterFlags into types to remove any future circular dependencies.

### DIFF
--- a/cmd/nerdctl/image_pull.go
+++ b/cmd/nerdctl/image_pull.go
@@ -20,7 +20,6 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/containerd/nerdctl/v2/pkg/clientutil"
 	"github.com/containerd/nerdctl/v2/pkg/cmd/image"
-	"github.com/containerd/nerdctl/v2/pkg/imgutil"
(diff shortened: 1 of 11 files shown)


djdongjin (approved):
LGTM on CI green, thanks

sondavidb:
Want to confirm these are just flaky tests? Seems to be similar to other tests failing in the open PRs.

Also sorry @djdongjin I force pushed right after your review so I think it dismissed it 😅

djdongjin:
> Want to confirm these are just flaky tests? Seems to be similar to other tests failing in the open PRs.
> 
> Also sorry @djdongjin I force pushed right after your review so I think it dismissed it 😅

Yeah seems flaky tests, most errrors look related to cosign.

AkihiroSuda:
Thanks, could you rebase, then the CI should be green

sondavidb:
Rebase done, CI looks good.

AkihiroSuda (approved):
Thanks

PR #2906, opened 2024-03-31
Implement `nerdctl run --annotation` (introduced in Docker v24)

AkihiroSuda:
### Commit 1: Implement `nerdctl run --annotation` (introduced in Docker v24)

An OCI runtime (as well as `nerdctl internal oci-hook`) may consume an annotation and behave differently.
e.g., https://github.com/opencontainers/runc/blob/v1.1.12/docs/systemd.md#auxiliary-properties

nerdctl v1:
- `nerdctl run --annotation` was not implemented.
- `nerdctl run --label` is set as a containerd label and an OCI annotation.

nerdctl v2:
- `nerdctl run --annotation` is only set as an OCI annotation.
- `nerdctl run --label` is only set as a containerd label. A label with the `nerdctl/` prefix can no longer be set manually, with an exception for `nerdctl/bypass4netns`. The `nerdctl/bypass4netns` label is still allowed and is propagated to an OCI annotation, for sake of compatibility.

Docker v23:
- `docker run --annotation` was not implemented.
- `docker run --label` is only set as a Docker label.

Docker v24 (implemented in docker/cli#4156, moby/moby#45025):
- `docker run --annotation` is only set as an OCI annotation.
- `docker run --label` is only set as a Docker label.

(In a nutshell, `--annotation` may change the behavior, while `--label` should not.)

### Commit 2: annotations: add `nerdctl/bypass4netns-ignore-subnets` (`[]string`)

For experiments of additional `bypass4netns --ignore`

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ Major:
 - [P2P image distribution using IPFS](./docs/ipfs.md): `nerdctl run ipfs://CID` .
   P2P image distribution (IPFS) is completely optional. Your host is NOT connected to any P2P network, unless you opt in to [install and run IPFS daemon](https://docs.ipfs.io/install/).
 - [Cosign integration](./docs/cosign.md): `nerdctl pull --verify=cosign` and `nerdctl push --sign=cosign`, and [in Compose](./docs/cosign.md#cosign-in-compose)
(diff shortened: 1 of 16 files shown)


AkihiroSuda:
cc @naoki9911

yankay:
Thanks @AkihiroSuda 
/lgtm

PR #2939, opened 2024-04-15
[RFC] inspect: show mounts info from CRI/ctr containers

### LINKED ISSUE #2934: nerdctl container inspect can't show Mounts of container which is created by CRI or ctr

Containers are created by CRI or ctr with bind mounts.
`nerdctl inspect <container>` shows Mounts  is null.

> #nerdctl inspect 0298a4 | grep Mounts
        "Mounts": null,
0298a4  is container ID which is created by CRI

However, `nerdctl inspect --mode native 0298a4` can show bind mounts which are actually in container OCI spec.

https://github.com/containerd/nerdctl/blob/main/pkg/inspecttypes/dockercompat/dockercompat.go#L251
It only shows mounts from label `nerdctl/mounts` .
In other words, `nerdctl inspect` only shows mounts of containers which are created by itself.

I'm not sure the result is by design or we can do some improvements.

My proposal:
1. inspect with dockercompat mode support showing container mounts both from CRI/ctr and nerdctl.
2. "nerdctl/mounts" in label has higher priority.
3. Only mounts in "bind" type from native are converted.  Native container shows all types of mounts, such as tmpfs, proc, sysfs.
4. ResolvConfPath and HostnamePath have the same issue, that should be also taken into consideration

### Additional context

_No response_

zjumoon01:
Fix #2934

Containers created by CRI or ctr with bind mounts
don't have label `nerdct/mounts`.

However, 'inspect' with dockercompat mode only parses mounts from label "nerdct/mounts".

After this patch, 'inspect' with dockercompat mode can show CRI and ctr container bind mounts.

ResolvConfPath, HostnamePath and Hostname are taken to the same consideration.

Further comments:
1. only show bind mount from CRI/ ctr containers. Because native container mounts contain all types of mount, such as tmpfs, proc, sysfs. That are not expected as dockercompat
2. `nerdctl/mounts` in label has higher priority. So do `nerdctl/hostname` and `nerdctl/state-dir`.  CRI/ctr mounts and label "nerdct/mounts" are not merged. It shouldn't happen that container created by nerdctl with bind mount doesn't inject mount info to label nerdctl/mounts 

How to test:

ctr: 

> ctr run -d --mount type=bind,src=/home/ttt/test,dst=/disk002,options=rbind:rw docker.io/library/ubuntu:22.04 test-c1 sleep 200

Before this patch
> #nerdctl -n default inspect test-c1 | grep -i mounts
>         "Mounts": null,

After this patch

> #/home/baijia.wr/nerdctl -n default inspect test-c1 | grep -i mounts -A 10
        "Mounts": [
            {
                "Type": "bind",
                "Source": "/home/ttt/test",
                "Destination": "/disk002",
                "Mode": "",
                "RW": true,
                "Propagation": ""
            }
        ],
        "Config": {

We can also test ResolvConf and hostname by crictl pod


zjumoon01:
test crictl container

pod.json
`{
    "metadata": {
        "name": "testpod",
        "namespace": "default"
    },
    "hostname": "myhostname",
    "log_directory": "/var/log/podlog"
}`

container.json
`{
  "metadata": {
      "name": "test-container"
  },
  "image":{
      "image": "docker.io/library/ubuntu:22.04"
  },
  "command": [
      "/bin/sh"
  ],
  "args": [
      "-c",
      "sleep infinity"
  ],
  "log_path":"ubuntu.0.log",
  "linux": {
  },
  "mounts": [
    {
        "container_path": "/tmp",
        "host_path": "/tmp"
    }
  ]
}`

```
podid=`crictl runp pod.json`
container=`crictl create $podid container.json pod.json`
crictl start $container
```
```
nerdctl inspect -f {{.Config.Hostname}} $container
myhostname
```
```
nerdctl inspect -f {{.HostnamePath}} $container
/home/containerd/io.containerd.grpc.v1.cri/sandboxes/cdeae27e72b9bd8de07b8f7f938d2606767a88342c6531c7913c8422bbb17065/hostname
```
```
nerdctl inspect -f {{.ResolvConfPath}} $container
/home/containerd/io.containerd.grpc.v1.cri/sandboxes/cdeae27e72b9bd8de07b8f7f938d2606767a88342c6531c7913c8422bbb17065/resolv.conf
```

AkihiroSuda:
Can we have a test?

zjumoon01:
> Can we have a test?

OK.
I have added a test and split into two commits.

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -324,6 +338,27 @@ func ImageFromNative(n *native.Image) (*Image, error) {
...
+
+// mountsFromNative only filters bind mount to transform from native container.
+// Because native container shows all types of mounts, such as tmpfs, proc, sysfs.
+func mountsFromNative(spMounts []specs.Mount) []MountPoint {
+	mountpoints := make([]MountPoint, len(spMounts))
+	for i, m := range spMounts {
+		var mp MountPoint
+		if m.Type != "bind" {
+			continue
+		}
```
djdongjin, line 350: This will cause the `mountpoints` will always be `len(spMounts)` and some zero-value in the middle (i.e. when `continue`d). Is that expected? If not we can use `mountpoints := make([]MountPoint, 0, len(spMounts))` and `mountpoints = append(mountpoints, mp)` at line 356.

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -232,15 +233,25 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
...
 			c.AppArmorProfile = p.ApparmorProfile
 		}
+		c.Mounts = mountsFromNative(sp.Mounts)
+		for _, mount := range c.Mounts {
+			if mount.Destination == "/etc/resolv.conf" {
+				c.ResolvConfPath = mount.Source
+			}
+			if mount.Destination == "/etc/hostname" {
+				c.HostnamePath = mount.Source
+			}
```
djdongjin, line 243: nit: we can use `else if` at line 241 to save an unnecessary `if`.

pkg/inspecttypes/dockercompat/dockercompat.go
zjumoon01: Fixed. Thanks

pkg/inspecttypes/dockercompat/dockercompat.go
zjumoon01: Fixed.

cmd/nerdctl/container_inspect_linux_test.go
```diff
@@ -156,6 +157,25 @@ func TestContainerInspectContainsLabel(t *testing.T) {
...
+	defer base.Cmd("rm", "-f", testContainer).Run()
+
+	base.Cmd("run", "-d", "--name", testContainer, "--mount", "type=bind,src=/tmp,dst=/app,readonly=false,bind-propagation=rprivate", testutil.NginxAlpineImage).AssertOK()
+	base.EnsureContainerStarted(testContainer)
+	inspect := base.InspectContainer(testContainer)
+	lbs := inspect.Config.Labels
+
+	// TODO: add more internal labels testcases
+	labelMount := lbs[labels.Mounts]
+	expectedLabelMount := "[{\"Type\":\"bind\",\"Source\":\"/tmp\",\"Destination\":\"/app\",\"Mode\":\"rprivate,rbind\",\"RW\":true,\"Propagation\":\"rprivate\"}]"
```
djdongjin, line 175: nit: it'll be easier to visualize if you use raw string, eg. 

```golang
expectedLabelMount := `[{"Type":"bind","Source":"/tmp",...}]`
```

cmd/nerdctl/container_inspect_linux_test.go
```diff
@@ -156,6 +157,25 @@ func TestContainerInspectContainsLabel(t *testing.T) {
...
+
+	base := testutil.NewBase(t)
+	defer base.Cmd("rm", "-f", testContainer).Run()
+
+	base.Cmd("run", "-d", "--name", testContainer, "--mount", "type=bind,src=/tmp,dst=/app,readonly=false,bind-propagation=rprivate", testutil.NginxAlpineImage).AssertOK()
+	base.EnsureContainerStarted(testContainer)
+	inspect := base.InspectContainer(testContainer)
+	lbs := inspect.Config.Labels
+
+	// TODO: add more internal labels testcases
```
djdongjin, line 173: do you still have more tests to add?

djdongjin (approved):
LGTM other than some minor comments, thanks

cmd/nerdctl/container_inspect_linux_test.go
zjumoon01: > do you still have more tests to add?

No. 
There were never tests for internal labels, such as “containerd.io/restart.status” and "nerdctl/hostname", that are beyond the scope of this commit.

djdongjin (approved):
LGTM, thanks

AkihiroSuda (approved):
Thanks

PR #2954, opened 2024-04-25
fix: error check in the ensureFile func

testwill:


fahedouch (approved):
Thanks

pkg/dnsutil/hostsstore/hostsstore.go
```diff
@@ -56,7 +56,7 @@ func ensureFile(path string) error {
 		return err
 	}
 	f, err := os.OpenFile(path, os.O_CREATE, 0644)
```
AkihiroSuda, line 58: Why not use `os.WriteFile`?

fahedouch: no need to write here, IMO `os.OpenFile` was fine

AkihiroSuda (approved):
Thanks

PR #2984, opened 2024-05-10
Fix some /etc/ file permissions (see containerd#2684) 

### LINKED ISSUE #2684: The `/etc/resolv.conf` file in the container is not world readable because the umask on the host machine is applied on launch
### Description

I have been running into an issue running containers when setting umask on the host to something more restrictive than the default of `0022`.  What I am observing is that the `/etc/resolv.conf` file in the container conforms to the umask on the host which is not what I would expect. I would like for that file in the container to be world readable but when using a more restrictive umask on the host that file's permissions in the container are also restricted. You can observe this behavior by following the steps below. Note that the container image has a default non-root user of `buildozer`.

**Default Umask**

```bash
$ umask 0022; nerdctl run --rm -it alpinelinux/build-base:latest nslookup www.wikipedia.org

--- WORKS ---
```

**Restrictive Umask**

```bash
$ umask 0027; nerdctl run --rm -it alpinelinux/build-base:latest nslookup www.wikipedia.org

--- DOES NOT WORK ---
```

**Restrictive Umask as Root**

```bash
$ umask 0027; nerdctl run --rm -it -u root alpinelinux/build-base:latest nslookup www.wikipedia.org

--- WORKS ---
```

You can observe the file permissions when running umask differently.

```bash
$ umask 0022; nerdctl run --rm -it -u root alpinelinux/build-base:latest ls -l /etc/resolv.conf
$ umask 0027; nerdctl run --rm -it -u root alpinelinux/build-base:latest ls -l /etc/resolv.conf
```

Note that with `0022` the file is world readable while with `0027` it is not.

I observe that the `/etc/resolv.conf` file in the container is not world readable. I would expect that the umask on the host would not propagate to the container environment in such a way that the `/etc/resolv.conf` file's permissions becomes restricted. This restrictions causes DNS queries to fail for non-root users. I suspect this is related to some kind of copy operation occurring on the host which applies the umask settings.

1.7.0

None

### Host information

_No response_

apostasie:
@AkihiroSuda @rafi-gh

This intend to fix #2684 

From local testing with a 0077 umask:
```
apo@oliphant:~ $ sudo ./nerdctl run -ti --network host --rm debian ls -lA
-rw------- 1 root root 241 May 10 15:28 /etc/resolv.conf
apo@oliphant:~ $ sudo ./nerdctl-patch run -ti --network host --rm debian ls -lA /etc/resolv.conf
-rw-r--r-- 1 root root 241 May 10 15:28 /etc/resolv.conf
```

The patch is very simple, and just chmod all three files after writing.

Let me know if this is acceptable in this state or if a more involved refactoring is in order.

### DIFF
--- a/pkg/containerutil/container_network_manager.go
+++ b/pkg/containerutil/container_network_manager.go
@@ -398,12 +398,18 @@ func withDedupMounts(mountPath string, defaultSpec oci.SpecOpts) oci.SpecOpts {
 	}
 }
 
+// copyFileContent copies a file and sets world readable permissions on it, regardless of umask.
+// This is used solely for /etc/resolv.conf and /etc/hosts
 func copyFileContent(src string, dst string) error {
 	data, err := os.ReadFile(src)
 	if err != nil {
 		return err
 	}
 	err = os.WriteFile(dst, data, 0644)
+	if err != nil {
+		return err
+	}
+	err = os.Chmod(dst, 0644)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


apostasie:
CI is happy.
PTAL.

Thanks!

AkihiroSuda:
The commit looks good, but it's title should be more descriptive

apostasie:
Fixed. Let me know if this is better.

Thank you @AkihiroSuda

apostasie:
Not sure what the failure is about.
Previous run was fine.
Is this flaky in any way? Or did I actually mess something here?

AkihiroSuda (approved):
Thanks

PR #2995, opened 2024-05-13
Re-acquire name in onStartContainer (see #2992)

### LINKED ISSUE #2992: nerdctl sometimes fail to check if a container name already exists and ends-up creating duplicates
### Description

Title says all

Unfortunately, I do not have a reproducer and this seems to happen under certain circumstances only.

This is definitely confirmed though:

```
apo@oliphant:~ $ sudo ./nerdctl ps
CONTAINER ID    IMAGE                                                         COMMAND                   CREATED              STATUS    PORTS                     NAMES
5e7d6291a423    docker.io/library/debian:latest                               "sleep 3600"              4 minutes ago        Up                                  restarttest
b579c38e3c55    docker.io/library/debian:latest                               "sleep 123"               6 seconds ago        Up                                  restarttest

apo@oliphant:~ $ sudo ./nerdctl rm restarttest
FATA[0000] 1 errors:
multiple IDs found with provided prefix: restarttest

```

Note error message above ^

This should never happen.

1.7.6

None

### Host information

_No response_

apostasie:
@AkihiroSuda 

This is solution A (fix #2992 ).

It does make things much better of course, but yes, it is still racy.

With `while true; do sudo ./nerdctl-patch run -d --restart always --name whatevername debian bash; done` it will race out the container restart about 1 time out of 30 (on an old rpi).

IMHO, we should go ahead and merge this anyway - because it is drastically improving the situation.

### DIFF
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -477,6 +477,15 @@ func onCreateRuntime(opts *handlerOpts) error {
 }
 
 func onStartContainer(opts *handlerOpts) error {
+	name := opts.state.Annotations[labels.Name]
+	ns := opts.state.Annotations[labels.Namespace]
+	namst, err := namestore.New(opts.dataStore, ns)
+	if err != nil {
+		log.L.WithError(err).Error("failed opening the namestore in onStartContainer")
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #3001, opened 2024-05-14
Add support for DHCP CNI plugin IPAM options and set the DHCP host-name option properly

### LINKED ISSUE #1103: Supply container hostname in CNI_ARGS

When running containers with `ipam: { "type": "dhcp" }`   , the dhcp request have a empty hostname.
Makes life painful when using dhcp to find out on DHCP server side which container have that lease.

the "dhcp" type already have a mechanism to send information in the DHCP request:

> provide (dict, optional): Options providing to DHCP server when acquire leases.
> option (string, optional): String or number representation of the DHCP option.
> value (string, optional): String representation of the value. Directly sent to server.
> fromArg (string, optional): Get value from CNI_ARGS by given argument name.

(https://www.cni.dev/plugins/current/ipam/dhcp/)

Thus supplying the container hostname in a given key under CNI_ARGS would solve the issue.

### Additional context

Just for information, podman supplies the following CNI_ARGS on my tests: `IgnoreUnknown=1;K8S_POD_NAME=elastic_wilbur`

I can use the following configuration for my ipam:
```
...
"type": "dhcp",
"provide": [
  {
    "option": "host-name",
    "fromArg": "K8S_POD_NAME"
  }
]
...
```

Also, If there's no option supplied, the value doesn't get overriden (on my tests, CNI v1.1.1) . For example:
```
...
"type": "dhcp",
"provide": [
  {
    "option": "host-name",
    "value": "foo"
  },
  {
    "option": "host-name",
    "fromArg": "K8S_POD_NAME"
  }
]
...
```
Will sent the DHCP option "host-name" with value `foo` when `K8S_POD_NAME` is not present on CNI_ARGS.
This way one can have a default value for a given option (useful for vendor-class-identifier).
So anything as key could be used .. (i.e.: NERDCTL_HOSTNAME, etc.). Logic of which key goes where would stay in the ipam config.

apostasie:
Fixes #1103.

cc @AkihiroSuda @theodiem

### DIFF
--- a/docs/cni.md
+++ b/docs/cni.md
@@ -90,6 +90,29 @@ an easier way is to use DHCP to assign the IP:
 
 Using `--driver ipvlan` can create `ipvlan` network, the default mode for IPvlan is `l2`.
 
+## DHCP host-name and other DHCP options
+
+Nerdctl automatically sets the DHCP host-name option to the hostname value of the container.
+
+Furthermore, on network creation, nerdctl supports the ability to set other DHCP options through `--ipam-options`.
+
+Currently, the following options are supported by the DHCP plugin:
+```
+dhcp-client-identifier
+subnet-mask
+routers
+user-class
+vendor-class-identifier
+```
+
+For example:
+```
+# nerdctl network create --driver macvlan \
+    --ipam-driver dhcp \
+    --ipam-opt 'vendor-class-identifier={"type": "provide", "value": "Hey! Its me!"}' \
+    my-dhcp-net
+```
+
 ## Custom networks
 
 You can also customize your CNI network by providing configuration files.
--- a/pkg/netutil/cni_plugin_unix.go
+++ b/pkg/netutil/cni_plugin_unix.go
@@ -144,10 +144,25 @@ func newHostLocalIPAMConfig() *hostLocalIPAMConfig {
 	}
 }
 
-// https://github.com/containernetworking/plugins/blob/v1.1.0/plugins/ipam/dhcp/main.go#L43-L54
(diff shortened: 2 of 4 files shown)


docs/cni.md
```diff
@@ -90,10 +90,33 @@ an easier way is to use DHCP to assign the IP:
...
+    --ipam-driver dhcp \
+    --ipam-opt 'vendor-class-identifier={"type": "provide", "value": "Hey! Its me!"}' \
+    my-dhcp-net
+```
+
 ## Custom networks
 
 You can also customize your CNI network by providing configuration files.
-For example you have one configuration file(`/etc/cni/net.d/10-mynet.conf`)
+For exemple you have one configuration file(`/etc/cni/net.d/10-mynet.conf`)
```
AkihiroSuda, line 119: ?

pkg/netutil/cni_plugin_unix.go
```diff
@@ -18,7 +18,9 @@
 
 package netutil
 
-import "github.com/containerd/nerdctl/v2/pkg/rootlessutil"
+import (
+	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
+)
```
AkihiroSuda, line 23: ?

apostasie: Sorry. Pushed too fast. Both typo are fixed.

apostasie:
Rebased.
The cni plugins sha failure is now gone.

PTAL.

Still some (random?) CI failure - unless I am responsible for these, it seems to me these are just flaky. It is really annoying.
I want these problems gone. Will have a look at them once done with image inspect.

AkihiroSuda (approved):
Thanks

Issue #2335, opened 2023-06-28
nerdctl inspect to show State.StartedAt


austinvazquez:
### What is the problem you're trying to solve

Docker CLI has both `Created` and `State.StartedAt` fields on inspect.

Currently nerdctl supports `Created` and `State.FinishedAt` but not `State.StartedAt`

### Describe the solution you'd like

It would be great to add this field to the information for objects.

austinvazquez:
Dug a little deeper to understand the gap here. TLDR; containerd does not support `StartedAt` universally. Somewhat makes sense because it is a little difficult to maintain this state.

The CRI plugin supports this by taking the timestamp after `task.Start`. We could do something similar but would need to maintain that state somewhere.

Happy to close this as not planned unless others have another idea how to implement.

AkihiroSuda:
Probably this should be implemented in the containerd daemon

fahedouch:
IMO State can be easily maintained via `annotation`

apostasie:
> IMO State can be easily maintained via `annotation`

@fahedouch Can you set annotations from inside ocihooks though?

fahedouch:
> > IMO State can be easily maintained via `annotation`
> 
> @fahedouch Can you set annotations from inside ocihooks though?

no because it is happening before the startup of the container. I think the good place to handle this in the `containerd-shim` side

apostasie:
> > > IMO State can be easily maintained via `annotation`
> > 
> > 
> > @fahedouch Can you set annotations from inside ocihooks though?
> 
> no because it is happening before the startup of the container. I think the good place to handle this is the `containerd-shim` .

Sure, that would be ideal if containerd would allow to query that.
We can do it here though, as proposed in the PR (inside ocihooks).

Totally understand if you would rather wait to see it implemented over there though.

apostasie:
Fixed by #3003

@AkihiroSuda can you close this one?

AkihiroSuda:
Thanks

PR #3003, opened 2024-05-14
[Design Review] StartedAt for containers

### LINKED ISSUE #3002: container inspect finishedAt date set to zero
### Description

If a container has not exited yet, the FinishedAt time on inspect will be set to "0001-01-01T00:00:00Z"

This does not make much sense IMHO.

Suggesting to test `if ExitTime.IsZero()` and only set FinishedAt if not zero.

Let me know if you are ok with this and I ll batch that into my dockercompat cleanup work.

1.
2.
3.

FinishedAt to not display a bogus date is expected.

1.7.6

None

### Host information

_No response_

apostasie:
This is a proposal to implement StartedAt (see #2335).

Short of containerd exposing this, the right place to implement is likely inside `ocihook.onStart` (to account for containers with a restart policy).

I do not know how or if it is possible to modify annotations from inside ocihooks. If it is, that would be a better solution and would love to see an example of how to do that.

So, instead, I am maintaining a new state file inside the stateDir.
For that, there is a `LifecycleState` struct that we can leverage in the future to store more information that has to be updated inside ocihooks.

This PR also contains a two lines minor change to fix #3002.

Let me know what you folks think.
As usual, happy to adapt to feedback on the right way to do this.

### DIFF
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	gocni "github.com/containerd/go-cni"
 	"github.com/containerd/log"
@@ -35,6 +36,7 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/namestore"
 	"github.com/containerd/nerdctl/v2/pkg/netutil"
 	"github.com/containerd/nerdctl/v2/pkg/netutil/nettype"
+	"github.com/containerd/nerdctl/v2/pkg/ocihook/state"
 	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
 	types100 "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -284,7 +286,7 @@ func getNetNSPath(state *specs.State) (string, error) {
 		return netNsPath, nil
 	}
 
-	if state.Pid == 0 && !netNsFound {
+	if state.Pid == 0 {
 		return "", errors.New("both state.Pid and the netNs annotation are unset")
 	}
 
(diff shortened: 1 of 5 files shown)


apostasie:
I did bust a few tests - at least `TestContainerFromNative` which expect these 0 dates.

If the design here is fine with you folks, I will fix the tests.

pkg/ocihook/ocihook.go
```diff
@@ -284,7 +286,7 @@ func getNetNSPath(state *specs.State) (string, error) {
 		return netNsPath, nil
 	}
 
-	if state.Pid == 0 && !netNsFound {
```
apostasie, line 287: This is unnecessary (always true)

pkg/ocihook/state/state.go
```diff
@@ -0,0 +1,85 @@
...
+	err := lockutil.WithDirLock(lf.stateDir, fun)
+	if err != nil {
+		return fmt.Errorf("failed to lock state dir: %w", err)
+	}
+
+	return nil
+}
+
+func (lf *LifecycleState) Read() error {
+	data, err := os.ReadFile(filepath.Join(lf.stateDir, "lifecycle.json"))
```
AkihiroSuda, line 61: "lifecycle.json" should be a `const`, and should be documented in https://github.com/containerd/nerdctl/blob/main/docs/dir.md

pkg/ocihook/state/state.go
```diff
@@ -0,0 +1,85 @@
...
+// This is meant to store stateful informations about containers that we receive from ocihooks
+// We are storing them inside the container statedir
+// Note that you MUST use WithLock to perform any operation (like Read or Write).
+// Typically:
+// lf.WithLock(func ()error {
+//   lf.Read()
+//   // Modify something on the object
+//   lf.StartedAt = ...
+//   lf.Write()
+// })
```
AkihiroSuda, line 38: `Read()` and `Write()` should be named like `Load()` and `Store()` ?

apostasie: Ok.
Opting for Load / Save.

AkihiroSuda (approved):
Thanks

pkg/ocihook/state/state.go
```diff
@@ -0,0 +1,89 @@
...
+
+func NewLifecycleState(stateDir string) *LifecycleState {
+	return &LifecycleState{
+		stateDir: stateDir,
+	}
+}
+
+type LifecycleState struct {
+	stateDir  string
+	StartedAt int64 `json:"started_at"`
```
AkihiroSuda, line 52:
```suggestion
	StartedAt time.Time `json:"started_at"`
```

apostasie: Ok.
Last version now serializes and unserialize `time.Time` instead of timestamp.

@AkihiroSuda PTAL

apostasie:
Reviewers: moby seems really inconsistent with their use of UTC vs. Local.

I personally think we should be consistent (and always display local) - we might want to review that across the entire codebase.

AkihiroSuda (approved):
Thanks

Issue #3006, opened 2024-05-15
nerdctl inspect should not return a stream of json array, but a single array


apostasie:
### Description

In a number of different occasions, but specifically when querying different objects in one call, nerdctl will return a stream of json arrays.

In comparison, docker returns one array.

### Steps to reproduce the issue

1. nerdctl inspect some_container some_image
2. docker inspect some_container some_image

### Describe the results you received and expected

na

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #3007, opened 2024-05-15
image inspect fixes & cleanup

### LINKED ISSUE #3004: Current implementation of images inspect is faulty (eg: RepoTags)
### Description

nerdctl does not deal properly with images bounds to multiple tags.

Compare:
```
sudo ./nerdctl pull dubodubonduponey/testouille
sudo ./nerdctl pull dubodubonduponey/testouille:2
sudo ./nerdctl inspect dubodubonduponey/testouille
sudo ./nerdctl inspect XXXXXXX
```

with

```
docker pull dubodubonduponey/testouille
docker pull dubodubonduponey/testouille:2
docker inspect dubodubonduponey/testouille
docker inspect XXXXXXX
```

1. when querying by reference, nerdctl returned result has in RepoTags ONLY the name used to query, regardless of the fact other tags may hold the same image
2. when querying by id (that is bound to multiple tags), nerdctl returns multiple distinct images instead of just one

Compare with docker behavior.

1.7.6

None

### Host information

_No response_

apostasie:
This overall enhance image inspect:
- multiple images with the same digest are now coalesced properly into one, with multiple RepoTags
- multiple images resulting for multiple arguments in the call to `image inspect` will coalesce into one array, instead of a stream of arrays
- image from native has been cleaned-up and simplified, making room for later addition of the missing TODO

In the process we stop calling on the image Walker which serves little to no purpose here, hence significantly reducing the complexity.

This mainly fixes #3004 and part of #3006.

After this, more work is needed overall for `inspect`, and then some specifically for `image inspect`.

### DIFF
--- a/pkg/cmd/image/inspect.go
+++ b/pkg/cmd/image/inspect.go
@@ -19,58 +19,74 @@ package image
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/containerd/nerdctl/v2/pkg/formatter"
-	"github.com/containerd/nerdctl/v2/pkg/idutil/imagewalker"
 	"github.com/containerd/nerdctl/v2/pkg/imageinspector"
+	"github.com/containerd/nerdctl/v2/pkg/imgutil"
 	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/dockercompat"
+	"github.com/containerd/nerdctl/v2/pkg/referenceutil"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/cmd/image/inspect.go
```diff
@@ -19,58 +19,74 @@ package image
...
-		if formatErr := formatter.FormatSlice(options.Format, options.Stdout, f.entries); formatErr != nil {
+	if len(entries) > 0 {
+		if formatErr := formatter.FormatSlice(options.Format, options.Stdout, entries); formatErr != nil {
 			log.G(ctx).Error(formatErr)
 		}
 	}
-	return err
-}
-
-type imageInspector struct {
```
apostasie, line 73: This does not serve any purpose.

Issue #3011, opened 2024-05-16
nerdctl images displays the wrong image id


apostasie:
### Description

Whenever nerdctl displays lists of images (nerdctl images, nerdctl image list), the ID being reported are the repo sha, not the image ID.

Compare with Docker.

### Steps to reproduce the issue

1. nerdctl images
2. docker images

### Describe the results you received and expected

Expected: Docker behavior - see the image ID.

Received: repo digest

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
I see.

Was there a reason for that choice?

The problems I see:

- it does make nerdctl and docker fundamentally disagree on what defines the "same image".

- it will also lead to internal inconsistent behavior with nerdctl wrt to commands like `rmi`.

Specifically:

```
sudo nerdctl pull debian >/dev/null 2>&1
sudo nerdctl tag debian apostasie/nerdctl-test
sudo nerdctl push apostasie/nerdctl-test

shortid="$(sudo nerdctl inspect apostasie/nerdctl-test | jq -rc .[0].RepoDigests[0])"
shortid="${shortid##*:}"
shortid="${shortid:0:8}"

sudo nerdctl rmi -f "$shortid"

sudo nerdctl images # <- empty

# Now...

sudo nerdctl pull debian  >/dev/null 2>&1
sudo nerdctl pull apostasie/nerdctl-test  >/dev/null 2>&1

shortid="$(sudo nerdctl inspect apostasie/nerdctl-test | jq -rc .[0].RepoDigests[0])"
shortid="${shortid##*:}"
shortid="${shortid:0:8}"

sudo nerdctl rmi -f "$shortid"

sudo nerdctl images # <- not empty
```

This is obviously due to the fact that the RepoDigest will be different for the very same image depending on whether it has been pushed&pulled or not.

I can fix maybe fix it, by querying and keying images during lookup along the "docker" Id, but it is going to make things more complicated and possibly lead to some other types of discrepancies.

AkihiroSuda:
containerd's image store is designed to use a single object for multi-platform image index, while Docker's image store is designed to allocate an object per platform.

AkihiroSuda:
BTW `nerdctl image inspect` has been using Docker-compatible ID which is incompatible with the ID shown in `nerdctl images`:
[...]

Issue #2061, opened 2023-03-01
`hello-world` and `hello-world@sha256:...` are treated as different images


AkihiroSuda:
### Description

`hello-world` and `hello-world@sha256:...` are treated as different images

### Steps to reproduce the issue

```console
$ nerdctl pull hello-world
docker.io/library/hello-world:latest:                                             resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:6e8b6f026e0b9c419ea0fd02d3905dd0952ad1feea67543f525c73a0a790fefb:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4: done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:feb5d9fea6a5e9606aa995e879d862b825965ba48de054caab5ef356dc6b3412:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:2db29710123e3e53a794f2694094b9b4338aa9ee5c40b930cb8063a1be392c54:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 5.3 s 

$ nerdctl image inspect hello-world@sha256:6e8b6f026e0b9c419ea0fd02d3905dd0952ad1feea67543f525c73a0a790fefb
FATA[0000] 1 errors:
no such image: hello-world@sha256:6e8b6f026e0b9c419ea0fd02d3905dd0952ad1feea67543f525c73a0a790fefb

$ nerdctl pull hello-world@sha256:6e8b6f026e0b9c419ea0fd02d3905dd0952ad1feea67543f525c73a0a790fefb
docker.io/library/hello-world@sha256:6e8b6f026e0b9c419ea0fd02d3905dd0952ad1feea67543f525c73a0a790fefb: resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:6e8b6f026e0b9c419ea0fd02d3905dd0952ad1feea67543f525c73a0a790fefb:                         done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:f54a58bc1aac5ea1a25d796ae155dc228b3f0e11d046ae276b39c4bf2f13d8c4:                      done           |++++++++++++++++++++++++++++++++++++++| 
[...]

PR #3017, opened 2024-05-17
Image inspect rework

### LINKED ISSUE #2988: docker / nerdctl inconsistent return value from `image inspect --format json`
### Description

Docker returns an array with one or more values.

Nerdctl returns a stream of single values.

1. sudo ./nerdctl image inspect --format json debian debian
2. docker image inspect --format json debian debian

nerdctl will return multiple lines of objects.

docker returns an array of objects.

1.7.6

None

### Host information

_No response_

### LINKED ISSUE #3005: nerdctl inspect does not allow inspecting networks, volumes
### Description

docker inspect allows inspecting networks, volumes, images, containers.

nerdctl only allows containers and images

1. nerdctl inspect bridge
2 docker inspect bridge.

.

1.7.6

None

### Host information

_No response_

### LINKED ISSUE #3016: nerdctl rmi can be broken by using tag names with repo short sha
### Description

When an image uses a tag that is a short version of another (unrelated) image repo sha, nerdctl rmi will break down.

This is another aspect of one of the many issues with our current image lookup implementation.

```
nerdctl pull busybox
nerdctl pull alpine
shortid="$(nerdctl inspect busybox | jq -rc .[0].RepoDigests[0])"
shortid="${shortid##*:}"
shortid="${shortid:0:8}"
nerdctl tag alpine "$shortid"
nerdctl rmi "$shortid"
```

```
FATA[0000] 1 errors:
multiple IDs found with provided prefix: 5eef5ed3
```

Should have just worked and removed the tagged version of alpine.

1.7.6

None

### Host information

_No response_

apostasie:
Apologies for the big commit here, but I do not believe we can fix image inspect incrementally, as some of the core assumptions are wrong (specifically in the way we walk images, in the returned data types, etc).

Furthermore, there is more work needed here.
Other command which are doing images lookup (rmi) are also faulty.
Fixing these as well would turn this PR into a large monster, and I would rather do that separately as a follow-up once this one is reviewed and accepted.

So, this PR fixes the following:
- image inspect (#3004 among other things):
    - allow querying all sort of references:
        - repo short name
        - repo full name
        - repo with tag and/or digest
        - digest, with or without sha prefix
        - short ids
    - properly populate RepoTags and RepoDigests
    - always return a JSON array, instead of a stream, for example when multiple references are provided (fixes #2988)
- dockercompat:
    - implement Parent (if it has been set on the image - note: I am not sure nerdctl does add the label when committing, but this is a different problem)
    - implement Variant
    - add GraphDriverData struct for future implementation
    - updated Image struct to the latest version at https://github.com/moby/moby/blob/v26.1.2/api/types/types.go#L34-L140
    - implement (deprecated) VirtualSize for compatibility

Obviously, the reworked `image inspect` should be wired into `inspect` at a later point, and the overall revised lookup mechanics should be generalized and used by other methods (like `rmi`) - again, as a follow-up.
Overall the current ImageWalker should either be removed or replaced with the updated workflow here.

Please note that our implementation also diverges from Docker, for good reasons:
For example, when retrieving an image using a digest and a tagged name, we do verify the tag name against the result, unlike Moby which only verifies the image name (because of their use of ParseDockerRef, which should be banned from existence IMHO…).
Because of that, Moby will allow inspecting images using random bogus tag names as long as the digest matches and return a different image than the requested one altogether.
This has had far reaching consequences: specifically, mind-bending bugs for build caching invalidation.
I do believe Moby must fix it, but it’s not my problem.

What is left overall:
- #3011 is a problem - we need to look into it and this might have consequences for the lookup implementation here - specifically wrt ContentStore
- generalizing the lookup method so that other commands can leverage it (will fix #3016 for example)
- rewrite overall `inspect` to fix #3005, #3006

Also, this PR adds a bunch of integration tests on the above.

PTAL

cc @AkihiroSuda

### DIFF
--- a/cmd/nerdctl/image_inspect_test.go
+++ b/cmd/nerdctl/image_inspect_test.go
@@ -17,10 +17,15 @@
 package main
 
 import (
+	"encoding/json"
+	"runtime"
+	"strings"
 	"testing"
 
-	"github.com/containerd/nerdctl/v2/pkg/testutil"
 	"gotest.tools/v3/assert"
+
+	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/dockercompat"
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
 )
 
 func TestImageInspectContainsSomeStuff(t *testing.T) {
@@ -39,9 +44,134 @@ func TestImageInspectWithFormat(t *testing.T) {
 	base := testutil.NewBase(t)
 
 	base.Cmd("pull", testutil.CommonImage).AssertOK()
+
 	// test RawFormat support
 	base.Cmd("image", "inspect", testutil.CommonImage, "--format", "{{.Id}}").AssertOK()
 
 	// test typedFormat support
 	base.Cmd("image", "inspect", testutil.CommonImage, "--format", "{{.ID}}").AssertOK()
+}
+
+func inspectImageHelper(base *testutil.Base, identifier ...string) []dockercompat.Image {
+	args := append([]string{"image", "inspect"}, identifier...)
+	cmdResult := base.Cmd(args...).Run()
+	assert.Equal(base.T, cmdResult.ExitCode, 0)
+	var dc []dockercompat.Image
+	if err := json.Unmarshal([]byte(cmdResult.Stdout()), &dc); err != nil {
+		base.T.Fatal(err)
+	}
+	return dc
+}
+
+func TestImageInspectDifferentValidReferencesForTheSameImage(t *testing.T) {
+	testutil.DockerIncompatible(t)
+
+	if runtime.GOOS == "windows" {
+		t.Skip("Windows is not supported for this test right now")
+	}
+
+	base := testutil.NewBase(t)
+
+	// Overall, we need a clean slate before doing these lookups.
+	// More specifically, because we trigger https://github.com/containerd/nerdctl/issues/3016
+	// we cannot do selective rmi, so, just nuke everything
+	ids := base.Cmd("image", "list", "-q").Out()
+	allIds := strings.Split(ids, "\n")
+	for _, id := range allIds {
+		id = strings.TrimSpace(id)
+		if id != "" {
+			base.Cmd("rmi", "-f", id).Run()
+		}
+	}
+
+	base.Cmd("pull", "alpine", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "busybox", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "busybox:stable", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "registry-1.docker.io/library/busybox", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "registry-1.docker.io/library/busybox:stable", "--platform", "linux/amd64").AssertOK()
+
+	tags := []string{
+		"",
+		":latest",
+		":stable",
+	}
+	names := []string{
+		"busybox",
+		"library/busybox",
+		"docker.io/library/busybox",
+		"registry-1.docker.io/library/busybox",
+	}
+
+	// Build reference values for comparison
+	reference := inspectImageHelper(base, "busybox")
+	assert.Equal(base.T, 1, len(reference))
+	// Extract image sha
+	sha := strings.TrimPrefix(reference[0].RepoDigests[0], "busybox@sha256:")
+
+	differentReference := inspectImageHelper(base, "alpine")
+	assert.Equal(base.T, 1, len(differentReference))
+
+	// Testing all name and tags variants
+	for _, name := range names {
+		for _, tag := range tags {
+			t.Logf("Testing %s", name+tag)
+			result := inspectImageHelper(base, name+tag)
+			assert.Equal(base.T, 1, len(result))
+			assert.Equal(base.T, reference[0].ID, result[0].ID)
+		}
+	}
+
+	// Testing all name and tags variants, with a digest
+	for _, name := range names {
+		for _, tag := range tags {
+			t.Logf("Testing %s", name+tag+"@"+sha)
+			result := inspectImageHelper(base, name+tag+"@sha256:"+sha)
+			assert.Equal(base.T, 1, len(result))
+			assert.Equal(base.T, reference[0].ID, result[0].ID)
+		}
+	}
+
+	// Testing repo digest and short digest with or without prefix
+	for _, id := range []string{"sha256:" + sha, sha, sha[0:8], "sha256:" + sha[0:8]} {
+		t.Logf("Testing %s", id)
+		result := inspectImageHelper(base, id)
+		assert.Equal(base.T, 1, len(result))
+		assert.Equal(base.T, reference[0].ID, result[0].ID)
+	}
+
+	// Demonstrate image name precedence over digest lookup
+	// Using the shortened sha should no longer get busybox, but rather the newly tagged Alpine
+	t.Logf("Testing (alpine tagged) %s", sha[0:8])
+	// Tag a different image with the short id
+	base.Cmd("tag", "alpine", sha[0:8]).AssertOK()
+	result := inspectImageHelper(base, sha[0:8])
+	assert.Equal(base.T, 1, len(result))
+	assert.Equal(base.T, differentReference[0].ID, result[0].ID)
+
+	// Prove that wrong references with an existing digest do not get retrieved when asking by digest
+	for _, id := range []string{"doesnotexist", "doesnotexist:either", "busybox:bogustag"} {
+		t.Logf("Testing %s", id+"@"+sha)
+		args := append([]string{"image", "inspect"}, id+"@"+sha)
+		cmdResult := base.Cmd(args...).Run()
+		assert.Equal(base.T, cmdResult.ExitCode, 0)
+		assert.Equal(base.T, cmdResult.Stdout(), "")
+	}
+
+	// Prove that invalid reference return no result without crashing
+	for _, id := range []string{"∞∞∞∞∞∞∞∞∞∞", "busybox:∞∞∞∞∞∞∞∞∞∞"} {
+		t.Logf("Testing %s", id)
+		args := append([]string{"image", "inspect"}, id)
+		cmdResult := base.Cmd(args...).Run()
+		assert.Equal(base.T, cmdResult.ExitCode, 0)
+		assert.Equal(base.T, cmdResult.Stdout(), "")
+	}
+
+	// Retrieving multiple entries at once
+	t.Logf("Testing %s", "busybox busybox busybox:stable")
+	result = inspectImageHelper(base, "busybox", "busybox", "busybox:stable")
+	assert.Equal(base.T, 3, len(result))
+	assert.Equal(base.T, reference[0].ID, result[0].ID)
+	assert.Equal(base.T, reference[0].ID, result[1].ID)
+	assert.Equal(base.T, reference[0].ID, result[2].ID)
+
 }
--- a/pkg/cmd/image/inspect.go
+++ b/pkg/cmd/image/inspect.go
@@ -19,58 +19,182 @@ package image
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/images"
 	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/containerd/nerdctl/v2/pkg/formatter"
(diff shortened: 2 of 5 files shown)


pkg/referenceutil/referenceutil.go
```diff
@@ -32,6 +32,20 @@ type Reference interface {
...
+	if scheme, ref, err := ParseIPFSRefWithScheme(rawRef); err == nil {
+		return Reference(stringRef{scheme: scheme, s: ref}), nil
+	}
+	if c, err := cid.Decode(rawRef); err == nil {
+		return c, nil
+	}
+	return refdocker.ParseAnyReference(rawRef)
+}
+
+// Deprecated
```
apostasie, line 48: Propose that we stop using ParseDockerRef in favor of ParseAnyReference. The former has some nasty side effects by dropping the tag in case there is a digest AFAIK.

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -294,48 +305,51 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
 	return c, nil
 }
 
-func ImageFromNative(n *native.Image) (*Image, error) {
-	i := &Image{}
-
-	imgoci := n.ImageConfig
+func ImageFromNative(nativeImage *native.Image) (*Image, error) {
```
apostasie, line 308: Light cleanup:
- improve variable name readability
- introduce variant and virtual size
- populate repodigest

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -47,28 +47,39 @@ import (
...
+	// TODO: GraphDriver	GraphDriverData
+
 	RootFS   RootFS
 	Metadata ImageMetadata
+
+	// Deprecated: TODO: Container   string
+	// Deprecated: TODO: ContainerConfig *container.Config
+}
+
+// From: https://github.com/moby/moby/blob/v26.1.2/api/types/graph_driver_data.go
```
apostasie, line 79: Provisional for future implementation

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -47,28 +47,39 @@ import (
 	"github.com/tidwall/gjson"
 )
 
-// Image mimics a `docker image inspect` object.
-// From https://github.com/moby/moby/blob/v20.10.1/api/types/types.go#L340-L374
+// From https://github.com/moby/moby/blob/v26.1.2/api/types/types.go#L34-L140
 type Image struct {
```
apostasie, line 51: Updated with latest definition and reordered a but for readability.
- moved down deprecated fields
- added VirtualSize, Parent, Variant

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -37,7 +37,7 @@ import (
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/runtime/restart"
-	gocni "github.com/containerd/go-cni"
```
apostasie, line 40: Unnecessary.

README.md
```diff
@@ -268,6 +268,10 @@ docker build -t test-integration --target test-integration .
 docker run -t --rm --privileged test-integration
 ```
 
+To run a single integration test (in this case, `image_inspect_test`):
```
apostasie, line 271: Just a note for newcomers.

cmd/nerdctl/image_inspect_test.go
```diff
@@ -33,15 +37,137 @@ func TestImageInspectContainsSomeStuff(t *testing.T) {
...
 	base.Cmd("pull", testutil.CommonImage).AssertOK()
+
 	// test RawFormat support
 	base.Cmd("image", "inspect", testutil.CommonImage, "--format", "{{.Id}}").AssertOK()
 
 	// test typedFormat support
 	base.Cmd("image", "inspect", testutil.CommonImage, "--format", "{{.ID}}").AssertOK()
 }
+
+func inspectImageHelper(base *testutil.Base, identifier ...string) []dockercompat.Image {
```
apostasie, line 55: Cannot use the one in testutil, as it assumes single entry array.

cmd/nerdctl/image_inspect_test.go
```diff
@@ -39,9 +43,132 @@ func TestImageInspectWithFormat(t *testing.T) {
...
+	assert.Equal(base.T, cmdResult.ExitCode, 0)
+	var dc []dockercompat.Image
+	if err := json.Unmarshal([]byte(cmdResult.Stdout()), &dc); err != nil {
+		base.T.Fatal(err)
+	}
+	return dc
+}
+
+func TestImageInspectDifferentValidReferencesForTheSameImage(t *testing.T) {
+	testutil.DockerIncompatible(t)
```
apostasie, line 66: If really necessary, I could split this in two parts - the part that is docker compatible and the part that is not.

Either way we need first to resolve #3011

apostasie:
Reviewer:

I don't want to introduce churn here by regularly adding more fixes.
The PR in its current state may have some minor issues (specifically there is one: Docker Hub names are not shortened properly before being added to RepoTags).
In the name of keeping our sanity, and if these minor issues are not a big deal for you, I will fix these in a follow-up once this is merged.

apostasie:
Windows CI failure is on me.
Will try to fix it.

apostasie:
Note this is likely fixing issues like #2061

Zheaoli:
I will convert this PR to a draft before we solve the Windows issue together. I will help to figure out what happened.

apostasie:
> https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/image_save_test.go#L35C22-L35C44
> 
> You can output the inspect.RepoDigests[0] in test during the debug, I think there one more `:` char in the string

Thanks @Zheaoli 

I believe the failure on save_test (TestSaveById) are unrelated, and are just flakyness.

The actual test failing is the one I introduced, and the issue here is that I am using test images that do not have a Windows version, so, pull fails on windows: https://github.com/containerd/nerdctl/actions/runs/9133230483/job/25136088395?pr=3017#step:8:569

apostasie:
More accurately, the issue is:
`time="2024-05-18T17:41:26Z" level=fatal msg="failed to extract layer sha256:d4fc045c9e3a848011de66f34b81f052d4f2c15a17bb196d637e526349601820: hcsshim::ProcessBaseLayer \\\\?\\C:\\ProgramData\\containerd\\root\\io.containerd.snapshotter.v1.windows\\snapshots\\106: The system cannot find the path specified.: unknown"`

And I know nothing about `hcsshim`, or why it would fail extracting an image.

apostasie:
@Zheaoli @AkihiroSuda 
CI is green.
I disabled the test on window for now.
Will revisit soon when I get a Windows environment to test on.

README.md
ktock, line 271: can be a separated PR

cmd/nerdctl/image_inspect_test.go
```diff
@@ -39,9 +44,136 @@ func TestImageInspectWithFormat(t *testing.T) {
...
+			base.Cmd("rmi", "-f", id).Run()
+		}
+	}
+
+	base.Cmd("pull", "alpine", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "busybox", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "busybox:stable", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "registry-1.docker.io/library/busybox", "--platform", "linux/amd64").AssertOK()
+	base.Cmd("pull", "registry-1.docker.io/library/busybox:stable", "--platform", "linux/amd64").AssertOK()
+
```
ktock, line 92: Could you use/update https://github.com/containerd/nerdctl/blob/c0adea84eb9aa77e8eeb4b0f73fbfe5455bacf57/pkg/testutil/testutil_linux.go#L32-L59

cmd/nerdctl/image_inspect_test.go
```diff
@@ -39,9 +44,136 @@ func TestImageInspectWithFormat(t *testing.T) {
...
+	// Build reference values for comparison
+	reference := inspectImageHelper(base, "busybox")
+	assert.Equal(base.T, 1, len(reference))
+	// Extract image sha
+	sha := strings.TrimPrefix(reference[0].RepoDigests[0], "busybox@sha256:")
+
+	differentReference := inspectImageHelper(base, "alpine")
+	assert.Equal(base.T, 1, len(differentReference))
+	// Extract image sha
+	// differentSha := strings.TrimPrefix(differentReference[0].RepoDigests[0], "alpine@sha256:")
```
ktock, line 114: Unneeded line?

cmd/nerdctl/image_inspect_test.go
```diff
@@ -39,9 +44,136 @@ func TestImageInspectWithFormat(t *testing.T) {
...
+	for _, id := range []string{"doesnotexist", "doesnotexist:either", "busybox:bogustag"} {
+		t.Logf("Testing %s", id+"@"+sha)
+		args := append([]string{"image", "inspect"}, id+"@"+sha)
+		cmdResult := base.Cmd(args...).Run()
+		assert.Equal(base.T, cmdResult.ExitCode, 0)
+		assert.Equal(base.T, cmdResult.Stdout(), "")
+	}
+
+	// Prove that invalid reference return no result without crashing
+	for _, id := range []string{"∞∞∞∞∞∞∞∞∞∞", "busybox:∞∞∞∞∞∞∞∞∞∞"} {
```
ktock, line 163: Can we use more readable string like `!` or a very long tag?

FYI: https://docs.docker.com/reference/cli/docker/image/tag/#description

> The tag must be valid ASCII and can contain lowercase and uppercase letters, digits, underscores, periods, and hyphens. It can't start with a period or hyphen and must be no longer than 128 characters.

pkg/cmd/image/inspect.go
```diff
@@ -19,58 +19,182 @@ package image
...
-	walker := &imagewalker.ImageWalker{
-		Client: client,
-		OnFound: func(ctx context.Context, found imagewalker.Found) error {
-			ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
-			defer cancel()
 
-			n, err := imageinspector.Inspect(ctx, client, found.Image, options.GOptions.Snapshotter)
+	// Set a timeout
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
```
ktock, line 103: Rather than hard-coded timeout, it should be cancellable from the user with a signal.

cmd/nerdctl/image_inspect_test.go
apostasie: I prefer something outside of the ASCII range if you don't mind, as this tends to uncover issues not necessarily seen without.

pkg/cmd/image/inspect.go
apostasie: The timeout was there already, hard-coded, with the same value.
I do agree we should do better, but maybe as a follow-up PR?

LMK your thoughts.

cmd/nerdctl/image_inspect_test.go
apostasie: Yes. Apologies.
Will fix.

cmd/nerdctl/image_inspect_test.go
apostasie: I need to be able to test shortnames and other variants that resolve to Docker Hub, not just the fully qualified (ghcr) names.

I can of course add them as well to `testutil_linux`.
Should I?

pkg/referenceutil/referenceutil.go
ktock: I don't think it doesn't need to be `Deprecated` as it's still used in the codebase. But I agree with ParseAnyReference should be used for necessary places.

pkg/cmd/image/inspect.go
ktock: OK, can be a following up.

cmd/nerdctl/image_inspect_test.go
ktock: > I need to be able to test shortnames and other variants that resolve to Docker Hub, not just the fully qualified (ghcr) names.

:+1:

> I can of course add them as well to testutil_linux.
> Should I?

Then I think it's OK as is.

ktock (approved):
LGTM on green

pkg/referenceutil/referenceutil.go
apostasie: Let me remove the deprecated tag.

Thanks a lot for your review @ktock 
Really appreciated!

Zheaoli (approved):
LGTM

AkihiroSuda:
@ktock @Zheaoli You can merge PRs

Issue #3019, opened 2024-05-18
image history issues


apostasie:
### Description

This ticket is meant to list all current issues with `nerdctl image history` when compared with `docker image history`.

* 1. [ ] fails if multiple tags point to the same image

Resolution here ties partly into the overall issue of image IDs (see #3011), and partly into the way we lookup images (which is currently being reworked in #3017).

These are fundamentally the same image, with different tags.

* 2. [ ] outputs a spurious "Snapshot" property

We can just remove this - or leave it if it serves a purpose.

* 3. [ ] does not has "CreatedAt"

Easy to do, as we already have the info (used for CreatedSince)

* 4. [ ] size formatting leaves to be desired
- spurious space between number and unit "XXX B"
- does not strip  irrelevant digits (eg: `0.0 B`)

Responsible package is in containerd (progress/humaans.go), which actually uses `docker/go-units` under the hood.
We are using it a few other places.
We could just drop the intermediary dependency and use go-units directly for better display.

* 5. [ ] does not list `ID`
This one also ties into #3011 and #3007, as might be more involved to get rid.

* 6. [ ] fails when using "format" and "quiet" together

* 7. [ ] does not implement the "human" flag - and overall doesn't display local time

* 8. [ ] "Snapshot" takes a lot of space with the table display, and should be truncated unless --no-trunc or --quiet

### Steps to reproduce the issue

For 1:
```
nerdctl pull debian
nerdctl tag debian testtag
nerdctl image history fac2c0fd33e8

> FATA[0000] 1 errors:
multiple IDs found with provided prefix: fac2c0fd33e8
```

For 2:
```
nerdctl image history debian | jq .Snapshot
```

For 3:
```
nerdctl image history debian | jq .CreatedAt
```

For 4:
```
nerdctl image history debian | jq .Size
```

For 5:
```
nerdctl image history debian | jq .ID
```

For 6:
```
nerdctl image history --format json --quiet debian
```

### Describe the results you received and expected

Expect to behave like docker.

I will send a PR that fixes the easy / non-controversial ones.
Let me know what you think about the other ones.

Thanks.

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Some of the issues fixed by #3020

PR #3020, opened 2024-05-18
Enhance image history

apostasie:
This fixes some of the issues described in #3019.

Specifically:
- implements the human flag
- display local dates
- implement CreatedAt
- enhances size display
- do not fail when format and quiet are passed together
- truncate snapshot unless quiet or no-trunc is passed

This PR stops using `containerd/pkg/progress` and instead goes directly to `docker/go-units` for better control.

Formatting was done in the image walker, and has been moved down to the printing part instead for better separation of concerns.

What is NOT fixed from #3019:
- any issue related to Id, or image lookup (still uses the walker for now)
- snapshot has been left untouched

### DIFF
--- a/cmd/nerdctl/image_history.go
+++ b/cmd/nerdctl/image_history.go
@@ -23,17 +23,18 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"text/tabwriter"
 	"text/template"
 	"time"
 
 	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/pkg/progress"
 	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/clientutil"
 	"github.com/containerd/nerdctl/v2/pkg/formatter"
 	"github.com/containerd/nerdctl/v2/pkg/idutil/imagewalker"
 	"github.com/containerd/nerdctl/v2/pkg/imgutil"
+	"github.com/docker/go-units"
 	"github.com/opencontainers/image-spec/identity"
 	"github.com/spf13/cobra"
 )
@@ -58,11 +59,16 @@ func addHistoryFlags(cmd *cobra.Command) {
 		return []string{"json"}, cobra.ShellCompDirectiveNoFileComp
 	})
 	cmd.Flags().BoolP("quiet", "q", false, "Only show numeric IDs")
+	cmd.Flags().BoolP("human", "H", true, "Print sizes and dates in human readable format (default true)")
 	cmd.Flags().Bool("no-trunc", false, "Don't truncate output")
 }
 
 type historyPrintable struct {
+	creationTime *time.Time
+	size         int64
+
 	Snapshot     string
+	CreatedAt    string
 	CreatedSince string
 	CreatedBy    string
 	Size         string
@@ -101,7 +107,7 @@ func historyAction(cmd *cobra.Command, args []string) error {
 			}
 			var historys []historyPrintable
 			for _, h := range configHistories {
-				var size string
+				var size int64
 				var snapshotName string
 				if !h.EmptyLayer {
 					if len(diffIDs) <= layerCounter {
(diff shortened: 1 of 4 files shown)


cmd/nerdctl/image_history.go
```diff
@@ -58,11 +59,16 @@ func addHistoryFlags(cmd *cobra.Command) {
 		return []string{"json"}, cobra.ShellCompDirectiveNoFileComp
 	})
 	cmd.Flags().BoolP("quiet", "q", false, "Only show numeric IDs")
+	cmd.Flags().BoolP("human", "H", true, "Print sizes and dates in human readable format (default true)")
```
AkihiroSuda, line 62: Please sync https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md

cmd/nerdctl/image_history.go
```diff
@@ -208,31 +218,47 @@ func printHistory(cmd *cobra.Command, historys []historyPrintable) error {
 	return nil
 }
 
-func (x *historyPrinter) printHistory(p historyPrintable) error {
+func (x *historyPrinter) printHistory(printable historyPrintable) error {
```
yankay, line 221: HI @apostasie 

Is it better to add a test case for it :-)

cmd/nerdctl/image_history.go
apostasie: > Please sync https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md

Yep, always forget about these. Sorry about that. Next push will have it.

cmd/nerdctl/image_history.go
apostasie: But but but but. Who needs tests! Testing is doubting! :D :D :D 


Ok - I'll have a look and get tests in with the next push. Thanks for keeping me honest :-).

apostasie:
Hey folks @yankay @AkihiroSuda 

- docs has been added for the `--human` flag
- added tests for image history overall

Pending CI happiness, let me know what you think and if you want some of that stuff modified.

apostasie:
Ok, windows is going to give me grief again.

How do you folks test locally on windows?

cmd/nerdctl/image_history_test.go
```diff
@@ -0,0 +1,145 @@
...
+	//     possibly one is unpacked on the filessystem while the other is the tar file size?
+	// - we do not truncate ids when --quiet has been provided
+	//     this is a conscious decision here - truncating with --quiet does not make much sense
+	testutil.DockerIncompatible(t)
+
+	base := testutil.NewBase(t)
+
+	// XXX the results here are obviously platform dependent - and it seems like windows cannot pull a linux image?
+	// Disabling for now - also ensuring that we are on arm64
+	if runtime.GOOS != "windows" && runtime.GOARCH == "arm64" {
```
AkihiroSuda, line 102:
```go
if runtime.GOOS != "linux" {
  t.Skipf("Needs linux")
}

if runtime.GOARCH != "amd64" {
  t.Skipf("Needs amd64")
}
```

apostasie: Done. Pending CI.

apostasie:
@AkihiroSuda @yankay though all your comments have been addressed, this is currently blocked, as `image history` fails to inspect a non-native platform (this should be a different bug, and this PR does not fix it) - thus making my newly introduced test fail.

I can weaken the test for now (eg: not checking `CreatedAt`) so that we can use a native image, and then when we fix this issue, we can reintroduce the more thorough test.

Thoughts?

apostasie:
@AkihiroSuda @yankay I disabled the test on select platforms and CI is green.
PTAL at you convenience.

Thanks a lot for your help!

yankay:
Thanks @apostasie 

/lgtm

apostasie:
> Thanks @apostasie
> 
> /lgtm

Wondering if "/lgtm" with the slash does act as intended.
Merging still show as blocked.
Could you try with just "LGTM"?

Thanks a ton @yankay !

yankay (approved):

AkihiroSuda (approved):
Thanks

PR #3021, opened 2024-05-18
Implement `inspect --size`, and Container SizeRootFs and SizeRw

### LINKED ISSUE #2194: inspect command is missing --size option

I was trying the `docker/logs-explorer-extension` with `nerdctl` instead of `docker` and found that it would call `nerdctl inspect '--format={{json .}}' --size=false ...`, which fails because `nerdctl` doesn't support `--size`.

The option is documented at https://docs.docker.com/engine/reference/commandline/inspect/#size

It is rather low priority, but it would be nice if the option was implemented for `docker` compatibility.

### Additional context

Note that adding support for `--size` alone would still not allow using the `docker/logs-explorer-extension` with `nerdctl`; it would also require #54 because it also runs `nerdctl events --filter type=container --format '{{json .}}'`

apostasie:
Fixes #2194

### DIFF
--- a/cmd/nerdctl/container_inspect.go
+++ b/cmd/nerdctl/container_inspect.go
@@ -38,6 +38,8 @@ func newContainerInspectCommand() *cobra.Command {
 		SilenceErrors:     true,
 	}
 	containerInspectCommand.Flags().String("mode", "dockercompat", `Inspect mode, "dockercompat" for Docker-compatible output, "native" for containerd-native output`)
+	containerInspectCommand.Flags().BoolP("size", "s", false, "Display total file sizes")
+
 	containerInspectCommand.RegisterFlagCompletionFunc("mode", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"dockercompat", "native"}, cobra.ShellCompDirectiveNoFileComp
 	})
@@ -71,10 +73,16 @@ func processContainerInspectOptions(cmd *cobra.Command) (opt types.ContainerInsp
 		return
 	}
 
+	size, err := cmd.Flags().GetBool("size")
+	if err != nil {
+		return
+	}
+
 	return types.ContainerInspectOptions{
 		GOptions: globalOptions,
 		Format:   format,
 		Mode:     mode,
+		Size:     size,
 		Stdout:   cmd.OutOrStdout(),
 	}, nil
 }
--- a/pkg/cmd/container/inspect.go
+++ b/pkg/cmd/container/inspect.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/snapshots"
 	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/containerd/nerdctl/v2/pkg/containerinspector"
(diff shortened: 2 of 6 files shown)


cmd/nerdctl/container_inspect.go
```diff
@@ -38,13 +38,16 @@ func newContainerInspectCommand() *cobra.Command {
 		SilenceErrors:     true,
 	}
 	containerInspectCommand.Flags().String("mode", "dockercompat", `Inspect mode, "dockercompat" for Docker-compatible output, "native" for containerd-native output`)
+	containerInspectCommand.Flags().BoolP("size", "s", false, "Display total file sizes")
```
AkihiroSuda, line 41: Could you also update https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-blue_square-nerdctl-inspect

cmd/nerdctl/container_inspect.go
```diff
@@ -38,13 +38,16 @@ func newContainerInspectCommand() *cobra.Command {
...
+
 	containerInspectCommand.RegisterFlagCompletionFunc("mode", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"dockercompat", "native"}, cobra.ShellCompDirectiveNoFileComp
 	})
 	containerInspectCommand.Flags().StringP("format", "f", "", "Format the output using the given Go template, e.g, '{{json .}}'")
 	containerInspectCommand.RegisterFlagCompletionFunc("format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"json"}, cobra.ShellCompDirectiveNoFileComp
 	})
+
```
AkihiroSuda, line 50: unnecessary

pkg/cmd/container/inspect.go
```diff
@@ -47,12 +50,45 @@ func Inspect(ctx context.Context, client *containerd.Client, containers []string
...
 type containerInspector struct {
-	mode    string
-	entries []interface{}
+	mode        string
+	size        bool
+	snapshotter snapshots.Snapshotter
+	entries     []interface{}
+}
+
+func resourceTotal(ctx context.Context, snapshotter snapshots.Snapshotter, resourceID string) (snapshots.Usage, snapshots.Usage, error) {
```
AkihiroSuda, line 64: Could you add a comment to explain the returned values?

pkg/cmd/container/inspect.go
```diff
@@ -47,12 +50,45 @@ func Inspect(ctx context.Context, client *containerd.Client, containers []string
...
+	var first snapshots.Usage
+	var total snapshots.Usage
+	next := resourceID
+
+	for next != "" {
```
AkihiroSuda, line 69: This could be written as `for next := resourceID; next != ""; next = info.Parent` ?

cmd/nerdctl/container_inspect.go
apostasie: I see.
I guess it is a matter of preference (I am fine with it either way - with a slight preference for separating returns)

Do we have a coding style here?
IIRC, golang-ci has a plugin to enforce these (one way or the other) and maybe we could add it?

cmd/nerdctl/container_inspect.go
apostasie: Yes.

pkg/cmd/container/inspect.go
apostasie: Yes.

pkg/cmd/container/inspect.go
apostasie: That seems fair to me. Will update.

apostasie: Fair.
Will update.

apostasie:
@AkihiroSuda all your comments integrated. PTAL.

Thanks a lot for all your help here.

AkihiroSuda (approved):
Thanks

Issue #2996, opened 2024-05-13
nerdctl does not create labels defined by image


apostasie:
### Description

This image here for example:
https://hub.docker.com/layers/dubodubonduponey/airplay/latest/images/sha256-b0cba69f13efc088530e96daefff5da4bd28031535ac974cfcd042d183128679?context=explore

Does define a number of LABEL.

When ran with Docker, these labels are created on the container.

Nerdctl does not.

### Steps to reproduce the issue

Just run the image:
```
docker run -d --name foo --entrypoint bash dubodubonduponey/airplay
sudo ./nerdctl run -d --name foo --entrypoint bash dubodubonduponey/airplay
```

Then inspect it:
```
docker inspect foo
sudo ./nerdctl inspect foo
```

### Describe the results you received and expected

Only the docker version shows the series of `org.opencontainers.image` labels.

Nerdctl should show them as well.

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #3023, opened 2024-05-20
feat: Support creating container with labels defined by image

yankay:
Support creating container with labels defined by image.
So, the image label can be inspected by `nerdctl inspect container-name`

fix: https://github.com/containerd/nerdctl/issues/2996


AkihiroSuda (approved):
Thanks

apostasie:
Thanks a lot @yankay !

Issue #3027, opened 2024-05-22
Snapshotter Stat and Usage calls are likely expensive and should be cached when dealing with many objects


apostasie:
### What is the problem you're trying to solve

Currently, whenever we list images (or containers), we are calling on snapshotter.Stat and snapshotter.Usage for every single layer.

This is likely a costly operation, that is also implementation dependent.

Potentially, there might be a lot of shared resources between images (eg: if they use the same base image for example).

This is related to #809 - albeit the benefits of a solution here would apply solely in cases where there is "duplication".

### Describe the solution you'd like

Our suggestion is to have a caching mechanism that implements the Snapshotter interface and can be used interchangeably.

PR coming soon.

apostasie:
Fixed by #3029

Issue #809, opened 2022-02-14
Benchmark and optimization of `nerdctl images`


AkihiroSuda:
>`nerdctl images` might be slower than before (which has been already slow).
>
>We should do some benchmark and probably have `--quick` (`--fast`?, idk) flag to skip slow size calculation

_Originally posted by @AkihiroSuda in https://github.com/containerd/nerdctl/issues/789#issuecomment-1038903152_

ningmingxiao:
I think  nerdctl image  -a or -o wide  should show blobsize,at default it is not necessary to show it.

fahedouch:
benchmark 

Only `Unpacked Size`:

```
ntimes -n 10 _output/nerdctl images

REPOSITORY    TAG       IMAGE ID        CREATED           PLATFORM       SIZE         BLOB SIZE
alpine        latest    21a3deaa0d32    24 hours ago      linux/amd64    5.9 MiB      0.0 B
debian        latest    fb45fd4e25ab    14 minutes ago    linux/amd64    134.5 MiB    0.0 B
ubuntu        latest    669e010b58ba    13 minutes ago    linux/amd64    77.9 MiB     0.0 B

real average: 164.16197ms, max: 185.0302ms, min: 137.0472ms, std dev: 16.171173ms
real 99 percentile: 185.0302ms, 95 percentile: 185.0302ms, 50 percentile: 165.57285ms
user average: 54.0097ms, max: 64.725ms, min: 41.953ms, std dev: 7.904296ms
sys  average: 99.1182ms, max: 119.68ms, min: 79.722ms, std dev: 13.591908ms
flaky: 0%
```

Only  `Blob Size`:

```
REPOSITORY    TAG       IMAGE ID        CREATED           PLATFORM       SIZE     BLOB SIZE
alpine        latest    21a3deaa0d32    24 hours ago      linux/amd64    0.0 B    2.7 MiB
debian        latest    fb45fd4e25ab    17 minutes ago    linux/amd64    0.0 B    52.4 MiB
ubuntu        latest    669e010b58ba    16 minutes ago    linux/amd64    0.0 B    27.2 MiB

real average: 157.88058ms, max: 236.3843ms, min: 142.0371ms, std dev: 28.001696ms
real 99 percentile: 236.3843ms, 95 percentile: 236.3843ms, 50 percentile: 149.58565ms
user average: 53.0145ms, max: 79.649ms, min: 43.332ms, std dev: 11.046308ms
[...]

Issue #325, opened 2021-08-26
nerdctl ps slows down and errors with 350+ containers


mac-chaffee:
I'm running a load test on my kubernetes cluster using [ClusterLoader2](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2), which just runs a bunch of `pause` containers on each node. When I have around 350 containers, the performance of `nerdctl ps` is affected compared to `ctr`: 

```
$ time sudo /usr/local/bin/nerdctl --debug-full -n k8s.io ps
FATA[0020] container "4690dc5561d03a5c89453546a3c5d7c0a7ce7c3938cac1560cf358a2c6c040e9" in namespace "k8s.io": not found

real	0m20.374s
user	0m0.284s
sys	0m0.117s
$ time sudo /usr/local/bin/nerdctl --debug-full -n k8s.io container ls
FATA[0034] container "4a08c9bdba3dc064ad82fbd583f992d409dd8ee3346bfd413ea010cae1f43030" in namespace "k8s.io": not found

real	0m34.376s
user	0m0.284s
sys	0m0.132s
$ time sudo ctr -n k8s.io c ls | wc -l
354

real	0m0.126s
user	0m0.081s
sys	0m0.091s
```

Also notice that the command fails due to a container being removed while the command was running (which becomes more likely the longer the command takes).

I think the race condition is caused by `ps.go` calling  `c.Spec` on each container after fetching the list of containers, meaning that if a container is removed before we can inspect it, the command will error. Could be fixed by skipping the removed container rather than erroring if the error is "not found": https://github.com/containerd/nerdctl/blob/cee3b6a4840db6b5dd4019ef343af7bf4ba5c940/cmd/nerdctl/ps.go#L118-L122

Not sure what to do about the performance issue though if we have to make `O(n)` requests to `Spec` each container. Maybe we could do some of those requests in parallel?

yardenshoham:
I looked a bit into this. In the Docker implementation, only one call to the daemon happens. Here:
https://github.com/docker/cli/blob/3dad26ca2d418092b8c4e01b03d0455d583bec86/cli/command/container/list.go#L122

In the nerdctl implementation, we make `O(n)` calls to `c.Spec` to achieve the same. My question: is `c.Spec` a call to containerd? If it's not, it shouldn't slow this operation down. The only place I'm sure makes `O(n)` calls to containerd is 
[...]

PR #3028, opened 2024-05-22
Snapshotter with caching

apostasie:
This is a proof of concept / draft for #3027

Average, very un-scientific measurements for`nerdctl images`:

On 500 identical images at 270MB:
- without cache: 5.2 seconds
- with cache: 3.3 seconds

On 500 identical images (buysbox, one layer, 4.174MB):
- without cache: 5.0 seconds
- with cache: 4.8 seconds

For about 250 containers (`nerdctl ps -as`) started from the same image (the big one above), results are in the same ballpark: about 50% gain with cache.

This PR also consolidates code and reduce duplication I did introduce... 

Let me know if this is something we should consider.

If yes, we can iterate on design / code location and possibly simplify the signature of `imgutil.SnapshotServiceWithCache` for eg.

### DIFF
--- a/pkg/imgutil/snapshotter.go
+++ b/pkg/imgutil/snapshotter.go
@@ -17,12 +17,15 @@
 package imgutil
 
 import (
+	"context"
 	"strings"
 
 	socisource "github.com/awslabs/soci-snapshotter/fs/source"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/mount"
 	ctdsnapshotters "github.com/containerd/containerd/pkg/snapshotters"
+	"github.com/containerd/containerd/snapshots"
 	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/containerd/nerdctl/v2/pkg/imgutil/pull"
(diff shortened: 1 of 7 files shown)


apostasie:
May have an impact on #809 and #325 when a lot of layers are shared.

pkg/imgutil/snapshotter.go
```diff
@@ -117,3 +120,71 @@ func stargzExtraLabels(f func(images.Handler) images.Handler, rFlags types.Remot
...
+}
+
+type CachingSnapshotter struct {
+	containerdSnapshotter snapshots.Snapshotter
+	statCache             map[string]snapshots.Info
+	usageCache            map[string]snapshots.Usage
+}
+
+func (snap *CachingSnapshotter) Stat(ctx context.Context, key string) (snapshots.Info, error) {
+	if stat, ok := snap.statCache[key]; ok {
```
ktock, line 139: `snapshots.Info.Updated` might be updated. I think this can't be cached.

apostasie: Hey @ktock ! Thanks for looking at this.

The snapshotter here is NOT meant to do caching across successive calls.

This is meant solely to be used when calling things `image list` and walking through the resulting collection of images, and NEVER for two successive, separate in time, calls.

Furthermore, there is no persistent caching mechanism - same reason.

ktock: Thanks for clarification. At least it should be documented that this should be used only for short term usage.

But why does it wrap the entire snapshotter API whereas it only patches Stat and Usage? I think this doesn't need to be a wrapper for the entire snapshotter and can be implemented at the caller.

apostasie: You welcome! Yes, I will document.

As for the other question: it does implement the Snapshotter interface so it can easily be swapped in / out in place of the containerd snapshotter.
Of course we can just implement a smaller struct, but that would change the signature.

I am fine with either approach.

ktock: >  Of course we can just implement a smaller struct, but that would change the signature.

SGTM

apostasie:
@ktock I am closing this in favor of #3029 which is no longer a draft and has a complete, cleaner version and caching optimizations that can benefit any method dealing with bulks of objects (`ps`, `images`).

PR #3029, opened 2024-05-22
[Proof of concept] Image list optimization

apostasie:
This is being detailed here: [#809](https://github.com/containerd/nerdctl/issues/809#issuecomment-2125926308)

It also contains the caching snapshotter from #3028, a caching ReadBlobber, and a full rewrite of image list replacing all containerd helper methods calls that tend to duplicate filesystem access with content ReadBlob where we need it.

There is some basic code instrumentation to measure execution time.

Any recommendations on a better way to measure that is of course welcome.

At this point, this is not ready to merge - needs some cleanup - it is provided here to gather feedback and see if this is worth pursuing or not.

Non-scientific testing showed the largest boost in cases where images share layers, or where several different tags point to the same image.

Against a bunch of random images from the Hub, this divide the execution time of printImages by about 3x.


pkg/cmd/container/inspect.go
```diff
@@ -61,39 +63,6 @@ type containerInspector struct {
 	entries     []interface{}
 }
 
-// resourceTotal will return:
-// - the Usage value of the resource referenced by ID
-// - the cumulative Usage value of the resource, and all parents, recursively
-// Typically, for a running container, this will equal the size of the read-write layer, plus the sum of the size of all layers in the base image
-func resourceTotal(ctx context.Context, snapshotter snapshots.Snapshotter, resourceID string) (snapshots.Usage, snapshots.Usage, error) {
```
apostasie, line 68: Promoting this method into `pkg/imgutil` as `ResourceUsage` as it is necessary both for `ps` and `images`.

pkg/cmd/container/list.go
```diff
@@ -107,6 +107,7 @@ func (x *ListItem) Label(s string) string {
 
 func prepareContainers(ctx context.Context, client *containerd.Client, containers []containerd.Container, options types.ContainerListOptions) ([]ListItem, error) {
 	listItems := make([]ListItem, len(containers))
+	snapshottersCache := map[string]containerdutil.Snapshotter{}
```
apostasie, line 110: Here, we are potentially dealing with multiple snapshotters, so, keeping them around for the duration of the function.

By the way, we are currently not dealing with mutiple different snapshotters for images. Should we?

pkg/cmd/image/list.go
```diff
@@ -241,28 +326,17 @@ func (x *imagePrinter) printImageSinglePlatform(ctx context.Context, img images.
...
 		CreatedSince: formatter.TimeSinceInHuman(img.CreatedAt),
 		Digest:       img.Target.Digest.String(),
 		ID:           img.Target.Digest.String(),
 		Repository:   repository,
 		Tag:          tag,
 		Name:         img.Name,
-		Size:         progress.Bytes(size).String(),
-		BlobSize:     progress.Bytes(blobSize).String(),
-		Platform:     platforms.Format(ociPlatform),
+		Size:         units.HumanSize(float64(size)),
```
apostasie, line 337: Uniformizing with docker, and what we do in other places.

pkg/cmd/image/list.go
```diff
@@ -188,50 +189,126 @@ type imagePrinter struct {
...
 	tmpl                                   *template.Template
 	client                                 *containerd.Client
-	contentStore                           content.Store
+	contentStore                           containerdutil.Provider
 	snapshotter                            snapshots.Snapshotter
 }
 
-func (x *imagePrinter) printImage(ctx context.Context, img images.Image) error {
-	ociPlatforms, err := images.Platforms(ctx, x.contentStore, img.Target)
+type Image struct {
```
apostasie, line 196: This and following is the biggest part of the refactoring of `images`.
As mentioned earlier, what we currently have is making many unnecessary calls / iterations.
This is closer to what an image is on the store, and minimizes reads and roundtrips to the store.

This Image struct here is ad-hoc - but we might consider making these methods into utils and revamping some of our abstractions later on.

pkg/containerdutil/content.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"context"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/content"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type Provider = content.Provider
+type ReaderAt = content.ReaderAt
```
apostasie, line 32: Copying the interfaces so that code using this package can avoid having to reference `containerd/content` altogether, minimizing possible confusion.

pkg/imgutil/imgutil.go
```diff
@@ -377,27 +377,36 @@ func ParseRepoTag(imgName string) (string, string) {
...
-	if err != nil {
-		return err
-	}
-
-	usage.Add(u)
+// ResourceUsage will return:
+// - the Usage value of the resource referenced by ID
+// - the cumulative Usage value of the resource, and all parents, recursively
+// Typically, for a running container, this will equal the size of the read-write layer, plus the sum of the size of all layers in the base image
+func ResourceUsage(ctx context.Context, snapshotter snapshots.Snapshotter, resourceID string) (snapshots.Usage, snapshots.Usage, error) {
```
apostasie, line 384: Function ported over from previous private implementation in container list.

pkg/imgutil/imgutil.go
```diff
@@ -409,23 +418,7 @@ func UnpackedImageSize(ctx context.Context, s snapshots.Snapshotter, img contain
...
-			return 0, nil
-		}
-		return 0, err
-	}
-
-	info, err := s.Stat(ctx, chainID)
-	if err != nil {
-		return 0, err
-	}
+	_, total, err := ResourceUsage(ctx, s, chainID)
```
apostasie, line 421: Simplifying all this code by relying on ResourceUsage instead.

apostasie:
@AkihiroSuda this one is good to go.
CI is green.

PTAL at your convenience.

pkg/cmd/container/list.go
djdongjin, line 110: I think it's still better to use containerd's `snapshots.Snapshotter` here, which is more common/well-known than `containerdutil.Snapshotter`.

pkg/cmd/image/list.go
```diff
@@ -188,50 +189,126 @@ type imagePrinter struct {
...
+		size:     size,
+		platform: plt,
+		config:   &manifest.Config,
+	}, nil
 }
 
-func makePlatformKey(platform v1.Platform) string {
-	if platform.OS == "" {
-		return "unknown"
+func ReadIndex(ctx context.Context, provider containerdutil.Provider, snapshotter snapshots.Snapshotter, desc ocispec.Descriptor) (map[string]*Image, error) {
```
djdongjin, line 257: Does this need to be exported?

pkg/cmd/image/list.go
```diff
@@ -188,50 +189,126 @@ type imagePrinter struct {
...
 	}
+	return nil, fmt.Errorf("nothing")
+}
 
-	image := containerd.NewImageWithPlatform(x.client, img, platMC)
-	desc, err := image.Config(ctx)
-	if err != nil {
-		log.G(ctx).WithError(err).Warnf("failed to get config of image %q for platform %q", img.Name, platforms.Format(ociPlatform))
+func (x *imagePrinter) printImage(ctx context.Context, img images.Image) error {
+	candidateImages, _ := Read(ctx, x.contentStore, x.snapshotter, img.Target)
```
ktock, line 300: Error shouldn't be ignored

pkg/cmd/image/list.go
```diff
@@ -188,50 +189,126 @@ type imagePrinter struct {
...
+	var config ocispec.Image
+	if err := json.Unmarshal(configData, &config); err != nil {
+		log.G(ctx).Error("Error unmarshaling config")
+		return nil, err
+	}
+
+	// If we are here, the image exists and is valid, so, do our size lookups
+
+	// Aggregate the blob size from the config and layers
+	blobSize := manifest.Config.Size
```
ktock, line 234: Can we add the size of manifest blob as well?

pkg/containerdutil/content.go
```diff
@@ -0,0 +1,75 @@
...
+func ContentStore(client *containerd.Client) Provider {
+	return &providerWithCache{
+		client.ContentStore(),
+		make(map[string]content.ReaderAt),
+	}
+}
+
+type providerWithCache struct {
+	native content.Provider
+	cache  map[string]content.ReaderAt
```
ktock, line 44: Closed `RedaerAt` shouldn't be reused. I don't think it's reusable.

pkg/containerdutil/snapshotter.go
```diff
@@ -0,0 +1,63 @@
...
+		client.SnapshotService(snapshotterName),
+		map[string]snapshots.Info{},
+		map[string]snapshots.Usage{},
+	}
+}
+
+type snapshotterWithCache struct {
+	snapshots.Snapshotter
+	statCache  map[string]snapshots.Info
+	usageCache map[string]snapshots.Usage
```
ktock, line 40: They can change on every read (e.g. Info.Updated). I don't think they can be cached.

apostasie: Thanks a lot for the review @ktock.

This is not meant to be used for *successive* operations.
The purpose of this PR and new objects overall is *solely* for use for single, atomic methods retrieving a bunch of objects at once (`nerdctl ps`, `merdctl images`), to avoid useless, successive reads of the same objects (as discussed in the linked issues and initial comment).

I did put a note here about that: https://github.com/containerd/nerdctl/pull/3029/files/8eba956203463c6e2ccce9afea8dae31a9149908#diff-c9a808f7f845733683410c3156f957135eba140a8783144049c9f2aea14251ccR17

Specifically, when multiple different tags point to the same image, it does not make sense to re-read `usage` once per tag (even if it would have changed between the time we read for the first tag)- and if we would, it does not make sense either to return different size information...

Hope that clarifies.

pkg/cmd/container/list.go
apostasie: Well, the point of this PR is specifically to solve performance issues with `ps` and `images`.
This method (`List`) is used solely by `ps` and is at the core of the issue.

Do you have specific concerns about `containerdutil.Snapshotter` that I could address?

Thanks for your review @djdongjin

pkg/cmd/image/list.go
apostasie: I believe as a follow-up PR we should move these methods to some `util` package and make use of them instead of what we are currently doing.
This will definitely benefit other commands (like `rmi`) which lookup methods are currently faulty (see for example #3016 (that should stay open btw)).

If you prefer me to un-export these for now, until we move these for more general use, I am of course happy to do that.

pkg/cmd/image/list.go
apostasie: Yep, not sure why it escaped me.
Thanks a lot for catching it.
Next push will fix it.

pkg/cmd/image/list.go
apostasie: Yes, I missed this one.
Thanks for catching it.
Next push will have it.

pkg/containerdutil/content.go
apostasie: In that case, given the code paths we are hitting here, I believe this is fine.
However, agreed it is risky, and you are right.
Next push will change that so we get a new ReaderAt every time - while still caching size.

apostasie:
- rebased against latest main
- resolved comments have been addressed
- answers provided on the three remaining (unresolved) comments - lmk your thoughts there

Thanks again for your help!

@ktock @djdongjin PTAL at your convenience.

pkg/containerdutil/content.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/content"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type Provider = content.Provider
+type ReaderAt = content.ReaderAt
```
djdongjin, line 32: nit: I think it's okay to just use `content.Provider/ReaderAt`. It's more clear that we're just use the original containerd interface.

(i.e., if these aliases are used in other nerdctl pkgs, it's not clear it's a nerdctl-customized `Provider` interface or just aliases.) Same for other type alias.

pkg/containerdutil/content.go
```diff
@@ -0,0 +1,79 @@
...
+
+// ContentStore should be called to get a Provider with caching
+func ContentStore(client *containerd.Client) Provider {
```
djdongjin, line 35: Suggest to call this `NewProvider`. `content.Store` is a separate interface.

pkg/containerdutil/helpers.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+
+	"github.com/containerd/containerd/content"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type ReadBlob func(ctx context.Context, provider Provider, desc ocispec.Descriptor) ([]byte, error)
+
+func WithCache() ReadBlob {
```
djdongjin, line 28: The name is not very clear. It's used as `containerdutil.WithCache` but not sure what the return can do. Suggest to call it `ReadBlobWithCache`.

pkg/cmd/container/list.go
djdongjin: I mean `containerdutil.Snapshotter` is just an alias to `snapshots.Snapshotter` from containerd right? Personally I prefer to use the original containerd interface directly instead of alias.

When aliases are used in other packages (e.g., here). It's not clear if `containerdutil.Snapshotter` is some nerdctl-customized interface or it's the same as containerd `snapshots.Snapshotter`. Also leave the comment somewhere else.

djdongjin (commented):
Thanks for the contribution. Left some other comments.

Also this PR introduces caches in 3 places `providerWithCache`/`WithCache()`/`snapshotterWithCache`, which one contributes the largest improvement?

pkg/cmd/container/list.go
apostasie: Oh I see your point now.
Misread your initial comment. Thanks for clarifying.
And point taken in the other comment.
I am fine removing these aliases and using the containerd interfaces instead.

apostasie:
> Also this PR introduces caches in 3 places `providerWithCache`/`WithCache()`/`snapshotterWithCache`, which one contributes the largest improvement?


Thanks a lot @djdongjin 

There are two major angles here:
1. the fact that we rely on a succession of calls to containerd methods that will independently re-read (or stat) the same files multiple times (see for example https://github.com/containerd/nerdctl/issues/809#issuecomment-2125926308 which provides details)
2. the fact that when dealing with numerous objects, there usually is duplication (shared layers, or even "same" image with different tags)

For containers, the current PR only enables cache when `--size` is being passed.

In that case, the benefit comes from `snapshotter`, as each layer of every container triggers a call to `snapshotter.Usage` and to `snapshotter.Stat`.
Benefit is then directly correlated to how much layers are shared across containers (whether because they are from the same image, or from distinct images sharing a common base).

For images, the full rewrite of the `images` logic is actually remediating **most** of point 1 (duplicated IO due to containerd methods) by bypassing most of them entirely.

Nevertheless, getting the info for a typical image still requires as far as I can tell: reading the index, the config, the platform specific manifest (3 ReaderAt/Size, 3 ReadBlob), and as many calls to `snapshotter.Usage` and `snapshotter.Stat` as there is layers.

So, to summarize:
- as soon as you have layers duplication (point 2), the snapshotter is what gives you most bang for the buck
- if our lookups have not been carefully rewritten to avoid calling containerd helpers (point 1), you do get most of the benefit from caching of `ReadBlob` and `ReaderAt`.

One more note here (I will also address your inline comments).
Unfortunately, my latest push did neuter some of the benefits we get from caching the Size of ReaderAt, as a reviewer raised the very valid point that we shouldn't keep io Readers after Close in the cache. 
I will rewrite this part imminently so that we get that benefit back.

pkg/containerdutil/helpers.go
apostasie: Agreed.
Will come with next push.

pkg/containerdutil/content.go
apostasie: Coming with next push.

pkg/cmd/image/list.go
apostasie: Un-exported with next push.
We can always make them public when need be later on.

apostasie:
@djdongjin addressed all your inline review comments with this push.

Note that I did change the implementation of the ReaderAt cache as mentioned above in my comment.
ReaderAt are now fully cached again, and remove themselves from the cache on `Close`.

pkg/cmd/image/list.go
```diff
@@ -188,50 +190,129 @@ type imagePrinter struct {
...
-func (x *imagePrinter) printImage(ctx context.Context, img images.Image) error {
-	ociPlatforms, err := images.Platforms(ctx, x.contentStore, img.Target)
+type image struct {
+	blobSize int64
+	size     int64
+	platform platforms.Platform
+	config   *ocispec.Descriptor
+}
+
+var ReadBlob = containerdutil.ReadBlobWithCache()
```
djdongjin, line 204: can we make this part of `imagePrinter` and pass it to `read/readManifest/readIndex` as an argument? I think that's better than global var?

pkg/cmd/image/list.go
```diff
@@ -188,50 +190,129 @@ type imagePrinter struct {
...
+		if err != nil {
+			return nil, err
+		}
+		descs := map[string]*image{}
+		descs[platforms.FormatAll(manifest.platform)] = manifest
+		return descs, nil
+	case images.MediaTypeDockerSchema2ManifestList, ocispec.MediaTypeImageIndex:
+		return readIndex(ctx, provider, snapshotter, desc)
 	}
+	return nil, fmt.Errorf("nothing")
```
djdongjin, line 297: `fmt.Errorf("unknown media type: %s", desc.MediaType)`

pkg/cmd/image/list.go
apostasie: Mmm...

What about instead we instanciate ReadBlob cache inside containerdutil/helpers and use it just as a drop in replacement for content.ReadBlob?

The point is: ReadBlob is used solely for reading image manifest, index and config objects, which never change for a given digest.
There is no reason to ever invalidate the cache on this one.

pkg/cmd/image/list.go
apostasie: Yep... good one, thanks.

pkg/cmd/image/list.go
apostasie: Doing that ^ with the latest push (also with your other comment fixed).
Let me know what you think (I will leave this comment unresolved for now).

apostasie: Pushed.

PTAL at your convenience.

Thanks @djdongjin !

apostasie: @djdongjin also adding another minor change: https://github.com/containerd/nerdctl/compare/1b401474ae01df7956eeb51c837c9b33f0d7ab55..80731ff39a21db7b9b58d4faebaf6ef8009a5cc8

Replacing testing media types with using existing containerd helpers for that.
Makes it slightly more future proof.

pkg/containerdutil/helpers.go
```diff
@@ -0,0 +1,45 @@
...
+package containerdutil
+
+import (
+	"context"
+
+	"github.com/containerd/containerd/content"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+var ReadBlob = readBlobWithCache()
```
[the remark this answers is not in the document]
apostasie: On ReadBlob specifically, it does not matter, and is safe to use in any and all cases.

apostasie:
Thanks a lot for all your help @djdongjin 

@ktock PTAL at your convenience.

cc @AkihiroSuda @yankay

pkg/containerdutil/helpers.go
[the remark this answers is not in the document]
djdongjin: why is that? If it's imported and used by a long running process, there is no mechanism to clean up the underlying cache.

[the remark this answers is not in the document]
apostasie: ReadBlob only ever reads image manifest files, from their `desc.Digest`.
They are content-addressable and never change.
Even containerd allows for that behavior by returning directly from `desc.Data` if it is set, instead of actually reading the file.

In that context, the behavior of our caching mechanism is the same as if we would keep descriptors around (with the `Data` property set) - if the underlying object gets removed, your descriptor is dangling and will gladly return data for something that doesn't exist anymore.

Nevertheless, yeah, I ll add the note to document that behavior - you are right.

Thank you @djdongjin .

AkihiroSuda (approved):
Thanks

PR #3034, opened 2024-05-24
[WORK IN PROGRESS] CI

apostasie:
Ignore for now.


apostasie:
I am going to close this for now.
I want to focus first on cleaning up our tests and race conditions so that we can paralelize, then we can come back to this.

PR #3041, opened 2024-05-30
CI: .md files changes should not trigger tests

apostasie:


.github/workflows/test.yml
```diff
@@ -6,6 +6,9 @@ on:
       - main
       - 'release/**'
   pull_request:
+    paths-ignore:
+      - '**/*.md'
```
sondavidb, line 10: I think you can just do '**.md' to account for all markdown files in the project

apostasie: Always confused about globs - where they work which way.
Pushing now.
Thanks!

apostasie:
@sondavidb fixed

AkihiroSuda (approved):
Thanks

PR #3045, opened 2024-05-31
Support CI on ARM64 

yankay:
To support the ARM64 better, add it to the CI.  And to make the CI pass, there are some changes:

* Because the [fluentd](https://hub.docker.com/r/fluent/fluentd/tags?page=&page_size=&ordering=&name=v1.14) image is not a multi-arch image, so using the arm64 image name
* Because the arm64 system is compatible with armv7, the multi_platform_linux_test.go behavior is different.

### DIFF
--- a/pkg/testutil/testutil_linux.go
+++ b/pkg/testutil/testutil_linux.go
@@ -40,7 +40,7 @@ var (
 	WordpressIndexHTMLSnippet   = "<title>WordPress &rsaquo; Installation</title>"
 	MariaDBImage                = mirrorOf("mariadb:10.5")
 	DockerAuthImage             = mirrorOf("cesanta/docker_auth:1.7")
-	FluentdImage                = mirrorOf("fluent/fluentd:v1.14-1")
+	FluentdImage                = "fluent/fluentd:v1.17.0-debian-1.0"
 	KuboImage                   = mirrorOf("ipfs/kubo:v0.16.0")
 	SystemdImage                = "ghcr.io/containerd/stargz-snapshotter:0.15.1-kind"
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
         run: go test -v ./pkg/...
 
   test-integration:
-    runs-on: "ubuntu-${{ matrix.ubuntu }}"
+    runs-on: "${{ matrix.runner }}"
     timeout-minutes: 40
     strategy:
       fail-fast: false
@@ -64,10 +64,16 @@ jobs:
         include:
           - ubuntu: 20.04
             containerd: v1.6.33
+            runner: "ubuntu-20.04"
           - ubuntu: 22.04
             containerd: v1.7.19
+            runner: "ubuntu-22.04"
           - ubuntu: 24.04
             containerd: v2.0.0-rc.3
+            runner: "ubuntu-24.04"
+          - ubuntu: 24.04
+            containerd: v2.0.0-rc.3
+            runner: github-arm64-2c-8gb
     env:
       UBUNTU_VERSION: "${{ matrix.ubuntu }}"
       CONTAINERD_VERSION: "${{ matrix.containerd }}"


yankay:
ref: https://github.com/containerd/containerd/issues/10466

Dockerfile
```diff
@@ -356,4 +356,8 @@ FROM test-integration AS test-integration-ipv6
 CMD ["gotestsum", "--format=testname", "--rerun-fails=2", "--packages=github.com/containerd/nerdctl/v2/cmd/nerdctl/...", \
   "--", "-timeout=30m", "-args", "-test.kill-daemon", "-test.ipv6"]
 
+FROM test-integration AS test-integration-arm64
```
AkihiroSuda, line 359: Doesn't need to be a separate stage AFAICS

cmd/nerdctl/container_run_linux_test.go
```diff
@@ -398,6 +398,13 @@ func TestRunSigProxy(t *testing.T) {
 	}
 }
 
+func fluentdImage() string {
+	if runtime.GOARCH == "arm64" {
+		return "fluent/fluentd:v1.14-arm64-debian"
```
AkihiroSuda, line 403: This should be defined as a const in `testutil`

cmd/nerdctl/multi_platform_linux_test.go
```diff
@@ -159,6 +166,10 @@ RUN uname -m > /usr/share/nginx/html/index.html
 		"http://127.0.0.1:8081": "aarch64",
 		"http://127.0.0.1:8082": "armv7l",
 	}
+	// using aarch64 instead of armv7l because the arm64 system is compatible with armv7.
```
apostasie, line 169: Hey @yankay I do not understand why we should do that / what is it bringing us.

yankay: HI @apostasie 

Ref to https://github.com/containerd/containerd/issues/10466. The `ctr` has the same behavior as `nerdctl`.
When we run armv7 on x86, the `binfmt` uses qemu to emulate. So the `uname -m` commands show `armv7` .
But when we run armv7 on aarch64, it does not need to emulate the Linux runs the binary directly. So the `uname -m` commands show `aarch64` .

Ref to: https://github.com/docker/buildx/issues/198

apostasie: My host is on arm64 (with qemu installed, as we do on the CI):
```
$ uname -m
aarch64
$ nerdctl run --rm --platform arm alpine uname -m
armv7l
```

apostasie: Tonis comment (in https://github.com/docker/buildx/issues/198#issuecomment-561252903 ) makes it clear that this is without qemu.

But here *we do* enable qemu - which is why I am confused at how this could work (especially since it does not, locally)

yankay: MY test environment is:
```
uname -a
Linux iZj6c145pjnolrzzpcuvx2Z 5.15.0-76-generic #83-Ubuntu SMP Thu Jun 15 19:21:56 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
```
Does your host MacBook arm64 ?

apostasie: Clearly the call to binfmt on the CI did not install the arm emulator, while it did on my machine. As to the why, my money is on the fact that the M1 may not have the magic native execution fallback thing.

Anyhow:
a. we should pin toniis image instead of using latest
b. looking at the source of it, it has clearly grown in complexity and trying too hard to do "magic" IMHO - `--install all` clearly no longer means `--install all`, it is now "install what it thinks you need". I would suggest we stop using it entirely or at least we stop doing `--install-all` and we instead explicitly install the architectures we need

yankay: Thanks @apostasie . It works. 👍

pkg/testutil/testutil_linux.go
```diff
@@ -40,7 +40,7 @@ var (
 	WordpressIndexHTMLSnippet   = "<title>WordPress &rsaquo; Installation</title>"
 	MariaDBImage                = mirrorOf("mariadb:10.5")
 	DockerAuthImage             = mirrorOf("cesanta/docker_auth:1.7")
-	FluentdImage                = mirrorOf("fluent/fluentd:v1.14-1")
+	FluentdImage                = "fluent/fluentd:v1.17.0-debian-1.0"
```
apostasie, line 43: I assume we generally avoid Docker Hub because of the dreaded 429? (which I guess is why we mirror stuff on ghcr?)
I also think this approach is problematic (as obviously publishing there is a process we cannot control from this repo).
I would prefer a solution where we cache the images on the CI.
Also, we might want to start pulling by digest for these.

Anyhow, this is fine ^ - and these problems can be addressed later.

apostasie (approved):
LGTM @yankay !

Glad we are testing on arm64!

apostasie:
@AkihiroSuda this here looks fine to me now (no more platform specialization / monkeying with the binfmt force install patch).

Do you think we could merge this?

apostasie:
@fahedouch @djdongjin 

Akihiro seems to be busy / out in the past few days.

Do you think we could get this reviewed?

djdongjin:
I can take a look today

djdongjin (approved):

AkihiroSuda (approved):
Thanks, merging, but please consider using GHCR to avoid rate limit

PR #3050, opened 2024-06-02
Remove spurious debugging leftover

apostasie:
These were inadvertently committed in during image history cleanup.

Apologies for that.

cc @AkihiroSuda


AkihiroSuda (approved):
Thanks

Issue #2987, opened 2024-05-11
docker / nerdctl inconsistent value type for Labels returned by `container list --format json`


apostasie:
### Description

docker returns Labels as a string with coma separated `key=value`, while nerdctl return them as an object.

I appreciate that nerdctl wants to mimic docker behavior as much as possible and I am usually on board with that, but in that case, I think Docker is just wrong and nerdctl should keep the current behavior, even at the cost of deviating.

I am opening the issue so that we can have a discussion on this and decide on a resolution.

My vote obviously is to just document it and not change the current behavior.

Thoughts?

### Steps to reproduce the issue

1. docker container list --all --format json
2. sudo ./nerdctl container list --all --format json

### Describe the results you received and expected

Docker returns:
```
{
  "Labels": "org.opencontainers.image.version=f40ece5,org.opencontainers.image.documentation=https://github.com/dubo-dubon-duponey/docker-dns/blob/master/README.md,org.opencontainers.image.ref.name=bookworm-2024-03-01,org.opencontainers.image.revision=f40ece567f4d6a0b1c8249f32ab0a82f5594efca,org.opencontainers.image.created=2024-03-27T12:15:26-0700,org.opencontainers.image.licenses=The MIT License (MIT),org.opencontainers.image.source=https://github.com/dubo-dubon-duponey/docker-dns/tree/master,org.opencontainers.image.vendor=dubodubonduponey,org.hadron.core.version=v0.1-dev,org.hadron.plan.tag=2024/04/02-17:19:16-5F6228B8-4152-4C7B-8D44-4E9293B1C59D,org.hadron.plan.sha=3c8d9485ffe7f0c91b837942b98be71d3d65a632029ae579143387eda499a137,org.opencontainers.image.authors=Dubo Dubon Duponey <dubo-dubon-duponey@farcloser.world>,org.opencontainers.image.description=So image. Much DBDBDP. Such description.,org.opencontainers.image.title=A DBDBDP image,org.opencontainers.image.url=https://github.com/dubo-dubon-duponey/docker-dns,org.hadron.plan.description=some_plan_descriptor,org.hadron.plan.name=plan_name"
}
```

Nerdctl returns:

```
  "Labels": {
    "containerd.io/restart.loguri": "binary:///home/apo/nerdctl-patch?_NERDCTL_INTERNAL_LOGGING=%2Fvar%2Flib%2Fnerdctl%2F1935db59",
    "containerd.io/restart.policy": "always",
[...]

PR #3058, opened 2024-06-05
Fix: inconsistent value type for Labels returned by `container list --format json` 

yankay:
fix: https://github.com/containerd/nerdctl/issues/2987

### DIFF
--- a/pkg/cmd/container/list.go
+++ b/pkg/cmd/container/list.go
@@ -97,12 +97,14 @@ type ListItem struct {
 	Status    string
 	Runtime   string // nerdctl extension
 	Size      string
-	Labels    map[string]string
+	Labels    string
+	LabelsMap map[string]string `json:"-"`
+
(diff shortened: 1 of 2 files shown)


pkg/cmd/container/list.go
```diff
@@ -97,14 +97,10 @@ type ListItem struct {
...
 	Size      string
-	Labels    map[string]string
+	Labels    string
 	// TODO: "LocalVolumes", "Mounts", "Networks", "RunningFor", "State"
 }
 
-func (x *ListItem) Label(s string) string {
-	return x.Labels[s]
-}
-
```
yankay, line 107: HI @AkihiroSuda 

The `Label` method is added by https://github.com/containerd/nerdctl/commit/2af4cef9e7e7ca2f36b3f81aa33f10b619e51e6b#diff-81c503cae22717c2cb5fb3e7c1dbee8d9d2a82d746f05d9b115cd251178e19dbR104 .

It not be used, so it may be removed :-)

PR #3074, opened 2024-06-07
Add --pull option to build command

### LINKED ISSUE #3073: Allow option to use locally cached images when building images

When building an image, I sometimes use local images that are not up to date with the tagged version in remote, but nerdctl will automatically pull from remote when it is out of date.

I would like to be able to choose whether or not nerdctl pulls from remote. Currently, it uses buildkit's default, which for me is pulling from remote.

### Additional context

_No response_

sondavidb:
Fixes #3073 

`docker build` initially had a `--pull` command that forced remote image lookup when building an image. This has been changed when docker switched to buildkit to be default behavior, with an [option](https://github.com/moby/buildkit/blob/efcde2fcfbc4299ad8f07727fd65562409608c88/frontend/dockerui/attr.go#L30) to not check for updates if the image is already cached.

Forcing `nerdctl build` to use local images may be desirable for a few edge cases, so I've implemented a solution via a `--pull` flag in `nerdctl build` that can be set to `false` to enable this behavior, and `true` to always pull from remote. Without it, it will use the buildkit default.

There's a caveat here, which is that buildkit must be set up to use the same namespace as the locally cached image, which the user must set up in their buildkit config file.

Tested in a similar workflow as the integration test:
1. Have a Dockerfile with just `FROM ubuntu:latest`
2. Pull an older image tag (for me, I chose `ubuntu:18.04`) into your buildkit namespace (default should be `buildkit`)
3. Tag it to latest version
4. `nerdctl build --pull false .`
5. Check image sha in output, it should be the same as the one locally cached
6. `nerdctl build --pull true .`
7. Check image sha in output, it should be a different one and should match the same as latest tag.

### DIFF
--- a/cmd/nerdctl/builder_build.go
+++ b/cmd/nerdctl/builder_build.go
@@ -51,6 +51,7 @@ If Dockerfile is not present and -f is not specified, it will look for Container
 	buildCommand.Flags().StringP("output", "o", "", "Output destination (format: type=local,dest=path)")
 	buildCommand.Flags().String("progress", "auto", "Set type of progress output (auto, plain, tty). Use plain to show container output")
 	buildCommand.Flags().String("provenance", "", "Shorthand for \"--attest=type=provenance\"")
+	buildCommand.Flags().Bool("pull", false, "On true, always attempt to pull latest image version from remote. Default uses buildkit's default.")
 	buildCommand.Flags().StringArray("secret", nil, "Secret file to expose to the build: id=mysecret,src=/local/secret")
 	buildCommand.Flags().StringArray("allow", nil, "Allow extra privileged entitlement, e.g. network.host, security.insecure")
 	buildCommand.RegisterFlagCompletionFunc("allow", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
@@ -133,6 +134,14 @@ func processBuildCommandFlag(cmd *cobra.Command, args []string) (types.BuilderBu
 	if err != nil {
 		return types.BuilderBuildOptions{}, err
 	}
+	var pull *bool
+	if cmd.Flags().Changed("pull") {
+		pullFlag, err := cmd.Flags().GetBool("pull")
+		if err != nil {
+			return types.BuilderBuildOptions{}, err
+		}
+		pull = &pullFlag
+	}
 	secret, err := cmd.Flags().GetStringArray("secret")
 	if err != nil {
 		return types.BuilderBuildOptions{}, err
@@ -205,6 +214,7 @@ func processBuildCommandFlag(cmd *cobra.Command, args []string) (types.BuilderBu
 		BuildArgs:            buildArgs,
 		Label:                label,
 		NoCache:              noCache,
+		Pull:                 pull,
 		Secret:               secret,
 		Allow:                allow,
 		Attest:               attest,
--- a/cmd/nerdctl/builder_linux_test.go
+++ b/cmd/nerdctl/builder_linux_test.go
@@ -19,9 +19,13 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
 	"testing"
 
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"gotest.tools/v3/assert"
 )
 
 func TestBuilderDebug(t *testing.T) {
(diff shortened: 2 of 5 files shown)


sondavidb:
Does buildkit output to stderr? Integration tests were failing with `AssertOutContains` and passed with `AssertErrContains`, but it seems strange that buildkit is outputting to stdout instead of stderr. Makes me worried that I did something wrong.

Zheaoli:
FYI https://github.com/docker/buildx/issues/802

cmd/nerdctl/builder_build.go
```diff
@@ -51,6 +51,7 @@ If Dockerfile is not present and -f is not specified, it will look for Container
 	buildCommand.Flags().StringP("output", "o", "", "Output destination (format: type=local,dest=path)")
 	buildCommand.Flags().String("progress", "auto", "Set type of progress output (auto, plain, tty). Use plain to show container output")
 	buildCommand.Flags().String("provenance", "", "Shorthand for \"--attest=type=provenance\"")
+	buildCommand.Flags().String("pull", "", "On true, always attempt to pull latest image version from remote. Default uses buildkit's default.")
```
AkihiroSuda, line 54: Could you also update https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-build

cmd/nerdctl/builder_build_test.go
```diff
@@ -592,3 +592,58 @@ func TestBuildAttestation(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestBuildWithPull(t *testing.T) {
+	testutil.RequiresBuild(t)
+
+	oldImage := "ubuntu:18.04"
```
AkihiroSuda, line 599: Any reason to use this very old version of Ubuntu?

cmd/nerdctl/builder_build_test.go
```diff
@@ -592,3 +592,58 @@ func TestBuildAttestation(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			base := testutil.NewBase(t)
+			defer base.Cmd("builder", "prune").AssertOK()
+			base.Cmd("image", "prune", "--force", "--all").AssertOK()
+
+			base.Cmd("--namespace=buildkit", "pull", oldImage)
```
Zheaoli, line 624: I think we don't need `namespace` here

cmd/nerdctl/builder_build_test.go
sondavidb: Just for testing purposes. The idea is to pull an old version of ubuntu, tag it as latest, and see if `--pull=false` uses local image while `--pull=true` uses the remote image (latest vesion).

sondavidb:
Last force-push revamped the integration test, so it should work properly. IMO it's a little bit ugly, since we're using `os.Exec()` to write directly to the config, but it's the only way I could think of approaching this, as buildctl doesn't support anything similar to a `--config` or a way to force the containerd worker instead of the OCI worker.

cmd/nerdctl/builder_build.go
ktock, line 54: Can we make it a bool flag and just check if it's set when parsing?

cmd/nerdctl/builder_build_test.go
ktock: Can we just use the mirror registry, with updating the tag with another image before checking --pull behaviror?

pkg/api/types/builder_types.go
```diff
@@ -69,6 +69,9 @@ type BuilderBuildOptions struct {
 	ExtendedBuildContext []string
 	// NetworkMode mode for the build context
 	NetworkMode string
+	// Pull determines if we should attempt to pull the latest image from remote.
+	// "true" will pull from remote while "false" will always use locally cached image
+	Pull string
```
ktock, line 74: Can we just make it *bool if we only take true/false/default ?

ktock (changes requested):
See comments

cmd/nerdctl/builder_build.go
sondavidb: Is this possible? From my understanding, the bool flag can only be true or false, and if none are provided then it will just set a default value.

Once parsed from a string, though, I can store it into a *bool instead of a string.

ktock: Have you tried `Changed`? https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Changed

cmd/nerdctl/builder_build_test.go
sondavidb: Makes sense, I switched to using much slimmer images as well so this test will now run a little bit faster

cmd/nerdctl/builder_build.go
sondavidb: This is perfect, thanks. Just pushed a new set of changes.

sondavidb:
Had to move the test to `builder_linux_test.go` to make use of testutil. It probably makes more logical sense there, anyway.

cmd/nerdctl/builder_build.go
```diff
@@ -51,6 +51,7 @@ If Dockerfile is not present and -f is not specified, it will look for Container
 	buildCommand.Flags().StringP("output", "o", "", "Output destination (format: type=local,dest=path)")
 	buildCommand.Flags().String("progress", "auto", "Set type of progress output (auto, plain, tty). Use plain to show container output")
 	buildCommand.Flags().String("provenance", "", "Shorthand for \"--attest=type=provenance\"")
+	buildCommand.Flags().Bool("pull", true, "On true, always attempt to pull latest image version from remote. Default uses buildkit's default.")
```
ktock, line 54: Default value should be false.

cmd/nerdctl/builder_linux_test.go
```diff
@@ -36,3 +40,82 @@ CMD ["echo", "nerdctl-builder-debug-test-string"]
...
+namespace = "%s"`, testutil.Namespace)
+
+	cleanup := useBuildkitConfig(t, buildkitConfig)
+	defer cleanup()
+
+	testCases := []struct {
+		name string
+		pull bool
+	}{
+		{
```
ktock, line 66: Can you add a test case for the default behaviour? (i.e. not specifying --pull)

sondavidb: Done

ktock (approved):
Thanks

Issue #3072, opened 2024-06-07
Login & authentication related issues


apostasie:
### Description

This is meant to collect all issues impacting login or overall authn/authz

[ ] #3044
[ ] #3046
[ ] #3047
[ ] #3052
[ ] #3068
[ ] #3069
[ ] #3070
[ ] #3071

Unconfirmed / need assessment:

[ ] #920
[ ] #427
[ ] #1675
[ ] #2840
[ ] #2844

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
@AkihiroSuda if you feel this is helpful, just assign any login related issue to me (including this one).
I am going with a rewrite and would like to fix them all.

PR #3076, opened 2024-06-09
Revamp registry test

### LINKED ISSUE #3044: TestLoginWithPlainHttp appears to be broken
### Description

The test *always* fails locally for me.

At this point, I believe it never worked.
The issue is that the auth server is configured over https, while the registry is on plain http.
Then, since the CA is not trusted, requests to the auth server will all fail.

`level=error msg="failed to call tryLoginWithRegHost" error="failed to call rh.Authorizer.Authorize: failed to fetch oauth token: Post \"https://10.4.0.1:5001/auth\": tls: failed to verify certificate: x509: certificate signed by unknown authority" i=0`

This was referenced in passing in #2607

If the registry used was over https, the CA gets appended to the tlsconfig roots, and likely reused when contacting the auth server.

However, on a retry during the same overall run, the test apparently passes on the CI.
I am not completely sure why at this point - my gut is that some other test (using the HTTPS registry) are not isolated properly from this one, and somehow the CA is still trusted afterwards?

Anyhow, this is clearly one of these cases where retrying a test is actually a bad idea.

We need to fix this test which seems to be doing nothing right now.
There is also a large amount of duplication in `testregistry_linux.go` which may be responsible for part of the confusion here.

I am also interested in better understanding why we apparently can't specify a `hosts.toml` for the auth server (from a cursory look, that might be somewhere in `containerd/remotes/docker/resolver`).
Appreciate that a token server is not a registry, but then, how do we specify a CA for these then (beside the obvious, adding to the system store, which is a non-starter)?
Maybe the use-case is nonsensical anyhow... (http registry + https auth server, or both over https using different third-party CAs...)

Finally, I do believe that retrying tests until they work is a bad idea - if a test fails, it should be looked into - if it's flaky, it should be fixed - retrying just ignores the underlying problem and makes things harder to understand all while creating weird side-effects and giving us a false confidence that tests are actually doing something while they very well might not. I do appreciate that flakyness is a PITA, but I do not see a better solution for it than to get to the bottom of it one by one...

Run integration tests (specifically `cmd/nerdctl/login_linux_test.go`)

Fails every time.

1.7.6

None

### Host information

_No response_

### LINKED ISSUE #3071: Tests manipulating credentials are not properly isolated from one another
### Description

All test manipulating credentials are loading and saving into the default docker config location.

This increases significantly the risk of cross-test pollution, false positives / negatives, and definitely prevents parallelization of tests.

Each test that relies on reading or saving credentials must use a dedicated temp location instead.

n.a.

n.a.

1.7.6

None

### Host information

_No response_

apostasie:
The fundamental motivation for this is to fix #3044 

Furthermore, before embarking on revamping login code to fix everything in #3072 we do need a clean working ready-to-go test tooling and environment (specifically fix #3071).

This PR does revamp `testregistry`, `testca` and adds a couple of files.

In a shell:
- `testca` now allows additional names
-  `portlock` allows for "locking" a port (or finding a free one), that is not racy and safe to use with `parallel` ~~across distinct invocations~~. This allows tests to not have to specify an explicit port anymore (if they don't want a specific one), which in turn allow for parralelization without risks of conflicts
- `testregistry` and `login_linux_test` have been largely rewritten and fix:
   * guarantee that cleanups are always done
   * add support for Basic Authentication
   * prevent cross registry pollution by using distinct credentials per registry
   * prevent credential pollution between tests by privatizing `.docker/config.json` 
   * reduce duplication in `testregistry`
   * better host toml coverage for variants (w/o ports)
   * Auth server can now serve on http or https
   * parallelize as much as feasible
   * adding a larger variety of tests
 
Pending green CI, I would suggest we merge this ASAP so that further test cleanup can happen (and more parallelization) and actual work can begin on cleaning-up login


cmd/nerdctl/image_convert_linux_test.go
```diff
@@ -48,18 +51,20 @@ func TestImageConvertNydus(t *testing.T) {
 		t.Skip("Nydusify check is not supported rootless mode.")
 	}
 
-	// skip if nydusify is not installed
+	// skip if nydusify and nydusd are not installed
 	testutil.RequireExecutable(t, "nydusify")
+	testutil.RequireExecutable(t, "nydusd")
```
apostasie, line 56: Adding missing requirement

cmd/nerdctl/image_convert_linux_test.go
```diff
@@ -48,18 +51,20 @@ func TestImageConvertNydus(t *testing.T) {
...
 
 	// setup local docker registry
-	registryPort := 15000
-	registry := testregistry.NewPlainHTTP(base, registryPort)
-	defer registry.Cleanup()
+	registry := testregistry.NewWithNoAuth(base, 0, false)
+	remoteImage := fmt.Sprintf("%s:%d/nydusd-image:test", "localhost", registry.Port)
+	t.Cleanup(func() {
```
apostasie, line 61: Cleanup is better to use than defer

apostasie:
CI is green

@AkihiroSuda and team PTAL at your convenience.

apostasie:
Apologies for the re-push.
Did mistakenly commit files from another PR.

apostasie:
@AkihiroSuda if this looks fine to you folks, I would love to see this merged in as I have more CI and tests changes that would benefit from it.

I believe the windows failure is a fluke, so, if you can poke it that would be lovely.

cmd/nerdctl/compose_run_linux_test.go
```diff
@@ -432,12 +432,13 @@ func TestComposePushAndPullWithCosignVerify(t *testing.T) {
 	keyPair := newCosignKeyPair(t, "cosign-key-pair")
 	defer keyPair.cleanup()
 
-	reg := testregistry.NewPlainHTTP(base, 5000)
-	defer reg.Cleanup()
+	reg := testregistry.NewWithNoAuth(base, 0, false)
```
AkihiroSuda, line 435: Why renamed?

cmd/nerdctl/compose_run_linux_test.go
```diff
@@ -432,12 +432,13 @@ func TestComposePushAndPullWithCosignVerify(t *testing.T) {
...
+	defer reg.Cleanup(nil)
+
 	localhostIP := "127.0.0.1"
 	t.Logf("localhost IP=%q", localhostIP)
 	testImageRefPrefix := fmt.Sprintf("%s:%d/",
-		localhostIP, reg.ListenPort)
+		localhostIP, reg.Port)
```
AkihiroSuda, line 441: Why renamed?

apostasie: "ListenIP" makes sense (0.0.0.0) - as in "the address the registry process is binding to" - and also does "IP" (the exposed ip you can access it with).
"ListenPort" feels the wrong way - since this is NOT the port the registry process is actually listening on (still 5000), but instead the port you can access it with.

It seems better to me that way.

Either way, this is minor. I can reverse it if you dont like it.

Thanks @AkihiroSuda

cmd/nerdctl/compose_run_linux_test.go
apostasie: Methods have been simplified / renamed:
- there is no longer separate helpers for http vs. https (you pass a tls bool instead) - so, the previous name (NewPlainHTTP, that does not specific what type of auth it is using)
- all helpers names now follow the same pattern:
  * NewWithTokenAuth
  * NewWithBasicAuth
  * NewWithNoAuth

Let me know if you don't like it and I can revert this one if you wish.

AkihiroSuda (approved):
Thanks

PR #3079, opened 2024-06-11
Make zstd convert safe to use concurrently

### LINKED ISSUE #3078: ZstdLayerConvertFunc unsafe to use concurrently
### Description

`convert zstd` is not safe to use concurrently, as there is no guarantee that the temporary uncompressed resource is still there by the time we read it.

https://github.com/containerd/nerdctl/blob/main/pkg/imgutil/converter/zstd.go#L54

Furthermore, using `uncompress.LayerConvertFunc` seems to create useless churn (and is at the root of the problem), as we will read and write twice instead of once.

From a quick read, it seems to me that we should instead just avoid the containerd Uncompress helper altogether and just get a DecompressStream directly and roll with it:

```golang

	readerAt, err := cs.ReaderAt(ctx, desc)
	sr := io.NewSectionReader(readerAt, 0, desc.Size)
	newR, err := compression.DecompressStream(sr)

```

Thoughts?

Use convert zstd concurrently.

Fail because the temporary uncompressed layer has already been deleted by the concurrent execution.

Should not fail.

1.7.6

None

### Host information

_No response_

apostasie:
Fix #3078 by removing use of intermediary uncompressed layer.

This incidentally gives us a small (?) perf boost (as we no longer read and write resources twice).

This will make convert tests able to be run concurrently.

@AkihiroSuda and team - PTAL at your convenience

### DIFF
--- a/pkg/imgutil/converter/zstd.go
+++ b/pkg/imgutil/converter/zstd.go
@@ -21,12 +21,12 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/containerd/containerd/archive/compression"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/images/converter"
 	"github.com/containerd/containerd/images/converter/uncompress"
-	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/klauspost/compress/zstd"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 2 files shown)


apostasie:
About the "perf boost": un-scientific measurement show 30% speed-up at minimum compression for Alpine - and ~15% at max compression.

pkg/imgutil/converter/zstd.go
```diff
@@ -40,34 +41,35 @@ func ZstdLayerConvertFunc(options types.ImageConvertOptions) (converter.ConvertF
...
+		var oldReader io.Reader
+		// If it is compressed, get a decompressed stream
+		if !uncompress.IsUncompressedType(desc.MediaType) {
+			decompStream, err := compression.DecompressStream(sectionReader)
+			if err != nil {
+				return nil, err
+			}
+			defer decompStream.Close()
+			oldReader = decompStream
+			delete(labelsMap, labels.LabelUncompressed)
```
apostasie, line 68: Note that I am not quite sure I understand why containerd is doing that in the helper - but I wanted to make sure we have iso behavior.

apostasie:
Adding cleaned-up test as well (now that we can parallelize).

pkg/imgutil/converter/zstd.go
ktock: Please do not delete this but update this value if the actual uncompressed digest differs from the original one. 

This is defined at: https://github.com/containerd/containerd/blob/4a18adcfca4e37e829c0d3174bd04d4cbf222f9a/pkg/labels/labels.go#L21
This is used several use cases for example diff computation: https://github.com/containerd/containerd/blob/4a18adcfca4e37e829c0d3174bd04d4cbf222f9a/core/images/diffid.go#L47 
Other clients also might use it.

apostasie: Ok.

Curious what your take is on the following in that context - as I said, not sure what there is to gain in doing it containerd?
https://github.com/containerd/containerd/blob/v1.7.18/images/converter/uncompress/uncompress.go#L80

Anyhow, re-pushed.

apostasie:
> [#3079 (comment)](https://github.com/containerd/nerdctl/pull/3079#discussion_r1634447937)

Fixed.

ktock:
Could you update labels.LabelUncompressed with the actual uncompressed digest, using Digester of go-digest? e.g.: https://github.com/containerd/containerd/blob/4a18adcfca4e37e829c0d3174bd04d4cbf222f9a/core/diff/apply/apply.go#L90

ktock:
> Can you clarify why we should change this behavior?
> Is the current logic faulty on that?

The current logic isn't faulty but it's good to make sure that the content has the correct LabelUncompressed (this can be used for fast path as mentioned in https://github.com/containerd/nerdctl/pull/3079#discussion_r1634447937 ). But yes, we can do that as a following up so this patch looks good for now.

apostasie:
Thanks a lot @ktock 
Appreciate the fast review on this!

Issue #3095, opened 2024-06-16
`networks` do not appear to support --namespace


apostasie:
### Description

@AkihiroSuda is there a reason we do not support namespaces for networks?
Or is this just TBD?

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

AkihiroSuda:
This is because CNI has no concept of namespaces

apostasie:
Fixed by #3096

PR #3096, opened 2024-06-16
[Proposal] Implement namespacing for networks

apostasie:
Hey folks.

From the conversation at #3095 I appreciate that plugins networks cannot be fully "isolated" from one another, and as such that there is no concept of namespacing in the spec.

Nevertheless, not having namespace for networks commands is very surprising for the user - especially since we do not complain at all about the use of the global `--namespace` flag alongside the `network` command. Furthermore, I certainly see use cases where it would make a lot of sense as a feature.

This (relatively simple) proof of concept does implement "namespace-ing", effectively preventing one namespace from listing or manipulating directly networks of another.

Let me know what you are thinking about this - and if this would be a welcome feature or not.

### DIFF
--- a/cmd/nerdctl/network_inspect_test.go
+++ b/cmd/nerdctl/network_inspect_test.go
@@ -66,3 +66,28 @@ func TestNetworkInspect(t *testing.T) {
 	}
 	assert.DeepEqual(base.T, expectedIPAM, got.IPAM)
 }
+
+func TestNetworkWithNamespace(t *testing.T) {
+	testutil.DockerIncompatible(t)
+
+	t.Parallel()
+
+	tID := testutil.Identifier(t)
+	base := testutil.NewBase(t)
+	baseOther := testutil.NewBaseWithNamespace(t, tID)
+
+	tearDown := func() {
+		base.Cmd("network", "rm", tID).Run()
+		baseOther.Cmd("namespace", "remove", tID).Run()
+	}
+	tearDown()
+	t.Cleanup(tearDown)
+
+	base.Cmd("network", "create", tID).AssertOK()
+
+	// Other namespace cannot inspect, prune, see, or remove this network
+	baseOther.Cmd("network", "inspect", tID).AssertFail()
+	baseOther.Cmd("network", "prune", "-f").AssertOutNotContains(tID)
+	baseOther.Cmd("network", "ls").AssertOutNotContains(tID)
+	baseOther.Cmd("network", "remove", tID).AssertFail()
+}
--- a/docs/cni.md
+++ b/docs/cni.md
@@ -116,7 +116,16 @@ For example:
 ## Custom networks
 
 You can also customize your CNI network by providing configuration files.
-For example you have one configuration file(`/etc/cni/net.d/10-mynet.conf`)
+
+When rootful, the expected root location is `/etc/cni/net.d`.
+For rootless, the expected root location is `~/.config/cni/net.d/`
+
+Configuration files (like `10-mynet.conf`) can be placed either in the root location,
+or under a subfolder.
+If in the root location, this network will be available to all nerdctl namespaces.
+If placed in a subfolder, it will be available only to the identically named namespace.
+
+For example, you have one configuration file(`/etc/cni/net.d/10-mynet.conf`)
 for `bridge` network:
 
 ```json
@@ -138,7 +147,7 @@ for `bridge` network:
 ```
 
 This will configure a new CNI network with the name `mynet`, and you can use
-this network to create a container:
+this network to create a container in any namespace:
 
 ```console
 # nerdctl run -it --net mynet --rm alpine ip addr show
--- a/pkg/cmd/network/create.go
+++ b/pkg/cmd/network/create.go
@@ -33,7 +33,7 @@ func Create(options types.NetworkCreateOptions, stdout io.Writer) error {
 		options.Subnets = []string{""}
 	}
 
-	e, err := netutil.NewCNIEnv(options.GOptions.CNIPath, options.GOptions.CNINetConfPath)
+	e, err := netutil.NewCNIEnv(options.GOptions.CNIPath, options.GOptions.CNINetConfPath, netutil.WithNamespace(options.GOptions.Namespace))
 	if err != nil {
 		return err
 	}
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -46,6 +46,7 @@ import (
 type CNIEnv struct {
 	Path        string
 	NetconfPath string
+	Namespace   string
 }
 
 type CNIEnvOpt func(e *CNIEnv) error
@@ -132,6 +133,16 @@ func WithDefaultNetwork() CNIEnvOpt {
 	}
 }
 
+func WithNamespace(namespace string) CNIEnvOpt {
+	return func(e *CNIEnv) error {
+		if err := os.MkdirAll(filepath.Join(e.NetconfPath, namespace), 0755); err != nil {
+			return err
+		}
+		e.Namespace = namespace
+		return nil
+	}
+}
+
 func NewCNIEnv(cniPath, cniConfPath string, opts ...CNIEnvOpt) (*CNIEnv, error) {
 	e := CNIEnv{
 		Path:        cniPath,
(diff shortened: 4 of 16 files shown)


apostasie:
Note that the current situation has weird UX because of the inconsistency here.
Specifically, trying to remove a network used by containers in another namespace will return the ids of said containers, which of course cannot be inspected nor listed.

apostasie:
Just a note that I will wait for feedback on the idea and design before engaging with CI / test fixing.

pkg/netutil/netutil.go
```diff
@@ -193,7 +204,7 @@ func (e *CNIEnv) FilterNetworks(filterf func(*NetworkConfig) bool) ([]*NetworkCo
 }
 
 func (e *CNIEnv) getConfigPathForNetworkName(netName string) string {
-	return filepath.Join(e.NetconfPath, "nerdctl-"+netName+".conflist")
+	return filepath.Join(e.NetconfPath, e.Namespace, "nerdctl-"+netName+".conflist")
```
AkihiroSuda, line 207: Does CNI support directories in netconf path?

AkihiroSuda:
Thanks, the design looks good if it can pass the CI.

We may want to have something like `nerdctl network create --global-namespace` to keep supporting unnamespaced networks

apostasie:
> Thanks, the design looks good if it can pass the CI.
> 

Ok.

I'll fix whatever is needed on CI/test to make this ready.

> We may want to have something like `nerdctl network create --global-namespace` to keep supporting unnamespaced networks

With the current design, unnamespaced networks will show up in all namespaces.
That should guarantee no disruption / migration required for users, and also acts as an interesting feature.

But we can talk about the details here when the PR is fully ready and with adequate tests.

Will ping here when done.

Thanks!

Zheaoli:
I think maybe we should make this feature optional or experimental. Users can config it in the global config. I think this is a huge change for user

apostasie:
> I think maybe we should make this feature optional or experimental. Users can config it in the global config. I think this is a huge change for user

That works for me too if you prefer to minimize impact for now.

That being said, this change is transparent for the user and in most cases they would not notice it.

apostasie:
Clarification on how this is working:

```
nerdctl --namespace=XXXX network ls
```

Will lookup in:
- `prefix/cni/net.d/`: any network found there will show in ALL namespaces - this means this will give users the exact same behavior they currently have
- `prefix/cni/net.d/XXXX`: this is the new, namespaced location - any network in there is only available to that exact namespace

When the default network is not found and needs to be created, it will be created in the same usual location with NO namespace subfolder.
This is also the exact same behavior we currently have.

When any other network is being created, it will be created under the namespaced folder instead of the global folder.
This is the one difference in term of user experience.

Finally, when the user is deleting a network, it will first try to find it in the namespaced folder. If it cannot be found there, it will search and possibly delete in the global folder. Here as well, this aligns with the current behavior.

The CI is now green - I think this is ready for formal review now (I am re-pushing right now but this is just md / documentation).

Note that I am not yet adding specific tests for the new behaviors - I am currently working on cleanup of all network tests, and would like to send a PR for that soon + new network namespace tests built on top of that cleaned test suite if you don't mind.

Kindly let me know your thoughts on all this, and if you want it experimental or mainline.

@AkihiroSuda @Zheaoli

docs/cni.md
```diff
@@ -116,7 +116,16 @@ For example:
 ## Custom networks
 
 You can also customize your CNI network by providing configuration files.
-For example you have one configuration file(`/etc/cni/net.d/10-mynet.conf`)
+
+When rootful, the expected root location is `/etc/cni/net.d`.
+For rootless, the expected root location is `~/.config/cni/net.d/`
```
AkihiroSuda, line 121: Could you also update https://github.com/containerd/nerdctl/blob/main/docs/dir.md#cni

pkg/cmd/network/create.go
```diff
@@ -33,7 +33,7 @@ func Create(options types.NetworkCreateOptions, stdout io.Writer) error {
 		options.Subnets = []string{""}
 	}
 
-	e, err := netutil.NewCNIEnv(options.GOptions.CNIPath, options.GOptions.CNINetConfPath)
+	e, err := netutil.NewCNIEnv(options.GOptions.CNIPath, options.GOptions.CNINetConfPath, netutil.WithNamespace(options.GOptions.Namespace))
```
AkihiroSuda, line 36: Do we want to have an integration test to ensure that the created network is not leaked to other namespaces?

docs/cni.md
apostasie: Done

pkg/cmd/network/create.go
apostasie: Yeah, I add some in another dev-branch focused on test cleanup but I just took them out and added them here.

apostasie:
@AkihiroSuda last push fixed your two comments (tests+doc).

Pending CI (green locally).

PTAL at your convenience.

Thanks!

cmd/nerdctl/network_inspect_test.go
```diff
@@ -66,3 +66,24 @@ func TestNetworkInspect(t *testing.T) {
 	}
 	assert.DeepEqual(base.T, expectedIPAM, got.IPAM)
 }
+
+func TestNetworkInspectWithNamespace(t *testing.T) {
+	if testutil.GetTarget() == testutil.Docker {
```
AkihiroSuda, line 71: `testutil.DockerIncompatible(t)`

cmd/nerdctl/network_inspect_test.go
```diff
@@ -66,3 +66,24 @@ func TestNetworkInspect(t *testing.T) {
...
+	}
+
+	t.Parallel()
+
+	base := testutil.NewBase(t)
+	baseOther := testutil.NewBaseWithNamespace(t, "nerdctl-other")
+	networkName := testutil.Identifier(t)
+
+	tearDown := func() {
+		base.Cmd("network", "rm", networkName).Run()
```
AkihiroSuda, line 82: The `nerdctl-other` namespace should be pruned

AkihiroSuda: (The namespace name will need to have an extra suffix to avoid racing)

cmd/nerdctl/network_inspect_test.go
```diff
@@ -66,3 +66,24 @@ func TestNetworkInspect(t *testing.T) {
 	}
 	assert.DeepEqual(base.T, expectedIPAM, got.IPAM)
 }
+
+func TestNetworkInspectWithNamespace(t *testing.T) {
```
AkihiroSuda, line 70: Probably this can be safely merged to `TestNetworkLsWithNamespace`, but no strong opinion

cmd/nerdctl/network_inspect_test.go
apostasie: - changed from `nerdctl-other` to `tID`.
- added a prune for it in the tearDown function

cmd/nerdctl/network_inspect_test.go
apostasie: Good idea. I merged them all actually - much more compact.

cmd/nerdctl/network_inspect_test.go
```diff
@@ -66,3 +67,31 @@ func TestNetworkInspect(t *testing.T) {
...
+	t.Cleanup(tearDown)
+
+	base.Cmd("network", "create", tID).AssertOK()
+
+	// Other namespace cannot inspect, prune, see, or remove this network
+	baseOther.Cmd("network", "inspect", tID).AssertFail()
+	baseOther.Cmd("network", "prune", "-f").AssertOutNotContains(tID)
+	baseOther.Cmd("network", "ls").AssertOutNotContains(tID)
+
+	if !rootlessutil.IsRootless() {
```
AkihiroSuda, line 94: Why?

cmd/nerdctl/network_inspect_test.go
```diff
@@ -66,3 +67,31 @@ func TestNetworkInspect(t *testing.T) {
...
+func TestNetworkWithNamespace(t *testing.T) {
+	testutil.DockerIncompatible(t)
+
+	t.Parallel()
+
+	tID := testutil.Identifier(t)
+	base := testutil.NewBase(t)
+	baseOther := testutil.NewBaseWithNamespace(t, tID)
+
+	tearDown := func() {
```
AkihiroSuda, line 80: The namespace should be teared down

apostasie: Oh! I see - of course.

cmd/nerdctl/network_inspect_test.go
apostasie: I actually do not know.
This was already in the original test: https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/network_remove_linux_test.go#L53

Is there any condition under which remove would not work for rootless?

apostasie: Looks like it was from here: https://github.com/containerd/nerdctl/commit/a4b51e2842d95800262fb162de43c2cab26fce1f

so - recent.

Maybe it was just a fluke? I removed it and it is ok locally.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
`nerdctl namespace ls` will have to show the number of the networks, but it can be done later

PR #3105, opened 2024-06-17
Add mechanism to prevent Cmd from executing multiple times

### LINKED ISSUE #3102: Some tests seem to run multiple times the command they are testing because of the way AssertOutContains is used
### Description

For example
https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run_network_linux_test.go#L491

Since we first call AssertOK, then AssertOutContains, the command is run twice.

This is quite problematic.
If we want to make these helpers friendlier / easier to use for people, we should maintain a state that would prevent multiple runs.
At least, we should do a quick review of the codebase and make sure we fix this specific pattern. 

na

na

1.7.6

None

### Host information

_No response_

apostasie:
Fix #3102 

After a quick review of the codebase, it does feel more realistic to prevent multiple-runs instead of chasing after mis-uses.

### DIFF
--- a/pkg/testutil/testutil.go
+++ b/pkg/testutil/testutil.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -343,11 +344,26 @@ func (b *Base) EnsureContainerExited(con string, expectedExitCode int) {
 type Cmd struct {
 	icmd.Cmd
 	*Base
+	runResult *icmd.Result
+	mu        sync.Mutex
 }
 
 func (c *Cmd) Run() *icmd.Result {
 	c.Base.T.Helper()
-	return icmd.RunCmd(c.Cmd)
+	c.mu.Lock()
+	c.runResult = icmd.RunCmd(c.Cmd)
+	c.mu.Unlock()
+	return c.runResult
+}
+
+func (c *Cmd) runIfNecessary() *icmd.Result {
+	c.Base.T.Helper()
+	c.mu.Lock()
+	if c.runResult == nil {
+		c.runResult = icmd.RunCmd(c.Cmd)
+	}
+	c.mu.Unlock()
+	return c.runResult
 }
 
 func (c *Cmd) CmdOption(cmdOptions ...func(*Cmd)) *Cmd {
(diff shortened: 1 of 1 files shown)


apostasie:
CI is green.

@AkihiroSuda and team, PTAL at your convenience.

pkg/testutil/testutil.go
```diff
@@ -343,11 +343,15 @@ func (b *Base) EnsureContainerExited(con string, expectedExitCode int) {
 type Cmd struct {
 	icmd.Cmd
 	*Base
+	runResult *icmd.Result
```
AkihiroSuda, line 346: Maybe protect this with a mutex?

pkg/testutil/testutil.go
```diff
@@ -343,11 +343,15 @@ func (b *Base) EnsureContainerExited(con string, expectedExitCode int) {
...
 }
 
 func (c *Cmd) Run() *icmd.Result {
```
AkihiroSuda, line 349: Should be renamed to something like RunCached, and the existing Run should be kept uncached?

pkg/testutil/testutil.go
apostasie: Yep, makes sense. Will fix. Thanks.

pkg/testutil/testutil.go
apostasie: Is there any case where we would like it to be able to run multiple times? (which is breaking things like WithStdIn, for eg)

apostasie: Alternatively I could make Run to always run, but all the assert func to call on runWithCache so that they dont trigger an exec if one has run already. Do you prefer this?

AkihiroSuda: I guess it should panic when one has run already?

apostasie: We can make it so - but I guess that would mean we would have to to fix it everywhere in the codebase where there are a succession of cmd.Assert on the same cmd object?

AkihiroSuda: Right

apostasie:
@AkihiroSuda latest push:
- adds a mutex
- now has a private method runIfNecessary that all the asserts are using - on the other hand, Run() still execute everytime, which will preserve the previous behavior for public consumer.

PTAL and let me know what you think.

PR #3106, opened 2024-06-17
Add doc note about --ip and bridge

### LINKED ISSUE #3101: nerdctl / docker behavior divergence on `--net bridge --ip-address`
### Description

This is the aftermath of #3097 and #3088.

Docker does not allow specifying an ip-address on non-user defined networks.
However, nerdctl does allow it for bridge.

Previously this was tolerated by docker (although it was a no-op as the ip was simply ignored), but now, it is a hard error.

We need to decide what to do here:
- option A: do nothing - this is a difference between nerdctl and docker - this is my preference
- option B: align behavior with Docker and bail out if an ip address is specified on a non-user defined network (which makes sense for none and host, but then, our concept of bridge is different from Docker)

Thoughts?

```
docker-24 run  -d --name foobar --ip 172.17.0.42 alpine sleep 1000
docker-26 run  -d --name foobar --ip 172.17.0.42 alpine sleep 1000
nerdctl run  -d --name foobar --ip 10.4.0.42 alpine sleep 1000
```

na

1.7.6

None

### Host information

_No response_

apostasie:
Fix #3101 

### DIFF
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -183,7 +183,7 @@ Network flags:
 - :whale: `-h, --hostname`: Container host name
 - :whale: `--add-host`: Add a custom host-to-IP mapping (host:ip). `ip` could be a special string `host-gateway`,
 - which will be resolved to the `host-gateway-ip` in nerdctl.toml or global flag.
-- :whale: `--ip`: Specific static IP address(es) to use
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #3116, opened 2024-06-19
Aftermath docker migration: fix racyness in tests

apostasie:
Following on fixes and updated tests to align with Docker v26, `createWithMacAddress` has become flaky for the docker target.

There was a couple of wrongs in my code:
- `start + logs` seems to be racy - replaced that by `start -a`
- with `none` network, of course `eth0` does not exist
- when `expected` is "", the resolution was always true
- after fixing ^ this, it became clear that decision from #3101 hasn't been handled yet and needed to

Hopefully this will fix flakyness on the docker target, and does fix the logic of the test for the docker bridge case (albeit the result is the same).

cc @AkihiroSuda assuming this passes green, can we merge this ASAP?

### DIFF
--- a/cmd/nerdctl/container_create_linux_test.go
+++ b/cmd/nerdctl/container_create_linux_test.go
@@ -106,24 +106,25 @@ func TestCreateWithMACAddress(t *testing.T) {
 			if expect == passedMac {
 				expect = macAddress
 			}
-			tt.Cleanup(func() {
+			tearDown := func() {
 				base.Cmd("rm", "-f", containerName).Run()
-			})
+			}
+			tearDown()
+			tt.Cleanup(tearDown)
 			// This is currently blocked by https://github.com/containerd/nerdctl/pull/3104
 			// res := base.Cmd("create", "-i", "--network", network, "--mac-address", macAddress, testutil.CommonImage).Run()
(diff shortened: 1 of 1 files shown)


apostasie:
@AkihiroSuda from the failing node:
```
https://public.ecr.aws/v2/docker/library/ubuntu/manifests/sha256:fd7fe639db24c4e005643921beea92bc449aac4f4d40d60cd9ad9ab6456aec01: 429 Too Many Requests - Server message: toomanyrequests: Data limit exceeded
```

AkihiroSuda:
> @AkihiroSuda from the failing node:
> 
> ```
> https://public.ecr.aws/v2/docker/library/ubuntu/manifests/sha256:fd7fe639db24c4e005643921beea92bc449aac4f4d40d60cd9ad9ab6456aec01:
> 429 Too Many Requests - Server message: toomanyrequests: Data limit exceeded
> ```
> 
> I assume there is specific reason why these are not on ghcr.io?

The image seem to be introduced for testing SOCI: https://github.com/containerd/nerdctl/commit/4c82cde7d594d3952344d815c0645205215cfec7

I guess it can be moved to ghcr.io , if the rate limit error happens frequently.
cc @ktock @ningziwen

AkihiroSuda (approved):
Thanks

PR #3117, opened 2024-06-19
Bump compose-go to v2

apostasie:


apostasie:
This is not too bad.
Basic functionality is here.

TBD
* [  ] .env files are bust
* [ ] Profiles are bust

apostasie:
@AkihiroSuda 

Feel like this one could go in or do we need some form of soak time?
CI is green but our test coverage might be so-so.

Appreciate this is a "big" change (not many lines of code / not complex, but certainly impactful).

IMHO we should go for it and not stay behind longer (we can fix bugs as they come) - it will obviously only get worse the more we wait.

pkg/composer/serviceparser/serviceparser_test.go
```diff
@@ -174,8 +174,8 @@ volumes:
 	assert.Assert(t, in(wp1.RunArgs, "--log-driver=json-file"))
 	assert.Assert(t, in(wp1.RunArgs, "--log-opt=max-size=5K"))
 	assert.Assert(t, in(wp1.RunArgs, "--log-opt=max-file=2"))
-	assert.Assert(t, in(wp1.RunArgs, "--add-host=test.com:172.19.1.1"))
-	assert.Assert(t, in(wp1.RunArgs, "--add-host=test2.com:172.19.1.2"))
+	assert.Assert(t, in(wp1.RunArgs, "--add-host=test.com:[172.19.1.1]"))
+	assert.Assert(t, in(wp1.RunArgs, "--add-host=test2.com:[172.19.1.2]"))
```
AkihiroSuda, line 178: This seems to generate an invalid `/etc/hosts`

```
$ nerdctl run --rm --add-host=example.com:[192.168.12.34] busybox cat /etc/hosts
# <nerdctl>
127.0.0.1       localhost localhost.localdomain
::1             localhost localhost.localdomain
[192.168.12.34] example.com
10.4.0.45       d39e31c01c00 busybox-d39e3
# </nerdctl>
```

apostasie: Interesting.

Let me look into it.

apostasie: So, compose now return []string instead of string.

I have no idea why oO. Is the hope that you would get some form of cheapo load balancing?

Passing multiple times the same host with different ips is not going to work.

apostasie:
@AkihiroSuda fixed the bizarro change in compose wrt host/ip.
At least we do now honor the fact it is an array.
Still not completely sure why / what is expected to happen here.

AkihiroSuda (approved):
Thanks

PR #3121, opened 2024-06-19
[release/1.7 backport] Add --pull option to build command

sondavidb:
Backport of #3074. Also took some helper functions for `pkg/testutil/testutil.go` from https://github.com/containerd/nerdctl/commit/f853ab8542dfd85a8a53c435ebf39d1d1b10cd85.

Per https://github.com/containernetworking/plugins/issues/1038, the binaries were modified as of the last commit so this change also brings in the new shasums so that CI runs properly.

Mainline also requires Golang 1.22 so cherry-picked https://github.com/containerd/nerdctl/commit/4e615922c484817898c81f4c70cbf6abeb8a89f9.

Feel free to close if we are not taking features into release/1.7.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
     - uses: actions/checkout@v4.1.1
     - uses: actions/setup-go@v5
       with:
-        go-version: 1.21.x
+        go-version: 1.22.x
     - name: "Compile binaries"
       run: make artifacts
     - name: "SHA256SUMS"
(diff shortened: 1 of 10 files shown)


AkihiroSuda:
I'm not sure we should backport new features to v1.7, but still negotiable if this is urgent for Finch.

sondavidb:
>still negotiable if this is urgent for Finch.

It's not urgent, and we should prepare for 2.0 upgrade anyway in the future. Closing for now, thanks for your input!

PR #3125, opened 2024-06-20
Volume QA, "volume remove" (fixes and tests)

apostasie:
This fixes #3124 and a few other discrepancies between docker and nerdctl.

This also expands testing significantly.

Outside of the tests, I will comment inline as to "why" certain changes.

PR #2902, opened 2024-03-27
[WIP] go.mod: github.com/containerd/containerd/v2 v2.0.0-rc.0

AkihiroSuda:
```console
$ go build ./...
# github.com/containerd/nerdctl/v2/pkg/imgutil
pkg/imgutil/imgutil.go:231:53: cannot use imgcryptUnpackOpt (variable of type containerd.UnpackOpt) as "github.com/containerd/containerd/v2/client".UnpackOpt value in array or slice literal
pkg/imgutil/snapshotter.go:114:9: cannot use source.AppendExtraLabelsHandler(prefetchSize, f) (value of type func("github.com/containerd/containerd/images".Handler) "github.com/containerd/containerd/images".Handler) as func("github.com/containerd/containerd/v2/core/images".Handler) "github.com/containerd/containerd/v2/core/images".Handler value in return statement
pkg/imgutil/snapshotter.go:114:55: cannot use f (variable of type func("github.com/containerd/containerd/v2/core/images".Handler) "github.com/containerd/containerd/v2/core/images".Handler) as func("github.com/containerd/containerd/images".Handler) "github.com/containerd/containerd/images".Handler value in argument to source.AppendExtraLabelsHandler
pkg/imgutil/snapshotter.go:118:9: cannot use socisource.AppendDefaultLabelsHandlerWrapper(rFlags.SociIndexDigest, f) (value of type func(f "github.com/containerd/containerd/images".Handler) "github.com/containerd/containerd/images".Handler) as func("github.com/containerd/containerd/v2/core/images".Handler) "github.com/containerd/containerd/v2/core/images".Handler value in return statement
pkg/imgutil/snapshotter.go:118:78: cannot use f (variable of type func("github.com/containerd/containerd/v2/core/images".Handler) "github.com/containerd/containerd/v2/core/images".Handler) as func("github.com/containerd/containerd/images".Handler) "github.com/containerd/containerd/images".Handler value in argument to socisource.AppendDefaultLabelsHandlerWrapper
```

Depends on:
- [ ] https://github.com/containerd/imgcrypt/pull/148
- [ ] https://github.com/containerd/stargz-snapshotter/pull/1545
- [ ] soci-snapshotter

PR #3133, opened 2024-06-20
Volume QA part II: fixes and tests

### LINKED ISSUE #3123: Volume code may be racy
### Description

While poking // tests, I did hit this:

```
level=fatal msg="mkdir /home/dmp.linux/.local/share/nerdctl/1935db59/volumes/nerdctl-test/nerdctl-testvolumeremove: file exists"
```

We should have a look at volume creation code and figure out if it is locking properly or if the test for existence can be raced out.

na

na

1.7.6

None

### Host information

_No response_

### LINKED ISSUE #3132: `usedVolumes` (called in volume remove and prune)  can be raced by container creation, and other sources of racyness in volumes
### Description

Volume Store does have a lock mechanism preventing race conditions between volume creation and volume removal / pruning.

However, calls to `usedVolumes`, which will decide which volumes are safe to remove, are not protected.
From a casual reading of the source, it seems to me that container creation or removal could race a call to `usedVolumes`, resulting in removal or pruning to either fail to remove volumes that should have been (this case is ok IMHO), but also possibly removing volumes that should not be as they are NOW in use (this one is probably really bad).

Solution could be that we enforce the volume lock mechanism so that container `creation` has to get an exclusive lock, and then we also put `usedVolumes` inside the locked function.

That would definitely solve the bad scenario:
- usedVolume check -> container is created with volume -> volume destroy

Trying to lock as well on container remove or prune is unlikely to bring joy - but again, in that case, it is not too bad (we just fail to remove stuff that could have been).

Furthermore, other calls to `volStore.Get` with the intent of creating volumes dependent on results are also susceptible to the same problem.

Since we now changed the behavior of `Create` to align with docker, I will fix the simplest / most egregious cases ("get" then "create") and overall revamp `Get` which is not safe.

But for the rest (locking volStore on container create):
Is this reading overall accurate?
Thoughts?

na

na

main

None

### Host information

_No response_

apostasie:
This is part 2 of #3125 - same idea.

The original motivation was to significantly increase testing coverage & quality, convergence with docker behavior, and stability wrt concurrency for anything related to volume lifecycle.

In the process of doing so, several issues cropped-up, that this PR also fix.

Tests:
- parallelized (which does also serve as a very good way to uncover raciness)
- expanded
- cleaned-up / normalized
- added namespace testing
- test more thoroughly, both in terms of cases and validating content of stdout / stderr
- ~~move tests from _linux to all platforms (not sure why it was linux only in the first place)~~ - wish I had a windows env to test

Code changes:
- docker align: "no labels" should be `nil`, not `[]string{}`
- docker align: `--label ""` should error
- docker align: multi-inspect that is partly failing, and other partly failing conditions
- remove useless / duplicated code (specifically around validation of identifiers that was performed multiple times)
- make sure all methods of the volume store are locking (to avoid racy behavior)
- replace broken/racy logic "`get` then `create`" by just `create` (since behavior has been recently changed and now always returns success)
- make the underlying filesystem implementation private (eg: no longer expose `Path`) to prevent leaking implementation to the consumer (see for example `cmd/nerdctl/namespace.go b/cmd/nerdctl/namespace.go` that was reading the filesystem directly instead of calling `List`)
- added a "persistent" `Lock` and `Unlock` methods on the store, for operations that need to call multiple methods successively or need guarantees that volumes cannot be modified at all until they are done
- cleaned-up and expanded `lockutil`, so that we don't expose anymore `Flock` (unimplemented for non Unix), and instead expose new lock/unlock methods
- reviewed all the codebase for any cases where that global lock call is necessary and added them
- reviewed codebase to fix #3132 (and fix #3123 hopefully)

While this is a significant PR in term of apparent code change size, the added testing should give us a good confidence boost on what's happening wrt volumes.

### DIFF
--- a/cmd/nerdctl/namespace.go
+++ b/cmd/nerdctl/namespace.go
@@ -113,11 +113,11 @@ func namespaceLsAction(cmd *cobra.Command, args []string) error {
 		}
 		numImages = len(images)
 
-		volStore, err := volumestore.Path(dataStore, ns)
+		volStore, err := volumestore.New(dataStore, ns)
 		if err != nil {
 			log.L.Warn(err)
 		} else {
-			volEnts, err := os.ReadDir(volStore)
+			volEnts, err := volStore.List(false)
 			if err != nil {
 				if !os.IsNotExist(err) {
 					log.L.Warn(err)
--- a/cmd/nerdctl/system_prune_linux_test.go
+++ b/cmd/nerdctl/system_prune_linux_test.go
@@ -32,6 +32,10 @@ import (
 
 func TestSystemPrune(t *testing.T) {
 	testutil.RequiresBuild(t)
+	// FIXME: using a dedicated namespace does not work with rootful (because of buildkit running)
+	// t.Parallel()
+	// namespaceID := testutil.Identifier(t)
+	// base := testutil.NewBaseWithNamespace(t, namespaceID)
 	base := testutil.NewBase(t)
 	base.Cmd("container", "prune", "-f").AssertOK()
 	base.Cmd("network", "prune", "-f").AssertOK()
--- a/cmd/nerdctl/volume_create.go
+++ b/cmd/nerdctl/volume_create.go
@@ -17,6 +17,9 @@
 package main
 
 import (
+	"fmt"
+
+	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/containerd/nerdctl/v2/pkg/cmd/volume"
 
@@ -45,6 +48,12 @@ func processVolumeCreateOptions(cmd *cobra.Command) (types.VolumeCreateOptions, 
 	if err != nil {
 		return types.VolumeCreateOptions{}, err
 	}
+	for _, label := range labels {
+		if label == "" {
+			return types.VolumeCreateOptions{}, fmt.Errorf("labels cannot be empty (%w)", errdefs.ErrInvalidArgument)
+		}
+	}
+
 	return types.VolumeCreateOptions{
 		GOptions: globalOptions,
 		Labels:   labels,
--- a/cmd/nerdctl/volume_prune_linux_test.go
+++ b/cmd/nerdctl/volume_prune_linux_test.go
@@ -17,30 +17,133 @@
 package main
 
 import (
-	"fmt"
+	"strings"
 	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/icmd"
 
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 )
 
 func TestVolumePrune(t *testing.T) {
-	base := testutil.NewBase(t)
-	tID := testutil.Identifier(t)
-	base.Cmd("volume", "prune", "-a", "-f").Run()
+	// Volume pruning cannot be parallelized for Docker, since we need namespaces to do that in a way that does interact with other tests
+	if testutil.GetTarget() != testutil.Docker {
+		t.Parallel()
+	}
 
-	vID := base.Cmd("volume", "create").Out()
-	base.Cmd("volume", "create", tID+"-1").AssertOK()
-	base.Cmd("volume", "create", tID+"-2").AssertOK()
+	// FIXME: for an unknown reason, when testing ipv6, calling NewBaseWithNamespace per sub-test, in the tearDown/tearUp methods
+	// will actually panic the test (also happens with target=docker)
+	// Calling base here *first* so that it can skip NOW - does seem to workaround the problem
+	// If you have any idea how to properly do this, feel free to remove the following line and fix the underlying issue
+	testutil.NewBase(t)
 
-	base.Cmd("run", "-v", fmt.Sprintf("%s:/volume", tID+"-1"), "--name", tID, testutil.CommonImage).AssertOK()
-	defer base.Cmd("rm", "-f", tID).Run()
+	subTearUp := func(tID string) {
+		base := testutil.NewBaseWithNamespace(t, tID)
+		res := base.Cmd("volume", "create").Run()
+		anonIDBusy := res.Stdout()
+		base.Cmd("volume", "create").Run()
+		base.Cmd("volume", "create", tID+"-busy").AssertOK()
+		base.Cmd("volume", "create", tID+"-free").AssertOK()
+		base.Cmd("run", "--name", tID,
+			"-v", tID+"-busy:/whatever",
+			"-v", anonIDBusy, testutil.CommonImage).AssertOK()
+	}
 
-	base.Cmd("volume", "prune", "-f").AssertOutContains(vID)
-	base.Cmd("volume", "prune", "-a", "-f").AssertOutContains(tID + "-2")
-	base.Cmd("volume", "ls").AssertOutContains(tID + "-1")
-	base.Cmd("volume", "ls").AssertOutNotContains(tID + "-2")
+	subTearDown := func(tID string) {
+		base := testutil.NewBaseWithNamespace(t, tID)
+		base.Cmd("rm", "-f", tID).Run()
+		base.Cmd("namespace", "remove", "-f", tID).Run()
+	}
 
-	base.Cmd("rm", "-f", tID).AssertOK()
-	base.Cmd("volume", "prune", "-a", "-f").AssertOK()
-	base.Cmd("volume", "ls").AssertOutNotContains(tID + "-1")
+	testCases := []struct {
+		description        string
+		command            func(tID string) *testutil.Cmd
+		tearUp             func(tID string)
+		tearDown           func(tID string)
+		expected           func(tID string) icmd.Expected
+		inspect            func(t *testing.T, stdout string, stderr string)
+		dockerIncompatible bool
+	}{
+		{
+			description: "prune anonymous only",
+			command: func(tID string) *testutil.Cmd {
+				base := testutil.NewBaseWithNamespace(t, tID)
+				return base.Cmd("volume", "prune", "-f")
+			},
+			tearUp:   subTearUp,
+			tearDown: subTearDown,
+			expected: func(tID string) icmd.Expected {
+				return icmd.Expected{
+					ExitCode: 0,
+				}
+			},
+			inspect: func(t *testing.T, stdout string, stderr string) {
+				tID := testutil.Identifier(t)
+				base := testutil.NewBaseWithNamespace(t, tID)
+				assert.Assert(base.T, !strings.Contains(stdout, tID+"-free"))
+				base.Cmd("volume", "inspect", tID+"-free").AssertOK()
+				assert.Assert(base.T, !strings.Contains(stdout, tID+"-busy"))
+				base.Cmd("volume", "inspect", tID+"-busy").AssertOK()
+				// TODO verify the anonymous volumes status
+			},
+		},
+		{
+			description: "prune all",
+			command: func(tID string) *testutil.Cmd {
+				base := testutil.NewBaseWithNamespace(t, tID)
+				return base.Cmd("volume", "prune", "-f", "--all")
+			},
+			tearUp:   subTearUp,
+			tearDown: subTearDown,
+			expected: func(tID string) icmd.Expected {
+				return icmd.Expected{
+					ExitCode: 0,
+				}
+			},
+			inspect: func(t *testing.T, stdout string, stderr string) {
+				tID := testutil.Identifier(t)
+				base := testutil.NewBaseWithNamespace(t, tID)
+				assert.Assert(t, !strings.Contains(stdout, tID+"-busy"))
+				base.Cmd("volume", "inspect", tID+"-busy").AssertOK()
+				assert.Assert(t, strings.Contains(stdout, tID+"-free"))
+				base.Cmd("volume", "inspect", tID+"-free").AssertFail()
+				// TODO verify the anonymous volumes status
+			},
+		},
+	}
+
+	for _, test := range testCases {
+		currentTest := test
+		t.Run(currentTest.description, func(tt *testing.T) {
+			if currentTest.dockerIncompatible {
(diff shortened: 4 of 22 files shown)


apostasie:
@AkihiroSuda what's the timeline for v2?

Do I still have time for more or are we frozen for v2 already?

I would have liked to cleanup / fix all bugs for `login` overall, and maybe some network work, but there is only so much I can do today - so, if you intend on freezing, lmk.

AkihiroSuda:
> v2

Planned to be synced with containerd v2
https://github.com/orgs/containerd/projects/9

apostasie:
> > v2
> 
> Planned to be synced with containerd v2 https://github.com/orgs/containerd/projects/9

Cool.

So I guess we also need this in then: https://github.com/containerd/stargz-snapshotter/pull/1545

And #2902 to pass.

Lmk if there is something I can do to help with ^.

AkihiroSuda:
nerdctl v2 will have containerd v2 binaries in the `nerdctl-full` bundle, but I guess it is not necessary to use Go packages of containerd v2

> Lmk if there is something I can do to help with ^.

Thanks, I'd appreciate if you can help #2902 and its dependencies (stargz, soci, etc.) 🙇

cmd/nerdctl/volume_create.go
```diff
@@ -45,6 +48,12 @@ func processVolumeCreateOptions(cmd *cobra.Command) (types.VolumeCreateOptions,
 	if err != nil {
 		return types.VolumeCreateOptions{}, err
 	}
+	for _, label := range labels {
```
apostasie, line 51: Align with docker behavior. Empty labels are not allowed.

cmd/nerdctl/namespace.go
```diff
@@ -113,11 +113,11 @@ func namespaceLsAction(cmd *cobra.Command, args []string) error {
 		}
 		numImages = len(images)
 
-		volStore, err := volumestore.Path(dataStore, ns)
+		volStore, err := volumestore.New(dataStore, ns)
 		if err != nil {
 			log.L.Warn(err)
 		} else {
-			volEnts, err := os.ReadDir(volStore)
+			volEnts, err := volStore.List(false)
```
apostasie, line 120: Replace direct access to underlying filesystem implementation by a proper api call.

pkg/cmd/container/create.go
```diff
@@ -151,7 +164,7 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
 	}
 
 	var mountOpts []oci.SpecOpts
-	mountOpts, internalLabels.anonVolumes, internalLabels.mountPoints, err = generateMountOpts(ctx, client, ensuredImage, options)
+	mountOpts, internalLabels.anonVolumes, internalLabels.mountPoints, err = generateMountOpts(ctx, client, ensuredImage, volStore, options)
```
apostasie, line 167: Since we are now locking the volume store, we need to pass it along.

pkg/cmd/volume/create.go
```diff
@@ -32,9 +31,6 @@ func Create(name string, options types.VolumeCreateOptions) (*native.Volume, err
 		name = stringid.GenerateRandomID()
 		options.Labels = append(options.Labels, labels.AnonymousVolumes+"=")
 	}
-	if err := identifiers.Validate(name); err != nil {
-		return nil, fmt.Errorf("malformed name %s: %w", name, err)
-	}
```
apostasie, line 37: Useless, as this is performed by the volume store.

pkg/cmd/volume/inspect.go
```diff
@@ -17,23 +17,40 @@
...
-	result := make([]interface{}, len(volumes))
+	result := []interface{}{}
 
-	for i, name := range volumes {
+	warns := []error{}
+	for _, name := range volumes {
 		var vol, err = volStore.Get(name, options.Size)
 		if err != nil {
-			return err
+			warns = append(warns, err)
```
apostasie, line 39: Align with docker behavior

pkg/cmd/volume/prune.go
```diff
@@ -26,29 +26,47 @@ import (
...
+	defer volStore.Unlock()
 
+	// Get containers and see which volumes are used
 	containers, err := client.Containers(ctx)
 	if err != nil {
 		return err
 	}
-	usedVolumes, err := usedVolumes(ctx, containers)
+
+	usedVolumesList, err := usedVolumes(ctx, containers)
```
apostasie, line 47: Better variable name...

pkg/cmd/volume/prune.go
```diff
@@ -26,29 +26,47 @@ import (
...
+
+	// Iterate through the known volumes, making sure we do not remove in-use volumes
+	// but capture as well anon volumes (if --all was passed)
 	for _, volume := range volumes {
-		if _, ok := usedVolumes[volume.Name]; ok {
+		if _, ok := usedVolumesList[volume.Name]; ok {
 			continue
 		}
 		if !options.All {
+			if volume.Labels == nil {
```
apostasie, line 67: Now that labels can be nil, need to check for that.

pkg/cmd/volume/rm.go
```diff
@@ -32,47 +32,61 @@ import (
...
+		return err
+	}
+
+	volumeNames := []string{}
+	cannotRemove := []error{}
+
 	for _, name := range volumes {
-		if _, ok := usedVolumes[name]; ok {
-			return fmt.Errorf("volume %q is in use", name)
+		if _, ok := usedVolumesList[name]; ok {
```
apostasie, line 62: Align with docker behavior

pkg/composer/up_volume.go
```diff
@@ -41,6 +41,9 @@ func (c *Composer) upVolume(ctx context.Context, shortName string) error {
 
 	// shortName is like "db_data", fullName is like "compose-wordpress_db_data"
 	fullName := vol.Name
+	// FIXME: this is racy. By the time we get below to creating the volume, there is no guarantee that things are still fine
+	// Furthermore, volStore.Get no longer errors if the volume already exists (docker behavior), so, the purpose of this
+	// call needs to be assessed (it might still error if the name is malformed, or if there is a filesystem error)
 	volExists, err := c.VolumeExists(fullName)
```
apostasie, line 47: I do not understand the compose code. Why do we shell out?
Leaving these untouched for now.

pkg/lockutil/lockutil_unix.go
```diff
@@ -32,18 +32,18 @@ func WithDirLock(dir string, fn func() error) error {
...
-		if err := Flock(dirFile, unix.LOCK_UN); err != nil {
+		if err := flock(dirFile, unix.LOCK_UN); err != nil {
 			log.L.WithError(err).Errorf("failed to unlock %q", dir)
 		}
 	}()
 	return fn()
 }
 
-func Flock(f *os.File, flags int) error {
+func flock(f *os.File, flags int) error {
```
apostasie, line 46: This should not be public. It is unix only.

pkg/mountutil/mountutil.go
```diff
@@ -203,16 +201,11 @@ func handleAnonymousVolumes(s string, volStore volumestore.VolumeStore) (volumeS
 func handleNamedVolumes(source string, volStore volumestore.VolumeStore) (volumeSpec, error) {
 	var res volumeSpec
 	res.Name = source
-	vol, err := volStore.Get(res.Name, false)
+
+	// Create returns an existing volume or creates a new one if necessary.
+	vol, err := volStore.Create(res.Name, nil)
```
apostasie, line 206: No reason for any of that anymore.

pkg/mountutil/mountutilmock/volumestore.mock.go
```diff
@@ -106,16 +106,26 @@ func (m *MockVolumeStoreMockRecorder) Remove(names any) *gomock.Call {
 	return m.mock.ctrl.RecordCallWithMethodType(m.mock, "Remove", reflect.TypeOf((*MockVolumeStore)(nil).Remove), names)
 }
 
-// Dir mocks the Dir method of VolumeStore
-func (m *MockVolumeStore) Dir() string {
+// Lock mocks the Lock method of VolumeStore
```
apostasie, line 109: The mock stuff. Still have no idea what is this and what to do here.
Merely ensuring we fullfill the interface here, but if you have info on what I should do here, please speak.

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -31,21 +31,33 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/strutil"
 )
 
-// Path returns a string like `/var/lib/nerdctl/1935db59/volumes/default`.
-func Path(dataStore, ns string) (string, error) {
```
apostasie, line 35: This should not be public.
Consumer should only use the interface methods and not fiddle with the underlying (filesystem based) implementation.

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -55,175 +67,246 @@ func New(dataStore, ns string) (VolumeStore, error) {
...
 
-type volumeStore struct {
-	// dir is a string like `/var/lib/nerdctl/1935db59/volumes/default`.
-	// dir is guaranteed to exist.
-	dir string
+	vs.locked = dirFile
+	return nil
 }
 
-func (vs *volumeStore) Dir() string {
```
apostasie, line 78: See other comment on publicizing these.

cmd/nerdctl/system_prune_linux_test.go
```diff
@@ -32,6 +32,11 @@ import (
 
 func TestSystemPrune(t *testing.T) {
 	testutil.RequiresBuild(t)
+	// FIXME: for some reason, using a dedicated namespace does work well with rootless, but fails with rootful... :/
```
apostasie, line 35: No idea what's going on here, and why this happens only with rootful.
Will investigate later when I get to system tests...

cmd/nerdctl/volume_prune_linux_test.go
```diff
@@ -17,30 +17,137 @@
...
 func TestVolumePrune(t *testing.T) {
-	base := testutil.NewBase(t)
-	tID := testutil.Identifier(t)
-	base.Cmd("volume", "prune", "-a", "-f").Run()
+	// Volume pruning cannot be parallelized for Docker, since we need namespaces to do that in a way that does interact with other tests
+	if testutil.GetTarget() != testutil.Docker {
+		t.Parallel()
+	}
+
+	// FIXME: for an unknown reason, when testing ipv6, calling NewBaseWithNamespace per sub-test, in the tearDown/tearUp methods
```
apostasie, line 35: No idea.........
I guess it is another thing to add to the list for whomever is going to cleanup our testing helpers eventually.

apostasie:
@AkihiroSuda this is finally ready - sorry it took so long to get there.

Also it is bigger that I wanted it to be - sorry about that as well. I guess problems just kept on piling up as I parallelized more...

Pure code changes are relatively small and targeted.
The only really new / interesting bit is the introduction of the "durable" locking mechanism for `VolumeStore` that allow operations like `container create` or `container remove` to lock the store for the duration of their operation.

The rest is merely small bug fixes - and a boatload of new tests.

I have added a bunch of github comments alongside the code to facilitate review. 

PTAL at your convenience.

Thanks!

cmd/nerdctl/volume_create.go
```diff
@@ -17,6 +17,9 @@
 package main
 
 import (
+	"fmt"
+
+	"github.com/containerd/containerd/errdefs"
```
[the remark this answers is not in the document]
AkihiroSuda: Yes, but probably it should be a separate PR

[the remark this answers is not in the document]
apostasie: @djdongjin I do not think this will work, unless we update everything (including dependencies) to v2, as both packages will yield different results wrt `errors.Is`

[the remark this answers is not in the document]
apostasie: Correction: may actually work with 1.7.19 - still, certainly better as a separate PR.

pkg/composer/up_volume.go
AkihiroSuda: > shell out

IIRC it was just for ease of quick implementation

cmd/nerdctl/system_prune_linux_test.go
```diff
@@ -32,6 +32,11 @@ import (
...
+	// Maybe something related to buildctl/kit?
```
AkihiroSuda, line 36: Probably because BuildKit is not enabled for rootless CI?

cmd/nerdctl/volume_prune_linux_test.go
```diff
@@ -17,30 +17,137 @@
...
+			"-v", anonIDBusy, testutil.CommonImage).AssertOK()
+	}
+
+	subTearDown := func(tID string) {
+		base := testutil.NewBaseWithNamespace(t, tID)
+		base.Cmd("rm", "-f", tID).Run()
+		//base.Cmd("volume", "rm", "-f", tID+"-busy").Run()
+		// base.Cmd("volume", "rm", "-f", tID+"-free").Run()
+		// base.Cmd("volume", "rm", "-f", anonIDFree).Run()
+		// base.Cmd("volume", "rm", "-f", anonIDBusy).Run()
```
AkihiroSuda, line 59: Is this WIP?

apostasie: No. This should have been cleaned-up - it just escaped me.

Let me do one last go through the diff and make sure there is no leftovers.

Thanks for catching it.

apostasie:
Rebased.

@AkihiroSuda last tiny push does remove the spurious commented leftovers, extra line, and updated the comment on rootful/buidkit.

Just did another proofread of the diff -  we should be good to go (let's make sure the CI is happy of course).

AkihiroSuda (approved):
Thanks

apostasie:
> Thanks

Thanks a lot @AkihiroSuda 
Appreciate your time and effort reviewing these.

PR #3135, opened 2024-06-21
pkg/defaults: add defaults_darwin.go 

sondavidb:
Before this change, people using nerdctl as a library could not build if their dependency tree took `pkg/defaults`. This adds a dummy file that implements the missing necesssary variables so that it can compile on Mac.

I understand we do not natively support building on Mac, but as this is a library as well as a binary I felt this change might be helpful for a few people.

### DIFF
--- a/pkg/defaults/defaults_darwin.go
+++ b/pkg/defaults/defaults_darwin.go
@@ -1,0 +1,45 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+// This is a dummy file to allow usage of library functions
+// on Darwin-based systems.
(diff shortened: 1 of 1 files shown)


sondavidb:
Wonder if we should just add a dummy file in pkg/defaults to avoid this in the future

pendo324:
Makes sense to me if the intention of `pkg/api/types` is to separate all of the non-platform specific code so that its more broadly usable

AkihiroSuda:
Not sure if this should be in "api".

What fails on macOS?

sondavidb:
> Not sure if this should be in "api".

All the other structs are in `pkg/api/types`, so this was my reason for putting it here. Do you think it makes less sense here?

> What fails on macOS?

If we pull any libraries that requires `pkg/defaults` down the dependency tree, it will fail with `build constraints exclude all Go files in /Users/dummyuser/go/pkg/mod/github.com/containerd/nerdctl/vs@v2.0.0-beta.5/pkg/defaults`. On looking closer it seems that the only non-platform specific or binary-related function call to this library is in the config, hence my suggested refactor.

Another thought I had was to just make a dummy file `defaults_darwin.go` and implement all the necessary functions with a no-op. Would this be preferred and more logical to you? It would also not break any other changes, while the proposed solution could be breaking.

AkihiroSuda:
> api

IIUC, the api package was created for third party projects that import Go packages of nerdctl.
So, unless the config is consumed by a third party project, it doesn't make much sense to move the config to the "api" package.

> defaults_darwin.go

SGTM

sondavidb:
Thanks. Changed commit title and PR title + message.

AkihiroSuda (approved):
Thanks

PR #3138, opened 2024-06-22
Add `shellcheck` lint to CI

abitrolly:
`find . -name '*.sh` loosk for filenames with .sh extension and passes the list to `shellcheck`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,9 +31,9 @@ jobs:
           version: v1.59.1
           args: --verbose
       - name: yamllint-lint
-        run: yamllint .
+        run: make lint-yaml
       - name: shellcheck
-        run: shellcheck $(find . -name '*.sh')
+        run: make lint-shell
 
   test-unit:
     runs-on: ubuntu-24.04
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,8 @@ jobs:
           args: --verbose
       - name: yamllint-lint
         run: yamllint .
+      - name: shellcheck
+        run: shellcheck $(find . -name '*.sh')
 
   test-unit:
     runs-on: ubuntu-24.04


fahedouch:
thank you @abitrolly , `shellcheck` is failing

abitrolly:
@fahedouch fixed all the warnings.

abitrolly:
Any ideas why rootless tests fail?

abitrolly:
Reverting https://github.com/abitrolly/nerdctl/commit/87d83faa5a1281c87946afdb5319ba4e60ca5061 didn't help.

abitrolly:
Reverting 4191964f849c03c8864c38efb604f0c5a65a7804 doesn't seem to help either.

@fahedouch lint warnings fixed. Any ideas why rootless tests fail?

apostasie:
Hey @abitrolly 

I have been working on something similar, introducing shellcheck as a Makefile target: #3213 - and fixing some of the issues here: #3214

This is clearly duplicating your effort here - I am really sorry about that, as I had not noticed your PR.

I am of course happy to close my PRs if you want to carry yours forward and fix the rootless issue.

LMK your thoughts.

apostasie:
> Reverting [4191964](https://github.com/containerd/nerdctl/commit/4191964f849c03c8864c38efb604f0c5a65a7804) doesn't seem to help either.
> 
> @fahedouch lint warnings fixed. Any ideas why rootless tests fail?

Yes.
You are quoting the extra arguments for the containerd, which will break the startup call with systemd.

examples/nerdctl-ipfs-registry-kubernetes/ipfs-cluster/bootstrap.yaml.sh
```diff
@@ -28,7 +28,7 @@ done
 
 TMPIDFILE=$(mktemp)
 BOOTSTRAP_KEY=$(ipfs-key 2>"${TMPIDFILE}" | base64 -w 0)
-ID=$(cat "${TMPIDFILE}" | grep "ID " | sed -E 's/[^:]*: (.*)/\1/')
+ID=$(< "${TMPIDFILE}" grep "ID " | sed -E 's/[^:]*: (.*)/\1/')
```
apostasie, line 31: `grep "ID " "$TMPIDFILE"` might be more readable / easier to understand for most folks

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -29,19 +29,15 @@ set -eu
 
 # utility functions
 INFO() {
-	# https://github.com/koalaman/shellcheck/issues/1593
-	# shellcheck disable=SC2039
-	/bin/echo -e "\e[104m\e[97m[INFO]\e[49m\e[39m ${*}"
+	printf "\e[104m\e[97m[INFO]\e[49m\e[39m %s\n" "$@"
```
apostasie, line 32: That is changing behavior.
We should stick with "$*" to keep the way this is working.

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -29,19 +29,15 @@ set -eu
...
 }
 
 WARNING() {
-	# shellcheck disable=SC2039
-	/bin/echo >&2 -e "\e[101m\e[97m[WARNING]\e[49m\e[39m ${*}"
+	printf "\e[101m\e[97m[WARNING]\e[49m\e[39m %s\n" "$@" >&2
```
apostasie, line 36: See above.

Also, I personally prefer `>&2 printf` (redirect first) in that case as it makes it more clear immediately we are printing on stderr.

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -29,19 +29,15 @@ set -eu
...
 }
 
 ERROR() {
-	# shellcheck disable=SC2039
-	/bin/echo >&2 -e "\e[101m\e[97m[ERROR]\e[49m\e[39m ${*}"
+	printf "\e[101m\e[97m[ERROR]\e[49m\e[39m %s\n" "$@" >&2
```
apostasie, line 40: See above.

extras/rootless/containerd-rootless.sh
```diff
@@ -145,17 +145,17 @@ if [ -z $_CONTAINERD_ROOTLESS_CHILD ]; then
...
+		--net="$net" --mtu="$mtu" \
+		--slirp4netns-sandbox="$CONTAINERD_ROOTLESS_ROOTLESSKIT_SLIRP4NETNS_SANDBOX" \
+		--slirp4netns-seccomp="$CONTAINERD_ROOTLESS_ROOTLESSKIT_SLIRP4NETNS_SECCOMP" \
+		--disable-host-loopback --port-driver="$CONTAINERD_ROOTLESS_ROOTLESSKIT_PORT_DRIVER" \
 		--copy-up=/etc --copy-up=/run --copy-up=/var/lib \
 		--propagation=rslave \
-		$CONTAINERD_ROOTLESS_ROOTLESSKIT_FLAGS \
-		$0 $@
+		"$CONTAINERD_ROOTLESS_ROOTLESSKIT_FLAGS" \
+		"$0" "$@"
```
apostasie, line 156: This "$CONTAINERD_ROOTLESS_ROOTLESSKIT_FLAGS" here is the reason you are failing to start containerd in rootless.
You need to leave that one unquoted so that it does expand.

abitrolly:
@apostasie you clearly possess more knowledge than I in `bash` shenanigans. So feel free to commit another more clear PR and I'll close this. Or maybe you will find it easier to commit here, if you're among maintainers.

I am just blindly obeying `shellcheck` instructions. )

apostasie:
@abitrolly 
Yeah, bash and I have been in a love/hate/want-to-kill-you type of relationship for a long time :P 

I am not a maintainer so cannot merge. Let see what I can get in on my PRs and reconvene here.

Thanks again!

abitrolly:
Rebased. So now there is only CI command, but also a lot of empty commits that are not removed by `rebase --force --empty drop`.

examples/nerdctl-ipfs-registry-kubernetes/ipfs-cluster/bootstrap.yaml.sh
abitrolly: Well, I agree. :D

extras/rootless/containerd-rootless-setuptool.sh
abitrolly: I am not sure how "$@" is different from "${*}".

extras/rootless/containerd-rootless-setuptool.sh
abitrolly: I had to learn that syntax, because it is not intuitive. It is like shell is breaking command parameters to extract its own instructions from here. I was more comfortable thinking that ">&2" is some kind of chaining option that only comes at the end of a command, and also marks the end.

extras/rootless/containerd-rootless.sh
abitrolly: Makes sense. Looks like it contains multiple flags, so it should be expanded.

abitrolly:
After rebasing with `--no-keep-empty`, the PR now only contains CI code for enabling `shellcheck`.

.github/workflows/test.yml
```diff
@@ -32,6 +32,8 @@ jobs:
           args: --verbose
       - name: yamllint-lint
         run: yamllint .
+      - name: shellcheck
+        run: shellcheck $(find . -name '*.sh')
```
apostasie, line 36: We could use `make lint-shell` - and also for yamllint above `make lint-yaml`.

The point is to keep the same behavior on the CI, and on people laptops.

LMK what you think.

Thanks!

apostasie: LGTM otherwise - and we do need this on the CI for sure.

cc @AkihiroSuda

extras/rootless/containerd-rootless-setuptool.sh
apostasie: `"$@"` will expand to `"$1" "$2"` etc
`"$*"` will expand to `"$1 $2"`

and `${*}` is just `$*`

For eg, try:


```
for i in "$@"; do
        echo "> $i"
done

for i in "$*"; do
        echo "> $i"
done

for i in "${*}"; do
        echo "> $i"
done
```

and call the above script with `./script.sh 1 2 3`

.github/workflows/test.yml
abitrolly: I would prefer CI to be the single source of truth and commands instead of `make` with its tabs vs spaces, but such tooling is not there yet. Fixed.

apostasie (approved):

AkihiroSuda (approved):
Thanks

.github/workflows/test.yml
apostasie: @abitrolly have a look at https://dagger.io/ - clearly CI is a hot steaming pile of mess right now and we definitely need better - dagger is onto that mission.

PR #3155, opened 2024-06-27
pkg/cmd/container: fix wrong var name in create

sondavidb:
Looking at the [commit](https://github.com/containerd/nerdctl/commit/18f29fb006ebb79590acdb07253bdc75527cdcf5) associated with this change, it seems the wrong variable was used when copying the above code, resulting in the wrong OCI hook being used. This change should fix this.


sondavidb:
Related, opened #3156 to add testing to catch something like this.

sondavidb:
Oh, seems like I was wrong. Apologies. Will keep above mentioned issue up unless we think current tests are sufficient.

Issue #3134, opened 2024-06-21
`nerdctl system prune -a` removes volumes in use too


amalthundiyil:
### Description

`nerdctl` removes an image from private registry on `nerdctl system prune -a`

```
root@machine:~# nerdctl ps
CONTAINER ID    IMAGE                             COMMAND                   CREATED         STATUS    PORTS                     NAMES
fe0ba1bdaf95    docker.io/library/registry:2.7    "/entrypoint.sh /etc…"    16 hours ago    Up        0.0.0.0:5000->5000/tcp    registry
root@machine:~# curl http://localhost:5000/v2/_catalog
{"repositories":["python"]}
root@machine:~# nerdctl system prune -af
root@machine:~# curl http://localhost:5000/v2/_catalog
{"repositories":[]}
```

### Steps to reproduce the issue

1. `nerdctl run -d -p 5000:5000 --restart=always --name registry registry:2.7`
2. `nerdctl pull registry.hub.docker.com/library/python:3.9`
3. `nerdctl image tag registry.hub.docker.com/library/python:3.9  localhost:5000/python:3.9`
4. `nerdctl push localhost:5000/python:3.9`
5. `nerdctl system prune -af`
6. `curl http://localhost:5000/v2/_catalog`

### Describe the results you received and expected

`nerdctl` shouldn't remove images from _inside of a private container registry_.

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

```
Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v1.7.16
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Log: fluentd journald json-file syslog
  Storage: aufs native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile: builtin
 Kernel Version: 5.4.0-88-generic
 Operating System: Ubuntu 20.04.3 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 7.109GiB
 Name: machine
 ID: 165e5604-e251-4d53-aeda-4ef52c811b35
```

apostasie:
@AkihiroSuda I ll take this one

apostasie:
@AkihiroSuda the faulty part seems to be `container prune`.

Reading the code, I just do not get it.
[...]

PR #3163, opened 2024-07-01
Rewrite container remove

apostasie:
This is initially motivated by #3134 

Currently `container.Remove` is faulty as it starts removing resources (specifically anonymous volumes) before even confirming that the container **should** be removed.

There are other likely issues with the current implementation - there seem to be cases where we delete the container, but do not cleanup networking on windows - and likely different behavior wrt snapshot cleanups dependent on container status.

Overall, this function is complex to reason about, because of the defer, the named return value, the confusing logic in what should happen when and what is a hard error or not.
All these reasons explain why the mistakes / issues were made in the first place.

This PR is a suggested rewrite, with the following objectives:
- make it (a little bit) simpler to reason about
- comment generously
- make variables names easier to read
- fix the issues

I appreciate this is impactful, but I also do not think we are quite done here and there may be more bugs in the overall container lifecycle when it comes to destruction.

Nevertheless - this is definitely fixing the core problem which is pretty bad (deletion of anonymous volumes for containers that should not be deleted).

This is draft until the CI is green (can't test windows here).

### DIFF
--- a/cmd/nerdctl/container_prune_linux_test.go
+++ b/cmd/nerdctl/container_prune_linux_test.go
@@ -26,15 +26,28 @@ func TestPruneContainer(t *testing.T) {
 	base := testutil.NewBase(t)
 	tID := testutil.Identifier(t)
 
-	base.Cmd("run", "-d", "--name", tID+"-1", testutil.CommonImage, "sleep", "infinity").AssertOK()
-	defer base.Cmd("rm", "-f", tID+"-1").Run()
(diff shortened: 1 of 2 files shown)


apostasie:
cc @AkihiroSuda PTAL at your convenience.
Windows is failing for now but does not seem related? (can you poke it?)

AkihiroSuda (approved):
Thanks, can we have a test?

apostasie:
@AkihiroSuda sure - sorry for not doing it in the first place, I was in a rush.

Added a test for #3134

AkihiroSuda (approved):
Thanks

PR #3170, opened 2024-07-04
Cleanup login command (remove global var and normalize to other commands

apostasie:
Been annoyed by this global var for a while.
I do not like the risk of collision on the name `options` with other commands.

Unless I missed the reason for `login` being different from other commands, suggestion here is to align with the syntax we are using everywhere else - remove the global var, do not use `FlagsVar`, remove the useless struct and normalize function names and order.

Let me know if this is desirable.

### DIFF
--- a/cmd/nerdctl/login.go
+++ b/cmd/nerdctl/login.go
@@ -28,15 +28,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks, nice catch

PR #3177, opened 2024-07-05
Make systemctl happy by not passing an explicit signal for kill

### LINKED ISSUE #3147: `-test.kill-daemon` does not work on Ubuntu 24.04 (`Failed to kill unit containerd.service: Failed to send signal SIGKILL to auxiliary processes: Invalid argument\n`)
https://github.com/containerd/nerdctl/blob/a5445322e9ed39ce18d82c8391fdbc7a84545855/pkg/testutil/testutil.go#L564-L573

apostasie:
That should fix #3147.

See discussion there for the rationale.


AkihiroSuda (approved):
Thanks

PR #3178, opened 2024-07-05
Move to 24.04 for docker tests

### LINKED ISSUE #3147: `-test.kill-daemon` does not work on Ubuntu 24.04 (`Failed to kill unit containerd.service: Failed to send signal SIGKILL to auxiliary processes: Invalid argument\n`)

apostasie:
I am assuming this will fail until we fix the issue in #3147 but curious what else is there to fix.

Draft for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -253,7 +253,7 @@ jobs:
         run: GO_VERSION="$(echo ${{ matrix.go-version }} | sed -e s/.x//)" make binaries
 
   test-integration-docker-compatibility:
-    runs-on: ubuntu-22.04  # TODO: ubuntu-24.04
+    runs-on: ubuntu-24.04
     timeout-minutes: 45
     steps:
       - uses: actions/checkout@v4.1.7
--- a/cmd/nerdctl/container_logs_test.go
+++ b/cmd/nerdctl/container_logs_test.go
@@ -209,7 +209,12 @@ func TestLogsWithForegroundContainers(t *testing.T) {
 }
 
 func TestTailFollowRotateLogs(t *testing.T) {
-	t.Parallel()
+	// FIXME this is flaky by nature... 5 lines is arbitrary, 2000 ms is arbitrary, and both are some sort of educated
+	// guess that things will mostly always kinda work maybe...
+	// Furthermore, parallelizing will put pressure on the daemon which might be even slower in answering, increasing
+	// the risk of transient failure.
+	// This test needs to be rethought entirely
+	// t.Parallel()
 	if runtime.GOOS == "windows" {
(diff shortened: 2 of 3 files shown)


apostasie:
Rebasing on top of #3177
Let's see what goes.

apostasie:
List of amendments:

* fixing `--kill.daemon` (other PR)

* TestRunCgroupParent - invalid cgroup name to compare with (faulty assumption that nerdctl was the only one of the two with systemd cgroup)

* TestTailFollowRotateLogs - somewhat unrelated, but a significant source of flakyness it seems - using Combined to test is a mistake, too short a timeout, not checking for empty lines - also, our code needs to be fixed as well (applied a bandaid in the meantime)

pkg/logging/json_logger.go
```diff
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/containerd/containerd/runtime/v2/logging"
+	"github.com/containerd/errdefs"
```
AkihiroSuda, line 32:
```suggestion
	"github.com/containerd/containerd/errdefs"
```
(Should be changed to github.com/containerd/errdefs later in another PR)

apostasie: Nice catch, thanks.

apostasie:
Rebasing against main for good measure.

apostasie:
1 failure - IPFS

I think this is good to go @AkihiroSuda

pkg/logging/json_logger.go
```diff
@@ -129,7 +130,15 @@ func (jsonLogger *JSONLogger) PostProcess() error {
 func viewLogsJSONFile(lvopts LogViewOptions, stdout, stderr io.Writer, stopChannel chan os.Signal) error {
```
AkihiroSuda, line 130: Is this function called in docker tests?

apostasie: Indirectly, yes

apostasie: Do you mean "is it called with target=docker"?

AkihiroSuda: Right, probably this should have another GitHub issue and PR?

apostasie: Oh, I see what you mean. Yes, I guess I can split it out and submit a separate PR.

apostasie: Pushed again without the above (will submit as separate PR).

apostasie:
@AkihiroSuda done

AkihiroSuda (approved):
Thanks

PR #3179, opened 2024-07-05
Workaround flaky save

apostasie:
Let's see what the CI says...

This one has been driving me bonkers for a while.

The issue is obviously still here (and I believe it is a containerd bug), but the workaround seems to work.

### DIFF
--- a/cmd/nerdctl/image_save_test.go
+++ b/cmd/nerdctl/image_save_test.go
@@ -25,7 +25,11 @@ import (
 )
 
 func TestSaveById(t *testing.T) {
-	base := testutil.NewBase(t)
+	// See detailed comment in TestRunCustomRootfs for why we need a separate namespace.
+	base := testutil.NewBaseWithNamespace(t, testutil.Identifier(t))
+	t.Cleanup(func() {
+		base.Cmd("namespace", "remove", testutil.Identifier(t)).Run()
+	})
 	base.Cmd("pull", testutil.CommonImage).AssertOK()
 	inspect := base.InspectImage(testutil.CommonImage)
 	var id string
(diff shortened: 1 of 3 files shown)


apostasie:
Ok, let's not jinx it, but for me locally, it DOES fix #2327 and with four pushes, we never failed once the dreaded `TestRunCustomRootfs`:
https://github.com/containerd/nerdctl/actions
(other stuff failing though...)

I am convinced the underlying issue is still here though - I also reproduced it with nerdcl/mod-v2 & containerd-v2 - so, it is here to stay.

@AkihiroSuda I suggest we merge this ASAP.

Worse case scenario it does nothing (this is good hygiene anyway to have that in a separate namespace). Best case scenario, it does make our lives significantly better...

Tagging @yankay and @fahedouch - who had some discussion about the underlying issue in #2327 .

For good measure, I am namespacing as well the other save tests.

cmd/nerdctl/image_save_test.go
```diff
@@ -25,7 +25,10 @@ import (
 )
 
 func TestSaveById(t *testing.T) {
-	base := testutil.NewBase(t)
+	base := testutil.NewBaseWithNamespace(t, testutil.Identifier(t))
```
AkihiroSuda, line 28: Could you add a comment about namespacing?

apostasie: Done and pushed.

AkihiroSuda:
Please squash the commits, then LGTM

apostasie:
@AkihiroSuda done. Thanks.

apostasie:
Wait. Something wrong with the project checks?

apostasie:
@AkihiroSuda 

Ok, so.
Apparently, containerd project-checks github action does bundle and depend on Vincent's https://github.com/vbatts/git-validation which by default calls `git show --check` which does enforce some form of linting on whitespaces.

Whitespaces "issues" have been flagged in my commit about `testing` from today (the ones that modified just markdown files).

The reason it did escape us is that we recently disabled running CI test for markdown-only changes...
And clearly we should still run `Project Checks` for these.

I'll take care of the CI right now (splitting out project checks and re-enabling them for md files changes), but unfortunately... it seems that we have to rewrite the commit now - which I assume only you can do.

The problematic commits are:
- the merge commit 402205749a5be864030352a184aa1071f7f7f22c
- and the original 5267e4ef5d2abf6c7cb332d2b76efe0fa87bd7b5

Sorry about this mess.
This `testing` doc is cursed...

AkihiroSuda (approved):
Thanks

PR #3180, opened 2024-07-05
Re-enable project checks for .md only changes

apostasie:
#3041 was too broad and should not have removed project checks for .md-only changesets.

This fixes that.

(details in #3179).


AkihiroSuda (approved):
Thanks

PR #3181, opened 2024-07-05
Fix IPFS container start routine

apostasie:
TestIPFSAddress and TestIPFSComposeUp have been failing regularly on the CI (with a rather long timeout).

Furthermore, the initialization code was using static ports and container name, making it impossible to parallelize.

This PR rewrites the initialization code of the IPFS registry container after the regular registry container start method - giving guarantees that there is no port or name collision, a sensible timeout in case of failure, an actual test to verify the registry is up and running.
It also removes the unnecessary testing of `detached-ns` (apparently triggering failures with rootlesskit 1).

CI is green on push (except for 1 unrelated failure involving systemctl that we have seen somewhere else).

Along with the other changes from today, this PR should increase CI reliability quite a bit.

PTAL at your convenience @AkihiroSuda @ktock .

Of course would love to see this in ASAP as failing CI has been a massive PITA, but lmk evidently if there is anything you would like tweaked.

Cheers.

### DIFF
--- a/pkg/testutil/testregistry/testregistry_linux.go
+++ b/pkg/testutil/testregistry/testregistry_linux.go
@@ -55,6 +55,7 @@ type TokenAuthServer struct {
 func EnsureImages(base *testutil.Base) {
 	base.Cmd("pull", testutil.RegistryImage).AssertOK()
 	base.Cmd("pull", testutil.DockerAuthImage).AssertOK()
+	base.Cmd("pull", testutil.KuboImage).AssertOK()
 }
 
 func NewAuthServer(base *testutil.Base, ca *testca.CA, port int, user, pass string, tls bool) *TokenAuthServer {
@@ -112,6 +113,8 @@ acl:
 	port, err = portlock.Acquire(port)
 	assert.NilError(base.T, err, fmt.Errorf("failed acquiring port: %w", err))
 	containerName := fmt.Sprintf("auth-%s-%d", name, port)
+	// Cleanup possible leftovers first
+	base.Cmd("rm", "-f", containerName).Run()
 
 	cleanup := func(err error) {
 		result := base.Cmd("rm", "-f", containerName).Run()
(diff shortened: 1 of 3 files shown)


apostasie:
Status of current failures:
- [x]unrelated  TestRunCustomRootfs - now rebased against merged workaround - should not happen anymore 
- [ ] unrelated TestCreateWithMACAddress is flaky (also seen on other PRs) - investigating now
- [ ] IPFS server seems to randomly not start - will increase timeout and see if it helps

apostasie:
@AkihiroSuda @ktock I can use an opinion here.

After a bunch of CI fixes, and the changes here that randomize the port being used, I end-up with a fairly stable build, EXCEPT systematically failing for rootlesskit v1.1.1 ONLY.

Is there something I should know about networking with rootlesskit 1?
Obviously my changes triggered this, so, I assume randomizing the port makes it unhappy somehow?

Thoughts?

pkg/testutil/testregistry/testregistry_linux.go
```diff
@@ -232,7 +235,78 @@ func (ba *BasicAuth) Params(base *testutil.Base) []string {
 	return ret
 }
 
+func NewIPFSRegistry(base *testutil.Base, ca *testca.CA, port int, auth Auth, boundCleanup func(error)) *RegistryServer {
+	EnsureImages(base)
+
+	name := testutil.Identifier(base.T)
+	// listen on 0.0.0.0 to enable 127.0.0.1
+	listenIP := net.ParseIP("0.0.0.0")
```
ktock, line 243: Why not  127.0.0.1?

apostasie: Then it will not allow access using (non-localhost) host-ip, reducing possible test scope (comment is slightly misleading, originating from previous code in the registry start function).

ktock (approved):
Thanks

apostasie:
Thanks a lot @ktock 
Appreciate your time and attentive review.

PR #3188, opened 2024-07-05
[PRIORITY] Improve tests debugging output

apostasie:
This is a small quality of life improvement to ease debugging of tests.

When commands are unexpectedly failing, output `.String()` which includes the actual command and the exit code instead of `.Combined()` which only shows stdout+stderr.


apostasie:
Failure is unrelated - `TestTailFollowRotateLogs` - which should get better with the band aid once #3178 gets merged.

Thanks.

AkihiroSuda (approved):
Thanks

PR #3189, opened 2024-07-05
[CANARY] Disable test retries

apostasie:
IMHO we cannot keep on hiding problems by "retrying tests" until it works.
This is a doomed approach.

I perfectly understand how painful it is to go and debug these flakyness / concurrency issues, and that it is much easier to ignore the problem - but we *have* to solve these - they are definitely indicative of deep quality issues and the only way to do so is to make it as obvious as possible instead of hiding it under the retry-carpet.

This is a canary PR that will very likely keep failing for some time - until we fix ALL these issues.

I will rebase it against main daily, and eventually we should get it green and stable.

Finally note that this canary is usually AHEAD of main - using commits from pending PRs fixing flakyness

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,12 +93,7 @@ jobs:
           docker run --privileged --rm tonistiigi/binfmt --install linux/arm64
           docker run --privileged --rm tonistiigi/binfmt --install linux/arm/v7
       - name: "Run integration tests"
-        uses: nick-fields/retry@v3
-        with:
-          timeout_minutes: 30
-          max_attempts: 2
-          retry_on: error
-          command: docker run -t --rm --privileged test-integration
+        run: docker run -t --rm --privileged test-integration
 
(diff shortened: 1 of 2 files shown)


apostasie:
Of course, please do not restart any of these manually.
It should be green on the first try immediately after push.

apostasie:
Green again, but #3225 is still a clear issue before we can consider merging this.

apostasie:
08/09/2024:
- TestLogsWithForegroundContainers
- TestIPFSComposeUp

Run: https://github.com/containerd/nerdctl/actions/runs/10323534574/job/28581121980

The logs one is perplexing, but far from being the only log problem in CI. We should really have a look at logs overall.

The IPFS one is also happening often - `use of closed network connection` may suggest low quality code / failure to handle errors? IPFS overall has been very flaky so same here, a serious rewrite might be in order.

apostasie:
Closing in favor of #3305

Issue #3186, opened 2024-07-05
Flaky test list


apostasie:
CI has been improved with the last few PRs, but there are still sources of flakyness.

Floodgates are open here: #3189

This ticket is to list them all, and get to the bottom of it and fix them all.

Feel free to add here any transient failure you see in the comments, with a copy of the logs and a link to the failing build (first failure in the run ONLY).

Fixed by #3192:
- [x] TestCreateWithMACAddress - fails regularly - no idea - `    container_create_linux_test.go:121: assertion failed: res.ExitCode is not 0: Command should have succeeded%!(EXTRA string=time="2024-07-05T16:36:12Z" level=fatal msg="container \"e06fb05e3174c1be8afedfc8d58981cdf90f14507b5d7040e6e6f2448823b7ca\" in namespace \"nerdctl-test\": not found"
        )` see https://github.com/containerd/nerdctl/actions/runs/9811180349/job/27092899863?pr=3178
- [x] TestExecEnv on windows - ` container_exec_test.go:79: assertion failed: res.ExitCode is not exitCode: time="2024-07-05T16:54:47Z" level=fatal msg="container \"30bbda4628649e094fee7f172c3aa7c16b1ddb032685f46bc297c9126fa9cefd\" in namespace \"nerdctl-test\": not found"` - looks strangely similar - not sure what is going on here - is there some concurrent prune happening? https://github.com/containerd/nerdctl/actions/runs/9811363482/job/27093391801?pr=3181
- [x] TestRunUserGID `2024-07-05T17:26:39Z" level=fatal msg="container \"d5eddc278faf40a910638e34c9186d5d1b4027c77be4e08f38192850bb194190\" in namespace \"nerdctl-test\": not found` (https://github.com/containerd/nerdctl/actions/runs/9811583654/job/27093967186?pr=3178)

apostasie:
These issues look a lot like #3092

Something really weird is going on.

apostasie:
Latest "no-retry" canary PR:

- [ ] TestTailFollowRotateLogs: 2 counts - this is very likely fixed by changes in #3178
- [ ] TestIPFSAddress: 1 count - probably fixed by #3181 
- [ ] TestNetworkWithNamespace: 4 counts - `task not found` 3 times - possibly another variant of #3167 - last count gives `container not created: not found` which is a bit more worrisome as this is from containerd
[...]

Issue #3092, opened 2024-06-16
network appears to not be immediately available after a call to `network create`


apostasie:
### Description

There seems to be latency between network create return, and availability of the network - or between the container creation return and the operation attaching it to the network.

Specifically:
- create a network
- run a container
- destroy the container
- destroy the network

^ will fail occasionally.

### Steps to reproduce the issue

Repeatedly create a network and immediately run a container attached to it, then destroy both in short succession.

### Describe the results you received and expected

Errors out complaining about the network not existing.

This is being run as part of our test suites (see `network remove` tests).

```
	//    network_remove_linux_test.go:80: assertion failed: res.ExitCode is not exitCode: time="2024-06-16T00:53:11Z" 
	//   level=fatal msg="failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: 
	//  error during container init: error running hook #0: error running hook: exit status 1, stdout: , 
	//  stderr: time=\"2024-06-16T00:53:11Z\" level=fatal msg=\"no such network: \\\"nerdctl-testnetworkremovewithstoppedcontainer\\\"\"\n
	//  Failed to write to log, write /var/lib/nerdctl/1935db59/containers/nerdctl-test/ce14cea63aab995e164f80901fbcb4bf7eedfa0a2b2da66a1c3923a90f52c474/oci-hook.startContainer.log: file already closed: unknown"

```

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Still not completely sure what is going on.
The error sometimes comes as:

`network_remove_linux_test.go:110: assertion failed: res.ExitCode is not exitCode: time="2024-06-16T00:39:09Z" level=fatal msg="container \"3a17eace71ae75ab7711011dc98e90e16d2bdf1af7ffc2b198399831236ef8b9\" in namespace \"nerdctl-test\": not found"`

This message is being sent by containerd metadata service, very likely called by `local.go` / getContainer.

It is unclear to me why this is happening. Hypothesis is that the container was not created properly, then since it is `--rm`, somehow containerd thinks it should be deleted, but it does not exist.
So maybe this message is indicative of a bug in containerd logic.

apostasie:
> Still not completely sure what is going on. The error sometimes comes as:
> 
[...]

PR #3192, opened 2024-07-05
[PRIORITY] Do not hard error if containers are gone

apostasie:
This is another instance of #3167 (see ticket for rationale - in a shell: do not assume containers are still there after a call to client.Containers...)

In that specific case, it will make `create` fail if another unrelated container got removed in a racy way.

It is very likely responsible for a large number of failures on the CI, and definitely strikes more with parallelization.

This very likely will address #3092, #3186, and possibly a number of others.

@AkihiroSuda @fahedouch if you are around - would appreciate a quick merge on this, preferably before other QA/CI PRs so that I can rebase and see what problems are left (specifically for the IPFS  one).

Of course, I will rebase #3189 as well ASAP which should give us a good hint.

Finally, if this is as bad as I think it is, merging this should give us a good speed boost on the CI (as a lot of "retries" will no longer be necessary).

Thanks a lot!

### DIFF
--- a/pkg/cmd/container/run_mount.go
+++ b/pkg/cmd/container/run_mount.go
@@ -19,6 +19,7 @@ package container
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -307,6 +308,12 @@ func generateMountOpts(ctx context.Context, client *containerd.Client, ensuredIm
 	for _, c := range containers {
 		ls, err := c.Labels(ctx)
 		if err != nil {
+			// Containerd note: there is no guarantee that the containers we got from the list still exist at this point
+			// If that is the case, just ignore and move on
+			if errors.Is(err, errdefs.ErrNotFound) {
+				log.G(ctx).Debugf("container %q is gone - ignoring", c.ID())
+				continue
+			}
 			return nil, nil, nil, err
 		}
 		_, idMatch := vfSet[c.ID()]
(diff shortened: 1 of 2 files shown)


apostasie:
Note: in the future, when we will be in a better position wrt CI, I will add tests to stress concurrency to ensure these problem patterns do not get reintroduced in the codebase.

pkg/cmd/container/run_mount.go
```diff
@@ -307,6 +308,9 @@ func generateMountOpts(ctx context.Context, client *containerd.Client, ensuredIm
 	for _, c := range containers {
 		ls, err := c.Labels(ctx)
 		if err != nil {
+			if errors.Is(err, errdefs.ErrNotFound) {
```
AkihiroSuda, line 311: Can we print a warning

apostasie: Mmm... I am on the fence.

It does not feel like an abnormal circumstance to me. It just means that a container that we retrieved from `client.Containers` a moment ago has been removed since.
Could happen naturally because a container was started with `--rm` and just exited - or because of concurrent operations (like we see on the CI).

Let me know your thoughts ^.

apostasie:
Note: failure seems entirely unrelated (systemd socket something)

pkg/cmd/container/run_mount.go
AkihiroSuda: What about Debugf?

apostasie: Sure.

apostasie: Debugf added (and also in the other place I fixed earlier) + a comment.

AkihiroSuda (approved):
Thanks

apostasie:
Last run failure on integ-rootless is IPFS - these should get fixed soon with #3181

PR #3194, opened 2024-07-06
Network prune should not fail if containers vanish

apostasie:
This is yet another instance of #3167

This here is less urgent than #3192 as it only affects `network prune`.

Issue was seen 3 times on this specific canary build: https://github.com/containerd/nerdctl/actions/runs/9815148442?pr=3189

Note that this is on top of #3192 - I will rebase on main as soon as it gets merged.


apostasie:
Back on main.

1 failure - still IPFS.

This one is good to go @AkihiroSuda

AkihiroSuda (approved):
Thanks

PR #3198, opened 2024-07-06
CI test fix: wait for systemd to come up

apostasie:
Seen many times with canary #3189 - see #3186 for full list of pain.

### DIFF
--- a/cmd/nerdctl/container_run_systemd_linux_test.go
+++ b/cmd/nerdctl/container_run_systemd_linux_test.go
@@ -46,7 +46,17 @@ func TestRunWithSystemdTrueEnabled(t *testing.T) {
 
 	base.Cmd("inspect", "--format", "{{json .Config.Labels}}", containerName).AssertOutContains("SIGRTMIN+3")
 
-	base.Cmd("exec", containerName, "systemctl", "list-jobs").AssertOutContains("jobs listed.")
+	base.Cmd("exec", containerName, "sh", "-c", "--", `tries=0
+until systemctl is-system-running >/dev/null 2>&1; do
+	>&2 printf "Waiting for systemd to come up...\n"
+	sleep 1s
+	tries=$(( tries + 1))
+	[ $tries -lt 10 ] || {
+		>&2 printf "systemd failed to come up in a reasonable amount of time\n"
+		exit 1
+	}
+done
+systemctl list-jobs`).AssertOutContains("jobs")
 }
 
 func TestRunWithSystemdTrueDisabled(t *testing.T) {


apostasie:
Windows failure is unrelated and likely fixed by #3133

PTAL at your convenience.

apostasie:
Latest push does increase the timeout to 10 seconds.

cmd/nerdctl/container_run_systemd_linux_test.go
```diff
@@ -46,7 +46,16 @@ func TestRunWithSystemdTrueEnabled(t *testing.T) {
 
 	base.Cmd("inspect", "--format", "{{json .Config.Labels}}", containerName).AssertOutContains("SIGRTMIN+3")
 
-	base.Cmd("exec", containerName, "systemctl", "list-jobs").AssertOutContains("jobs listed.")
+	base.Cmd("exec", containerName, "sh", "-c", "--", `tries=0
+until [ -e /run/systemd/io.system.ManagedOOM ]; do
```
AkihiroSuda, line 50: Would it be possible to use `systemctl is-system-running`?

apostasie: Of course. I am just not that familiar with systemd TBH, so, whatever is best in their world.

Will use your suggestion ^.

apostasie: Modified.

apostasie:
@AkihiroSuda changing test condition did change the timing slightly, then changing systemd response (`No jobs running.` vs `jobs listed`).

Latest push adapts to that......

This test is not really good honestly...

cmd/nerdctl/container_run_systemd_linux_test.go
```diff
@@ -46,7 +46,17 @@ func TestRunWithSystemdTrueEnabled(t *testing.T) {
...
+until systemctl is-system-running >/dev/null 2>&1; do
+	>&2 printf "Waiting for systemd to come up...\n"
+	sleep 1s
+	tries=$(( tries + 1))
+	[ $tries -lt 10 ] || {
+		>&2 printf "systemd failed to come up in a reasonable amount of time\n"
+		exit 1
+	}
+done
+systemctl list-jobs`).AssertOutContains("jobs")
```
AkihiroSuda, line 59: Maybe use `AssertOutWithFunc` to check `stdout == "No jobs running." || stdout == "jobs listed"`

apostasie: I do not see the point in testing stdout at all here.

We should not be in the business of testing systemd, especially on matching random mundane non-error messages.
What matters here is that systemd is up and running (exit code 0), and that we have a test that does not fail randomly (that is the point of this PR).

I would rather leave this stuff as-is for now - until we come back here to write more useful versions of these tests.

What do you think?

AkihiroSuda (approved):
Thanks

Issue #3199, opened 2024-07-07
 nerdctl will not output logs after log file is rotated


xyz-li:
### Description

`nerdctl -n k8s.log logs CONTAINER_ID -f ` won't output logs after kubelet rotate log filer of the container. 
Related issue with k8s https://github.com/kubernetes/kubernetes/issues/115701.

### Steps to reproduce the issue

1. Add the config  `containerLogMaxSize: 5Mi` to  `kubelet.yaml`. So kubelet will rotate the log soon.
2.  Build a image with this program
   ```
package main

 import (
        "fmt"
        "time"
 )

 func main() {
        i := 0
        for range time.Tick(10 * time.Microsecond) {
                fmt.Printf("print log to stand out %d\n", i)
                i++
        }
 }
   ```
3. Create a deploy with the previous image.
4. Run the command `nerdctl -n k8s.io logs containerd -f`

### Describe the results you received and expected

nerdctl output logs correctly after the log file is rotated. 

### What version of nerdctl are you using?

v2.0.0-rc.0

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #3200, opened 2024-07-07
fix: output log even log is rotated

xyz-li:
Fix the bug nerdctl not outputting log after log file rotation.
Fixes: #3199

### DIFF
--- a/pkg/logging/cri_logger.go
+++ b/pkg/logging/cri_logger.go
@@ -25,6 +25,7 @@ package logging
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -35,6 +36,7 @@ import (
 
 	"github.com/containerd/log"
 	"github.com/containerd/nerdctl/v2/pkg/logging/tail"
+	"github.com/fsnotify/fsnotify"
 )
 
 // LogStreamType is the type of the stream in CRI container log.
@@ -45,6 +47,9 @@ const (
 	Stdout LogStreamType = "stdout"
 	// Stderr is the stream type for stderr.
 	Stderr LogStreamType = "stderr"
+
+	// logForceCheckPeriod is the period to check for a new read
+	logForceCheckPeriod = 1 * time.Second
 )
 
 // LogTag is the tag of a log line in CRI container log.
@@ -89,7 +94,9 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
 	if err != nil {
 		return fmt.Errorf("failed to open log file %q: %v", logPath, err)
 	}
-	defer f.Close()
+	defer func() {
+		f.Close()
+	}()
 
 	// Search start point based on tail line.
 	start, err := tail.FindTailLineStartIndex(f, opts.Tail)
@@ -101,6 +108,8 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
 		return fmt.Errorf("failed to seek in log file %q: %v", logPath, err)
 	}
 
+	var watcher *fsnotify.Watcher
+
 	limitedMode := (opts.Tail > 0) && (!opts.Follow)
 	limitedNum := opts.Tail
 	// Start parsing the logs.
@@ -110,6 +119,9 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
 	isNewLine := true
 	writer := newLogWriter(stdout, stderr, opts)
 	msg := &logMessage{}
+	baseName := filepath.Base(logPath)
+	dir := filepath.Dir(logPath)
+
 	for {
 		select {
 		case <-stopChannel:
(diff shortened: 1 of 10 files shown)


pkg/logging/cri_logger.go
```diff
@@ -126,13 +140,48 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
...
+						if watcher, err = fsnotify.NewWatcher(); err != nil {
+							return fmt.Errorf("failed to create fsnotify watcher: %v", err)
+						}
+						defer watcher.Close()
+						if err := watcher.Add(dir); err != nil {
+							return fmt.Errorf("failed to watch directory %q: %w", dir, err)
+						}
+						// If we just created the watcher, try again to read as we might have missed
+						// the event.
+						continue
```
fahedouch, line 163: this can be handled in a `NewTail()` func.

other logger (e.g json logger) are using tail binary, IMO if you manage to make  `tail` implementation as utility func we can replace the legacy binary by this implementation

fahedouch, line 163: > // If we just created the watcher, try again to read as we might have missed
						// the event.

Sorry, I didn't understand the comment

pkg/logging/cri_logger.go
```diff
@@ -126,13 +140,48 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
...
+					}
+
+					var recreated bool
+					// Wait until the next log change.
+					continueNext, recreated, err = waitLogs(context.Background(), opts.ContainerID, baseName, watcher)
```
fahedouch, line 168:
```suggestion
					continueNext, recreated, err = startTail(context.Background(), opts.ContainerID, baseName, watcher)
```

pkg/logging/cri_logger.go
```diff
@@ -126,13 +140,48 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
...
+					if err != nil {
+						return err
+					}
+					if recreated {
+						newF, err := openFileShareDelete(logPath)
+						if err != nil {
+							if os.IsNotExist(err) {
```
fahedouch, line 175: the file is recreate, so it should absolutely exist otherwise we need to raise error

pkg/logging/cri_logger.go
```diff
@@ -126,13 +140,48 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
 					return fmt.Errorf("failed to read log file %q: %v", logPath, err)
 				}
 				if opts.Follow {
-
+					if !continueNext {
```
fahedouch, line 143: IMO continueNext is useless, the err returned by `waitLogs` is enough

pkg/logging/cri_logger.go
```diff
@@ -249,6 +298,36 @@ func (w *logWriter) write(msg *logMessage, addPrefix bool) error {
...
+// the error is error happens during waiting new logs.
+func waitLogs(ctx context.Context, id string, logName string, w *fsnotify.Watcher) (bool, bool, error) {
+	errRetry := 5
+	for {
+		select {
+		case <-ctx.Done():
+			return false, false, fmt.Errorf("context cancelled")
+		case e := <-w.Events:
+			switch e.Op {
+			case fsnotify.Write, fsnotify.Rename, fsnotify.Remove, fsnotify.Chmod:
```
fahedouch, line 312:
```suggestion
			case fsnotify.Write:
```

pkg/logging/cri_logger.go
xyz-li: I can make a new pr.

xyz-li: This  copy from this commit https://github.com/kubernetes/kubernetes/pull/73041/commits/2c30eee92f8a1b461848cd0ad2b2848d908a69de

xyz-li: If there are one or more WRITE before `watcher.Add` and the first event after `watcher.Add` is CREATE. If we remove the `continue`, then the reader will read data from the new file, and we may lose some data.

fahedouch: the watcher will detect only events after the `watcher.Add`;  the `continue` can't go back

xyz-li: > the watcher will detect only events after the `watcher.Add` (it can't watch in the past)

Yes. So if there is a WRITE before watcher.Add, apparently which can't be detected by watcher.Add. And the first event after watcher.Add is Create, then func `startTail` will return `recreated` as True, and close the old log file. At last, we may lose some logs.  So, we need the `continue`, let the reader read the old file again.

pkg/logging/cri_logger_test.go
```diff
@@ -230,3 +231,85 @@ func TestReadLogsLimitsWithTimestamps(t *testing.T) {
...
+				return
+			}
+
+			time.Sleep(20 * time.Millisecond)
+			newF := filepath.Join(dir, baseName)
+			if file, err = os.Create(newF); err != nil {
+				t.Errorf("unable to create new log file, error: %s", err.Error())
+				return
+			}
+		}
```
fahedouch, line 295: small suggestion; we have a logrotate module https://github.com/fahedouch/go-logrotate that you can use :)

pkg/logging/cri_logger_test.go
```diff
@@ -230,3 +231,85 @@ func TestReadLogsLimitsWithTimestamps(t *testing.T) {
...
+	}
+
+	// Finished writing into the file, close it, so we can delete it later.
```
fahedouch, line 298: where is it deleted ?

pkg/logging/json_logger_test.go
```diff
@@ -0,0 +1,173 @@
...
+		log.Log = fmt.Sprintf("line%d\n", line)
+		log.Stream = "stdout"
+		log.Time = time.Now().Format(time.RFC3339Nano)
+		time.Sleep(1 * time.Millisecond)
+		logData, _ := json.Marshal(log)
+		file.Write(logData)
+
+		if line == 5 {
+			file.Close()
+			// Pretend to rotate the log.
```
fahedouch, line 75: small suggestion; currently in nerdctl we use  https://github.com/fahedouch/go-logrotate to rotate logs :)

pkg/logging/cri_logger_test.go
xyz-li:
```
// TempDir returns a temporary directory for the test to use.
// The directory is automatically removed when the test and
// all its subtests complete.
// Each subsequent call to t.TempDir returns a unique directory;
// if the directory creation fails, TempDir terminates the test by calling Fatal.
func (c *common) TempDir() string {
```

xyz-li: The  `testing` framework will do it.

pkg/logging/json_logger_test.go
xyz-li: Rewriting using this library will be more complicated.

fahedouch:
please check https://github.com/containerd/nerdctl/issues/3225#issuecomment-2260598484

pkg/logging/cri_logger.go
```diff
@@ -126,13 +138,48 @@ func ReadLogs(opts *LogViewOptions, stdout, stderr io.Writer, stopChannel chan o
...
+					var recreated bool
+					// Wait until the next log change.
+					recreated, err = startTail(context.Background(), baseName, watcher)
+					if err != nil {
+						return err
+					}
+					if recreated {
+						newF, err := openFileShareDelete(logPath)
+						if err != nil {
+							if errors.Is(err, os.ErrNotExist) {
```
xyz-li, line 167: @fahedouch

fahedouch: thanks

fahedouch (approved):
Thanks

PR #3203, opened 2024-07-08
fix the created tasks cannot be deleted

Iceber:
[task.Delete](https://github.com/containerd/containerd/blob/6e67345ab7ed68d08155dcd39353344b9a098a49/task.go#L314) only allows direct deletion of tasks with PID 0.

fix: https://github.com/containerd/containerd/issues/10438

Related pr: https://github.com/containerd/containerd/pull/10442

### DIFF
--- a/pkg/cmd/container/remove.go
+++ b/pkg/cmd/container/remove.go
@@ -274,8 +274,17 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions 
 		if err == nil {
 			<-es
 		}
-	case containerd.Created, containerd.Stopped:
-		// Created and stopped containers always get removed
+	case containerd.Created:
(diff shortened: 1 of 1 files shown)


apostasie:
Moving the conversation back in nerdctl then (since the original ticket had been moved to containerd).

Finally stumbled upon a repro:
```
{ sleep 1; echo "\16\17"; } | /usr/bin/unbuffer -p nerdctl run -it --name foobar ghcr.io/stargz-containers/alpine:3.13-org

nerdctl rm -f foobar
ERRO[0000] 1 errors:
task must be stopped before deletion: created: failed precondition
```

@AkihiroSuda ^

AkihiroSuda (approved):
Thanks

PR #3207, opened 2024-07-09
Test fixes

apostasie:
Canary build #3189 (without any retry) has ran green for the first time today.

These here are the remaining fixes in the canary branch that have yet to be merged, and that should fix the last flaky tests.

In a shell:
- tests involving `builder prune` are not safe to be parallelized - disabling parallel on them
- increasing delay on `DelayOnceReader` - these things are inherently flaky, but the new extended delay seems to work better
- further increasing timeout for TestTailFollowRotateLogs - this one has been particularly painful - I think we should rewrite this test eventually, but for now, the new timeout seems to help

Once this is merged, I suggest we give it another week of running the Canary PR against main regularly to confirm we are constantly green, then we can consider merging it.

### DIFF
--- a/cmd/nerdctl/builder_build_test.go
+++ b/cmd/nerdctl/builder_build_test.go
@@ -152,7 +152,6 @@ CMD ["cat", "/hello2"]
 }
 
 func TestBuildFromStdin(t *testing.T) {
-	t.Parallel()
 	testutil.RequiresBuild(t)
 	base := testutil.NewBase(t)
 	defer base.Cmd("builder", "prune").Run()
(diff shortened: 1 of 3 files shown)


apostasie:
Green.

@AkihiroSuda and team, can you take a look at your convenience?

Thanks a lot.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Maybe this sort of PR should be split to 3 commits in the next time

apostasie:
> Maybe this sort of PR should be split to 3 commits in the next time

Will do.
Still trying to find my marks on these questions...

Thanks for the guidance. Appreciate it.

PR #3209, opened 2024-07-10
Adding a canary CI target to test against upcoming dependencies

### LINKED ISSUE #3208: Distribution v3 is coming

With distribution v3 coming (beta 1 just dropped), we should start testing against it to let them know what issues we are facing and to get ready ahead of their releases.

It might also be a good time to split out integration testing into groups so that we get more modular on what we test while avoiding growing the test matrix for the entire suite.

Example: we could have a distribution-latest target that would run *only* login / push / pull tests against distribution master (or latest alpha/beta/RC).

na

### Additional context

na

apostasie:
This PR introduces a new CI target (held in a separate file) that uses unreleased, upcoming, or otherwise `main` versions of our dependencies.

Specifically, it is setting us forward on:
- the next golang version
- containerd main
- distribution upcoming v3
- buildkit last rc

and a few others on main.

The purpose of this is to constantly have a "canary" target that will let us proactively identify and address upcoming issues in our dependency ecosystem.

Right now, it is not testing *everything*. Specifically ignored on this target are freebsd and ipv6 targets.

This will fix #3208

This PR also modifies slightly our "main" testing targets to pin containerd to an actual RC instead of main.

This is now running "green" - but of course would love feedback here so if you folks feel anything is missing / should be done differently, lmk and happy to discuss.

### DIFF
--- a/.github/workflows/test-canary.yml
+++ b/.github/workflows/test-canary.yml
@@ -1,0 +1,103 @@
+# This pipeline purpose is solely meant to run a subset of our test suites against upcoming or unreleased dependencies versions
+name: canary
+
+on:
+  push:
+    branches:
+      - main
+      - 'release/**'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+
+env:
+  UBUNTU_VERSION: "24.04"
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  linux:
+    runs-on: "ubuntu-24.04"
+    timeout-minutes: 40
+    steps:
+      - uses: actions/checkout@v4.1.7
+        with:
+          fetch-depth: 1
+      - name: "Prepare integration test environment"
+        run: |
+          ./hack/build-integration-canary.sh
+      - name: "Remove snap loopback devices (conflicts with our loopback devices in TestRunDevice)"
+        run: |
+          sudo systemctl disable --now snapd.service snapd.socket
+          sudo apt-get purge -y snapd
+          sudo losetup -Dv
+          sudo losetup -lv
+      - name: "Register QEMU (tonistiigi/binfmt)"
+        run: docker run --privileged --rm tonistiigi/binfmt --install all
+      - name: "Run unit tests"
+        run: go test -v ./pkg/...
+      - name: "Run integration tests"
+        continue-on-error: true
+        run: docker run -t --rm --privileged test-integration
+
+  windows:
+    runs-on: windows-latest
+    timeout-minutes: 30
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Set GO env
+        run: |
+          # Get latest containerd
+          args=(curl --proto '=https' --tlsv1.2 -fsSL -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28")
+          [ "${GITHUB_TOKEN:-}" == "" ] && {
+            >&2 printf "GITHUB_TOKEN is not set - you might face rate limitations with the Github API\n"
+          } || args+=(-H "Authorization: Bearer $GITHUB_TOKEN")
+          ctd_v="$("${args[@]}" https://api.github.com/repos/containerd/containerd/tags | jq -rc .[0].name)"
+          echo "CONTAINERD_VERSION=${ctd_v:1}" >> $GITHUB_ENV
+
+          # Get latest golang version and split it in components
+          norm=()
+          while read -r line; do
+            norm+=("$line")
+          done < \
+            <(sed -E 's/^go([0-9]+)[.]([0-9]+)([.]([0-9]+))?(([a-z]+)([0-9]+))?/\1.\2\n\4\n\6\n\7/i' \
+              <(curl -fsSL "https://go.dev/dl/?mode=json&include=all" | jq -rc .[0].version) \
+            )
+
+          # Serialize version, making sure we have a patch version, and separate possible rcX into .rc-X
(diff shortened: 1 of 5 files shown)


.github/workflows/test-canary.yml
```diff
@@ -0,0 +1,105 @@
...
+      - main
+      - 'release/**'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+
+env:
+  # If left blank, will default to latest stable
+  GO_VERSION: "1.23-rc1"
+  GO_VERSION_SETUP_GO: "1.23.0-rc.1"
```
AkihiroSuda, line 16: Maybe parse https://go.dev/dl/?mode=json&include=all to automatically get the latest version

Dockerfile
```diff
@@ -54,7 +54,7 @@ ARG SOCI_SNAPSHOTTER_VERSION=0.6.1
 FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.4.0 AS xx
 
 
-FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-bullseye AS build-base-debian
+FROM --platform=$BUILDPLATFORM golang:$GO_VERSION${GO_VERSION:+-}bullseye AS build-base-debian
```
AkihiroSuda, line 57: ?

.github/workflows/test-canary.yml
apostasie: I like that. Will do.

Dockerfile
apostasie: Right now, go is pinned (^) to the latest released RC, but in normal times, we might want to just be on `golang:bullseye` (or `golang:alpine`). With the current Dockerfile, this is not possible.
This change makes it so that an empty `GO_VERSION` will give us `golang:bullseye`, while a non-empty one will give us `golang:foo-bullseye`.

Of course if we do your suggestion above we will always set the go version to the latest available (detecting from https://go.dev/dl/?mode=json&include=all ) and this is less useful, so, I can remove that.

apostasie:
@AkihiroSuda 
With latest push:
- auto-detection of latest go release (+magic to convert it to a version acceptable to `actions/setup-go`)
- removed the stuff in Dockerfile to deal with empty GO_VERSION
- hook-up the variables into build-args

.github/workflows/test-canary.yml
```diff
@@ -0,0 +1,137 @@
...
+  BYPASS4NETNS_VERSION: "main"
+  KUBO_VERSION: "main"
+  GOTESTSUM_VERSION: "main"
+  # Set this to "next" for upcoming version (defined by `RegistryImageNext`)
+  DISTRIBUTION_VERSION: "next"
+  # FIXME. Currently pinned, see: https://github.com/containerd/nerdctl/pull/3153
+  RUNC_VERSION: "v1.1.13"
+
+  # Released versions - one can update these manually to override the Dockerfile pinned versions
+  BUILDKIT_VERSION: "v0.15.0-rc1"
```
AkihiroSuda, line 30: v0.15 is now GA

apostasie:
CI is still running, but nevertheless, this is ready for the most part.

What this is **now** doing:
- for all dependencies listed at the top of `./hack/build-integration-canary.sh`, retrieves the latest published version (alpha, beta, RC), and use that for integration test
- do the same thing for GO_VERSION

This is automatic - we do not have to manually maintain this file.
It is a bit slow (1 minute), and the script is a bit of a mouthful, but it does take care of everything (inc. the SHASUM files in case they are not here - which is fine, in that case).

I initially considered running (some) deps against their main / master, but after some tests, it seems like more trouble than reward.

Pending CI being green (or further comments here), PTAL at your convenience.

.github/workflows/test-canary.yml
apostasie: See last comment. New version is now automatically retrieving the latest published release (including alpha/beta/RC).

hack/build-integration-canary.sh
```diff
@@ -0,0 +1,438 @@
...
+# Github API helpers
+# Set NERDCTL_GITHUB_API_TOKEN to use authenticated requests to workaround limitations
+github::request(){
+  local endpoint="$1"
+  local args=(
+    -H "Accept: application/vnd.github+json"
+    -H "X-GitHub-Api-Version: 2022-11-28"
+  )
+
+  [ "${NERDCTL_GITHUB_API_TOKEN:-}" == "" ] || args+=(-H "Authorization: Bearer $NERDCTL_GITHUB_API_TOKEN")
```
AkihiroSuda, line 187: `NERDCTL_` prefix might not be needed?

apostasie: > `NERDCTL_` prefix might not be needed?

Yeah I was thinking about that. Might be more convenient for peeps. On the other hand, for these who like to segregate tokens, it is convenient to be able to use a different name.

Will look into this again tomorrow on a laptop

apostasie: Ok, renamed to just GITHUB_TOKEN and hooked-in the `secret` auto generated thing (hope that works as I think it does...).

apostasie:
@AkihiroSuda I think we are getting close.

Comment has been addressed, and now Windows also autodetect latest containerd.

I am now using the GITHUB_TOKEN magically injected secret.

PTAL at your convenience.

Sorry for the churn here. This one obviously needed a little bit of back and forth - and your suggestion to automate retrieval of latest Go prompted an interesting direction for the rest.

Thanks!

AkihiroSuda (approved):
Thanks

PR #3213, opened 2024-07-12
Makefile fixes and other spring cleanup

apostasie:
First commit removes DOCKER_BUILDKIT references, which have been the default for a while now.

Second commit focuses on the Makefile:
- most importantly, make it possible to run make targets outside of source tree - clarifying and fixing use of CURDIR vs. newly introduced MAKEFILE_DIR
- introducing `lint` tasks (golint & yamllint) - and also introducing the use of shellcheck
- remove references to GO111MODULE which are no longer useful in 2024
- some minor maintenance (variable-ize calls to `docker`, etc)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@
 # Licensed under the Apache License, Version 2.0
 # -----------------------------------------------------------------------------
 
+DOCKER ?= docker
 GO ?= go
 GOOS ?= $(shell $(GO) env GOOS)
 ifeq ($(GOOS),windows)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -52,63 +57,76 @@ help:
...
 	@echo " * 'binaries' - Build nerdctl."
 	@echo " * 'clean' - Clean artifacts."
+	@echo " * 'lint' - Run various linters."
 
 nerdctl:
-	$(GO_BUILD) $(GO_BUILD_FLAGS) $(VERBOSE_FLAG) -o $(CURDIR)/_output/nerdctl$(BIN_EXT) $(PACKAGE)/cmd/nerdctl
+	$(GO_BUILD) $(GO_BUILD_FLAGS) $(VERBOSE_FLAG) -o $(CURDIR)/_output/nerdctl$(BIN_EXT) ./cmd/nerdctl
 
 clean:
 	find . -name \*~ -delete
```
apostasie, line 66: I am not sure what or why these two are for and if we do need these.
Is that to cleanup leftover files from vi?

Let me know if this should be moved to `find $(MAKEFILE_DIR)`.

cmd/nerdctl/builder_build_test.go
```diff
@@ -45,7 +45,6 @@ CMD ["echo", "nerdctl-build-test-string"]
 	base.Cmd("build", buildCtx, "-t", imageName).AssertOK()
 	base.Cmd("run", "--rm", imageName).AssertOutExactly("nerdctl-build-test-string\n")
 
-	// DOCKER_BUILDKIT (v20.10): `Error response from daemon: exporter "docker" could not be found`
 	if base.Target == testutil.Nerdctl {
```
AkihiroSuda, line 48: The comment above should be kept, unless we can remove this base.Target check too

apostasie: Fair point.
Let me see about that.

apostasie: Base target check removed on latest push (works locally with docker 27 - pending CI).

apostasie:
@AkihiroSuda comment addressed (base check removed) - CI is green

AkihiroSuda (approved):
Thanks

PR #3214, opened 2024-07-12
Shellscript sanity / cleanup PR

apostasie:
This focuses on our shell scripts.

Fixes one major issue:
- `--detach-ns` is not set in some cases when it should, because of mis-assignment of `CONTAINERD_ROOTLESS_ROOTLESSKIT_FLAGS`

And a bunch of more minor ones (mainly quoting variables to prevent expansion).

Some simplification (removing useless loop, useless pipe / cat).

And some sanity - remove `echo -e` and replace with `printf`.

It is likely that none of the minor stuff would change behavior "in normal circumstances" (although it could definitely help with issues from unexpected values), but the `--detach-ns` will likely have an impact and we should double check what's going on with this (I am not super familiar with that part of the codebase yet).

Thanks.

### DIFF
--- a/extras/rootless/containerd-rootless.sh
+++ b/extras/rootless/containerd-rootless.sh
@@ -60,14 +60,14 @@ fi
 : "${XDG_DATA_HOME:=$HOME/.local/share}"
 : "${XDG_CONFIG_HOME:=$HOME/.config}"
 
-if [ -z $_CONTAINERD_ROOTLESS_CHILD ]; then
+if [ -z "$_CONTAINERD_ROOTLESS_CHILD" ]; then
 	if [ "$(id -u)" = "0" ]; then
 		echo "Must not run as root"
 		exit 1
 	fi
 	case "$1" in
 	"check" | "install" | "uninstall")
-		echo "Did you mean 'containerd-rootless-setuptool.sh $@' ?"
+		echo "Did you mean 'containerd-rootless-setuptool.sh $*' ?"
 		exit 1
 		;;
 	esac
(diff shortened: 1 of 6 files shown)


apostasie:
Failure is unrelated (transient docker hub 502).

This is ready @AkihiroSuda - pending your assessment on the impact of actually passing `--detach-ns` which was not set before.

extras/rootless/containerd-rootless.sh
```diff
@@ -102,12 +102,11 @@ if [ -z $_CONTAINERD_ROOTLESS_CHILD ]; then
...
 		fi
 	fi
-	if [ -z $mtu ]; then
+	if [ -z "$mtu" ]; then
 		mtu=1500
 	fi
 
-	_CONTAINERD_ROOTLESS_CHILD=1
-	export _CONTAINERD_ROOTLESS_CHILD
+	export _CONTAINERD_ROOTLESS_CHILD=1
```
AkihiroSuda, line 109: This bashism may not work with some /bin/sh implementation

apostasie: Yep... you are right... 

Not quite sure why we are exporting this one btw.

Anyhow, will reverse this line.

apostasie:
Comment about `export` addressed.

AkihiroSuda:
Now we should have shellcheck in the CI?

apostasie:
> Now we should have shellcheck in the CI?

Yes.

There is an ongoing PR here: #3138

PR #3215, opened 2024-07-12
Small curl cleanup

apostasie:
This is a very small PR, normalizing the way we use curl.

Specifically it enforces `-fsSL` and proto https, and force tls to v1.2.
- `fsSL` is obviously just good practice for scripts.
- the tls/https bits serve as a minor security posture that would prevent downgrade attacks (either from https to http or to a less recent tls version)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -146,7 +146,7 @@ COPY --from=build-runc /out/runc.${TARGETARCH:-amd64} /out/bin/runc
 RUN echo "- runc: ${RUNC_VERSION}" >> /out/share/doc/nerdctl-full/README.md
 ARG CNI_PLUGINS_VERSION
 RUN fname="cni-plugins-${TARGETOS:-linux}-${TARGETARCH:-amd64}-${CNI_PLUGINS_VERSION}.tgz" && \
-  curl -o "${fname}" -fSL "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/${fname}" && \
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -146,15 +146,15 @@ COPY --from=build-runc /out/runc.${TARGETARCH:-amd64} /out/bin/runc
 RUN echo "- runc: ${RUNC_VERSION}" >> /out/share/doc/nerdctl-full/README.md
 ARG CNI_PLUGINS_VERSION
 RUN fname="cni-plugins-${TARGETOS:-linux}-${TARGETARCH:-amd64}-${CNI_PLUGINS_VERSION}.tgz" && \
-  curl -o "${fname}" -fSL "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/${fname}" && \
+  curl -o "${fname}" -fsSL --proto '=https' --tlsv1.2 "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/${fname}" && \
```
AkihiroSuda, line 149: `-s` (`--silent`) might not be necessary

apostasie: Without `silent`, curl will output progress:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     9    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
```

This is a lot of useless noise - for example:
https://github.com/containerd/nerdctl/actions/runs/9913970558/job/27392100283?pr=3216#step:3:1581

I think we should have it.

What do you think?

AkihiroSuda (approved):
Thanks

PR #3216, opened 2024-07-12
Update go.mod

apostasie:
As we are going for v2, it might be a good time to shake things up and update all our go.mod dependencies.

Tentative PR here to see if anything breaks.


go.mod
```diff
@@ -37,6 +38,7 @@ require (
 	github.com/fatih/color v1.17.0
 	github.com/fluent/fluent-logger-golang v1.9.0
 	github.com/ipfs/go-cid v0.4.1
+	github.com/klauspost/compress v1.17.9
```
fahedouch, line 41: from indirect to direct dep .

go.mod
```diff
@@ -1,14 +1,14 @@
 module github.com/containerd/nerdctl/v2
 
-go 1.21
+go 1.21.0
```
fahedouch, line 3: I prefer enforcing the last patch version for 1.21 as minimum supported version

apostasie: Sure. Will do that with next push.

apostasie:
@fahedouch comment addressed and pushed go to 1.21.12.

fahedouch:
> @fahedouch comment addressed and pushed go to 1.21.12.

Why not just keeping `1.21` ?

apostasie:
> > @fahedouch comment addressed and pushed go to 1.21.12.
> 
> Why not just keeping `1.21` ?

I believe using the full version is recommended since go1.18. Specifically, `go get -u` will enforce it and update your go.mod file on its own (adding .0 if there is no patch component).

containerd/containerd does that too.

go.mod
```diff
@@ -1,14 +1,14 @@
 module github.com/containerd/nerdctl/v2
 
-go 1.21
+go 1.21.12
```
AkihiroSuda, line 3: Can dependabot keep updating this?
Otherwise probably this should be .0?

apostasie: I do not think dependabot will up this.
I also prefer .0 (as I do not think we should be in the business of forcing people to the last patch update), but @fahedouch prefers to use the last patch (https://github.com/containerd/nerdctl/pull/3216#discussion_r1676892285)

@fahedouch what are your thoughts at this point?

fahedouch: from a security perspective I think forcing people to use the last patch update is better. But since it's the supported version I am also ok for `.0`

fahedouch: > Can dependabot keep updating this?

I don't think so

fahedouch (approved):

go.mod
apostasie: Thanks a lot @fahedouch 

I honestly do not quite care either way - especially since 1.21 is soon to be out of support (and containerd v2 forces 1.22.0) - so, maybe we just leave it as-is (1.21.12) and revisit early August for go 1.23 release or our v2 release - whichever comes first.

AkihiroSuda (approved):
Thanks, merging.
The `go 1.21.12` line should be revisited on the EOL of Go 1.21

Issue #1274, opened 2022-07-29
Unable to Restart the Container via the `nerdctl restart` comamnd


harshanarayana:
### Description

As per the document, the `nerdctl restart` should let me restart one or more containers running on the system. However, when I try to do that with the `containerd` based infra, it runs into a fatal error. 

```bash
root@tardis-169-254-6-46:/var/run/docker/containerd/daemon/io.containerd.runtime.v2.task/k8s.io# nerdctl ps | grep metrics-server
3eb77b051f92    myregistry.gallifrey.com/gallifrey-docker/pause:3.6                                                                              "/pause"                  2 days ago        Up                 k8s://kube-system/metrics-server-6c8bbfdb57-gz6sx
ff39a309f6d8    myregistry.gallifrey.com/gallifrey-docker/metrics-server:2.0.3                                                               "/metrics-server --r…"    9 seconds ago     Up                     k8s://kube-system/metrics-server-6c8bbfdb57-gz6sx/metrics-server


root@tardis-169-254-6-46:/var/run/docker/containerd/daemon/io.containerd.runtime.v2.task/k8s.io# nerdctl restart ff39a309f6d8
FATA[0010] failed to start shim: mkdir /var/run/docker/containerd/daemon/io.containerd.runtime.v2.task/k8s.io/ff39a309f6d85bf50e98ca59600739b6bfbac4452d8f7cbca57f92e56d26dff5: file exists: unknown
```

### Steps to reproduce the issue

1. Create Kubernetes cluster with Containerd based CRI
2. Use `nerdctl restart` to restart any non pause container of the pod in the node

Below is the config of the `containerd` Daemon

```toml

version=2
root = "/data/docker/containerd/daemon"
state = "/var/run/docker/containerd/daemon"
oom_score = -500
required_plugins = ["io.containerd.grpc.v1.cri"]

[grpc]
  uid = 0
  gid = 0
  max_recv_message_size = 16777216
  max_send_message_size = 16777216

[debug]
  uid = 0
  gid = 0
  level = "info"

[metrics]
  address = "localhost:1414"
  grpc_histogram = true

[cgroup]
  path = ""

[plugins]
  [plugins."io.containerd.grpc.v1.cri"]
    stream_server_address = "127.0.0.1"
    max_container_log_line_size = 262144
    sandbox_image = "myregistry.gallifrey.com/gallifrey-docker/pause:3.6"
  [plugins."io.containerd.grpc.v1.cri".cni]
[...]

PR #3219, opened 2024-07-15
fix: restart lark loguri cause start failed

Howie59:
if the container was not created by nerdctl, by default logURI args is empty.
https://github.com/containerd/nerdctl/blob/d273fd202abdf8153cc7d8739fc1a2eda1a9efc1/pkg/cmd/container/create.go#L178
when exec nerdctl restart, it will use logURI start log path, then will cause
`FATA[0000] 1 errors:
failed to start binary process with cmdArgs []: exec: no command`

https://github.com/containerd/nerdctl/blob/d273fd202abdf8153cc7d8739fc1a2eda1a9efc1/pkg/cioutil/container_io.go#L179

### DIFF
--- a/pkg/taskutil/taskutil.go
+++ b/pkg/taskutil/taskutil.go
@@ -117,6 +117,8 @@ func NewTask(ctx context.Context, client *containerd.Client, container container
 			return nil, err
 		}
 		ioCreator = cio.LogURI(u)
+	} else if flagD && logURI == "" {
+		ioCreator = cio.NullIO
 	} else {
 		var in io.Reader
 		if flagI {


Howie59:
ref: https://github.com/containerd/nerdctl/issues/1274
i changed this code and encountered another problem!!
hope for response

AkihiroSuda:
Can we have a test?

pkg/taskutil/taskutil.go
```diff
@@ -117,6 +117,8 @@ func NewTask(ctx context.Context, client *containerd.Client, container container
 			return nil, err
 		}
 		ioCreator = cio.LogURI(u)
+	} else if flagD && logURI == "" {
```
fahedouch, line 120: Could you please add comment for this specific case!

AkihiroSuda: Ping @Howie59

apostasie:
Hey @Howie59 
Thanks for this!
Do you think you would be able to add a test for it and comment the code so that we can merge this?

Issue #2844, opened 2024-02-27
nerdctl pull from private registry mirror is missing Authorization header


networkhermit:
### Description

I tried to use nerdctl with [registry mirrors](https://github.com/containerd/nerdctl/issues/668) served by private harbor, but I got `401 Unauthorized` error.

After some investigation, I found that when using registry mirrors, nerdctl doesn't pass the `Authorization: Basic xxxx` header to `/service/token` endpoints, causing the `401 Unauthorized` error.

### Steps to reproduce the issue

1. configure containerd to use registry mirrors:

/etc/containerd/config.toml
```toml
    [plugins."io.containerd.grpc.v1.cri".registry]
      config_path = "/etc/containerd/certs.d"
```

/etc/containerd/certs.d/gcr.io/hosts.toml
```toml
server = "https://gcr.io"

[host."https://harbor.example.com:8443/v2/gcr.io"]
  capabilities = ["pull", "resolve"]
  override_path = true
```

2. use `nerdctl login` to harbor.example.com:8443, optionally setup mitmproxy to view the api traffic

3. directly pull the image from harbor is successfully
```console
❯ sudo HTTPS_PROXY=http://127.0.0.1:8080 nerdctl --debug-full pull harbor.example.com:8443/gcr.io/cadvisor/cadvisor:latest
DEBU[0000] verifying process skipped
DEBU[0000] Found hosts dir "/etc/containerd/certs.d"
DEBU[0000] Ignoring hosts dir "/etc/docker/certs.d"      error="stat /etc/docker/certs.d: no such file or directory"
DEBU[0000] The image will be unpacked for platform {"amd64" "linux" "" [] ""}, snapshotter "overlayfs".
DEBU[0000] fetching                                      image="harbor.example.com:8443/gcr.io/cadvisor/cadvisor:latest"
DEBU[0000] resolving                                     host="harbor.example.com:8443"
DEBU[0000] do request                                    host="harbor.example.com:8443" request.header.accept="application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, */*" request.header.user-agent=containerd/1.7.11+unknown request.method=HEAD url="https://harbor.example.com:8443/v2/gcr.io/cadvisor/cadvisor/manifests/latest"
[...]

Issue #920, opened 2022-03-23
Cannot pull image from private repository after successful login


catalinbahrin:
### Description

I am able to pull an image from a private repository when I do `lima nerdctl pull REPOSITORY\IMAGE`. However, when I try to build with `lima nerdctl compose up --build` a Dockerfile that uses the same image from a private repository, I get an error.

### Steps to reproduce the issue

1. lima nerdctl login
2. lima nerdctl compose up --build

### Describe the results you received and expected

`=> [internal] load build definition from Dockerfile                                                                                                                                                 0.0s
 => => transferring dockerfile: 1.51kB                                                                                                                                                               0.0s
 => [internal] load .dockerignore                                                                                                                                                                    0.1s
 => => transferring context: 166B                                                                                                                                                                    0.1s
 => ERROR [internal] load metadata for docker.io/REPOSITORY/image:tag:                                                                                                              2.8s

 > [internal] load metadata for docker.io/REPOSITORY/image:tag:

Dockerfile:1
   1 | >>> FROM REPOSITORY/image:tag
   2 |     
   3 |     # general environment settings

error: failed to solve: REPOSITORY/image:tag: pull access denied, repository does not exist or may require authorization: authorization status: 401: authorization failed
FATA[0003] unrecognized image format                    
FATA[0003] error while building image some_name: exit status 1 `

### What version of nerdctl are you using?

Client:
 Version:       v0.17.1
 Git commit:    20233c26d26f11ca73a9a775fba87ca884ab14d2

Server:
 containerd:
  Version:      v1.6.1
  GitCommit:    10f428dac7cec44c864e1b830a4623af27a9fc70

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Lima

### Host information

Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v1.6.1
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Log: json-file
  Storage: native overlayfs fuse-overlayfs stargz
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
  rootless
 Kernel Version: 5.13.0-35-generic
 Operating System: Ubuntu 21.10
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 3.827GiB
 Name: lima-default
 ID: 71cbe510-fc1b-4029-ac4a-c456e41560ff

[...]

Issue #299, opened 2021-07-22
Option --insecure-registry does not work with pull & push


Teo90:
Hi,

I'm trying to use nerdctl with a private harbor without TLS, while the global flag --insecure-registry does not really ignore the HTTP responses.

Here is what i didi:
pull
`nerdctl --insecure-registry pull 192.168.88.150/devops/kaniko-project-executor:debug`
![image](https://user-images.githubusercontent.com/23375797/126621062-846893d5-4f46-4b49-a8fa-bf62e9fcb225.png)

push
`nerdctl  -n k8s.io --insecure-registry push 192.168.88.150/k8s/coredns:v1.8.0`
![image](https://user-images.githubusercontent.com/23375797/126621114-9b6453b9-dcc0-4d57-8543-7d1fbe79ebd4.png)

AkihiroSuda:
The `no route to host` error does not seem to be related to nerdctl.
Can you ping 192.168.88.150?

Teo90:
Sure, because crictl succeeded in pulling harbor http protocol 
![image](https://user-images.githubusercontent.com/23375797/127113751-f8ab5f5f-c5c7-446b-80b1-f19b097906dd.png)

Teo90:
nerdctl added port 80 to download successfully

![image](https://user-images.githubusercontent.com/23375797/127116236-b637db3d-11d0-462c-9994-09ccef0f0394.png)

AkihiroSuda:
cc @ktock @fahedouch

fahedouch:
I agree with @AkihiroSuda . It looks like a network issue. Did you tried crictl from same network as nerdctl ?

fahedouch:
@Teo90 your `registry firewall rules` may not authorize port `443`. 
check if your `registry firewall` is blocking  connection on port `443`

beckend:
This is still an issue, it doesn't care about the --insecure-flag at all.

```shell
nerdctl --version
nerdctl version 2.2.0

sudo nerdctl images
REPOSITORY                                                                             TAG                       IMAGE ID        CREATED               PLATFORM       SIZE       BLOB SIZE
192.168.10.6/vtdx/drs-simulator-wrapper                                                latest                    bdfbab865dda    About a minute ago    linux/amd64    1.743GB    583.3MB
```

```shell
sudo nerdctl push --insecure-registry 192.168.10.6/vtdx/drs-simulator-wrapper
[...]

PR #3249, opened 2024-07-25
[WIP] Login rework

### LINKED ISSUE #3046: Even when explicitly passing https, registry requests to localhost are downgraded to http
### Description

`nerdctl login https://127.0.0.1:5000` will always be downgraded to http://127.0.0.1:5000

It seems to be because of the faulty logic in `remotes/docker/config/hosts.go` that apparently considers "the scheme is ambiguous" as the port is non-standard and the host is localhost.

Whether they fixed it for 2.0 is unclear to me at this point.

1. `nerdctl login https://127.0.0.1:5000`

Seeing: request downgraded to http://127.0.0.1:5000

Should see: not downgraded, the scheme is explicit

1.7.6 / main

None

### Host information

_No response_

### LINKED ISSUE #3047: hosts.toml file resolution does not seem to account for default port being ommitted
### Description

`nerdctl login https://foo:443`
vs.
`nerdctl login https://foo`

If the `hosts.toml` path contains the port, the second one will fail.
On the other hand, if it does not, the first one will fail.

Responsibility for walking the host dir is in `dockerconfigresolver.go` (NewHostOptions > HostDir()).
This does not try to be smart wrt what the host contains (port or no port).

We should patch this so:
- if scheme is https, passing :443 or not passing it should work
- ~~all of this should work with hosts.toml files path containing the port or not~~

See description

Fail / not fail.

1.7.6

None

### Host information

_No response_

### LINKED ISSUE #3052: `nerdctl login` should warn about passing along an explicit scheme and should ignore it entirely
### Description

While `pull` and `push` do not allow explicit schemes (as image identifiers do not), both docker and nerdctl `login` DO allow it, albeit docker simply ignores it entirely.

Nerdctl on the other hand (or containerd?), still handles the scheme to decide what should be the port, if it was not specified.

For example:
```
sudo nerdctl login http://192.168.5.15
```

will assume port 80, but of course ALSO assumes https... which is... comical.

```
"failed to call rh.Client.Do: Get \"https://192.168.5.15:80/v2/\": http: server gave HTTP response to HTTPS client
```

While docker just ignores the scheme entirely and assumes 443:

```
Get "https://192.168.5.15/v2/": dial tcp 192.168.5.15:443
```

Overall, the resolution logic (for hosts.toml, docker credentials, and unspecified ports) is complex enough without the scheme (albeit it can be reasoned about).
With the scheme, it becomes untractable, and leaves a lot of unspecified behavior.
Furthermore, it does not make much sense, since neither pull / push allow it anyhow.

Our suggestion is to:
- drop and ignore the scheme entirely on `nerdctl login` before doing anything with it
- warn the user that this syntax is only supported as a convenience and that the scheme is ignored (which docker should do too)

This will impact #3046 - which I believe still matters as there should be a way to talk TLS even with localhost.

See above

See above

1.7.6

None

### Host information

_No response_

### LINKED ISSUE #3068: nerdctl login used against a basic auth server will error in confusing ways after hammering the server
### Description

Passing wrong credentials against a basic auth registry will result in nerdctl trying repeatedly (ten times).

Then spit out:
`too many 401 (probably)`

Followed by a `failed to call rh.Client.Do: Get \"https://10.4.0.1:5000/v2/\": tls: failed to verify certificate: x509: certificate signed by unknown authority" i=1` which does not make sense here.

1. Start a registry with basic auth
2. Login using nerdctl

Received:
- confusing error message
- repeated attempts with the same wrong credentials instead of bailing out

Should have received:
- more clear error message
- try once and stop

1.7.6

None

### Host information

_No response_

### LINKED ISSUE #3245: Login to docker.io failed. (`expected acArg to be "docker.io", got "registry-1.docker.io"`)
### Description

I tried to use nerdctl to login to docker.io, but got:
```
nerdctl login docker.io --debug
Enter Username: lingdie
Enter Password:
DEBU[0005] Found hosts dir "/etc/containerd/certs.d"
DEBU[0005] Ignoring hosts dir "/etc/docker/certs.d"      error="stat /etc/docker/certs.d: no such file or directory"
DEBU[0005] len(regHosts)=1
ERRO[0007] failed to call tryLoginWithRegHost            error="failed to call rh.Authorizer.AddResponses: expected acArg to be \"docker.io\", got \"registry-1.docker.io\"" i=0
FATA[0007] failed to call rh.Authorizer.AddResponses: expected acArg to be "docker.io", got "registry-1.docker.io"
```

however, this works:
```
nerdctl login --debug
DEBU[0000] Found hosts dir "/etc/containerd/certs.d"
DEBU[0000] len(regHosts)=1
DEBU[0001] no scope specified for token auth challenge   host=index.docker.io
WARNING: Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
---
nerdctl login https://index.docker.io/v1 --debug
DEBU[0000] Found hosts dir "/etc/containerd/certs.d"
DEBU[0000] len(regHosts)=1
DEBU[0001] no scope specified for token auth challenge   host=index.docker.io
WARNING: Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
```

whats the differance betwen these commands? Is this a bug?

I need to ues nerdctl sdk to login and push images to registries, what should I do if I don't want to write hard codes for `docker.io`.

1. nertctl login docker.io
2. nertctl login
3. nertctl login https://index.docker.io/v1

`nertctl login docker.io` should works as well.

nerdctl version
Client:
 Version:	v1.5.0
 OS/Arch:	linux/amd64
 Git commit:	b33a58f288bc42351404a016e694190b897cd252
 buildctl:
  Version:	

Server:
 containerd:
  Version:	v1.7.15
  GitCommit:	926c9586fe4a6236699318391cd44976a98e31f1
 runc:
  Version:	1.1.12
  GitCommit:	v1.1.12-0-g51d5e946

None

### Host information

_No response_

apostasie:
WORK IN PROGRESS

Fixed:

- fix #3245 (cc @lingdie ):
  - `docker.io` `index.docker.io`, with or without port `443`, with or without scheme, will all be treated as meant for "docker hub" and use whatever credentials are associated with it
- fix #3046:
  - we now allow passing explicit `--insecure-registry=false` (and treating it differently from "no flag"), which allows the user to override the default behavior for localhost (default behavior being to treat localhost as always insecure)
- fix #3047 
  - existing "port-less" hosts.toml files are now considered and used when the registry being requested uses the default 443 port (or no port at all, which is considered equivalent)
- fix #3052
  - when an `http` scheme is explicitly used in front of the server url, we are now warning the user that the scheme is ignored entirely and that our behavior is governed by the use of the insecure flag (which is the current behavior)
- fix #3068
  - containerd remotes/docker/resolver should be fixed instead, but it was patchable here

Other changes:
- nerdctl now returns a user-agent string with version and OS
- removing `IsErrHTTPResponseToHTTPSClient`, as `http.ErrSchemeMismatch` is usable in place
- new errors

TODO:
- implement result of design discussion at #3265 
- cleanup logic in push / pull to use the new credentials / hosts resolver mechanisms
- add tests for TLS client auth
- add docs
- #2844

To be confirmed:
- #299

Currently NOT addressed:
- #3257
- #920
  - WFM - cannot produce
- #1675


apostasie:
@AkihiroSuda I am closing this as I am breaking it down and distilling it into smaller, separate PRs to ease review. Will keep the branch as long as there are things to merge, but otherwise there is no reason to keep this open.

PR #3258, opened 2024-07-30
Remove gjson dependency

apostasie:
See https://github.com/containerd/nerdctl/pull/3255#issuecomment-2257365418

Am I missing something?


AkihiroSuda (approved):
Thanks, I don't remember why we had to use gjson at that time

PR #3260, opened 2024-07-30
fix(workflow): remove unknown os/arch build packages

longkai:
In the currently oci image relase github action page, there is a unknown/unknown manifest which is confused and useless.

According to this PR https://github.com/docker/build-push-action/issues/820, we could remove it.

image


AkihiroSuda:
Disagree that provenance is useless 
https://slsa.dev/spec/v1.0/provenance

AkihiroSuda:
If you feel the web UI of the release page is confusing, you should negotiate with GitHub, Inc.

Issue #3262, opened 2024-07-30
nerdctl compose down should not forcefully kill the process


midnight-wonderer:
### The problem

`nerdctl compose down` does not respect my `compose.yaml`.

I have this in my service:
~~~YAML
services:
  application:
    stop_signal: SIGINT
    stop_grace_period: 120s
~~~
which works great with `nerdctl compose stop`.

But when I use `nerdctl compose down` or just press `Ctrl + C`, the process is forcefully terminated without a grace period and proper `SIGINT`.

### Why does it matter?

The container in my project has cleanup logic on shutdown when doing it gracefully.  
It doesn't matter if this logic only affects the container, but it writes something into volumes.
That's why the forced stop is problematic.  

For example, Rails delete `tmp/pids/server.pid` on shutdown, or it won't start properly next time.
This `tmp/pids/server.pid` can easily be in a bind-mounted volume in development environments.

`docker compose down` would try to stop the container gracefully before cleaning up but not `nerdctl`.

### Workaround
Right now, I issue `nerdctl compose stop` before `nerdctl compose down`. However, this workaround should not be needed.

### Steps to reproduce the issue

1. `nerdctl compose up -d`
2. `nerdctl compose down`

### Expected
The process in containers receives SIGINT and performs cleanup logic before shutting down.

### Get
`Stopping containers (forcibly)`

### What version of nerdctl are you using?

Client:
 Version:       v1.7.6
 OS/Arch:       linux/amd64
 Git commit:    845e989f69d25b420ae325fedc8e70186243fd93
 buildctl:
  Version:      v0.12.5
  GitCommit:    bac3f2b673f3f9d33e79046008e7a38e856b3dc6

Server:
 containerd:
  Version:      v1.7.10
  GitCommit:    4e1fe7492b9df85914c389d1f15a3ceedbb280ac
 runc:
  Version:      1.1.12
  GitCommit:    51d5e94601ceffbbd85688df1c928ecccbfa4685

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Lima

### Host information

Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v1.7.10
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Log: fluentd journald json-file syslog
  Storage: native overlayfs
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
[...]

PR #3263, opened 2024-07-30
gracefully terminate containers on nerdctl compose down

midnight-wonderer:
Fix for https://github.com/containerd/nerdctl/issues/3262

I am just an amateur copying code from:
https://github.com/containerd/nerdctl/blob/b030e1dc7d95d2147e5dd653de18afb57d67ab5f/pkg/composer/rm.go#L53-L56

I don't actually write any code. I wonder if this is enough.


AkihiroSuda (approved):
Thanks

PR #3266, opened 2024-07-31
Add SOCI handler wrapper

sondavidb:
Relevant: https://github.com/awslabs/soci-snapshotter/pull/1329

As part of upgrading nerdctl to containerd v2, we must update dependencies to depend on v2 as well. As SOCI maintainers do not wish to upgrade to containerd v2 yet, this is a workaround to avoid depending on SOCI library.

We can revert this change if we wish once SOCI updates to containerd v2.

### DIFF
--- a/pkg/snapshotterutil/socisource.go
+++ b/pkg/snapshotterutil/socisource.go
@@ -1,0 +1,110 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+/*
+   Copyright The Soci Snapshotter Authors.
+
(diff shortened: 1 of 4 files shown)


apostasie:
Thanks @sondavidb !

@AkihiroSuda wdyt?

I personally like this solution, and it feels like a good compromise.

pkg/snapshotterutil/socisource.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// targetImageLayersSizeLabel is a label which contains layer sizes contained in
+	// the target image.
+	targetImageLayersSizeLabel = "containerd.io/snapshot/remote/image.layers.size"
+
+	// TargetSociIndexDigestLabel is a label which contains the digest of the soci index.
+	TargetSociIndexDigestLabel = "containerd.io/snapshot/remote/soci.index.digest"
+)
+
+// AppendDefaultLabelsHandlerWrapper makes a handler which appends image's basic
```
apostasie, line 61: Nit: rename the comment to match `SociAppendDefaultLabelsHandlerWrapper`

sondavidb: Thanks, fixed.

sondavidb:
FWIW, we rarely change this function. Checking the [history](https://github.com/awslabs/soci-snapshotter/commits/main/fs/source), the most recent [commit](https://github.com/awslabs/soci-snapshotter/commit/629017fa718b85ae5e11910309ec3b55249ea3ac) that changed this was well over a year ago (2023/6/29). So, if we are worried about drift, it should not be a notable issue.

ktock (approved):

djdongjin (approved):
LGTM, thanks

PR #3270, opened 2024-08-03
enforce authentication realm for dockerhub login

fahedouch:
point dockerhub login addresses to the [authentication legacy default server ](https://github.com/containerd/nerdctl/blob/main/pkg/imgutil/dockerconfigresolver/dockerconfigresolver_util.go#L33)

fixing https://github.com/containerd/nerdctl/issues/3245

### DIFF
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -67,7 +67,7 @@ func Login(ctx context.Context, options types.LoginCommandOptions, stdout io.Wri
 		authConfig = &registry.AuthConfig{ServerAddress: serverAddress}
 	}
 	if err == nil && authConfig.Username != "" && authConfig.Password != "" {
-		//login With StoreCreds
+		// login With StoreCreds
 		responseIdentityToken, err = loginClientSide(ctx, options.GOptions, *authConfig)
 	}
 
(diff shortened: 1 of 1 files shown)


apostasie:
This is a good short term fix.
I think we should merge this regardless.

However, it will save independent credentials for different variants, which it should not (because normalization should happen before):
```
{
	"auths": {
		"docker.io": {
		},
		"https://index.docker.io/v1/": {
		},
		"index.docker.io": {
		}
	}
}
```

fahedouch:
> This is a good short term fix. I think we should merge this regardless.
> 
> However, it will save independent credentials for different variants, which it should not (because normalization should happen before):
> 
> ```
> {
> 	"auths": {
> 		"docker.io": {
> 		},
> 		"https://index.docker.io/v1/": {
> 		},
> 		"index.docker.io": {
> 		},
> 		"index.docker.io:443": {
> 		},
> 		"registry-1.docker.io": {
> 		}
> 	}
> }
> ```

good catch. 

Now authentification endpoints are more straightforward; they are redirected to the authentication realm.

pkg/cmd/login/login.go
```diff
@@ -124,7 +124,7 @@ func GetDefaultAuthConfig(checkCredStore bool, serverAddress string, isDefaultRe
 			return nil, err
 		}
 	}
-	var authconfig = dockercliconfigtypes.AuthConfig{}
+	authconfig := dockercliconfigtypes.AuthConfig{}
```
AkihiroSuda, line 127: Probably these cosmetic changes should be a separate commit

fahedouch: is it the gofumpt

AkihiroSuda (approved):
Thanks

Issue #3197, opened 2024-07-06
cp failure uses confusing error message


apostasie:
### Description

Destination of the copy being read-only:

```
nerdctl cp buildkitd:/usr/bin/buildctl .
/usr/bin/tar: buildctl: Cannot open: Read-only file system
/usr/bin/tar: Exiting with failure status due to previous errors
FATA[0000] could not find container: buildkitd, with error: failed to wait [/usr/bin/tar -x --no-same-owner -f -]: exit status 2
```

Of course the container `buildkitd` is here and fine.

We should clean-up this above and provide a helpful message to the user instead.

Good first issue.

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #3275, opened 2024-08-06
Add error message when container is exist when `nerdctl cp`

yankay:
Show different error message when the container exists to fix nerdctl cp failure uses confusing error message.

fix: https://github.com/containerd/nerdctl/issues/3197

### DIFF
--- a/pkg/cmd/container/cp_linux.go
+++ b/pkg/cmd/container/cp_linux.go
@@ -28,9 +28,11 @@ import (
 
 // Cp copies files/folders between a running container and the local filesystem.
 func Cp(ctx context.Context, client *containerd.Client, options types.ContainerCpOptions) error {
+	foundMatchCount := 0
 	walker := &containerwalker.ContainerWalker{
 		Client: client,
 		OnFound: func(ctx context.Context, found containerwalker.Found) error {
+			foundMatchCount = found.MatchCount
 			if found.MatchCount > 1 {
 				return fmt.Errorf("multiple IDs found with provided prefix: %s", found.Req)
 			}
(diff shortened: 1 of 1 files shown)


apostasie:
Hey @yankay :-)

So, this makes things better, but will not address the issue:

This below (inside a read-only location)
```bash
nerdctl cp foo:/etc/issue .
/usr/bin/tar: issue: Cannot open: Read-only file system
/usr/bin/tar: Exiting with failure status due to previous errors
FATA[0000] could not find /etc/issue in container foo
```

^ says "could not find" but this is not true - it was found, but we failed copying it.

pkg/cmd/container/cp_linux.go
```diff
@@ -48,7 +50,11 @@ func Cp(ctx context.Context, client *containerd.Client, options types.ContainerC
 	count, err := walker.Walk(ctx, options.ContainerReq)
 
 	if count < 1 {
-		err = fmt.Errorf("could not find container: %s, with error: %w", options.ContainerReq, err)
+		if foundMatchCount == 0 {
+			err = fmt.Errorf("could not find container: %s, with error: %w", options.ContainerReq, err)
+		} else {
+			err = fmt.Errorf("could not find %s in container %s", options.SrcPath, options.ContainerReq)
```
fahedouch, line 56: `containerwalker.ContainerWalker` is just searching for containers.

```
// Walk walks containers and calls w.OnFound .
// Req is name, short ID, or long ID.
// Returns the number of the found entries.
```

apostasie:
@yankay #3197 got fixed by #3323.

I think we could close this one here?

yankay:
Thanks @apostasie

Issue #2942, opened 2024-04-16
identifier "<identifier name>" greater than maximum length (76 characters)


Shubhranshu153:
### Description

Some of users of finch are reporting max length of 76 characters, which seems to stem from this validation
https://github.com/containerd/containerd/blob/92900bf7300b4872db2a43e82128fda9cde08a79/pkg/identifiers/validate.go#L58

Seeking some guidance what would be best way to handle it, Should we handle downstream (in finch) or have an implementation to generate UUID in cases length are longer than 76.

### Describe the results you received and expected

Handling of longer length identifiers.

### What version of nerdctl are you using?

1.7.3

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

Lima

apostasie:
Personal opinion: handle downstream (using annotations or labels to do what they want).
Swapping in UUID in certain cases would either cause UX issues or make lookups more complex.

AkihiroSuda:
Any string supported by Docker should be accepted in nerdctl too.

TODO: figure out regexp used by Docker

apostasie:
> Any string supported by Docker should be accepted in nerdctl too.
> 
> TODO: figure out regexp used by Docker

Related: #3115

Shubhranshu153:
if we are aligning with docker then we can probably remove the validate calls as docker doesnt do that particular check is this an acceptable fix? 

Doing any sort of manipulation of the string seem to need a two way conversion and probably a pretty large pr as it needs to take into account most of the commands

AkihiroSuda:
> if we are aligning with docker then we can probably remove the validate calls as docker doesnt do that particular check is this an acceptable fix?

No, docker does some check
```console
$ docker run -it --rm --name "foo bar" alpine
docker: Error response from daemon: Invalid container name (foo bar), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.
See 'docker run --help'.
```

Shubhranshu153:
i see, i was searching for the validate call.  :( 
they have different regexp for different calls, we can make it compatible.
Will be working on an pr for it.

secana:
@Shubhranshu153 any update on this? It currently blocks https://github.com/cross-rs/cross (default solution to compile Rust to different targets) from running with nerdctl, as cross generates some pretty long names. See the issue linked above.

Shubhranshu153:
@secana  picked up this week, will have a pr in the next couple of days.

apostasie:
[...]

PR #3281, opened 2024-08-08
Change VolumeStore to use digests instead of names on the filesystem

apostasie:
The motivation for this comes from #2942 - with some discussion in #3279

Restricting names to what the filesystem can do is very limiting.
Furthermore, user-controlled strings being used directly as paths components on the filesystem gives me the creeps.

This proposal here moves away from storing volumes inside `volume_root/volume_name` to `volume_root/digest(volume_name)`.

It is fully backward compatible: if a volume exists with the legacy layout, it will still be usable.

It is NOT entirely forward compatible though: although existing volumes will keep working even with older versions even after having been used by newer version, this is not true for entirely new volumes:
- for the older version, newly created volumes will have their name appear as a digest
- legacy volumes that have been touched by the new version will have a doppelganger which will appear with a name that is the digest

What do you folks think?

cc @AkihiroSuda and @Shubhranshu153 

FWIW docker IS using names as path components, and their regexp is: `[a-zA-Z0-9][a-zA-Z0-9_.-]`
They do enforce a limitation at max 254 characters as far as I can tell.

### DIFF
--- a/pkg/mountutil/volumestore/volumestore.go
+++ b/pkg/mountutil/volumestore/volumestore.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/containerd/containerd/identifiers"
 	"github.com/containerd/errdefs"
@@ -29,12 +31,28 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/native"
 	"github.com/containerd/nerdctl/v2/pkg/lockutil"
 	"github.com/containerd/nerdctl/v2/pkg/strutil"
+	"github.com/opencontainers/go-digest"
 )
 
 const (
 	dataDirName        = "_data"
 	volumeJSONFileName = "volume.json"
 )
+
+var (
+	validateRegExp = regexp.MustCompile("^[a-zA-Z0-9_+:/-]{1,256}$")
+)
+
+func validateName(name string) error {
+	if !validateRegExp.MatchString(name) {
+		return fmt.Errorf("identifier %q must match %v: %w", name, validateRegExp, errdefs.ErrInvalidArgument)
+	}
+	return nil
+}
+
+func validateLegacyName(name string) error {
+	return identifiers.Validate(name)
+}
 
 // VolumeStore allows manipulating containers' volumes
 // Every method is protected by a file lock, and is safe to use concurrently.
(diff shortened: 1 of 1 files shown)


pkg/mountutil/volumestore/volumestore.go
```diff
@@ -22,20 +22,38 @@ import (
...
 const (
 	dataDirName        = "_data"
 	volumeJSONFileName = "volume.json"
 )
 
+var (
+	validateRegExp = regexp.MustCompile("^[a-zA-Z0-9_+:/-]{1,256}$")
+)
+
+func validateName(name string) error {
```
apostasie, line 46: Purely provisional ^ - to be replaced by whatever we deem right.
I do not quite care what this does exactly, as this PR technically should free us from any limitations on names.

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -105,19 +123,63 @@ func (vs *volumeStore) Unlock() error {
 	return nil
 }
 
+func (vs *volumeStore) legacyEnsureLink(name string) error {
```
apostasie, line 126: If we have a legacy volume, calling this method will add the name inside `volFilePath` and create a symlink from `legacyVolPath` (with name) to `volPath` (with digest).

pkg/mountutil/volumestore/volumestore.go
```diff
@@ -176,40 +241,62 @@ func (vs *volumeStore) Create(name string, labels []string) (*native.Volume, err
 
 // Get retrieves a native volume from the store
 // Besides a possible locking error, it might return ErrInvalidArgument, ErrNotFound, or a filesystem error
-func (vs *volumeStore) Get(name string, size bool) (*native.Volume, error) {
-	if err := identifiers.Validate(name); err != nil {
-		return nil, fmt.Errorf("malformed volume name %q: %w", name, err)
+func (vs *volumeStore) Get(nameOrDigest string, size bool) (*native.Volume, error) {
```
apostasie, line 244: Given how `List` is working, `Get` will now be able to retrieve from either name or digest.

Shubhranshu153:
From what i tested it doesnt seem docker is controlling it, rather the mkdir is failing for volume create
```
docker volume create --name ksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkk
Error response from daemon: create ksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkk: error while creating volume root path '/local/docker/volumes/ ksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkk': mkdir /local/docker/volumes/ksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkkksafnlksnaslknglksanglknglkkkk: file name too long
```

for other identifier names like containers i am able to give a pretty long name. (> 255 chars)


having a sha256 as identifier i feel needs to be a 2 way hash rather compression otherwise it would be difficult to list it out. Not sure if we need to complicate it.

apostasie:
> for other identifier names like containers i am able to give a pretty long name. (> 255 chars)

Our NameStore implementation also uses the filesystem, so, same thing here: we need to rewrite it if we want to open it up.

> having a sha256 as identifier i feel needs to be a 2 way hash rather compression otherwise it would be difficult to list it out. Not sure if we need to complicate it.

`VolumeStore.Get` was already reading a data file (for labels) - just adding the name was trivial.
As for `List`, it does call `Get`, so.
The extra complexity here is having to deal with the legacy format transparently.

I do not see how we could lift our length limit with a two way hash (which I assume would have to be at least as long as the name it is hashing...).

AkihiroSuda:
Thanks, but I feel this is overengineering and hard to test

AkihiroSuda:
Something bad may happen when a legacy volume exists with hexadecimal name?

apostasie:
> Something bad may happen when a legacy volume exists with hexadecimal name?

It would fail `digest.Parse()`, and be treated as "legacy" name.
Since we do right now limit at 76 chars, 

The only case where we could possibly run into trouble is if someone has been using as a volume name a valid digest prefixed by sha256 (eg: `nerdctl volume create sha256_b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c`).

If this is a concern, we could certainly adapt the new layout to be over 76 characters long. Eg:
```
nerdctl volume create my_volume_name
# ===> on the filesystem: volume_sha256_93ff9de5cd31a50432dd2e39012e53ae78dd9a7ccb520c04f93828873813fef8
```

Since legacy identifiers cannot be over 76 chars, that would do it.

apostasie:
> Thanks, but I feel this is overengineering and hard to test

Ok, that's fine.

Issue #3284, opened 2024-08-09
Is file-locking on windows currently broken?


apostasie:
### Description

I have been scratching my head over suspicious CI failures that only affect Windows and that definitely involves volumes (hence file locking).

Closely reading source code, it looks like the behavior on unix is to lock exclusively:

https://github.com/containerd/nerdctl/blob/main/pkg/lockutil/lockutil_unix.go#L62

While on windows:

https://github.com/containerd/nerdctl/blob/main/pkg/lockutil/lockutil_windows.go#L54

We do not seem to request an exclusive lock (which would be LOCKFILE_EXCLUSIVE_LOCK = 2).

I will submit a PR now as this seems wrong - but I would like a confirmation from someone, as windows is not really my jam.

@AkihiroSuda opinion?

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

na

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
I am also confused at the values we are setting for nNumberOfBytesToLockLow and nNumberOfBytesToLockHigh.

Any idea why?

apostasie:
Fixed by #3285

PR #3285, opened 2024-08-09
Make windows filelocking exclusive

### LINKED ISSUE #3228: TestVolumeLsFilter is failing a lot on Windows
### Description

Title says all

na

na

main

None

### Host information

_No response_

apostasie:
See #3284 

Very likely fix #3228 and possibly a range of other issues...

### DIFF
--- a/pkg/lockutil/lockutil_windows.go
+++ b/pkg/lockutil/lockutil_windows.go
@@ -31,13 +31,12 @@ func WithDirLock(dir string, fn func() error) error {
 	}
 	defer dirFile.Close()
 	// see https://msdn.microsoft.com/en-us/library/windows/desktop/aa365203(v=vs.85).aspx
-	// 1 lock immediately
-	if err = windows.LockFileEx(windows.Handle(dirFile.Fd()), 1, 0, 1, 0, &windows.Overlapped{}); err != nil {
+	if err = windows.LockFileEx(windows.Handle(dirFile.Fd()), windows.LOCKFILE_EXCLUSIVE_LOCK, 0, ^uint32(0), ^uint32(0), new(windows.Overlapped)); err != nil {
 		return fmt.Errorf("failed to lock %q: %w", dir, err)
 	}
 
 	defer func() {
-		if err := windows.UnlockFileEx(windows.Handle(dirFile.Fd()), 0, 1, 0, &windows.Overlapped{}); err != nil {
+		if err := windows.UnlockFileEx(windows.Handle(dirFile.Fd()), 0, ^uint32(0), ^uint32(0), new(windows.Overlapped)); err != nil {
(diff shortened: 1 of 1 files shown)


apostasie:
Failure is unrelated.

Windows experts review highly welcome here, though the win-CI is green.

pkg/lockutil/lockutil_windows.go
```diff
@@ -31,13 +31,12 @@ func WithDirLock(dir string, fn func() error) error {
 	}
 	defer dirFile.Close()
 	// see https://msdn.microsoft.com/en-us/library/windows/desktop/aa365203(v=vs.85).aspx
-	// 1 lock immediately
```
apostasie, line 34: This was never "1 = lock immediately". It is "1 = The function returns immediately if it is unable to acquire the requested lock. Otherwise, it waits.". We obviously want "wait".

pkg/lockutil/lockutil_windows.go
```diff
@@ -31,13 +31,12 @@ func WithDirLock(dir string, fn func() error) error {
...
-	if err = windows.LockFileEx(windows.Handle(dirFile.Fd()), 1, 0, 1, 0, &windows.Overlapped{}); err != nil {
+	if err = windows.LockFileEx(windows.Handle(dirFile.Fd()), windows.LOCKFILE_EXCLUSIVE_LOCK, 0, ^uint32(0), ^uint32(0), new(windows.Overlapped)); err != nil {
```
jsturtevant, line 34: The docs say:

```
If an exclusive lock is requested for a range of a file that already has a shared or exclusive lock, the function
 returns the error ERROR_IO_PENDING. The system will signal the event specified in the [OVERLAPPED]
(https://learn.microsoft.com/en-us/windows/desktop/api/minwinbase/ns-minwinbase-overlapped) structure
 after the lock is granted.
```

So it might be possible that this fails if something already has the lock and the file was opened with asynchronous i/o.  It doesn't appear to be the case here so we should be good with this approach, I mention it just in case something changes in the future.

jsturtevant (approved):
LGTM

apostasie:
> LGTM

Thanks a lot for the extra info and the review @jsturtevant .

Honestly, I am very close to biting the bullet and writing tests for this stuff.
It is such a miserable experience to debug downstream problems caused by issues in this, as symptoms are by nature misleading / flaky / racy...

I wish we had something usable off the bat instead of having to maintain that ( https://github.com/golang/proposal/blob/master/design/33974-add-public-lockedfile-pkg.md ).

apostasie:
@AkihiroSuda do you think we can get that in?

AkihiroSuda (approved):
Yes, thanks

PR #3247, opened 2024-07-24
feat: adding filter infrastructure + status and event filter

CodeChanning:
Implementing `--filter` flag for `nerdctl events`, Addresses part of #54

This change adds infrastructure for adding filters. Similar to Podman's implementation of `--filter`.
 
I also added Status and Event filter and unit tests for those. 

**Context:**
- Listed as unimplemented in [nerdctl command reference](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-events)
- [Podman's implementation](https://github.com/containers/podman/blob/189d862d54b3824c74bf7474ddfed6de69ec5a09/libpod/events/filters.go#L11)

PR #3286, opened 2024-08-09
[release/1.7] Cherry Pick Commits required to support dev containers using nerdctl and finch

Shubhranshu153:
Based on the discussions on https://github.com/containerd/nerdctl/pull/3247
Backporting the contributions made to be able to support dev containers.

### DIFF
--- a/Dockerfile.d/SHA256SUMS.d/cni-plugins-v1.4.1
+++ b/Dockerfile.d/SHA256SUMS.d/cni-plugins-v1.4.1
@@ -1,2 +1,2 @@
-1511f6c003ace805eafeb1132727791326283cff88a923d76329e1892bba7a10  cni-plugins-linux-amd64-v1.4.1.tgz
-72644e13557cda8a5b39baf97fc5e93d23fdf7baba7700000e7e9efd8bdf9234  cni-plugins-linux-arm64-v1.4.1.tgz
+2a0ea7072d1806b8526489bcd3b4847a06ab010ee32ba3c3d4e5a3235d3eb138  cni-plugins-linux-amd64-v1.4.1.tgz
+56fe62d73942cffd8f119d2b8ecb6a062e85f529a3dbfc7aa5cd83c2c01929a7  cni-plugins-linux-arm64-v1.4.1.tgz
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,8 +82,6 @@ jobs:
             containerd: v1.7.16
           - ubuntu: 22.04
             containerd: v1.7.16
-          - ubuntu: 22.04
-            containerd: main
     env:
       UBUNTU_VERSION: "${{ matrix.ubuntu }}"
       CONTAINERD_VERSION: "${{ matrix.containerd }}"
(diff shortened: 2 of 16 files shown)


AkihiroSuda:
> CI update: remove containerd main test

Why?

Shubhranshu153:
That commit needs a bit of rework and want to add test for containerd 1.7.20 . 
From what i understand
containerd: main is analogous to   containerd: v2.0.0-rc.X and we are having the main which would be supported by nerdctl main : 2.x releases.
As this is for a 1.7 branch release it would support containerd 1.7.x releases which is not in main.

if this understanding is wrong would revert it.

AkihiroSuda:
> As this is for a 1.7 branch release it would support containerd 1.7.x releases which is not in main.

nerdctl v1.7 should continue to work with containerd v2.
If something is not working and its not mentioned in containerd v2 release note, it is likely to be a bug. Please open an issue in that case.

Dockerfile.d/SHA256SUMS.d/cni-plugins-v1.4.1
```diff
@@ -1,2 +1,2 @@
-1511f6c003ace805eafeb1132727791326283cff88a923d76329e1892bba7a10  cni-plugins-linux-amd64-v1.4.1.tgz
-72644e13557cda8a5b39baf97fc5e93d23fdf7baba7700000e7e9efd8bdf9234  cni-plugins-linux-arm64-v1.4.1.tgz
+2a0ea7072d1806b8526489bcd3b4847a06ab010ee32ba3c3d4e5a3235d3eb138  cni-plugins-linux-amd64-v1.4.1.tgz
+56fe62d73942cffd8f119d2b8ecb6a062e85f529a3dbfc7aa5cd83c2c01929a7  cni-plugins-linux-arm64-v1.4.1.tgz
```
AkihiroSuda, line 2: Probably CNI plugins (and also other components) should be just updated to the latest version

Shubhranshu153: Will create a separate profile to update this branch and then do the cherry pick. Thanks for the comments

Shubhranshu153:
Yeah will update the test and probably do some back porting as I see even for nerdctl to 1.7.20 there are some changes in the mainline.

AkihiroSuda:
Do you also want to cherry-pick:
- https://github.com/containerd/nerdctl/pull/3318

Dockerfile.d/SHA256SUMS.d/cni-plugins-v1.4.1
AkihiroSuda: What do you mean by "profile"?

Shubhranshu153: Sorry that was auto from the keyboard, I meant pr

Shubhranshu153:
Yes I do want to cherry pick that for the network response. For the identifiers PR I am not sure if i should add to it,  technically the issue I get is there are few of cases where the volume name is long and that leads to failure. But the identifiers PR it has a backward incompatibility for the min length (1). Without that Dev containers would work but with errors in certain cases.

AkihiroSuda:
The identifier one seems risky for backporting.

Shubhranshu153:
I do agree that one is risky. Will keep that out of the backport

AkihiroSuda:
You may also want to update runc
https://github.com/opencontainers/runc/releases/tag/v1.1.14

PR #3292, opened 2024-08-11
Minor cleanup

apostasie:
A handful of maintenance changes

- use `http.ErrSchemeMismatch` instead of our ad-hoc plain text parsing method now that go exposes it
- move away from abandoned `mapstructure` dependency (https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc) and replace it by blessed fork
- get rid of naked returns (naked-rets are true evil) and enable linter
- remove `contains` method and use go `slices.Contains` (go 1.21)
- move `projectloader.Load` to `testutil`


pkg/errutil/errors_check.go
```diff
@@ -16,16 +16,9 @@
...
-// "http: server gave HTTP response to HTTPS client"
-func IsErrHTTPResponseToHTTPSClient(err error) bool {
-	// The error string is unexposed as of Go 1.16, so we can't use `errors.Is`.
-	// https://github.com/golang/go/issues/44855
-	const unexposed = "server gave HTTP response to HTTPS client"
-	return strings.Contains(err.Error(), unexposed)
-}
+import (
+	"strings"
+)
```
AkihiroSuda, line 21: Parenthesis are not needed

.golangci.yml
```diff
@@ -10,7 +10,7 @@ linters:
     - govet
     - ineffassign
     - misspell
-    # - nakedret
+    - nakedret
```
AkihiroSuda, line 13: What is the benefit?

apostasie: It is a matter of choice, and deciding if we want to allow naked returns or not.

I (clearly) think we should not.

Naked returns IMHO are a very contentious language feature (https://github.com/golang/go/issues/21291 ).

They make code harder to read and understand, more confusing to newcomers (have to look back to the function signature to figure out what is being returned), and they are likely error prone, for example when err gets shadowed (named returns are already problematic in that regard when used with defer).

We only had two in the codebase.
Adding the linter will ensure we won't get any in the future.

WDYT?

pkg/errutil/errors_check.go
apostasie: Thanks. Fixed.

AkihiroSuda (approved):
Thanks

PR #3268, opened 2024-08-01
fix: add unpack logic after container commit to images.

lingdie:
May fix this: https://github.com/containerd/nerdctl/issues/827

Issue #827, opened 2022-02-19
Cannot commit and push container image from Kubernetes pods


nettoclaudio:
Hello folks! :wave:

I'm experiencing a strange behavior while committing a container (from Kubernetes pod) and pushing it to some container registry after. I always get the same error from the push command: `content digest sha256:898c46f3b1a1f39827ed135f020c32e2038c87ae0690a8fe73d94e5df9e6a2d6: not found`.

I'm filing this issue here but there's a chance the problem is on containerd side because I've already seen it working in an earlier version. I'm trying to find out which version it stopped to work.

I'm still investigating but no idea so far... suggestions are welcome. 

### Steps to reproduce:

1. Create an arbitrary Kubernetes pod (for simplicity w/ just one container).
```
$ kubectl run --image tsuru/go:latest my-app -- sleep Inf
```

2. Execute arbitrary commands to generate changes in the container image (to be committed):
```
$ kubectl exec my-app -- mkdir -p /tmp/foo/bar
```

3. Commit the container to a local image:
```bash
$ CONTAINER_ID=$(kubectl get pods my-app -o 'jsonpath={ .status.containerStatuses[0].containerID }' | sed -E 's|(.+)://||g')
$ nerdctl --namespace k8s.io commit ${CONTAINER_ID} registry.example.com/my-app:v1
```
4. Push to local registry (no credentials needed):
```
$ nerdctl --namespace k8s.io push registry.example.com/my-app:v1
FATA[0000] failed to create a tmp single-platform image "registry.example.com/my-app:v1-tmp-reduced-platform": content digest sha256:898c46f3b1a1f39827ed135f020c32e2038c87ae0690a8fe73d94e5df9e6a2d6: not found
```

### Environment information: 

* nerdctl:  `0.17.0`
* containerd: `v1.4.12 7b11cfaabd73bb80907dd23182b9347b4245eb5d`
* OS: `Buildroot 2021.02.4` (from minikube)
* Kernel: `4.19.202`
* Kubernetes version: `v1.20.6`
* kubectl version: `v1.20.6`
* Minikube: `v1.25.1`

Junnplus:
I can't reproduce this problem on containerd with latest and 1.5.10 version.

```
jun@lima-k8s-test:~$ kubectl version
[...]

PR #3296, opened 2024-08-12
Kube on the CI

### LINKED ISSUE #3282: Add a Kube testing environment?

We recently faced issues involving kubernetes (#827) and corresponding fixes (#3268) that cannot be tested, because we do not have a kube testing environment.

Have a Kube testing environment.

We should be able to easily do that by either sharing or copying the work that has been done on that front in Nydus: https://github.com/containerd/nydus-snapshotter/tree/main/tests/helpers

### Additional context

_No response_

apostasie:
This is meant to bring in kube testing (fix #3282):
- using `kind` to spin-up `kube`
   - script usable locally or on the CI indifferently
   - most of it comes from what I committed recently to nydus
- mimic the ipv6 mechanism: `-test.kube` will run only "kube-compatible" tests - see testutils
- example / placeholder test inside container_commit_test, since this is where we will need the first one

Tagging @lingdie because of recent work on kube issues with #827 and #3268 and facing that question of how to test on kube in here.

@fahedouch @AkihiroSuda and other folks - thoughts/feeback?

### DIFF
--- a/.github/workflows/test-kube.yml
+++ b/.github/workflows/test-kube.yml
@@ -1,0 +1,27 @@
+# This pipeline purpose is solely meant to run a subset of our test suites against a kube cluster
+name: kube
+
+on:
+  push:
+    branches:
+      - main
+      - 'release/**'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+
+env:
+  ROOTFUL: true
+
+jobs:
+  linux:
+    runs-on: "ubuntu-24.04"
+    timeout-minutes: 40
+    steps:
+      - uses: actions/checkout@v4.1.7
+        with:
+          fetch-depth: 1
+      - name: "Run Kube integration tests"
+        run: |
+          ./hack/build-integration-kube.sh
+          sudo ./_output/nerdctl exec nerdctl-test-control-plane bash -c -- 'export TMPDIR="$HOME"/tmp; mkdir -p "$TMPDIR"; cd /nerdctl-source; /usr/local/go/bin/go test ./cmd/nerdctl/ -test.only-kube'
--- a/cmd/nerdctl/container_commit_linux_test.go
+++ b/cmd/nerdctl/container_commit_linux_test.go
@@ -1,0 +1,70 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/icmd"
+
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+)
+
+func TestKubeCommitPush(t *testing.T) {
+	t.Parallel()
+
+	base := testutil.NewBaseForKube(t)
+	tID := testutil.Identifier(t)
+
+	var containerID string
+
+	setup := func() {
+		testutil.KubectlHelper(base, "run", "--image", testutil.CommonImage, tID, "--", "sleep", "Inf").
+			AssertOK()
+
+		testutil.KubectlHelper(base, "wait", "pod", tID, "--for=condition=ready", "--timeout=1m").
+			AssertOK()
+
+		testutil.KubectlHelper(base, "exec", tID, "--", "mkdir", "-p", "/tmp/whatever").
+			AssertOK()
+
+		cmd := testutil.KubectlHelper(base, "get", "pods", tID, "-o", "jsonpath={ .status.containerStatuses[0].containerID }")
+		cmd.Run()
+		containerID = strings.TrimPrefix(cmd.Out(), "containerd://")
+	}
+
+	tearDown := func() {
+		testutil.KubectlHelper(base, "delete", "pod", tID).Run()
+	}
+
+	tearDown()
+	t.Cleanup(tearDown)
+	setup()
+
+	t.Run("test commit / push on Kube (https://github.com/containerd/nerdctl/issues/827)", func(t *testing.T) {
+		t.Log("This test is meant to verify that we can commit / push an image from a pod." +
+			"Currently, this is broken, hence the test assumes it will fail. Once the problem is fixed, we should just" +
+			"change the expectation to 'success'.")
+
+		base.Cmd("commit", containerID, "registry.example.com/my-app:v1").AssertOK()
+		base.Cmd("push", "registry.example.com/my-app:v1").Assert(icmd.Expected{
+			ExitCode: 1,
+			Err:      "failed to create a tmp single-platform image",
+		})
+	})
+}
--- a/hack/build-integration-canary.sh
+++ b/hack/build-integration-canary.sh
@@ -16,6 +16,10 @@
 
 # shellcheck disable=SC2034,SC2015
 set -o errexit -o errtrace -o functrace -o nounset -o pipefail
+root="$(cd "$(dirname "${BASH_SOURCE[0]:-$PWD}")" 2>/dev/null 1>&2 && pwd)"
+readonly root
+# shellcheck source=/dev/null
+. "$root/scripts/lib.sh"
 
 ######################
 # Definitions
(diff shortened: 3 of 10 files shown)


.github/workflows/test-kube.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches:
+      - main
+      - 'release/**'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+
+env:
+  UBUNTU_VERSION: "24.04"
```
AkihiroSuda, line 14: Where is this consumed?

hack/build-integration-kube.sh
```diff
@@ -0,0 +1,83 @@
...
+# shellcheck source=/dev/null
+. "$root/scripts/lib.sh"
+
+GO_VERSION=1.22
+
+
+# Note that these are ONLY used to bootstrap the cluster, and are only installed if not already on the system
+KIND_VERSION=v0.23.0
+# Note that 1.7.6 does not work. No particular need to keep this in sync though.
+NERDCTL_VERSION=2.0.0-rc.0
```
AkihiroSuda, line 14: Shouldn't this be compiled from the source?

.github/workflows/test-kube.yml
```diff
@@ -0,0 +1,31 @@
...
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
apostasie: Yep, sorry - was still WIP.

hack/build-integration-kube.sh
apostasie: This is only used by kind to bootstrap, so, it does not matter much.
I'll change it though - we have the binaries, I can just force the PATH for kind.

hack/build-integration-canary.sh
```diff
@@ -67,141 +71,6 @@ STARGZ_SNAPSHOTTER_CHECKSUM=linux
 # We specifically want the static ones
 TINI_CHECKSUM=static
 
-
```
apostasie, line 70: Sharing that with the new kube script (moved to lib.sh)

pkg/testutil/testutil.go
```diff
@@ -51,6 +51,8 @@ type Base struct {
 	DaemonIsKillable bool
 	EnableIPv6       bool
 	IPv6Compatible   bool
+	EnableKube       bool
+	KubeCompatible   bool
```
AkihiroSuda, line 55: This does not seem to be used by tests yet?

apostasie: This is set by NewBaseForKube (which also set the namespace to k8s.io).
Right now, a single test is using it: https://github.com/containerd/nerdctl/pull/3296/files#diff-cb510ac1ecc8b03a6de395f8bfde3d161432dfbbf3d26814eec378cb350b41b7R61 (this test does nothing - it is a placeholder in the location where we will need a test once we fix the infamous commit error).

apostasie:
FreeBSD failure is unrelated. Last push just removes two debugging print that are not consequential.

pkg/testutil/testutil.go
apostasie: Maybe I understood your point: would you like ALL tests to run on kube, in addition to future Kube specific tests?

AkihiroSuda: Sorry, hard to grasp the design until seeing an actual test that uses `NewBaseForKube`.

I guess we don't need to run all tests on kube.

cmd/nerdctl/container_commit_test.go
```diff
@@ -53,3 +55,12 @@ func TestCommit(t *testing.T) {
 		base.Cmd("rmi", testImage).Run()
 	}
 }
+
+func TestKubeCommit(t *testing.T) {
+	t.Parallel()
+	_ = testutil.NewBaseForKube(t)
+
+	t.Run("not an actual test", func(t *testing.T) {
+		assert.Assert(t, true, "this is a non-test")
```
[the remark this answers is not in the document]
apostasie: No. This is provisional, as an example to implement an actual test for #827 when it will get fixed.

If you prefer, I can remove this though.

pkg/testutil/testutil.go
apostasie: Let me add a concrete, actual test involving kubectl.

apostasie:
@AkihiroSuda 

To run kube tests:

```
go test ./cmd/nerdctl/ -test.kube
```

This must be run in the (containerd) context where kube containers have been created (by kubectl for example).
In the case of `kind` (on the CI) , I am running that command inside the kind control plane container.

There is now one test in https://github.com/containerd/nerdctl/pull/3296/files#diff-cb510ac1ecc8b03a6de395f8bfde3d161432dfbbf3d26814eec378cb350b41b7

It is an "anti" test - it assumes failure (because right now our code is broken), but MUST be transformed to assume success as soon as we fix the problem.

The gist of things is:
- create stuff with kubectl
- test that nerdctl can perform operations on these containers

`BaseForKube` puts you in the `k8s.io` namespace, where kube created containers are expected to live.

pkg/testutil/testutil.go
apostasie: Concrete test added (see last comment for more details and lmk what you think). Thanks Akihiro!

cmd/nerdctl/container_commit_test.go
```diff
@@ -53,3 +56,56 @@ func TestCommit(t *testing.T) {
 		base.Cmd("rmi", testImage).Run()
 	}
 }
+
+func kubeCtlHelper(base *testutil.Base, args ...string) *testutil.Cmd {
```
AkihiroSuda, line 60: nit:
```suggestion
func kubectlHelper(base *testutil.Base, args ...string) *testutil.Cmd {
```

This should be also moved to testutil

cmd/nerdctl/container_commit_test.go
```diff
@@ -53,3 +56,56 @@ func TestCommit(t *testing.T) {
...
+
+func TestKubeCommitPush(t *testing.T) {
+	t.Parallel()
+
+	base := testutil.NewBaseForKube(t)
+
+	var containerID string
+
+	setup := func() {
+		kubeCtlHelper(base, "run", "--image", "debian", "my-app", "--", "sleep", "Inf").
```
AkihiroSuda, line 78: Can we use testutil.BusyboxImage ?

AkihiroSuda: "my-app" should be prefixed with "nerdctl-", or probably moved to "nerdctl-test" namespace (`kubectl --namespace=nerdctl-test`)

cmd/nerdctl/container_commit_test.go
```diff
@@ -53,3 +56,56 @@ func TestCommit(t *testing.T) {
...
+			AssertOK()
+
+		kubeCtlHelper(base, "wait", "po", "my-app", "--for=condition=ready", "--timeout=1m").
```
AkihiroSuda, line 81:
```suggestion
		kubeCtlHelper(base, "wait", "pod", "my-app", "--for=condition=ready", "--timeout=1m").
```

.github/workflows/test-kube.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: "ubuntu-24.04"
+    timeout-minutes: 40
+    steps:
+      - uses: actions/checkout@v4.1.7
+        with:
+          fetch-depth: 1
+      - name: "Run Kube integration tests"
+        run: |
+          ./hack/build-integration-kube.sh
+          sudo ./_output/nerdctl exec nerdctl-test-control-plane bash -c -- 'export TMPDIR="$HOME"/tmp; mkdir -p "$TMPDIR"; cd /nerdctl-source; /usr/local/go/bin/go test ./cmd/nerdctl/ -test.kube'
```
AkihiroSuda, line 27: Maybe:
```suggestion
          sudo ./_output/nerdctl exec nerdctl-test-control-plane bash -c -- 'export TMPDIR="$HOME"/tmp; mkdir -p "$TMPDIR"; cd /nerdctl-source; /usr/local/go/bin/go test ./cmd/nerdctl/ -test.kube -test.run TestKube'
```

cmd/nerdctl/container_commit_test.go
apostasie: Moved to testutil, renamed to `KubectlHelper`.

cmd/nerdctl/container_commit_test.go
apostasie: Moved the namespace to `nerdctl-test-k8s` for both Kubectl helper and BaseKube.

apostasie: Also moved to testutil.BusyboxImage

cmd/nerdctl/container_commit_test.go
apostasie: Afaik both syntax work? Anyhow, changed to `pod` (better, more explicit).
Thanks.

.github/workflows/test-kube.yml
apostasie: Not sure I understand why?
`-test.kube` is already going to limit the run to whichever test is explicitly enabled for Kube, so, wouldn't that additional `-test.run TestKube` be moot in that context?

Am I missing something?

apostasie:
One of the adjustments from the comments above does prevent the pod from starting. Debugging now.

cmd/nerdctl/container_commit_test.go
apostasie: Unless I misunderstand something, it does not look like kubectl namespaces are containerd namespaces.

```
kubectl get namespace
NAME                 STATUS   AGE
default              Active   5m22s
kube-node-lease      Active   5m22s
kube-public          Active   5m22s
kube-system          Active   5m22s
local-path-storage   Active   5m18s
nerdctl-test-k8s     Active   3m44s
```

```
nerdctl namespace list'
NAME      CONTAINERS    IMAGES    VOLUMES    LABELS
k8s.io    18            34        0
```

apostasie:
@AkihiroSuda from the comments above - most of them addressed, BUT:
- `kubectl wait` will timeout with a busybox image - no idea why, and not particularly interested in debugging kube - I thus switched the image to `testutil.CommonImage` instead
- kubectl namespaces are not containerd namespaces as far as I can tell, so, for now, we will operate on the standard namespace `k8s.io` (not sure how to change that)
- moved the kube test to a linux only test file, as linking to testutil images would fail the tests on windows

Kube-CI is green.

.github/workflows/test-kube.yml
AkihiroSuda: > -test.kube is already going to limit the run to whichever test is explicitly enabled for Kube

This looks confusing and inconsistent with `-test.kill-daemon`

apostasie: Oh!
I see what you mean now.

I actually just mimicked the mechanism we have for `ipv6`: https://github.com/containerd/nerdctl/blob/main/Dockerfile#L357

https://github.com/containerd/nerdctl/blob/main/pkg/testutil/testutil.go#L548

Should I replace both flags?

- `-test.kube` => `-with-kube`
- `-test.ipv6` => `-with-ipv6`

Or something else?

AkihiroSuda: The `-test.ipv6` flag should be updated to follow the behavior of `-test.kill-daemon`?

If it is not possible the flag name should be like `-test.ipv6-only`?

AkihiroSuda:
> kubectl namespaces are not containerd namespaces as far as I can tell, so, for now, we will operate on the standard namespace k8s.io (not sure how to change that)

The containerd objects should stay in containerd's `k8s.io` namespace, but the Kubernetes objects (Pods, etc.) should be in  Kubernetes API's `nerdctl-test` namespace

.github/workflows/test-kube.yml
apostasie: > The `-test.ipv6` flag should be updated to follow the behavior of `-test.kill-daemon`?

Yeah, I think what we want for both ipv6 and kube is to run these tests *only* - instead of *in addition* (like kill-daemon).


What about this for full clarity on these flags:

```
-test.enable-kill-daemon
-test.only-ipv6
-test.only-kube
```

?

apostasie:
Latest push:
- rebased, conflict resolved
- uses k8s namespace for kubectl created objects
- tentative rename for the flags:
   - test.allow-kill-daemon
   - test.only-ipv6
   - test.only-kube

Let me know if you would like these flag names tweaked to something else.

.github/workflows/test-kube.yml
AkihiroSuda: SGTM

apostasie:
CI is green.

AkihiroSuda (approved):
Thanks

pkg/testutil/testutil.go
```diff
@@ -681,31 +689,46 @@ func NewBaseWithNamespace(t *testing.T, ns string) *Base {
...
+		KubeCompatible:   kubeCompatible,
 		Env:              os.Environ(),
 	}
 	if base.EnableIPv6 && !base.IPv6Compatible {
 		t.Skip("runner skips non-IPv6 compatible tests in the IPv6 environment")
 	} else if !base.EnableIPv6 && base.IPv6Compatible {
 		t.Skip("runner skips IPv6 compatible tests in the non-IPv6 environment")
 	}
+	if base.EnableKube && !base.KubeCompatible {
+		t.Skip("runner skips non-kube compatible tests in the kube environment")
```
fahedouch, line 728: isn't it the opposite? 

`runner skips kube compatible tests in the kube environment` ?

fahedouch:
Sorry for late review! thanks for effort I recommend using `k8s` rather then `kube` to be more explicit.

cmd/nerdctl/container_commit_linux_test.go
```diff
@@ -0,0 +1,70 @@
...
+		cmd := testutil.KubectlHelper(base, "get", "pods", tID, "-o", "jsonpath={ .status.containerStatuses[0].containerID }")
+		cmd.Run()
+		containerID = strings.TrimPrefix(cmd.Out(), "containerd://")
+	}
+
+	tearDown := func() {
+		testutil.KubectlHelper(base, "delete", "pod", tID).Run()
+	}
+
+	tearDown()
```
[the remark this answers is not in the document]
fahedouch: we can keep either a pre-start clean or `t.Cleanup()`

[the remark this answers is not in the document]
apostasie: We do need both `tearDown()` AND `Cleanup(tearDown)`.

There are some scenarios where `Cleanup` will not execute.
When it does not, you end-up with leftovers that will make the next run fail.
This is (of course) particularly true locally when testing repeatedly.

cmd/nerdctl/container_commit_linux_test.go
```diff
@@ -0,0 +1,70 @@
...
+	t.Cleanup(tearDown)
+	setup()
+
+	t.Run("test commit / push on Kube (https://github.com/containerd/nerdctl/issues/827)", func(t *testing.T) {
+		t.Log("This test is meant to verify that we can commit / push an image from a pod." +
```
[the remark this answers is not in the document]
apostasie: Do you mean issue #827 can be reproduced in a non-kube environment?

pkg/testutil/testutil.go
apostasie: Arrrr, double negatives hell :D 

So, I guess what we have here is: `if base has KubeEnabled (kube environment), and if the test is not kube-compatible, we skip it`.

Seems correct?

cmd/nerdctl/container_commit_linux_test.go
```diff
@@ -0,0 +1,70 @@
...
+		testutil.KubectlHelper(base, "exec", tID, "--", "mkdir", "-p", "/tmp/whatever").
+			AssertOK()
+
+		cmd := testutil.KubectlHelper(base, "get", "pods", tID, "-o", "jsonpath={ .status.containerStatuses[0].containerID }")
+		cmd.Run()
+		containerID = strings.TrimPrefix(cmd.Out(), "containerd://")
+	}
+
+	tearDown := func() {
+		testutil.KubectlHelper(base, "delete", "pod", tID).Run()
```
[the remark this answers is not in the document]
apostasie: Can you paste an example here?
Thanks!

apostasie:
> Sorry for late review! thanks for effort I recommend using `k8s` rather then `kube` to be more explicit.

Thanks @fahedouch !

It is not too late - can definitely send a follow-up PR - and yes, I can rename `kube` -> `k8s`.

Just answered your comments above. Let me know what you think when you get a chance.

cmd/nerdctl/container_commit_linux_test.go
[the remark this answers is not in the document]
fahedouch: `kubectl delete pod -f podName`

cmd/nerdctl/container_commit_linux_test.go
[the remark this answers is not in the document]
fahedouch: Yes

ouyangningdong:
Has the issue of the save error prompt been resolved? Thank you very much! content digest sha256:a58ecd4f0c864650a4286c3c2d49c7219a3f2fc8d7a0bf478aa9834acfe14ae7: not found

cmd/nerdctl/container_commit_linux_test.go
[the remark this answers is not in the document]
apostasie: of course - facepalm

apostasie:
@ouyangningdong https://github.com/containerd/nerdctl/issues/827#issuecomment-2300594324

PR #3299, opened 2024-08-12
Ensure binfmt does install emulation for arm7 on the CI

apostasie:
Tonis binfmt `--install all` does NOT install all - but only the stuff that it thinks should be installed.

On most arm64 rigs, it will NOT install emulation for armv7l. On an M1 though, it WILL.

This leads to inconsistent behaviors that break our tests: https://github.com/containerd/nerdctl/pull/3045#pullrequestreview-2220234696

Replacing `--install all` on the CI by explicitly installing the architectures we need.

### DIFF
--- a/.github/workflows/test-canary.yml
+++ b/.github/workflows/test-canary.yml
@@ -32,7 +32,14 @@ jobs:
           sudo losetup -Dv
           sudo losetup -lv
       - name: "Register QEMU (tonistiigi/binfmt)"
-        run: docker run --privileged --rm tonistiigi/binfmt --install all
+        run: |
+          # `--install all` will only install emulation for architectures that cannot be natively executed
+          # Since some arm64 platforms do provide native fallback execution for 32 bits,
+          # armv7 emulation may or may not be installed, causing variance in the result of `uname -m`.
+          # To avoid that, we explicitly list the architectures we do want emulation for.
+          docker run --privileged --rm tonistiigi/binfmt --install linux/amd64
+          docker run --privileged --rm tonistiigi/binfmt --install linux/arm64
+          docker run --privileged --rm tonistiigi/binfmt --install linux/arm/v7
       - name: "Run unit tests"
         run: go test -v ./pkg/...
       - name: "Run integration tests"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,7 +84,14 @@ jobs:
           sudo losetup -Dv
           sudo losetup -lv
       - name: "Register QEMU (tonistiigi/binfmt)"
-        run: docker run --privileged --rm tonistiigi/binfmt --install all
+        run: |
+          # `--install all` will only install emulation for architectures that cannot be natively executed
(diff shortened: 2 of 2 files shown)


apostasie:
This is not going to change anything right now - on amd64 (which is our only CI) - it will only matter once @yankay PR to enable arm64 will land.

AkihiroSuda:
> Toniis binfmt `--install all` does NOT install all - but only the stuff that it thinks should be installed.
> 
> On most arm64 rigs, it will NOT install emulation for armv7l. On an M1 though, it WILL.

Should this be fixed on https://github.com/tonistiigi/binfmt ?

apostasie:
> > Toniis binfmt `--install all` does NOT install all - but only the stuff that it thinks should be installed.
> > On most arm64 rigs, it will NOT install emulation for armv7l. On an M1 though, it WILL.
> 
> Should this be fixed on https://github.com/tonistiigi/binfmt ?

IIRC toniis does not consider this a bug - but I may be wrong - (https://github.com/docker/buildx/issues/198#issuecomment-680302374 ).

The detection mechanism for "all" is rather intentional: if it executes, it does not install the emulator (which it does in this case on the CI because of the execution fallback)

.github/workflows/test-canary.yml
```diff
@@ -32,7 +32,10 @@ jobs:
           sudo losetup -Dv
           sudo losetup -lv
       - name: "Register QEMU (tonistiigi/binfmt)"
-        run: docker run --privileged --rm tonistiigi/binfmt --install all
+        run: |
```
AkihiroSuda, line 35: Could you add a comment to explain why "all" can't be used?

apostasie: Of course. Will do.

apostasie: Done

apostasie:
@AkihiroSuda by the way, and to be completely honest, I am not quite sure why we should give an f at all about armhf.

I do have a bunch of boards on that architecture, but I am running arm64 distros and arm64 containers on them, and it works just fine.
It is just a hobbyist platform, mostly found in iOT, and not for long...

My vote is to just stop worrying about it entirely (and obviously remove traces of it in our codebase, like this here - which is causing issues and complexity for very little reward). We obviously do not care about 386, or ppcle, for good reasons - and I think these reasons should apply to armhf as well.

wdyt?

AkihiroSuda:
We may consider removing armhf when containerd and Moby deprecate it

apostasie:
> We may consider removing armhf when containerd and Moby deprecate it

Sorry if I wasn't clear. I am not saying we should remove support for it - just that we should remove testing for it - as we are not testing either for any of the other platforms - arm/v6, s390x, risc, etc.

AkihiroSuda (approved):
Thanks

apostasie:
@yankay that should simplify things for your CI/ARM PR (hopefully, no longer need to mess around with the architecture).

PR #3304, opened 2024-08-13
Remove leading slash in Windows path

### LINKED ISSUE #2966: Running a Windows container with --detach and --tty results in "nerdctl.exe must be absolute" error message
### Description

Trying to start a Windows container using -d and -t at the same time results in an error as follows:

```
PS C:\> C:\usr\bin\nerdctl\nerdctl.exe run -d -t mcr.microsoft.com/powershell:7.3-windowsservercore-ltsc2022
time="2024-05-03T17:58:37-07:00" level=fatal msg="\"\\\\C:\\\\usr\\\\bin\\\\nerdctl\\\\nerdctl.exe\" must be absolute"
```

I discovered this because I have a Windows Jenkins agent and I'm using their Docker plugin's ".inside()" method, which calls starts a container using `docker run -d -t <whatever>`. So far all other functionality has been perfectly Docker compatible, but I can't find a workaround for this. I initially thought the problem was with Jenkins, and the fact that I was running a symlink called "docker.exe" through the Jenkins plugin, however I've been able to consistently replicate the behavior with several different images, calling the nerdctl executable directly, as shown above. I even moved the executable to a directory with no spaces in its name just to eliminate that as a possible issue.

1. Pull a compatible Windows image to a Windows machine
2. Do `nerdctl run -d -t <that image>`

I should be able to get a daemonized container with a TTY allocated, but instead I get the above error.

```
Client:
 Version:       v2.0.0-beta.5
 OS/Arch:       windows/amd64
 Git commit:    9236b9370203b7f3274ae8d1417cc6ddcbdd4511
 buildctl:
  Version:      v0.13.2
  GitCommit:    2e18d709fefdcc2db20853ee241c75b058189d39

Server:
 containerd:
  Version:      v1.7.13
  GitCommit:    7c3aca7a610df76212171d200ca3811ff6096eb8
 runc:
  Version:
```

None

### Host information

```
Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v1.7.13
 Storage Driver: windows
 Logging Driver: json-file
 Cgroup Driver:
 Cgroup Version:
 Plugins:
  Log: fluentd journald json-file syslog
  Storage: windows-lcow windows
 Security Options:
 Kernel Version:
 Operating System:
 OSType: windows
 Architecture:
 CPUs: 0
 Total Memory: 0B
 Name: <hostname>
 ID: <id>
```

TinaMor:
## PR Description

This PR resolves #2966

### Background

On Windows, the [LogURIGenerator()](https://github.com/containerd/containerd/blob/a406da962818ba9ca8f38b4ac5215753ceb1846c/pkg/cio/io.go#L295) function in [containerd/containerd](https://github.com/containerd/containerd) fails when it checks if the `logURI` path is absolute. This is due to the leading slash in the parsed logURI, which causes the check to fail and results in an error, ultimately leading to the command's failure.

The logURI parameter is a binary path formatted as a URI, for example:

```
"binary:///C:/Program Files/nerdctl.exe?_NERDCTL_INTERNAL_LOGGING=C%3A%5CProgramData%5Cnerdctl%5C052055e3"
```

When parsed with `url.Parse(logURI)`, the resulting `u.Path` is:

```
"/C:/Program Files/nerdctl.exe"
```

The leading slash is incorrect for Windows paths and causes issues when the path is later checked for being absolute.

### Solution
To resolve this issue, the leading slash must be removed from the `u.Path` before the absolute path check is performed. This ensures the path is correctly recognized as an absolute path on Windows systems.

### Additional tasks

Remove Windows check in [TestRunWithTtyAndDetached](https://github.com/containerd/nerdctl/blob/74f2755c0c3f965e97297b2dd53890724b17467e/cmd/nerdctl/container_run_test.go#L461-L464C3). 'json-file' logging for Windows implemented by PR #1468 .

https://github.com/containerd/nerdctl/blob/74f2755c0c3f965e97297b2dd53890724b17467e/cmd/nerdctl/container_run_test.go#L461-L464

### DIFF
--- a/pkg/taskutil/taskutil.go
+++ b/pkg/taskutil/taskutil.go
@@ -96,11 +96,15 @@ func NewTask(ctx context.Context, client *containerd.Client, container container
 		if len(args) != 2 {
 			return nil, errors.New("parse logging path error")
 		}
-		ioCreator = cio.TerminalBinaryIO(u.Path, map[string]string{
+		parsedPath := u.Path
+		// For Windows, remove the leading slash
+		if (runtime.GOOS == "windows") && (strings.HasPrefix(parsedPath, "/")) {
+			parsedPath = strings.TrimLeft(parsedPath, "/")
+		}
+		ioCreator = cio.TerminalBinaryIO(parsedPath, map[string]string{
 			args[0]: args[1],
 		})
 	} else if flagT && !flagD {
-
 		if con == nil {
 			return nil, errors.New("got nil con with flagT=true")
 		}
--- a/cmd/nerdctl/container/container_run_linux_test.go
+++ b/cmd/nerdctl/container/container_run_linux_test.go
@@ -498,3 +498,24 @@ func TestRunWithDetachKeys(t *testing.T) {
 	container := base.InspectContainer(containerName)
 	assert.Equal(base.T, container.State.Running, true)
 }
+
+func TestRunWithTtyAndDetached(t *testing.T) {
+	base := testutil.NewBase(t)
+	imageName := testutil.CommonImage
+	withoutTtyContainerName := "without-terminal-" + testutil.Identifier(t)
+	withTtyContainerName := "with-terminal-" + testutil.Identifier(t)
+
+	// without -t, fail
+	base.Cmd("run", "-d", "--name", withoutTtyContainerName, imageName, "stty").AssertOK()
+	defer base.Cmd("container", "rm", "-f", withoutTtyContainerName).AssertOK()
+	base.Cmd("logs", withoutTtyContainerName).AssertCombinedOutContains("stty: standard input: Not a tty")
+	withoutTtyContainer := base.InspectContainer(withoutTtyContainerName)
+	assert.Equal(base.T, 1, withoutTtyContainer.State.ExitCode)
+
+	// with -t, success
+	base.Cmd("run", "-d", "-t", "--name", withTtyContainerName, imageName, "stty").AssertOK()
+	defer base.Cmd("container", "rm", "-f", withTtyContainerName).AssertOK()
+	base.Cmd("logs", withTtyContainerName).AssertCombinedOutContains("speed 38400 baud; line = 0;")
+	withTtyContainer := base.InspectContainer(withTtyContainerName)
+	assert.Equal(base.T, 0, withTtyContainer.State.ExitCode)
+}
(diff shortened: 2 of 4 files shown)


TinaMor:
Waiting for the changes from [Set stderr to empty string when using terminal on Windows](https://github.com/containerd/containerd/pull/10412) to be available in the latest Containerd v2 release.

apostasie:
Hey @TinaMor 

2.0.0 rc4 just dropped and we moved nerdctl (and testing env) to it.

Curious if the change in containerd got in?

pkg/taskutil/taskutil.go
```diff
@@ -96,11 +96,15 @@ func NewTask(ctx context.Context, client *containerd.Client, container container
 		if len(args) != 2 {
 			return nil, errors.New("parse logging path error")
 		}
-		ioCreator = cio.TerminalBinaryIO(u.Path, map[string]string{
+		parsedPath := u.Path
+		// For Windows, remove the leading slash
+		if (runtime.GOOS == "windows") && (strings.HasPrefix(parsedPath, "/")) {
```
jsturtevant, line 101: Is this the right place to remove the extra slash?

From the report `"binary:///C:/Program Files/nerdctl.exe?_NERDCTL_INTERNAL_LOGGING=C%3A%5CProgramData%5Cnerdctl%5C052055e3"` it seems it has 3 slashes in front of the file path: `binary:///`

Where is that generated and why does it have three slashes?

TinaMor: The path is prefixed with a "/" in containerd/containerd [LogURIGenerator()](https://github.com/containerd/containerd/blob/a406da962818ba9ca8f38b4ac5215753ceb1846c/pkg/cio/io.go#L299C2-L308C3). See the comment in [LogURIGenerator()](https://github.com/containerd/containerd/blob/a406da962818ba9ca8f38b4ac5215753ceb1846c/pkg/cio/io.go#L299C2-L308C3).

jsturtevant: ah interesting, it is expected that url.parse returns `/c:/path` https://github.com/golang/go/commit/844b625ebcc7101e09fb87828a0e71db942a2416

This actually feels like a bug in `cio.TerminalBinaryIO` when it is checking abs path logic but I there might be some nuance that I am missing.

jsturtevant (commented):
LGTM

AkihiroSuda (approved):
Thanks

Issue #3310, opened 2024-08-14
`nerdctl inspect` response does not include `Ports: { }` for containers without published ports


chews93319:
### Description

When a container is launched without a published port, the response of `nerdctl inspect` omits the `Ports` key-value pair from the `NetworkSettings`.

Compared to the Docker scenario, when `docker inspect` evaluates a container without a published port, the `Ports` key-value pair is present, but the value is an empty structure.

This bug affects integration with VSCode DevContainers because DevContainers expects the `Ports` key-value pair to always be present.
When the `Ports` key-value pair is missing, a stack trace error reports: `TypeError: Cannot read properties of null (reading 'Ports')`

----------

According to [dockercompat line189](https://github.com/containerd/nerdctl/blob/74f2755c0c3f965e97297b2dd53890724b17467e/pkg/inspecttypes/dockercompat/dockercompat.go#L189) and the [logical evaluation of Port Annotations](https://github.com/containerd/nerdctl/blob/74f2755c0c3f965e97297b2dd53890724b17467e/pkg/inspecttypes/dockercompat/dockercompat.go#L440), the `Ports` are not assigned within the response when values are present within the Annotations.
Therefore, containers without published ports yield an `NetworkSettings` structure without a `Ports` key-value pair.

### Steps to reproduce the issue

1.  Using [finch](https://github.com/runfinch/finch), launch a container with specified ports
    a.  `finch run -p 8080:80 -d alpine:latest sleep infinity`, which invokes `nerdctl run -p 8080:80 -d alpine:latest sleep infinity`
1. Inspect the container to verify response payload
    a. `finch inspect --mode=dockercompat <container_id>`, which invokes `nerdctl inspect --mode=dockercompat <container_id>`
1. Confirm the presence of `Ports` entity within `NetworkSettings` struct
    ```
    "NetworkSettings": {
        "Ports": {
            "80/tcp": [
                {
                    "HostIp": "0.0.0.0",
                    "HostPort": "8080"
                }
            ]
        },
        "GlobalIPv6Address": "",
        "GlobalIPv6PrefixLen": 0,
        "IPAddress": "10.4.0.9",
        "IPPrefixLen": 24,
        "MacAddress": "xx:xx:xx:xx:xx:xx",
        "Networks": {
            "unknown-eth0": {
                "IPAddress": "10.4.0.9",
                "IPPrefixLen": 24,
                "GlobalIPv6Address": "",
                "GlobalIPv6PrefixLen": 0,
[...]

Issue #2999, opened 2024-05-13
nerdctl inspect does not list networks name properly


apostasie:
### Description

Title says all.

### Steps to reproduce the issue

```
sudo ./nerdctl run --name foo --network bridge -d debian sleep 3600
sudo ./nerdctl inspect foo
```

### Describe the results you received and expected

"bridge" is listed as "unknown-eth0"

```
        "NetworkSettings": {
            "GlobalIPv6Address": "",
            "GlobalIPv6PrefixLen": 0,
            "IPAddress": "10.4.0.50",
            "IPPrefixLen": 24,
            "MacAddress": "fa:b9:e3:9f:67:1b",
            "Networks": {
                "unknown-eth0": {
                    "IPAddress": "10.4.0.50",
                    "IPPrefixLen": 24,
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "fa:b9:e3:9f:67:1b"
                }
            }
        }
```

### What version of nerdctl are you using?

1.7.6.

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

Shubhranshu153:
Can confirm this output for dockercompat version
There is a todo in ContainerFromNative function

```
// TODO: set CNI name when possible
fakeDockerNetworkName := fmt.Sprintf("unknown-%s", x.Name)
```

apostasie:
@Shubhranshu153 
Looking at dockercompat.go now - looks like a lot of TODO is needed.
It is also quite outdated wrt current Moby.

Mind if I give it a go overall?

Shubhranshu153:
feel free to, i looked a bit but havent done any implementation.

newsworthy39:
When using, the inspect function, the hyphen (-), causes a a 'bad character'

```
nerdctl container inspect container --format '{{json .NetworkSettings.Networks.unknown-eth0 }}',
```

template: :1: bad character U+002D '-'

changing the character, to say, a underscore (_), as a temporary workaround, allows that

```
nerdctl container inspect container --format '{{json .NetworkSettings.Networks.unknown_eth0.IPAddress }}'
result: "10.0.92.34"
```

apostasie:
@newsworthy39 

`nerdctl container inspect alpine-56303 --format '{{ json (index .NetworkSettings.Networks "unknown-eth0") }}'`

dash is reserved in go templating.

PR #3311, opened 2024-08-14
Always include Ports within inspect response

chews93319:
Resolves https://github.com/containerd/nerdctl/issues/3310

New behavior will always initialize a PortMap for the inspect response. If published ports are available, then values will be included in the response.

### Background

When a container is launched without a published port, the response of `nerdctl inspect` omits the `Ports` key-value pair from the `NetworkSettings`.

Compared to the Docker scenario, when `docker inspect` evaluates a container without a published port, the `Ports` key-value pair is present, but the value is an empty structure.

This bug affects integration with VSCode DevContainers because DevContainers expects the `Ports` key-value pair to always be present.
When the `Ports` key-value pair is missing, a stack trace error reports: `TypeError: Cannot read properties of null (reading 'Ports')`

### Solution

To resolve this issue, the `PortMap` is initialized prior to logic evaluating "Port Annotations" to ensure that an empty PortMap is always included within the `inspect` response.
When "Port Annotations" are available, the new logic will add the key-value pairs into the initialized `PortMap` of the response.

### Additional Tasks

This bug is also present within `v1.7.6` and the change can be easily reconciled with the released version.

- [ ] **Please consider this PR for Back Porting into `Release/1.7` branch.**


apostasie:
@chews93319 Nice!

Two things:
- CI failures are unrelated - canary is bust for now - and the other is network related.
- maybe we could have a test or two?

chews93319:
@apostasie Thanks for taking a look.

I actually need to iterate on my PR implementation for another use-case (Refer to my updated [Issue 3310 Description](https://github.com/containerd/nerdctl/issues/3310) ), so I will also try to introduce a test against the method `networkSettingsFromNative`.
Interestingly, there were no tests against the method before, so it will be a good step to introduce something new.

apostasie:
> @apostasie Thanks for taking a look.
> 
> I actually need to iterate on my PR implementation for another use-case (Refer to my updated [Issue 3310 Description](https://github.com/containerd/nerdctl/issues/3310) ), so I will also try to introduce a test against the method `networkSettingsFromNative`. Interestingly, there were no tests against the method before, so it will be a good step to introduce something new.

That's very welcome. Inspect needs a lot of love...

There is a bunch of additional issues with it.
If you see anything low-hanging fruit in these while patching that code region:

- #3006 
- #3005
- #2999
- #2998 
- #2905
- #415

CI Canary failure got fixed (rebase should bring in the fix).

I am not a maintainer here - but lmk if I can help you with anything.

PR #3316, opened 2024-08-15
Add builder prune --all and --force flag support

austinvazquez:
### Description

This change adds support for builder prune `--all` and `--force`. A refactor
of previous tests which used builder prune functionality was required to
handle the new confirmation prompt. A test utility was added to handle
this for reuse by current and future tests.

### Testing

From `containerd/nerdctl`:
```
nerdctl builder build .
nerdctl builder prune --all --force
```

### Next steps
1. Test assertions are weak without a `builder du` implementation


cmd/nerdctl/builder.go
```diff
@@ -58,29 +58,83 @@ func newBuilderPruneCommand() *cobra.Command {
...
 	}
-	buildkitHost, err := getBuildkitHost(cmd, globalOptions.Namespace)
+
+	if !options.Force {
+		var (
+			confirm string
+			msg     string
+		)
+
+		if !options.All {
```
pendo324, line 79: nit: might be slightly easier to read if you reverse these conditionals since we need to handle both cases anyway

pendo324 (approved):

pkg/api/types/builder_types.go
```diff
@@ -82,4 +82,6 @@ type BuilderPruneOptions struct {
 	BuildKitHost string
 	// All will remove all unused images and all build cache, not just dangling ones
 	All bool
+	// Force will not prompt for confirmation.
+	Force bool
```
AkihiroSuda, line 86: Where is this consumed?

austinvazquez: Looks like it ended up being used on the CLI side only, so not much reason to pass it to the package backend.

AkihiroSuda (approved):
Thanks

PR #3318, opened 2024-08-16
Include init NetworkSettings within inspect response

chews93319:
Related to [containerd#3310](https://github.com/containerd/nerdctl/issues/3310)

Replacing [Previous Pull Request 3311](https://github.com/containerd/nerdctl/pull/3311)

New behavior will always initialize a NetworkSettings entity for the inspect response, including Ports child member.

If inspecting a running container with published ports, then all information will be correctly returned.
If inspecting a running container without published ports, then NetworkSettings will include an initialized Ports member. If inspecting a stopped/exited container,
then an entirely initialized, "empty-value" NetworkSettings is returned.

### Background

Research related to Scenario 1:
According to [dockercompat line189](https://github.com/containerd/nerdctl/blob/74f2755c0c3f965e97297b2dd53890724b17467e/pkg/inspecttypes/dockercompat/dockercompat.go#L189) and the [logical evaluation of Port Annotations](https://github.com/containerd/nerdctl/blob/74f2755c0c3f965e97297b2dd53890724b17467e/pkg/inspecttypes/dockercompat/dockercompat.go#L440), the `Ports` are not assigned within the response when values are present within the Annotations.
Therefore, containers without published ports yield an `NetworkSettings` structure without a `Ports` key-value pair.

Research related to Scenario 2:
According to [dockercompat line401](https://github.com/containerd/nerdctl/blob/74f2755c0c3f965e97297b2dd53890724b17467e/pkg/inspecttypes/dockercompat/dockercompat.go#L401), the `native.NetNS` of a stopped/exited container is a null pointer, which would return `nil`.

### Solution

To resolve this issue, the `NetworkSettings` and `PortMap` are initialized at the beginning of the method to ensure that at least "Empty-value" structs are always included within the `inspect` response.
When "Port Annotations" are available, the new logic will add the key-value pairs into the initialized `PortMap` of the response.

Since `Ports` should not be omitted, this change also removes the `json:",omitempty"` decorator for the `NetworkSettings` struct.

### Additional Tasks

This bug is also present within `v1.7.6` and the change can be easily reconciled with the released version.

- [ ] **Please consider this PR for Back Porting into `Release/1.7` branch.**

### DIFF
--- a/pkg/inspecttypes/dockercompat/dockercompat.go
+++ b/pkg/inspecttypes/dockercompat/dockercompat.go
@@ -186,7 +186,7 @@ type ContainerState struct {
 }
 
 type NetworkSettings struct {
-	Ports *nat.PortMap `json:",omitempty"`
+	Ports *nat.PortMap
 	DefaultNetworkSettings
 	Networks map[string]*NetworkEndpointSettings
 }
@@ -398,12 +398,15 @@ func statusFromNative(x containerd.Status, labels map[string]string) string {
 }
 
 func networkSettingsFromNative(n *native.NetNS, sp *specs.Spec) (*NetworkSettings, error) {
-	if n == nil {
-		return nil, nil
-	}
 	res := &NetworkSettings{
 		Networks: make(map[string]*NetworkEndpointSettings),
 	}
+	resPortMap := make(nat.PortMap)
+	res.Ports = &resPortMap
+	if n == nil {
+		return res, nil
+	}
+
 	var primary *NetworkEndpointSettings
 	for _, x := range n.Interfaces {
 		if x.Interface.Flags&net.FlagLoopback != 0 {
@@ -447,8 +450,11 @@ func networkSettingsFromNative(n *native.NetNS, sp *specs.Spec) (*NetworkSetting
 			if err != nil {
 				return nil, err
 			}
-			res.Ports = nports
+			for portLabel, portBindings := range *nports {
+				resPortMap[portLabel] = portBindings
+			}
 		}
+
 		if x.Index == n.PrimaryInterface {
 			primary = nes
 		}
--- a/pkg/inspecttypes/dockercompat/dockercompat_test.go
+++ b/pkg/inspecttypes/dockercompat/dockercompat_test.go
@@ -17,11 +17,13 @@
 package dockercompat
 
 import (
+	"net"
 	"os"
 	"path/filepath"
 	"runtime"
 	"testing"
 
+	"github.com/docker/go-connections/nat"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"gotest.tools/v3/assert"
 
@@ -90,6 +92,10 @@ func TestContainerFromNative(t *testing.T) {
 						"nerdctl/hostname":  "host1",
 					},
 					Hostname: "host1",
+				},
+				NetworkSettings: &NetworkSettings{
+					Ports:    &nat.PortMap{},
+					Networks: map[string]*NetworkEndpointSettings{},
 				},
 			},
 		},
@@ -172,6 +178,10 @@ func TestContainerFromNative(t *testing.T) {
 					// ignore sysfs mountpoint
 				},
 				Config: &Config{},
+				NetworkSettings: &NetworkSettings{
+					Ports:    &nat.PortMap{},
+					Networks: map[string]*NetworkEndpointSettings{},
+				},
 			},
 		},
 		// ctr container, mount /mnt/foo:/mnt/foo:rw,rslave; internal sysfs mount; hostname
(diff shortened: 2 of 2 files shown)


chews93319:
@apostasie 

I have added the requested test cases as well as caught an update to the pre-existing test, which I was responsible for.
Unfortunately, I did something to "close" the previous Pull Request, so this is a "fresh replacement".

PTAL at your convenience.

chews93319:
@AkihiroSuda 

PTAL at your convenience.
We would also like to discuss this change as another Back Port into Release/1.7 branch.

pkg/inspecttypes/dockercompat/dockercompat_test.go
```diff
@@ -226,12 +237,126 @@ func TestContainerFromNative(t *testing.T) {
...
+				NetworkSettings: &NetworkSettings{
+					Ports:    &nat.PortMap{},
+					Networks: map[string]*NetworkEndpointSettings{},
+				},
 			},
 		},
 	}
 
 	for _, tc := range testcase {
+		fmt.Fprintln(os.Stderr, tc.name)
```
Shubhranshu153, line 249: remove the printline.

pkg/inspecttypes/dockercompat/dockercompat_test.go
```diff
@@ -226,12 +237,126 @@ func TestContainerFromNative(t *testing.T) {
...
+						IPPrefixLen: 24,
+						MacAddress:  "xx:xx:xx:xx:xx:xx",
+					},
+				},
+			},
+		},
+		// Given native.NetNS with single Interface without Port Annotations, Return populated NetworkSettings
+		//   UseCase: Inspect a Running Container without published ports
+		{
+			name: "Given NetNS with single Interface without Port Annotations, Return populated NetworkSettings",
```
Shubhranshu153, line 325: nit: initialized network setting.

chews93319: This was a difficult comment and test case name to write.
Technically, a "Running Container without published ports" will still have a populated NetworkSettings entity that contains an "initialized Ports" child entity.

I will try to update the comment to reflect this statement, if this clarification makes sense to you.

pkg/inspecttypes/dockercompat/dockercompat_test.go
chews93319: This print statement was actually an attempt to clarify the sub-test cases, similar to the precedence of other unit tests. (See reference example below)

I will remove the print statements.
If I can easily iterate on the test to uniquely report the sub-case PASS, then I will do that too.

```
--- PASS: TestBuildKitFile (0.00s)
    --- PASS: TestBuildKitFile/only_Dockerfile_is_present (0.00s)
    --- PASS: TestBuildKitFile/only_Containerfile_is_present (0.00s)
    --- PASS: TestBuildKitFile/both_Dockerfile_and_Containerfile_are_present (0.00s)
    --- PASS: TestBuildKitFile/Dockerfile_and_Containerfile_have_different_contents (0.00s)
    --- PASS: TestBuildKitFile/Custom_file_is_specfied (0.00s)
    --- PASS: TestBuildKitFile/Absolute_path_is_specified_along_with_custom_file (0.00s)
    --- PASS: TestBuildKitFile/Absolute_path_is_specified_along_with_Docker_file (0.00s)
    --- PASS: TestBuildKitFile/Absolute_path_is_specified_with_Container_file_in_the_path (0.00s)
```

Shubhranshu153: ah ok.  its fine then

pkg/inspecttypes/dockercompat/dockercompat_test.go
Shubhranshu153: thanks

pkg/inspecttypes/dockercompat/dockercompat_test.go
chews93319: @Shubhranshu153  I successfully corrected the syntax to execute the sub-case test names.

pkg/inspecttypes/dockercompat/dockercompat_test.go
chews93319: @Shubhranshu153  within commit `adf4035`, I also updated the test description/name to reflect this conversation.

pkg/inspecttypes/dockercompat/dockercompat_test.go
```diff
@@ -226,12 +236,128 @@ func TestContainerFromNative(t *testing.T) {
...
+	for _, tc := range testcase {
+		t.Run(tc.name, func(tt *testing.T) {
+			d, _ := ContainerFromNative(tc.n)
+			assert.DeepEqual(tt, d, tc.expected)
+		})
+	}
+}
+
+func TestNetworkSettingsFromNative(t *testing.T) {
+	tempStateDir, err := os.MkdirTemp(t.TempDir(), "rw")
```
AkihiroSuda, line 256: t.TempDir can be used

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -186,7 +186,7 @@ type ContainerState struct {
 }
 
 type NetworkSettings struct {
-	Ports *nat.PortMap `json:",omitempty"`
+	Ports *nat.PortMap
```
AkihiroSuda, line 189: Docker seems to use `Ports nat.PortMap`
https://github.com/moby/moby/blob/v27.1.2/api/types/types.go#L302

AkihiroSuda (approved):
Thanks

PR #3319, opened 2024-08-17
Add image prune --filter support

austinvazquez:
### Description
This change adds nerdctl image prune `--filter` support including:

- `image prune --filter label=<key>=<value>`
- `image prune --filter until=<timestamp>`

### Testing

- Added unit tests for `pkg/imgutil` code
- Added integration tests for `cmd/nerdctl/image_prune` code
  - `TestImagePruneFilterUntil` does not currently work on Docker. Docker image's created timestamp is aligning with the timestamp of the base image. Docker supports the `until` filter; just didn't have any ideas for deterministically testing both behaviors.

### References

- https://docs.docker.com/reference/cli/docker/image/prune/#filter

### DIFF
--- a/cmd/nerdctl/image_prune_test.go
+++ b/cmd/nerdctl/image_prune_test.go
@@ -19,6 +19,7 @@ package main
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 )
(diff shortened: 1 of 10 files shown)


cmd/nerdctl/image_prune_test.go
```diff
@@ -71,3 +72,57 @@ func TestImagePruneAll(t *testing.T) {
...
+	buildCtx := createBuildContext(t, dockerfile)
+
+	base.Cmd("build", "-t", imageName, buildCtx).AssertOK()
+	base.Cmd("images", "--all").AssertOutContains(imageName)
+
+	base.Cmd("image", "prune", "--force", "--all", "--filter", "until=12h").AssertOK()
+	base.Cmd("images", "--all").AssertOutContains(imageName)
+
+	// Pause to ensure enough time has passed for the image to be clean on next prune.
+	time.Sleep(1 * time.Second)
```
AkihiroSuda, line 124: This might be too short?

pkg/imgutil/filtering_test.go
```diff
@@ -0,0 +1,467 @@
...
+	}{
+		{
+			name:          "ImageHasNoLabels",
+			imageLabels:   map[string]string{},
+			labelsToMatch: map[string]string{"foo": "bar"},
+			matches:       false,
+		},
+		{
+			name:          "SingleMatchingLabel",
+			imageLabels:   map[string]string{"org": "com.containerd.nerdctl"},
```
AkihiroSuda, line 438: Better to use `com.example...`

cmd/nerdctl/image_prune_test.go
austinvazquez: Tried increase the pause to 3 seconds and lowering the requirement to 10ms to provide two orders of magnitude here. _Should_ be pretty reliable. 😅

AkihiroSuda (approved):
Thanks

PR #3320, opened 2024-08-17
update Go (1.23), BuildKit (0.15.2), RootlessKit (2.3.1), and test deps

AkihiroSuda:
Will release v2.0.0-rc.1 after merging this

PR #3321, opened 2024-08-17
Add linter to Canary and fix go 1.23 linting issues

apostasie:
This adds a lint validation to the Canary target.
Currently done on windows (because of the go install on the host), but will move to an independent stage after cleaning up the go_version detection routine.

Once green, we should merge ASAP to help with #3320

### DIFF
--- a/.github/workflows/test-canary.yml
+++ b/.github/workflows/test-canary.yml
@@ -15,6 +15,27 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
+  lint:
+    runs-on: "ubuntu-24.04"
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4.1.7
+        with:
+          fetch-depth: 1
+      - name: Set GO env
+        run: |
+          . ./hack/build-integration-canary.sh
+          canary::golang::latest
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          check-latest: true
+          cache: true
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6.1.0
+        with:
+          args: --verbose
+
   linux:
     runs-on: "ubuntu-24.04"
     timeout-minutes: 40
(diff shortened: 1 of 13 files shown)


.github/workflows/test-canary.yml
```diff
@@ -92,6 +99,11 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
           cache: true
           check-latest: true
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6.1.0
+        with:
+          version: v1.59.1
```
AkihiroSuda, line 105: For canary probably this should be left blank?

.github/workflows/test.yml
```diff
@@ -222,7 +222,7 @@ jobs:
           if echo "${ROOTLESSKIT_VERSION}" | grep -q v1; then
             WORKAROUND_ISSUE_622=1
           fi
-          echo "WORKAROUND_ISSUE_622=${WORKAROUND_ISSUE_622}" >>$GITHUB_ENV
+          echo "WORKAROUND_ISSUE_622=${WORKAROUND_ISSUE_622}" >> "$GITHUB_ENV"
```
AkihiroSuda, line 225: ?

AkihiroSuda:
Thanks, in the commit message could you mention the fixed errors such as `SA4032: due to the file's build constraints, runtime.GOOS will never equal "windows" (staticcheck)`?

The commits can be also split for the each of the SA%d fixes.

AkihiroSuda:
> go 1.23 linting issues

IIUC this is the matter of golangci-lint rather than go itself?

.github/workflows/test.yml
apostasie: That must be quoted to prevent splitting.

apostasie:
> > go 1.23 linting issues
> 
> IIUC this is the matter of golangci-lint rather than go itself?

I think it is a combination of both? (eg: golangci-lint with that version of go)

Anyhow, thanks for the comments.

Will integrate all and split in different commits.

Right now struggling with windows (as usual), will re-request review when done.

apostasie:
@AkihiroSuda canary lint is green. All comments addressed.

AkihiroSuda (approved):
Thanks

PR #3323, opened 2024-08-17
Rewrite cp

### LINKED ISSUE #1257: Flaky: TestCopyToContainer

### LINKED ISSUE #3197: cp failure uses confusing error message

### LINKED ISSUE #3313: cp fails to detect readonly mount and fallback into tar error
### Description

This

https://github.com/containerd/nerdctl/blob/main/pkg/containerutil/cp_linux.go#L305

should surface error: "mount point %s is marked read-only"

This is two fold:
- error message
- more importantly, the logic in getContainerMountInfo seems to be broken

```
touch something
nerdctl volume create foo
nerdctl run -d --name testro -v foo:/bar:ro debian sleep Inf
nerdctl cp something testro:/bar
```

Received:
```
/usr/bin/tar: something: Cannot open: Read-only file system
/usr/bin/tar: Exiting with failure status due to previous errors
FATA[0000] could not find container: testro, with error: failed to wait [nsenter -t 1958627 -U --preserve-credentials -- /usr/bin/tar -x -f -]: exit status 2
```

Expected:

```
mount point %s is marked read-only
```

main

None

### Host information

Am I missing something or doing something wrong?

### LINKED ISSUE #3315: cp does not detect that a container is `--read-only`, and lets tar fail through in some cases
### Description

Title says all.

Start a container read-only. cp into it

Similarly to #3313 there is a mechanism to detect that in some cases - just not for when the container is running.

na

main

None

### Host information

_No response_

apostasie:
The motivation for this is to address all issues (and variants not listed as individual tickets) in:
- fix #3315
- fix #3313 
- (hopefully) fix #1257
- fix #3197

The fundamental problem with our current implementation of cp is that the logic to figure out if a destination is read-only is incomplete.
One of the key reason for it is that we do not follow symlinks, or relative paths, inside a container destination. For example, if we are copying to a path located into a readonly volume, we stop there, while the path may very well resolve to a completely different volume (or the rootfs).
Furthermore, there are other logic issues involving readonly rootfs, leading to situations where we DO copy while we should not, and conversely, where we do NOT copy while we should.

There is a also a fair amount of duplication, and finally, test isolation was problematic.

For this PR, the highlights are, for the testing part:
- tests rewritten from scratch to cleanly separate test cases and test rig (should make adding more tests cases easier)
- expanded test-suite to add more cases
- better isolation
- beside the "regular" expanded tests, added a "TestAcidCopy" test meant to reproduce corner/complicated cases and other "special" conditions that do not fit in the normal test rig

For the code part:
- provide a mechanism to fully resolve a *container* path while on the host when all we have is the mounted root snapshot
- provide a range of more expressive and informative errors for the user to diagnose issues
- should address all link resolution, relative paths and read-only concerns

Also note this is superseding #3275

I appreciate this is a somewhat sizable PR (although a very large part of it (about 1500 lines) are just for tests).

If there is a different (simpler) approach to this problem - that would pass the test-suite - I am happy to re-evaluate of course.

### DIFF
--- a/cmd/nerdctl/container/container_cp_linux.go
+++ b/cmd/nerdctl/container/container_cp_linux.go
@@ -115,16 +115,16 @@ func processCpOptions(cmd *cobra.Command, args []string) (types.ContainerCpOptio
 	}
 
 	container2host := srcSpec.Container != nil
-	var container string
+	var containerReq string
 	if container2host {
-		container = *srcSpec.Container
+		containerReq = *srcSpec.Container
 	} else {
-		container = *destSpec.Container
+		containerReq = *destSpec.Container
 	}
 	return types.ContainerCpOptions{
 		GOptions:       globalOptions,
 		Container2Host: container2host,
-		ContainerReq:   container,
+		ContainerReq:   containerReq,
 		DestPath:       destSpec.Path,
 		SrcPath:        srcSpec.Path,
 		FollowSymLink:  flagL,
--- a/pkg/cmd/container/cp_linux.go
+++ b/pkg/cmd/container/cp_linux.go
@@ -39,17 +39,22 @@ func Cp(ctx context.Context, client *containerd.Client, options types.ContainerC
 				ctx,
 				client,
 				found.Container,
-				options.Container2Host,
-				options.DestPath,
-				options.SrcPath,
-				options.GOptions.Snapshotter,
-				options.FollowSymLink)
+				options)
 		},
 	}
 	count, err := walker.Walk(ctx, options.ContainerReq)
 
-	if count < 1 {
-		err = fmt.Errorf("could not find container: %s, with error: %w", options.ContainerReq, err)
+	if count == -1 {
+		if err == nil {
+			panic("nil error and count == -1 from ContainerWalker.Walk should never happen")
+		}
+		err = fmt.Errorf("unable to copy: %w", err)
+	} else if count == 0 {
+		if err != nil {
+			err = fmt.Errorf("unable to retrieve containers with error: %w", err)
+		} else {
+			err = fmt.Errorf("no container found for: %s", options.ContainerReq)
+		}
 	}
 
 	return err
--- a/pkg/containerutil/cp_linux.go
+++ b/pkg/containerutil/cp_linux.go
@@ -17,163 +17,200 @@
 package containerutil
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
-	"io/fs"
 	"os"
 	"os/exec"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
 
-	securejoin "github.com/cyphar/filepath-securejoin"
-
 	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/core/containers"
 	"github.com/containerd/containerd/v2/core/mount"
 	"github.com/containerd/errdefs"
 	"github.com/containerd/log"
 
+	"github.com/containerd/nerdctl/v2/pkg/api/types"
 	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
 	"github.com/containerd/nerdctl/v2/pkg/tarutil"
 )
 
-// CopyFiles implements `nerdctl cp`.
 // See https://docs.docker.com/engine/reference/commandline/cp/ for the specification.
-func CopyFiles(ctx context.Context, client *containerd.Client, container containerd.Container, container2host bool, dst, src string, snapshotter string, followSymlink bool) error {
+
+var (
+	// Generic and system errors
+	ErrFilesystem             = errors.New("filesystem error") // lstat hard errors, etc
+	ErrContainerVanished      = errors.New("the container you are trying to copy to/from has been deleted")
+	ErrRootlessCannotCp       = errors.New("cannot use cp with stopped containers in rootless mode") // rootless cp with a stopped container
+	ErrFailedMountingSnapshot = errors.New("failed mounting snapshot")                               // failure to mount a stopped container snapshot
+
+	// CP specific errors
+	ErrTargetIsReadOnly           = errors.New("cannot copy into read-only location")                            // ...
+	ErrSourceIsNotADir            = errors.New("source is not a directory")                                      // cp SOMEFILE/ foo:/
+	ErrDestinationIsNotADir       = errors.New("destination is not a directory")                                 // * cp ./ foo:/etc/issue/bah
+	ErrSourceDoesNotExist         = errors.New("source does not exist")                                          // cp NONEXISTENT foo:/
+	ErrDestinationParentMustExist = errors.New("destination parent does not exist")                              // nerdctl cp VALID_PATH foo:/NONEXISTENT/NONEXISTENT
+	ErrDestinationDirMustExist    = errors.New("the destination directory must exist to be able to copy a file") // * cp SOMEFILE foo:/NONEXISTENT/
+	ErrCannotCopyDirToFile        = errors.New("cannot copy a directory to a file")                              // cp SOMEDIR foo:/etc/issue
+)
+
+// getRoot will tentatively return the root of the container on the host (/proc/pid/root), along with the pid,
+// (eg: doable when the container is running)
+func getRoot(ctx context.Context, container containerd.Container) (string, int, error) {
+	task, err := container.Task(ctx, nil)
+	if err != nil {
+		return "", 0, err
+	}
+
+	status, err := task.Status(ctx)
+	if err != nil {
+		return "", 0, err
+	}
+
+	if status.Status != containerd.Running {
+		return "", 0, nil
+	}
+	pid := int(task.Pid())
+
+	return fmt.Sprintf("/proc/%d/root", pid), pid, nil
+}
+
+// CopyFiles implements `nerdctl cp`
+// It currently depends on the following assumptions:
+// - linux only
+// - tar binary exists on the system
+// - nsenter binary exists on the system
+// - if rootless, the container is running (aka: /proc/pid/root)
+func CopyFiles(ctx context.Context, client *containerd.Client, container containerd.Container, options types.ContainerCpOptions) (err error) {
+	// We do rely on the tar binary as a shortcut - could also be replaced by archive/tar, though that would mean
+	// we need to replace nsenter calls with re-exec
 	tarBinary, isGNUTar, err := tarutil.FindTarBinary()
 	if err != nil {
 		return err
 	}
-	log.G(ctx).Debugf("Detected tar binary %q (GNU=%v)", tarBinary, isGNUTar)
-	var srcFull, dstFull, root, mountDestination, containerPath string
-	var cleanup func()
-	task, err := container.Task(ctx, nil)
-	if err != nil {
-		// FIXME: Rootless does not support copying into/out of stopped/created containers as we need to nsenter into the user namespace of the
-		// pid of the running container with --preserve-credentials to preserve uid/gid mapping and copy files into the container.
-		if rootlessutil.IsRootless() {
-			return errors.New("cannot use cp with stopped containers in rootless mode")
-		}
-		// if the task is simply not found, we should try to mount the snapshot. any other type of error from Task() is fatal here.
-		if !errdefs.IsNotFound(err) {
-			return err
-		}
-		if container2host {
-			containerPath = src
-		} else {
-			containerPath = dst
-		}
-		// Check if containerPath is in a volume
-		root, mountDestination, err = getContainerMountInfo(ctx, container, containerPath, container2host)
-		if err != nil {
-			return err
-		}
-		// if containerPath is in a volume and not read-only in case of host2container copy then handle volume paths,
-		// else containerPath is not in volume so mount container snapshot for copy
-		if root != "" {
-			dst, src = handleVolumePaths(container2host, dst, src, mountDestination)
-		} else {
-			root, cleanup, err = mountSnapshotForContainer(ctx, client, container, snapshotter)
-			if cleanup != nil {
-				defer cleanup()
-			}
-			if err != nil {
-				return err
-			}
-		}
-	} else {
-		status, err := task.Status(ctx)
-		if err != nil {
-			return err
-		}
-		if status.Status == containerd.Running {
-			root = fmt.Sprintf("/proc/%d/root", task.Pid())
-		} else {
-			if rootlessutil.IsRootless() {
-				return fmt.Errorf("cannot use cp with stopped containers in rootless mode")
-			}
-			if container2host {
-				containerPath = src
-			} else {
-				containerPath = dst
-			}
-			root, mountDestination, err = getContainerMountInfo(ctx, container, containerPath, container2host)
-			if err != nil {
-				return err
-			}
-			// if containerPath is in a volume and not read-only in case of host2container copy then handle volume paths,
-			// else containerPath is not in volume so mount container snapshot for copy
-			if root != "" {
-				dst, src = handleVolumePaths(container2host, dst, src, mountDestination)
-			} else {
-				root, cleanup, err = mountSnapshotForContainer(ctx, client, container, snapshotter)
-				if cleanup != nil {
-					defer cleanup()
-				}
-				if err != nil {
-					return err
-				}
-			}
-		}
-	}
-	if container2host {
-		srcFull, err = securejoin.SecureJoin(root, src)
-		dstFull = dst
-	} else {
-		srcFull = src
-		dstFull, err = securejoin.SecureJoin(root, dst)
-	}
-	if err != nil {
-		return err
-	}
-	var (
-		srcIsDir       bool
-		dstExists      bool
-		dstExistsAsDir bool
-		st             fs.FileInfo
-	)
-	st, err = os.Stat(srcFull)
-	if err != nil {
-		return err
-	}
-	srcIsDir = st.IsDir()
 
-	// dst may not exist yet, so err is negligible
-	if st, err := os.Stat(dstFull); err == nil {
-		dstExists = true
-		dstExistsAsDir = st.IsDir()
+	log.G(ctx).Debugf("Detected tar binary %q (GNU=%v)", tarBinary, isGNUTar)
+
+	// This can happen if the container being passed has been deleted since in a racy way
+	conSpec, err := container.Spec(ctx)
+	if err != nil {
+		return errors.Join(ErrContainerVanished, err)
 	}
-	dstEndsWithSep := strings.HasSuffix(dst, string(os.PathSeparator))
-	srcEndsWithSlashDot := strings.HasSuffix(src, string(os.PathSeparator)+".")
-	if !srcIsDir && dstEndsWithSep && !dstExistsAsDir {
-		// The error is specified in https://docs.docker.com/engine/reference/commandline/cp/
-		// See the `DEST_PATH does not exist and ends with /` case.
-		return fmt.Errorf("the destination directory must exists: %w", err)
+
+	// Try to get a running container root
+	root, pid, err := getRoot(ctx, container)
+	// If the task is "not found" (for example, if the container stopped), we will try to mount the snapshot
+	// Any other type of error from Task() is fatal here.
+	if err != nil && !errdefs.IsNotFound(err) {
+		return errors.Join(ErrContainerVanished, err)
 	}
-	if !srcIsDir && srcEndsWithSlashDot {
-		return fmt.Errorf("the source is not a directory")
+
+	log.G(ctx).Debugf("We have root %s and pid %d", root, pid)
+
+	// If we have no root:
+	// - bail out for rootless
+	// - mount the snapshot for rootful
+	if root == "" {
+		// FIXME: Rootless does not support copying into/out of stopped/created containers as we need to nsenter into
+		// the user namespace of the pid of the running container with --preserve-credentials to preserve uid/gid
+		// mapping and copy files into the container.
+		if rootlessutil.IsRootless() {
+			return ErrRootlessCannotCp
+		}
+
+		// See similar situation above. This may happen if we are racing against container deletion
+		var conInfo containers.Container
+		conInfo, err = container.Info(ctx)
+		if err != nil {
+			return errors.Join(ErrContainerVanished, err)
+		}
+
+		var cleanup func() error
+		root, cleanup, err = mountSnapshotForContainer(ctx, client, conInfo, options.GOptions.Snapshotter)
+		if cleanup != nil {
+			defer func() {
+				err = errors.Join(err, cleanup())
+			}()
+		}
+
+		if err != nil {
+			return errors.Join(ErrFailedMountingSnapshot, err)
+		}
+
+		log.G(ctx).Debugf("Got new root %s", root)
 	}
-	if srcIsDir && dstExists && !dstExistsAsDir {
-		return fmt.Errorf("cannot copy a directory to a file")
+
+	var sourceSpec, destinationSpec *pathSpecifier
+	var sourceErr, destErr error
+	if options.Container2Host {
+		sourceSpec, sourceErr = getPathSpecFromContainer(options.SrcPath, conSpec, root)
+		destinationSpec, destErr = getPathSpecFromHost(options.DestPath)
+	} else {
+		sourceSpec, sourceErr = getPathSpecFromHost(options.SrcPath)
+		destinationSpec, destErr = getPathSpecFromContainer(options.DestPath, conSpec, root)
 	}
-	if srcIsDir && !dstExists {
-		if err := os.MkdirAll(dstFull, 0o755); err != nil {
-			return err
+
+	if destErr != nil {
+		if errors.Is(destErr, errDoesNotExist) {
+			return ErrDestinationParentMustExist
+		} else if errors.Is(destErr, errIsNotADir) {
+			return ErrDestinationIsNotADir
+		}
+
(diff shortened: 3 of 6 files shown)


pkg/cmd/container/cp_linux.go
```diff
@@ -39,11 +39,7 @@ func Cp(ctx context.Context, client *containerd.Client, options types.ContainerC
 				ctx,
 				client,
 				found.Container,
-				options.Container2Host,
-				options.DestPath,
-				options.SrcPath,
-				options.GOptions.Snapshotter,
-				options.FollowSymLink)
+				options)
```
apostasie, line 42: Simplifying signature

pkg/testutil/testutil.go
```diff
@@ -58,6 +58,7 @@ type Base struct {
 	Binary           string
 	Args             []string
 	Env              []string
+	Dir              string
```
apostasie, line 61: Need this to test relative paths.
We cannot `os.Chdir` for the same reasons we cannot manipulate the environment directly.

cmd/nerdctl/container_cp_linux.go
```diff
@@ -111,16 +111,16 @@ func processCpOptions(cmd *cobra.Command, args []string) (types.ContainerCpOptio
 	}
 
 	container2host := srcSpec.Container != nil
-	var container string
+	var containerReq string
```
apostasie, line 114: Avoid shadowing import.

pkg/cmd/container/cp_linux.go
```diff
@@ -39,17 +40,22 @@ func Cp(ctx context.Context, client *containerd.Client, options types.ContainerC
...
+		if err == nil {
+			panic("nil error and count == -1 from ContainerWalker.Walk should never happen")
+		}
+		err = errors.Join(errors.New("unable to copy"), err)
+	} else if count == 0 {
+		if err != nil {
+			err = errors.Join(errors.New("unable to retrieve containers with error"), err)
+		} else {
+			err = fmt.Errorf("no container found for: %s", options.ContainerReq)
+		}
```
apostasie, line 58: Make sure we report the right error message to the user.

pkg/containerutil/cp_linux.go
```diff
@@ -17,163 +17,200 @@
...
+
+	// A file cannot be copied inside a non-existent directory with a trailing slash, or slash+dot
+	if !sourceSpec.isADir && !destinationSpec.exists && (destinationSpec.endsWithSeparator || destinationSpec.endsWithSeparatorDot) {
+		return ErrDestinationDirMustExist
 	}
-	if srcIsDir && !dstExists {
-		if err := os.MkdirAll(dstFull, 0o755); err != nil {
-			return err
+
+	// XXX FIXME: this seems wrong. What about ownership? We could be doing that inside a container
```
apostasie, line 197: I am not touching this right now.
This is previous code.
Am I mis-reading this, or is this a problem when doing that in a container target? (eg: shouldn't we do that as part of the nsenter/re-exec routine?)

pkg/containerutil/cp_linux.go
```diff
@@ -237,97 +277,74 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
-		return "", nil, err
-	}
-	snapKey := cinfo.SnapshotKey
+func mountSnapshotForContainer(ctx context.Context, client *containerd.Client, conInfo containers.Container, snapshotter string) (string, func() error, error) {
+	snapKey := conInfo.SnapshotKey
 	resp, err := client.SnapshotService(snapshotter).Mounts(ctx, snapKey)
 	if err != nil {
 		return "", nil, err
 	}
+
```
apostasie, line 330: I know you don't like these Akihiro, but would you reconsider? It does make code much easier to read for these old eyes here :-)

pkg/containerutil/cp_resolve.go
```diff
@@ -0,0 +1,436 @@
+/*
```
apostasie, line 1: That is the gist of it, that provides (proper) symlink and path resolution.

pkg/containerutil/cp_resolve_other.go
```diff
@@ -0,0 +1,24 @@
+/*
```
apostasie, line 1: If one day we support this on Windows - we will need the win version of volumeNameLen

pkg/imgutil/filtering_test.go
```diff
@@ -20,8 +20,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/containerd/containerd/v2/core/images"
```
apostasie, line 23: Unrelated.
I believe I fixed this in another recent PR on main, so, shouldn't be a concern?

cmd/nerdctl/container_cp_linux_test.go
```diff
@@ -25,396 +25,933 @@ import (
...
 
+// For the test matrix, see https://docs.docker.com/engine/reference/commandline/cp/
+// Obviously, none of this is fully windows ready - obviously `nerdctl cp` itself is not either, so, ok for now.
+const (
+	// Use this to poke the testing rig for improper path handling
+	// TODO: fuzz this more seriously
+	// FIXME: the following will break the test (anything that will evaluate on the shell, obviously):
+	// - `
+	// - $a, ${a}, etc
+	complexify = "" //  = "-~a0-_.(){}[]*#! \"'∞"
```
apostasie, line 43: Useful to stress test rig when modifying it. Deactivated by default as it makes things much harder to read.

apostasie:
CI is green.

PTAL at your convenience.

cmd/nerdctl/container_cp_acid_linux_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/containerd/nerdctl/v2/pkg/containerutil"
+	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+)
+
+// This is a separate set of tests for cp specifically meant to test corner or extreme cases that do not fit in the normal testing rig
+// because of their complexity
+
+func TestAcidCopy(t *testing.T) {
```
AkihiroSuda, line 36: Does Acid here mean https://en.wikipedia.org/wiki/ACID  ([atomicity](https://en.wikipedia.org/wiki/Atomicity_(database_systems)), [consistency](https://en.wikipedia.org/wiki/Consistency_(database_systems)), [isolation](https://en.wikipedia.org/wiki/Isolation_(database_systems)), [durability](https://en.wikipedia.org/wiki/Durability_(database_systems)))?

AkihiroSuda: Probably tests should have the `TestCopy...` prefix

cmd/nerdctl/container_cp_acid_linux_test.go
```diff
@@ -0,0 +1,180 @@
...
+				ExitCode: 1,
+				Err:      expectedErr,
+			})
+		})
+
+		t.Run("Cannot copy into a HOST read-only location", func(t *testing.T) {
+			t.Parallel()
+
+			if !rootlessutil.IsRootless() {
+				t.Skip("This test does not work rootful right now")
```
AkihiroSuda, line 168: Could you add the reason?

cmd/nerdctl/container_cp_acid_linux_test.go
apostasie: "Acid test" generally would be a really hard to pass, crucial test (https://www.merriam-webster.com/dictionary/acid%20test)

It is also a nod to the old browsers "acid tests" (https://en.wikipedia.org/wiki/Acid1 https://en.wikipedia.org/wiki/Acid2 https://en.wikipedia.org/wiki/Acid3)

I can rename to TestCopyAcid to facilitate `-run TestCopy*`.

cmd/nerdctl/container_cp_acid_linux_test.go
apostasie: Will do tomorrow.

cmd/nerdctl/container_cp_acid_linux_test.go
apostasie: The overall point is that there is two tiers of tests - "the happy path" kind of tests, and the "acid" kind.

"Acid" has very few tests, will very likely break on you very fast, and you can use it instead of the full test-suite to quickly **validate** if something is broken or not - but then it is probably really hard to debug if it fails.

"Happy path" should have many test, not very challenging, easy to debug, will run much longer, and is more useful for incremental debugging.

apostasie:
Rebased and comments addressed.

pkg/testutil/testutil.go
```diff
@@ -70,6 +71,7 @@ func WithStdin(r io.Reader) func(*Cmd) {
 func (b *Base) Cmd(args ...string) *Cmd {
 	icmdCmd := icmd.Command(b.Binary, append(b.Args, args...)...)
 	icmdCmd.Env = b.Env
+	icmdCmd.Dir = b.Dir
```
djdongjin, line 74: what's the behavior if we don't set `Dir` (which means both b.Dir and icmdCmd.Dir are "")?

pkg/cmd/container/cp_linux.go
```diff
@@ -39,17 +40,22 @@ func Cp(ctx context.Context, client *containerd.Client, options types.ContainerC
...
-				options.GOptions.Snapshotter,
-				options.FollowSymLink)
+				options)
 		},
 	}
 	count, err := walker.Walk(ctx, options.ContainerReq)
 
-	if count < 1 {
-		err = fmt.Errorf("could not find container: %s, with error: %w", options.ContainerReq, err)
+	if count == -1 {
```
djdongjin, line 48: I don't quite follow this change. What's the issue with the current logic? (also I don't think we should panic here).

pkg/containerutil/cp_linux.go
```diff
@@ -17,163 +17,200 @@
...
-		}
-		// Check if containerPath is in a volume
-		root, mountDestination, err = getContainerMountInfo(ctx, container, containerPath, container2host)
+
+		// See similar situation above. This may happen if we are racing against container deletion
+		var conInfo containers.Container
+		conInfo, err = container.Info(ctx)
 		if err != nil {
-			return err
+			return errors.Join(ErrShouldNeverHappen, err)
```
djdongjin, line 128: why is this a `ErrShouldNeverHappen` error?

pkg/containerutil/cp_linux.go
```diff
@@ -237,97 +277,74 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
+			return ErrTargetIsReadOnly
+		}
+
+		// Other errors, just put them back on stderr
+		_, fpErr := fmt.Fprint(os.Stderr, tarErr.String())
+		if fpErr != nil {
+			return errors.Join(fpErr, err)
+		}
+
+		return errors.Join(fmt.Errorf("failed to execute %v", tarXCmd.Args), err)
```
djdongjin, line 300: we can just use `fmt.Errorf("failed to execute %v: %w", tarXCmd.Args, err)`

pkg/containerutil/cp_linux.go
```diff
@@ -237,97 +277,74 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
+
 	tempDir, err := os.MkdirTemp("", "nerdctl-cp-")
 	if err != nil {
 		return "", nil, err
 	}
+
 	err = mount.All(resp, tempDir)
 	if err != nil {
-		return "", nil, fmt.Errorf("failed to mount snapshot with error %s", err.Error())
+		return "", nil, errors.Join(errors.New("failed to mount snapshot"), err)
```
djdongjin, line 338: `fmt.Errorf("failed to mount snapshot: %w", err)`

pkg/testutil/testutil.go
apostasie: > what's the behavior if we don't set `Dir` (which means both b.Dir and icmdCmd.Dir are "")?

I believe we then execute in whatever is `os.Getwd`.

pkg/cmd/container/cp_linux.go
apostasie: Current logic (count < 1) would treat identically two unrelated conditions:
- count == 0, which means we found no container
- count == -1, which means we errored 

https://github.com/containerd/nerdctl/blob/main/pkg/idutil/containerwalker/containerwalker.go#L47-L74

pkg/testutil/testutil.go
apostasie: https://go.dev/play/p/x_-0-uNtVd-

pkg/containerutil/cp_linux.go
apostasie: I think `errors.Join` is better.

With `fmt.Errorf`:
- we get a variety of format accross the codebase ("failed: %w", "failed (%w)", "failed with %w", "%w (hint: foo bla)")
- because of ^, things are less readable, as errors are clumped together with little indication which comes first

pkg/containerutil/cp_linux.go
apostasie: The only situation I can think of where you would get an error here is if the container that was still live at the time we called `CopyFiles` has since been deleted.
As mentioned above, while this is not impossible, it is incredibly unlikely.

You are right though that it is 0 probability, and I can wrap that in another condition instead.

pkg/containerutil/cp_linux.go
apostasie: (same as above)

apostasie:
Thanks a lot for the ongoing review @djdongjin !
Let me know your thoughts in the individual comments - will look again tomorrow.

apostasie:
Rebased.
Pending CI.

apostasie:
Obviously, the move to sub-packages is further exposing problems in our tests (chiefly because the parallel pool is greatly reduced - split in 15 buckets - hence the chance of leftovers is dramatically increased).

The introduction of a lot more / longer cp tests with this PR is likely tripping unrelated tests - and there seem to be something wrong with encrypt tests (starting with the fact that they are pruning), and even for the docker compat suite (with prune again).

That doesn't change things for this PR fundamentally, as the problem is not the code in here (which is very limited in its impact) - but obviously I'll send a couple of other PRs first to fix some of our tests issues so this here gets green.

apostasie:
Pending #3402

apostasie:
Reverting to draft until I figure out what's going on with testing.

apostasie:
Will wait for new tooling framework to land to update tests and get some clarity on what's going on.

apostasie:
As soon as #3492 is in, will rewrite tests with the new framework and get this in shape.

apostasie:
@djdongjin I paused this (and other feature / refactor work) 3 months ago because the CI failed randomly, as testing became unmanageable at that point, and I embarked on the new test framework and tests clean-up overall instead.

Rebasing this on top of all of that then, CI is now green.

I will do another final proof-read of this PR before turning it to "ready for review" - but otherwise, I am still interested in getting this in as it will fix a number of issues with cp, so, would love your updated thoughts on this if/when you have bandwidth for that.

cmd/nerdctl/container/container_cp_linux.go
```diff
@@ -115,16 +115,16 @@ func processCpOptions(cmd *cobra.Command, args []string) (types.ContainerCpOptio
 	}
 
 	container2host := srcSpec.Container != nil
-	var container string
+	var containerReq string
```
apostasie, line 118: Do not shadow package name.

pkg/containerutil/cp_linux.go
apostasie: Changing these to ErrContainerVanished

pkg/containerutil/cp_linux.go
apostasie: Removing this one as it is wrapped in the caller.

pkg/cmd/container/cp_linux.go
```diff
@@ -39,17 +40,22 @@ func Cp(ctx context.Context, client *containerd.Client, options types.ContainerC
...
+		if err == nil {
+			panic("nil error and count == -1 from ContainerWalker.Walk should never happen")
+		}
+		err = errors.Join(errors.New("unable to copy"), err)
```
fahedouch, line 52: `fmt.Errorf` is fine

pkg/cmd/container/cp_linux.go
```diff
@@ -39,17 +40,22 @@ func Cp(ctx context.Context, client *containerd.Client, options types.ContainerC
...
+	} else if count == 0 {
+		if err != nil {
+			err = errors.Join(errors.New("unable to retrieve containers with error"), err)
```
fahedouch, line 55: same behavior you should panic if `err != nil` with `count == 0`

pkg/containerutil/cp_resolve_linux.go
```diff
@@ -0,0 +1,444 @@
...
+	"slices"
+	"strings"
+	"syscall"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+
+	"github.com/containerd/containerd/v2/pkg/oci"
+)
+
+// volumeNameLen returns length of the leading volume name on Windows.
```
fahedouch, line 34: `Windows` ?

pkg/containerutil/cp_resolve_linux.go
```diff
@@ -0,0 +1,444 @@
...
+
+	path := originalPath
+
+	// Path may still be relative at this point. If it is, figure out getwd.
+	if !filepath.IsAbs(path) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return nil, errors.Join(errCannotResolvePathNoCwd, err)
+		}
+		path = cwd + string(os.PathSeparator) + path
```
fahedouch, line 77: `filepath.Join(cwd, path)`

pkg/containerutil/cp_resolve_linux.go
apostasie: Why not? One day we might support it there.

pkg/containerutil/cp_resolve_linux.go
apostasie: I do not fully remember the detailed reason of why in that case, but I am pretty sure it would not work. filepath.Join will clean the path, which we do not want in that case.

pkg/cmd/container/cp_linux.go
apostasie: Reading into this now. Thanks.

pkg/containerutil/cp_resolve_linux.go
apostasie: Re-reading. We absolutely do not want filepath.Join here. That would defeat the whole purpose of resolving the final path properly.

pkg/cmd/container/cp_linux.go
apostasie: I would say it is not, generally. It ends-up with wonky / un-uniform formatting across the codebase - making it hard for everybody where errors are coming from - inability to actually test against specific, typed-errors with errors.Is.
I am not going to fight for it here in this specific case (which is a bad hill to die on), but generally, I am strongly arguing that we should `Join` instead of `Errorf`.

pkg/containerutil/cp_resolve_linux.go
apostasie: filepath.Join will "clean" the path (eg: reduce relative path). The whole problem and point here is to perform *proper* resolution instead.

pkg/cmd/container/cp_linux.go
fahedouch: It is possible to error.Is by wrapping the error using the fmt pkg (we did some refacto in the past to switch nerdctl to fmt pkg)
The idea is:
-  to stick with current error formatting stragtegy.
- avoid import `errors` pkg

I think we can discuss the migration from fmt to errors.Join in a different PR

pkg/containerutil/cp_resolve_linux.go
fahedouch: Nvm, It was strange to see the word windows in a linux file

pkg/containerutil/cp_resolve_linux.go
fahedouch: IMO, we should reflect this intention in a comment.

pkg/containerutil/cp_resolve_linux.go
apostasie: Fair. Comment just below does clarify a bit.

pkg/cmd/container/cp_linux.go
apostasie: I don't think we should panic here.

pkg/cmd/container/cp_linux.go
apostasie: Sure, you can error.Is if you have a typed error to compare with. Anyhow, will not fight it in this case.
Will amend when I get my key.

pkg/containerutil/cp_resolve_linux.go
djdongjin: > filepath.Join will "clean" the path (eg: reduce relative path). The whole problem and point here is to perform proper resolution instead.

Why `clean` will be an issue here? do you have an example/test that can show the issue?

apostasie: Because it will clean-up the path purely lexically with no regards for the reality of the files.

Something like `/doesnotexist/foo/../../etc/passwd` clearly illustrates the issue.

It will, and must fail when copying or reading from it, as proper path resolution will fail for any actual filesystem operation (cp, read, write), but filepath.Join will "cleanup" to something that exist.

Conversely, with:
```
mkdir -p /tmp/target1/target2
ln -s /tmp/target1/target2 tmp/symlink
touch /tmp/target1/somefile
```

The path `/tmp/symlink/../somefile` will fail to resolve to an existing file using `filepath.Join`, while it SHOULD, and will, using any filesystem operation, since it is actually a valid file.

It is a common misunderstanding about lexical functions - they should not be used before resolving symlinks.

pkg/cmd/container/cp_linux.go
apostasie: Done.

apostasie:
Rebased.

Comments answered or addressed with latest push.

pkg/containerutil/cp_resolve_linux.go
fahedouch: > It is a common misunderstanding about lexical functions - they should not be used before resolving symlinks.

isn't safer to resolve the symlink before and then work with cleaned path ?

fahedouch: I see that you evaluate the symlink at this line https://github.com/containerd/nerdctl/pull/3323/files#diff-f8d3022dbe81c88660be3b5c26107695a00ef30d103f67558a218780e5da1dbaR81 so may be resolving the relative path just after it ?

apostasie: > > It is a common misunderstanding about lexical functions - they should not be used before resolving symlinks.
> 
> isn't safer to resolve the symlink before and then work with cleaned path ?

I am not completely sure what you mean - but here, the path may be relative, so in that case we do need to start resolving against cwd.

apostasie: > I see that you evaluate the symlink at this line https://github.com/containerd/nerdctl/pull/3323/files#diff-f8d3022dbe81c88660be3b5c26107695a00ef30d103f67558a218780e5da1dbaR81 so may be resolving the relative path just after it ?

We need to account for the fact that the path may not exist - which is handled a few lines below.
Again not completely sure what you mean here, but see previous comment ^ - we need to handle relative path cases (which is the reason we resolve links against cwd + path).

djdongjin: Hi @apostasie, I will take another look when I get some time. 

> So, point is: I am not interested in forcing anything here.
> We all have better things to do.
> If you folks are not comfortable with this PR, I would just accept that and close it.

I don't think folks imply you're forcing anything or not comfortable with your change. I think mostly it's because people (including me) may be slow at responding recently due to holidays approaching.

apostasie: All fine @djdongjin 
I want(ed) to make sure that people feel comfortable with a “no”
I am a simple and straightforward person 😁

Enjoy the holiday folks!

AkihiroSuda (approved):
Thanks

PR #3324, opened 2024-08-18
fix: filter container state without option "-a"

xyz-li:
`docker ps -f status=exited` will filter and output exited containers.

### DIFF
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -460,7 +460,7 @@ Flags:
   - :nerd_face: `--format=json`: Alias of `--format='{{json .}}'`
 - :whale: `-n, --last`: Show n last created containers (includes all states)
 - :whale: `-l, --latest`: Show the latest created container (includes all states)
-- :whale: `-f, --filter`: Filter containers based on given conditions
+- :whale: `-f, --filter`: Filter containers based on given conditions. When specifying the condition 'status', it filters all containers
   - :whale: `--filter id=<value>`: Container's ID. Both full ID and
     truncated ID are supported
   - :whale: `--filter name=<value>`: Container's name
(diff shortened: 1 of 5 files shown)


fahedouch:
Hi @xyz-li,

thank you for your PR.

as mention in the [doc](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-blue_square-nerdctl-ps) : 

```
-a, --all: Show all containers (default shows just running)
```

either we modify the doc to say that `-f` will enable all containers by default or you just append `-a` flag. We prefer to stick to docker behavior

xyz-li:
According to the docker implementation, it just enables all containers when specifying the filter `status`.
https://github.com/moby/moby/blob/master/daemon/list.go#L244-L251

fahedouch:
> According to the docker implementation, it enables all containers when specifying the filter `status`. https://github.com/moby/moby/blob/master/daemon/list.go#L244-L251

let's reflect this in the [doc](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-blue_square-nerdctl-ps)

apostasie:
LGTM (pending documentation change)

docs/command-reference.md
```diff
@@ -449,7 +449,7 @@ Usage: `nerdctl ps [OPTIONS]`
 
 Flags:
 
-- :whale: `-a, --all`: Show all containers (default shows just running)
+- :whale: `-a, --all`: Show all containers (default shows just running). When specifying the option '-f' with 'status', it filters through all containers
```
fahedouch, line 452: I would rather change the `-f` flag line 463

`:whale: -f, --filter: Filter all containers based on given conditions`

fahedouch (approved):
Thanks

PR #3327, opened 2024-08-18
Add builder OCI layout build context

austinvazquez:
### Description
This change adds building images with OCI layout build contexts.

### Testing
- Added unit tests
  - Negative test cases
- Added integration test (Linux only)
  - Test requires build functionality which excludes Windows platform testing.

### Additional references
- https://docs.docker.com/reference/cli/docker/buildx/build/#source-oci-layout
- https://github.com/opencontainers/image-spec/blob/main/image-layout.md
- https://docs.docker.com/build/exporters/oci-docker/
- https://docs.docker.com/build/builders/drivers/docker-container/

### DIFF
--- a/cmd/nerdctl/builder_build_linux_test.go
+++ b/cmd/nerdctl/builder_build_linux_test.go
@@ -1,0 +1,84 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+)
+
+func TestBuildContextWithOCILayout(t *testing.T) {
+	testutil.RequiresBuild(t)
+	testutil.RegisterBuildCacheCleanup(t)
+
+	var dockerBuilderArgs []string
+	if testutil.IsDocker() {
+		// Default docker driver does not support OCI exporter.
+		// Reference: https://docs.docker.com/build/exporters/oci-docker/
+		builderName := testutil.SetupDockerContainerBuilder(t)
+		dockerBuilderArgs = []string{"buildx", "--builder", builderName}
+	}
+
+	base := testutil.NewBase(t)
+	imageName := testutil.Identifier(t)
+	ociLayout := "parent"
+	parentImageName := fmt.Sprintf("%s-%s", imageName, ociLayout)
+
+	teardown := func() {
+		base.Cmd("rmi", parentImageName, imageName).Run()
+	}
+	t.Cleanup(teardown)
+	teardown()
+
+	dockerfile := fmt.Sprintf(`FROM %s
+LABEL layer=oci-layout-parent
+CMD ["echo", "test-nerdctl-build-context-oci-layout-parent"]`, testutil.CommonImage)
+	buildCtx := createBuildContext(t, dockerfile)
+
+	tarPath := fmt.Sprintf("%s/%s.tar", buildCtx, ociLayout)
+
+	// Create OCI archive from parent image.
+	base.Cmd("build", buildCtx, "--tag", parentImageName).AssertOK()
+	base.Cmd("image", "save", "--output", tarPath, parentImageName).AssertOK()
+
+	// Unpack OCI archive into OCI layout directory.
+	ociLayoutDir := t.TempDir()
+	err := extractTarFile(ociLayoutDir, tarPath)
+	assert.NilError(t, err)
+
+	dockerfile = fmt.Sprintf(`FROM %s
+CMD ["echo", "test-nerdctl-build-context-oci-layout"]`, ociLayout)
+	buildCtx = createBuildContext(t, dockerfile)
+
+	var buildArgs = []string{}
+	if testutil.IsDocker() {
+		buildArgs = dockerBuilderArgs
+	}
+
+	buildArgs = append(buildArgs, "build", buildCtx, fmt.Sprintf("--build-context=%s=oci-layout://%s", ociLayout, ociLayoutDir), "--tag", imageName)
+	if testutil.IsDocker() {
+		// Need to load the container image from the builder to be able to run it.
+		buildArgs = append(buildArgs, "--load")
+	}
+
+	base.Cmd(buildArgs...).AssertOK()
+	base.Cmd("run", "--rm", imageName).AssertOutContains("test-nerdctl-build-context-oci-layout")
+}
--- a/pkg/cmd/builder/build.go
+++ b/pkg/cmd/builder/build.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 
 	distributionref "github.com/distribution/reference"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/containerd/containerd/v2/core/images"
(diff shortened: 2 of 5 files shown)


cmd/nerdctl/builder_build_linux_test.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"gotest.tools/v3/assert"
+)
+
+func TestBuildContextWithOCILayout(t *testing.T) {
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 28: Please add a comment to explain this

cmd/nerdctl/builder_build_linux_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"gotest.tools/v3/assert"
+)
+
+func TestBuildContextWithOCILayout(t *testing.T) {
+	// Docker driver does not support OCI exporter.
```
AkihiroSuda, line 28: `docker buildx build` supports it though

austinvazquez: Thanks @AkihiroSuda, that helped me figure out the issue. I needed to setup a docker-container builder in order to be able to build the OCI tarball.

austinvazquez:
Looks like my test needs a little more bake time. Looking into it now.

austinvazquez:
Got it. Just forgot to tag the image before running. Test now passes on nerdctl and Docker.

cmd/nerdctl/builder_build_linux_test.go
```diff
@@ -0,0 +1,80 @@
...
+	if testutil.IsDocker() {
+		// Default docker driver does not support OCI exporter.
+		// Reference: https://docs.docker.com/build/exporters/oci-docker/
+		builderName := testutil.SetupDockerContainerBuilder(t)
+		dockerBuilderArgs = []string{"buildx", "--builder", builderName}
+	}
+
+	base := testutil.NewBase(t)
+	imageName := testutil.Identifier(t)
+	t.Cleanup(func() { base.Cmd("rmi", imageName).AssertOK() })
```
apostasie, line 41: Generally speaking, I would suggest you **also** run your cleanup before starting the test.

The reason is that `t.Cleanup` is not always called, and you may end-up in with a broken environment that fails test.

Pretty much

```
tearDown := func() { base.Cmd("rmi", imageName).AssertOK() }

t.Cleanup(tearDown)
tearDown()
```

In the case here

cmd/nerdctl/builder_build_linux_test.go
```diff
@@ -0,0 +1,80 @@
...
+	ociLayout := "parent"
+	tarPath := fmt.Sprintf("%s/%s.tar", buildCtx, ociLayout)
+
+	var buildArgs []string
+	if testutil.IsDocker() {
+		buildArgs = dockerBuilderArgs
+	}
+
+	buildArgs = append(buildArgs, "build", buildCtx, fmt.Sprintf("--output=type=oci,dest=%s", tarPath))
+	base.Cmd(buildArgs...).Run()
```
apostasie, line 57: `AssertOK()` instead of `Run`?

cmd/nerdctl/builder_build_linux_test.go
```diff
@@ -0,0 +1,80 @@
...
+		buildArgs = dockerBuilderArgs
+	}
+
+	buildArgs = append(buildArgs, "build", buildCtx, fmt.Sprintf("--build-context=%s=oci-layout://%s", ociLayout, ociLayoutDir), "-t", imageName)
+	if testutil.IsDocker() {
+		// Need to load the container image from the builder to be able to run it.
+		buildArgs = append(buildArgs, "--load")
+	}
+
+	base.Cmd(buildArgs...).Run()
```
apostasie, line 78: Same as above (`AssertOK`)?

pkg/cmd/builder/build.go
```diff
@@ -534,3 +545,61 @@ func parseContextNames(values []string) (map[string]string, error) {
 	}
 	return result, nil
 }
+
+var (
+	errOCILayoutPrefixNotFound = errors.New("OCI layout prefix not found")
```
apostasie, line 550: Generally, I think we should export these - to allow `errors.Is/As` testing.

pkg/cmd/builder/build.go
```diff
@@ -534,3 +545,61 @@ func parseContextNames(values []string) (map[string]string, error) {
...
+	}
+
+	return []string{
+		fmt.Sprintf("--oci-layout=parent-image-key=%s", abspath),
+		fmt.Sprintf("--opt=context:%s=oci-layout:parent-image-key@%s", name, digest),
+	}, nil
+}
+
+func readOCIIndexFromPath(path string) (*ocispec.Index, error) {
+	ociIndexJSONFile, err := os.Open(path + "/index.json")
```
apostasie, line 588: `filepath.Join` seems preferable.

cmd/nerdctl/builder_build_linux_test.go
apostasie: Also, it is fine to fail in the cleanup procedure. So, `Run` instead of `AssertOK` will do here.

cmd/nerdctl/builder_build_linux_test.go
```diff
@@ -0,0 +1,80 @@
...
+*/
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"gotest.tools/v3/assert"
```
apostasie, line 24: Normally, we enforce import ordering and this should have failed here - unfortunately the check is broken right now (fix in #3334).

What I would suggest is that you call `make lint-fix-imports` locally before commit. That will ensure all your imports are ordered according to local style.

pkg/cmd/builder/build_test.go
```diff
@@ -187,3 +187,42 @@ func TestIsBuildPlatformDefault(t *testing.T) {
...
+		expectedArgs  []string
+		errorIsNil    bool
+		expectedErr   string
+	}{
+		{
+			name:          "PrefixNotFoundError",
+			ociLayoutName: "unit-test",
+			ociLayoutPath: "/tmp/oci-layout/",
+			expectedArgs:  []string{},
+			expectedErr:   "OCI layout prefix not found",
```
apostasie, line 205: Maybe `errOCILayoutPrefixNotFound.Error()` would be better here?

apostasie:
@austinvazquez left a few comments / nits, but otherwise looks fine.

austinvazquez:
Thanks @apostasie, taking a look now. :)

cmd/nerdctl/builder_build_linux_test.go
austinvazquez: Sure, sounds good. Should be addressed in latest push.

apostasie: Feel free to hit "resolve" on these comments to hide them (just to keep the progress/conversation more clear).

cmd/nerdctl/builder_build_linux_test.go
austinvazquez: I made some progress here. Client was returning error due to nil manifest list. The workaround I applied was to use image save to create a properly formatted OCI archive. The trade-off was I needed to import distribution/distribution to match against Docker schema2 manifest; however, I think this is fine as the import is small. Alternatively could consider copying the constant if we don't want to take the dependency.

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/coreos/go-iptables v0.7.0
 	github.com/coreos/go-systemd/v22 v22.5.0
 	github.com/cyphar/filepath-securejoin v0.3.1
+	github.com/distribution/distribution v2.8.3+incompatible
```
AkihiroSuda, line 37: This dependency should not be needed

pkg/cmd/builder/build.go
```diff
@@ -534,3 +546,61 @@ func parseContextNames(values []string) (map[string]string, error) {
...
+	}
+
+	ociIndex, err := readOCIIndexFromPath(abspath)
+	if err != nil {
+		return []string{}, err
+	}
+
+	var digest string
+	for _, manifest := range ociIndex.Manifests {
+		if manifest.MediaType == schema2.MediaTypeManifest || manifest.MediaType == ocispec.MediaTypeImageManifest {
```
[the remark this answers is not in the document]
austinvazquez: 🙌🏼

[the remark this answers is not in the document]
austinvazquez: Thanks @AkihiroSuda, pushed the change to use containerd images pkg. Even found a helper function to call in place of the logic we pushed before.

pendo324 (approved):

AkihiroSuda (approved):
Thanks

PR #3329, opened 2024-08-18
ParentMain/nsenter: do not fail if getwd errors

### LINKED ISSUE #3328: getwd: should we really hard fail?
### Description

Investigating https://github.com/rootless-containers/usernetes/pull/327

Current intuition is that the error comes from:
https://github.com/containerd/nerdctl/blob/main/pkg/rootlessutil/parent_linux.go#L87

Do we really need to hard-fail when we have no cwd?

Put otherwise: is it important that nsenter gets `-w` set no matter what (especially in the case where cwd does not exist), or can we just "not set" `-w` when getwd fails?

cc @AkihiroSuda 

```
mkdir foo
cd foo
rmdir ../foo
nerdctl container run -d --name foo debian sleep Inf
FATA[0000] getwd: no such file or directory
```

na

main

None

### Host information

_No response_

apostasie:
To fix #3328 

It is not clear to me if there would be side-effects to this.

nsenter specialists opinion highly welcome.

### DIFF
--- a/pkg/rootlessutil/parent_linux.go
+++ b/pkg/rootlessutil/parent_linux.go
@@ -84,11 +84,6 @@ func ParentMain(hostGatewayIP string) error {
 	detachNetNSMode := detachedNetNSPath != ""
 	log.L.Debugf("RootlessKit detach-netns mode: %v", detachNetNSMode)
 
-	wd, err := os.Getwd()
-	if err != nil {
-		return err
-	}
-
 	// FIXME: remove dependency on `nsenter` binary
 	arg0, err := exec.LookPath("nsenter")
 	if err != nil {
@@ -96,13 +91,24 @@ func ParentMain(hostGatewayIP string) error {
 	}
 	// args are compatible with both util-linux nsenter and busybox nsenter
 	args := []string{
-		"-r/",     // root dir (busybox nsenter wants this to be explicitly specified),
-		"-w" + wd, // work dir
-		"--preserve-credentials",
+		"-r/", // root dir (busybox nsenter wants this to be explicitly specified),
+	}
+
+	// Only append wd if we do have a working dir
+	// - https://github.com/rootless-containers/usernetes/pull/327
+	// - https://github.com/containerd/nerdctl/issues/3328
+	wd, err := os.Getwd()
+	if err != nil {
+		log.L.WithError(err).Warn("unable to determine working directory")
+	} else {
+		args = append(args, "-w"+wd)
+	}
+
+	args = append(args, "--preserve-credentials",
 		"-m", "-U",
 		"-t", strconv.Itoa(childPid),
 		"-F", // no fork
-	}
+	)
 	if !detachNetNSMode {
 		args = append(args, "-n")
 	}


apostasie:
Definitely not working. Reverting to draft.

pkg/rootlessutil/parent_linux.go
```diff
@@ -84,25 +84,29 @@ func ParentMain(hostGatewayIP string) error {
...
+		"-r/", // root dir (busybox nsenter wants this to be explicitly specified),
+	}
+
+	// Only append wd if we do have a working dir
+	// - https://github.com/rootless-containers/usernetes/pull/327
+	// - https://github.com/containerd/nerdctl/issues/3328
+	wd, err := os.Getwd()
+	if err == nil {
+		args = append(args, "-w"+wd)
+	}
```
AkihiroSuda, line 103: Non-nil err should be still warned?

pkg/rootlessutil/parent_linux.go
```diff
@@ -84,25 +84,29 @@ func ParentMain(hostGatewayIP string) error {
...
-	}
-
 	// FIXME: remove dependency on `nsenter` binary
 	arg0, err := exec.LookPath("nsenter")
 	if err != nil {
 		return err
 	}
 	// args are compatible with both util-linux nsenter and busybox nsenter
 	args := []string{
-		"-r/",     // root dir (busybox nsenter wants this to be explicitly specified),
```
AkihiroSuda, line 99: Was this removed?

pkg/rootlessutil/parent_linux.go
apostasie: Done.

apostasie:
Comment addressed.
CI is green.

pkg/rootlessutil/parent_linux.go
```diff
@@ -84,25 +84,31 @@ func ParentMain(hostGatewayIP string) error {
...
-		"--preserve-credentials",
+		"-r/", // root dir (busybox nsenter wants this to be explicitly specified),
+	}
+
+	// Only append wd if we do have a working dir
+	// - https://github.com/rootless-containers/usernetes/pull/327
+	// - https://github.com/containerd/nerdctl/issues/3328
+	wd, err := os.Getwd()
+	if err != nil {
+		log.L.Warn("unable to determine working directory")
```
djdongjin, line 102: can we print the err as well in the log?

AkihiroSuda: What log?

djdongjin: This part `.WithError(err)`. @apostasie just added in the latest revision.

djdongjin (commented):
LGTM, thanks. btw did you verify if this resolves the issue? i.e. the repro in https://github.com/containerd/nerdctl/issues/3328 works now:

```shell
mkdir foo
cd foo
rmdir ../foo
nerdctl container run -d --name foo debian sleep Inf
FATA[0000] getwd: no such file or directory
```

apostasie:
> LGTM, thanks. btw did you verify if this resolves the issue? i.e. the repro in #3328 works now:
> 
> ```shell
> mkdir foo
> cd foo
> rmdir ../foo
> nerdctl container run -d --name foo debian sleep Inf
> FATA[0000] getwd: no such file or directory
> ```

Yes, it does fix the issue.
I am just mildly concerned about the behavior of nsenter in such a condition, but then... this is (somewhat) abnormal to start with.

pkg/rootlessutil/parent_linux.go
```diff
@@ -84,25 +84,31 @@ func ParentMain(hostGatewayIP string) error {
...
+	}
+
+	// Only append wd if we do have a working dir
+	// - https://github.com/rootless-containers/usernetes/pull/327
+	// - https://github.com/containerd/nerdctl/issues/3328
+	wd, err := os.Getwd()
+	if err != nil {
+		log.L.WithError(err).Warn("unable to determine working directory")
+	} else {
+		args = append(args, "-w"+wd)
```
fahedouch, line 104: IMO we may even set the working directory after nsenter with `--wdns`

djdongjin (approved):
LGTM, thanks

pkg/rootlessutil/parent_linux.go
apostasie: To what path?

apostasie: ah - you mean, in case getwd does NOT error, right?

Honestly, I would rather focus efforts on re-exec-ing instead of tweaking nsenter.

AkihiroSuda (approved):
Thanks

PR #3330, opened 2024-08-18
Remove Docker incompat from image prune and build tests

austinvazquez:
### Description
This change is a followup to #3319 to enable image prune filter until integration tests on Docker.

`SOURCE_DATE_EPOCH` is available in buildx since v0.10 which should be available in Docker 23 and greater.

### Testing
- Enables `TestImagePruneFilterUntil` on Docker
- Enables `TestBuildSourceDateEpoch` on Docker

### DIFF
--- a/cmd/nerdctl/image_prune_test.go
+++ b/cmd/nerdctl/image_prune_test.go
@@ -102,10 +102,10 @@ func TestImagePruneFilterUntil(t *testing.T) {
 	testutil.RequiresBuild(t)
 	testutil.RegisterBuildCacheCleanup(t)
 
-	// Docker image's created timestamp is set based on base image creation time.
-	testutil.DockerIncompatible(t)
-
 	base := testutil.NewBase(t)
+	// For deterministically testing the filter, set the image's created timestamp to 2 hours in the past.
+	base.Env = append(base.Env, fmt.Sprintf("SOURCE_DATE_EPOCH=%d", time.Now().Add(-2*time.Hour).Unix()))
+
 	imageName := testutil.Identifier(t)
 	teardown := func() {
 		// Image should have been pruned; but cleanup on failure.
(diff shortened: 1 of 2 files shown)


apostasie:
+1

cmd/nerdctl/image_prune_test.go
```diff
@@ -102,10 +102,8 @@ func TestImagePruneFilterUntil(t *testing.T) {
 	testutil.RequiresBuild(t)
 	testutil.RegisterBuildCacheCleanup(t)
 
-	// Docker image's created timestamp is set based on base image creation time.
-	testutil.DockerIncompatible(t)
-
 	base := testutil.NewBase(t)
+	base.Env = append(base.Env, fmt.Sprintf("SOURCE_DATE_EPOCH=%d", time.Now().Add(-2*time.Hour).Unix()))
```
AkihiroSuda, line 106: Needs a comment line to explain this

austinvazquez:
I was hoping we could remove the sleep and deterministically test with the SOURCE_DATE_EPOCH; however, through testing it appears images built with containerd v1.6 are not respecting the SOURCE_DATE_EPOCH.

apostasie:
Looks like 1.6 will be supported until at least Feb 2025... https://containerd.io/releases/

I am not sure it is worth the effort, but we could specialize this test for different versions of containerd?

cmd/nerdctl/image_prune_test.go
```diff
@@ -102,12 +102,15 @@ func TestImagePruneFilterUntil(t *testing.T) {
...
-
 	base := testutil.NewBase(t)
+	// For deterministically testing the filter, set the image's created timestamp to 2 hours in the past.
+	base.Env = append(base.Env, fmt.Sprintf("SOURCE_DATE_EPOCH=%d", time.Now().Add(-2*time.Hour).Unix()))
+
 	imageName := testutil.Identifier(t)
-	t.Cleanup(func() { base.Cmd("rmi", "--force", imageName) })
+	t.Cleanup(func() {
+		// Image should have been pruned; but cleanup on failure.
+		base.Cmd("rmi", "--force", imageName).AssertOK()
```
apostasie, line 112: Run here, instead of AssertOK, is preferable.
Cleanup is not meant to test anything - nor fail.

Also, generally it is good hygiene to run the cleanup routine before the test as well - in case there is any leftovers from a previous borked run.

PR #3340, opened 2024-08-21
Fix docker hub lag on golang versions breaking canary

### LINKED ISSUE #3306: Canary broken: follow-up to #3224
### Description

As described in #3224 docker hub has a delay before go images for latest releases show-up.
The patchy-patch in #3224 was just an ad-hoc solution to fix it for rc delays, but obviously this is breaking again.

go 1.23.0 is just out, but not yet available on Hub.

**Bottom-line: today, and until Hub has the latest go image, let's ignore "canary" failures.**
@AkihiroSuda ^

I will come up with a better solution shortly.

na

na

na

None

### Host information

_No response_

apostasie:
Fix #3306

### DIFF
--- a/hack/build-integration-canary.sh
+++ b/hack/build-integration-canary.sh
@@ -293,40 +293,58 @@ canary::build::integration(){
     docker_args+=(--build-arg "${shortsafename}_VERSION=$higher_readable")
   done
 
-
-  GO_VERSION="$(curl -fsSL "https://go.dev/dl/?mode=json&include=all" | jq -rc .[0].version)"
-  GO_VERSION="${GO_VERSION##*go}"
-  # If a release candidate, docker hub may not have the corresponding image yet.
-  # So, soften the version to just "rc", as they provide that as an alias to the latest available rc on their side
-  # See https://github.com/containerd/nerdctl/issues/3223
-  ! grep -Eq "rc[0-9]+$" <<<"$GO_VERSION" || GO_VERSION="${GO_VERSION%rc[0-9]*}-rc"
-  docker_args+=(--build-arg "GO_VERSION=$GO_VERSION")
+  hub_available_go_version="$(canary::golang::hublatest)"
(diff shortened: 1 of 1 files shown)


hack/build-integration-canary.sh
```diff
@@ -293,40 +293,55 @@ canary::build::integration(){
...
+canary::golang::hublatest(){
+  local hub_tags
+  local go_version
+  local available_version=""
+  local index
+
+  hub_tags="$(http::get /dev/stdout "https://registry-1.docker.io/v2/library/golang/tags/list" -H "Authorization: Bearer $(http::get /dev/stdout "https://auth.docker.io/token?service=registry.docker.io&scope=repository%3Alibrary%2Fgolang%3Apull" | jq  -rc .access_token)")"
+
+  index=0
+  while [ "$available_version" == "" ] && [ "$index" -lt 5 ]; do
```
ktock, line 315: nit: why `5`? Don't we need to handle cases of `available_version` isn't found?

apostasie: Limiting the number of tries to 5.
If docker hub does not have any of the past five versions, all bets are off.
Should hard error in that case though.
Will do.

apostasie: Taking that back re-reading the code. It is fine to fail to find a version. In that case, the dockerfile will just fallback to the hard-coded version.

apostasie: Latest push has the tweak.
Thanks @ktock 
Appreciate the sharp review.

apostasie: And: hard error is not necessary (hub is notorious for random failures). The new check will allow the dockerfile to still build with the hardcoded version instead.

ktock (approved):

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@ktock You may merge PRs

PR #3342, opened 2024-08-22
Rename rollback on error

apostasie:
This comes from #3341 

Failure to rename will leave the system in a broken state.
This PR implements a rollback mechanism to revert whatever we can.

This is not a solution for the problems linked to #3341, but at least it will stop bricking systems (hopefully...).

Issue #355, opened 2021-09-12
Import /etc/hosts entries automatically


AkihiroSuda:
Nerdctl should import entries of /etc/hosts on the host by default (except loopback addresses)

AkihiroSuda:
Link: an opposite proposal for Podman https://github.com/containers/podman/issues/11835

apostasie:
Interesting.

Maybe just for `--network=host` though? (that is docker behavior)

Issue #2993, opened 2024-05-12
nerdctl is leaking host files when a container fails to start


apostasie:
### Description

While investigating #2992, I noticed that `AllocHostsFile` is called before we verify that the requested `--name` for the container to be started is actually available.

If the name is not free, we bail out, but we leave the allocated host file in place.

### Steps to reproduce the issue

```
# Just repeatedly run containers with the exact same name
sudo ./nerdctl-patch run -d --name whatevername debian bash
sudo ./nerdctl-patch run -d --name whatevername debian bash
sudo ./nerdctl-patch run -d --name whatevername debian bash
```

### Describe the results you received and expected

Now, /var/lib/nerdctl/XXXX/etchosts/default will start filling up with orphaned files.

This is also true for `/var/lib/nerdctl/1935db59/containers/default`.

Looking at `create.go` shows why.

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Seems to me like the general issue here is that we just do not clean-up anything if the container fails to create.
Solving this is likely more involved than just shuffling things around.

apostasie:
Hey @haytok 

Yeah, I would say, just go for it and let's discuss details on your PR.

Generally, I would love to see a solution that takes care of all the places where we error in that function - not just this specific location.

apostasie:
> Therefore, it is assumed that this issue can be solved by specifying a function in the second return value that removes the following.

Be sure to use existing abstractions to do the cleanup (`hostsstore.DeallocHostsFile`) and do _not_ manipulate the filesystem directly.
The point is: we need to make sure this stays safe wrt concurrency.

haytok:
Thanks for comments !!!

> Generally, I would love to see a solution that takes care of all the places where we error in that function - not just this specific location.

Sure, the second return value of most of this function is `nil` and does not return a `gc`, which should also be corrected.

apostasie:
Fixed by #3457

Issue #1829, opened 2023-01-11
meta.json is not removed when the container process exits.


fahedouch:
### Description

container `Poststop` hook is not working as expected: `meta.json` is not removed when the container process exits.

### Steps to reproduce the issue

1. nerdctl run alpine
2. `/var/lib/nerdctl/<ADDRHASH>/etchosts/<NS>/<ID>/meta.json` is still present

### Describe the results you received and expected

`/var/lib/nerdctl/<ADDRHASH>/etchosts/<NS>/<ID>/meta.json` should be removed when container with id  `ID` exists

### What version of nerdctl are you using?

latest

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

fahedouch:
container `Poststop` is executed in the Runtime Namespace and not the container ns

apostasie:
Hey folks.

Wondering if we do really want this.
`meta.json` will get deleted when the container gets _removed_ (postStop).

Is removing at `stop` really bring anything valuable to the table?

apostasie:
@fahedouch ^

apostasie:
@AkihiroSuda @fahedouch shall we close this?
I do not think we should remove it when the container stops. It is removed when the container is removed though.

AkihiroSuda:
We may remove it when the container exits, but it’s not a critical requirement

apostasie:
> We may remove it when the container exits, but it’s not a critical requirement

I do not think it is feasible, as we do not get ocihooks events for container stop/start.
Generally speaking, we cannot ensure we do things when containers stop outside of us explicitly calling `stop`.

PR #3344, opened 2024-08-22
Fix host networking issues and rename incidental issue

### LINKED ISSUE #2591: add-host parameter not valid
### Description

nerdctl run -d --add-host=node01:172.16.148.76 --add-host=node02:172.16.148.77 --add-host=node03:172.16.148.78 --add-host=node04:172.16.148.79 XXX
image
add-host not added in container /etc/hosts
same add-host param run container , first container in nerdctl config file have it. but  container same not . all nerdctl config(meta.json) right. 

1. run a contianer by nerdctl  width -d(--detach)  
2.check nerdctl config file(hosts,meta.json)
3.in contianer, check /etc/hosts

add-host params write to container /etc/hosts

contianerd-1.6.24 nerdctl 1.6.2

Others

### Host information

centos7,

### LINKED ISSUE #2613: Parameter --add-host not working when using also --net=host
### Description

When I run container using `nerdctl run ...` the parameter `--add-host` is not working when I also use `--net host` parametr. This seems to be strange because the same is working well on docker.

1. run `nerdctl run --rm  --add-host=dakfgashfaohsfashfoasfno:2.1.3.7 --net=host ubuntu getent hosts dakfgashfaohsfashfoasfno`
2. run `nerdctl run --rm  --add-host=dakfgashfaohsfashfoasfno:2.1.3.7 ubuntu getent hosts dakfgashfaohsfashfoasfno`
3. compare results

First command is giving no output, second one gives an entry defined by `--add-host` parameter. I expect that both commands gives the same results.

1.6.2 installed in rootles mode, but I discoverd it on one of my hosts with version 1.4.0.

None

### Host information

Virtual machine on OpenStack cloud, running on Ubuntu-focal.

### LINKED ISSUE #2694: hostname is always overwritten by container id in container when using --network=host
### Description

Restart the container every time , `hostname` is always overwritten by container id in container when using --network=host，but the hostname in the container will change at any time, after restarting the container it changes to the container id AGAIN! 

1.`nerdctl run --network=host -v /etc/resolv.conf:/etc/resolv.conf -v /etc/hosts:/etc/hosts -v /etc/hostname:/etc/hostname --privileged=true images:v1.0.0 --name=test `  
2.`nerdctl exec -it test bash`
3.`hostname` : it shows container id `a7e60e651786`
4.`echo 'myhost' > /etc/hostname && hostname 'myhost'`
5.restart the container 
6.`nerdctl exec -it test bash`
7.`hostname`: it shows container id `a7e60e651786`  AGAIN!  not named 'myhost'

hostname :  it shows `myhost` 

1.7.0

None

### Host information

_No response_

### LINKED ISSUE #3341: nerdctl rename broken (with `--network host`, `none`)
### Description

```
nerdctl run --network host -d --name test1 debian sleep Inf
nerdctl rename test1 test2
```

```
FATA[0000] open /home/dmp.linux/.local/share/nerdctl/1935db59/etchosts/default/bda418243819bf46ac0f8a52f88310ba6ed77e4038febda296df158f65295889/meta.json: no such file or directory
```

Futhermore, this will leave the system in a broken state, where the destination name is locked.

This is likely related to #2613 which I was debugging before hitting this.

na

na

main

None

### Host information

_No response_

### LINKED ISSUE #3343: Message too big using internal logging cause the process to block using stdout/stderr
### Description

When the container's process use the stdout/stderr to log a message too big, the process hang without crashing and for eternity.

I have created a small container to debug this but originally was happening to a game server started with nerdctl.

1. Start a small Python Flask server `nerdctl run -d -p 5555:5555 cerberusqc/messager:1.2`
2. Navigate to http://localhost:5555
3. Enter 2-3 time 100000 bytes messages and the process hang completely

I expect the logs to be able to process any size of message passed to the stdout/stderr without blocking my main process.

I think that I have identified that the problem was probably coming from `pkg/logging/logging.go`
```go
        stdout := make(chan string, 10000)
	stderr := make(chan string, 10000)
	processLogFunc := func(reader io.Reader, dataChan chan string) {
		defer wg.Done()
		defer close(dataChan)
		scanner := bufio.NewScanner(reader)
		for scanner.Scan() {
			if scanner.Err() != nil {
				log.L.Errorf("failed to read log: %v", scanner.Err())
				return
			}
			dataChan <- scanner.Text()
		}
	}
```

I did a quick test and replacing by a dynamic buffer just to confirm
```go
	stdout := make(chan string)
	stderr := make(chan string)
	processLogFunc := func(reader io.Reader, dataChan chan string) {
		defer wg.Done()
		defer close(dataChan)

		buf := make([]byte, 0, 4096)  // A dynamic buffer
		byteBuf := make([]byte, 1024) // Read in chunks

		for {
			n, err := reader.Read(byteBuf)
			if n > 0 {
				// Append the new data to the buffer
				buf = append(buf, byteBuf[:n]...)

				// Process complete lines in the buffer
				for {
					index := bytes.IndexByte(buf, '\n')
					if index == -1 {
						// No complete line found, exit the inner loop
						break
					}

					// Extract the complete line (including the newline)
					line := buf[:index+1]

					// Send the line to the logger and channel
					dataChan <- string(line)

					// Remove the processed line from the buffer
					buf = buf[index+1:]
				}
			}

			if err != nil {
				if err != io.EOF {
					log.L.Errorf("Failed to read log: %v", err)
				}
				break // Exit the loop on error or EOF
			}
		}

		// Process any remaining data in the buffer
		if len(buf) > 0 {
			remaining := string(buf)
			dataChan <- remaining
		}
	}
```
Looks like it is now processing any size of logs without blocking the main process but I definitely don't have the experience in GO to know if this the real reason.

1.7.6

None

### Host information

Linode VM Dedicated 16 GB
8 vCPU
16 GB 
Ubuntu 22.04 LTS

CONTAINERD_VERSION="1.7.19"
RUNC_VERSION="1.1.13"
CNI_VERSION="1.5.1"

### LINKED ISSUE #3345: Container create panic (index out of range)
### Description

Title says all

```
nerdctl --namespace=nerdctl-test run --rm --add-host foo=1.2.3.4 ghcr.io/stargz-containers/alpine:3.13-org getent hosts foo
```

Not crash.

Argument is just malformed.

main

None

### Host information

_No response_

apostasie:
Purported to:
- fix #3345
- fix #3341
- fix #2591 
- fix #2613 
- (for the relevant part of) fix #2694 

Does NOT fix yet #355

The take-away are:
1. the modification to make network=host use the host hostname as a default is straightforward
2. host networking previously did not call `Acquire` on the hosts store, causing `Update` to fail during a rename operation. Note that the same is true for `none`. Rename has been modified to not hard fail anymore on hostsstore Update failures. Another PR (#3342) is also making rename able to rollback in case of failure
3. for the same reason as above, extraHosts were not added - this has now been done in SetupNetworking and CleanupNetworking for the hostNetworkManager

Note that calls to CleanupNetworking and SetupNetworking were previously gated by a platform check to apply only on Windows (although not everywhere).
Now, this does not seem to serve much of a purpose, as these methods did nothing for any of the networkManager (with the exception of the Windows implementation of CNI).
Is this some kind of unfinished refactor? 

Anyhow, I did remove the platform checks altogether, since the code decides already if something should happen.

### DIFF
--- a/cmd/nerdctl/container_run.go
+++ b/cmd/nerdctl/container_run.go
@@ -377,13 +377,8 @@ func runAction(cmd *cobra.Command, args []string) error {
 	id := c.ID()
 	if createOpt.Rm && !createOpt.Detach {
 		defer func() {
-			// NOTE: OCI hooks (which are used for CNI network setup/teardown on Linux)
-			// are not currently supported on Windows, so we must explicitly call
-			// network setup/cleanup from the main nerdctl executable.
-			if runtime.GOOS == "windows" {
-				if err := netManager.CleanupNetworking(ctx, c); err != nil {
-					log.L.Warnf("failed to clean up container networking: %s", err)
-				}
+			if err := netManager.CleanupNetworking(ctx, c); err != nil {
+				log.L.Warnf("failed to clean up container networking: %s", err)
 			}
 			if err := container.RemoveContainer(ctx, c, createOpt.GOptions, true, true, client); err != nil {
 				log.L.WithError(err).Warnf("failed to remove container %s", id)
--- a/cmd/nerdctl/container_run_network_linux_test.go
+++ b/cmd/nerdctl/container_run_network_linux_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"os"
 	"regexp"
 	"runtime"
 	"strings"
@@ -421,6 +422,24 @@ func TestRunDNS(t *testing.T) {
 	cmd.AssertOutContains("nameserver 8.8.8.8\n")
 	cmd.AssertOutContains("search test\n")
 	cmd.AssertOutContains("options attempts:10\n")
+}
+
+func TestRunNetworkHostHostname(t *testing.T) {
+	base := testutil.NewBase(t)
+
+	hostname, err := os.Hostname()
+	assert.NilError(t, err)
+	hostname = hostname + "\n"
+	base.Cmd("run", "--rm", "--network", "host", testutil.CommonImage, "hostname").AssertOutExactly(hostname)
+	base.Cmd("run", "--rm", "--network", "host", testutil.CommonImage, "sh", "-euxc", "echo $HOSTNAME").AssertOutExactly(hostname)
+	base.Cmd("run", "--rm", "--network", "host", "--hostname", "override", testutil.CommonImage, "hostname").AssertOutExactly("override\n")
+	base.Cmd("run", "--rm", "--network", "host", "--hostname", "override", testutil.CommonImage, "sh", "-euxc", "echo $HOSTNAME").AssertOutExactly("override\n")
+}
+
+func TestRunNetworkHost2613(t *testing.T) {
+	base := testutil.NewBase(t)
+
+	base.Cmd("run", "--rm", "--add-host", "foo:1.2.3.4", testutil.CommonImage, "getent", "hosts", "foo").AssertOutExactly("1.2.3.4           foo  foo\n")
 }
 
 func TestSharedNetworkStack(t *testing.T) {
--- a/pkg/cmd/container/create.go
+++ b/pkg/cmd/container/create.go
@@ -290,7 +290,7 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
 		parts := strings.SplitN(host, ":", 2)
 		// If the IP Address is a string called "host-gateway", replace this value with the IP address stored
 		// in the daemon level HostGateway IP config variable.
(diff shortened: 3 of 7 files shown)


apostasie:
Note that we will not clean-up meta.json if the container exits on its own.

This is similar to #2993 and #1829 and belongs to a more general class of problem with our network / container lifecycle management (be it CNI or not).

@fahedouch proposal on containerd never went anywhere unfortunately (https://github.com/containerd/containerd/issues/8163).

I think we need to re-evaluate our strategy wrt lifecycle cleanup.

AkihiroSuda (approved):
Thanks

PR #3356, opened 2024-08-23
Restart fixes

### LINKED ISSUE #3353: Calling stop on a created container should not error with "container not found"
### Description

Breakout of #3352

```
nerdctl create --name foo debian sleep Inf
nerdctl stop foo
docker create --name foo debian sleep Inf
docker stop foo
```

na

main

None

### Host information

_No response_

### LINKED ISSUE #3354: After a containerd restart, container fail re-acquiring their own name
### Description

Since the name never got released (as the container postStop hook does not get called when containerd stops), trying to restart the container will currently fail as unable to re-acquire the name.

Breakout of #3352 

_No response_

na

main

None

### Host information

_No response_

### LINKED ISSUE #3355: CNI bridge plugin broken on containerd restart for existing containers
### Description

This is a breakout of #3352 

```
nerdctl rm -f foo
nerdctl run --net bridge -d --name foo debian sleep Inf
systemctl --user stop containerd
systemctl --user start containerd
nerdctl start foo
```

```
failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: time="2024-08-23T14:07:01-07:00" level=fatal msg="failed to call cni.Setup: plugin type=\"bridge\" failed (add): failed to allocate for range 0: 10.4.0.237 has been allocated to default-0b6b3a64bf4039828fc9443354f39009befe0240c6d0d458cefe3685ac1996e8, duplicate allocation is not allowed": unknown
```

The bridge plugin refuses to allocate an ip to this container id, since there is already one allocated...

I do believe the plugin should INSTEAD return the existing already allocated ip, but I am out of my depth here, and would welcome opinion from people more savvy with CNI networking.

na

na

main

None

### Host information

_No response_

apostasie:
- fix #3353 (old issue affecting 1.7 as well)
- fix #3354 (regression compared to 1.7)
- fix #3355 (from the looks of it, this has probably been there for a while)

### DIFF
--- a/pkg/containerutil/containerutil.go
+++ b/pkg/containerutil/containerutil.go
@@ -37,6 +37,7 @@ import (
 	"github.com/containerd/containerd/v2/core/runtime/restart"
 	"github.com/containerd/containerd/v2/pkg/cio"
 	"github.com/containerd/containerd/v2/pkg/oci"
+	"github.com/containerd/errdefs"
 	"github.com/containerd/log"
 
 	"github.com/containerd/nerdctl/v2/pkg/consoleutil"
@@ -371,6 +372,13 @@ func Stop(ctx context.Context, container containerd.Container, timeout *time.Dur
 
 	task, err := container.Task(ctx, cio.Load)
 	if err != nil {
+		// NOTE: NotFound doesn't mean that container hasn't started.
+		// In docker/CRI-containerd plugin, the task will be deleted
+		// when it exits. So, the status will be "created" for this
+		// case.
+		if errdefs.IsNotFound(err) {
+			return nil
+		}
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/containerutil/containerutil.go
```diff
@@ -371,6 +372,13 @@ func Stop(ctx context.Context, container containerd.Container, timeout *time.Dur
 
 	task, err := container.Task(ctx, cio.Load)
 	if err != nil {
+		// NOTE: NotFound doesn't mean that container hasn't started.
+		// In docker/CRI-containerd plugin, the task will be deleted
+		// when it exits. So, the status will be "created" for this
+		// case.
+		if errdefs.IsNotFound(err) {
```
apostasie, line 379: Note that we possibly have similar problems in other places in our codebase, where we assume that a container does not exist because we fail retrieving the task.

AkihiroSuda (approved):
Thanks

Issue #3350, opened 2024-08-23
[v2 regression] `nerdctl start` fails after restarting the host (`msg="failed to lock state dir: unable to unmarshall lifecycle data: unexpected end of JSON input"`)


AkihiroSuda:
### Description

`nerdctl start` fails after restarting the host

### Steps to reproduce the issue

1. `sudo nerdctl run -d --name foo busybox sleep infinity`
2. Reboot the host
3.
```console
$ sudo nerdctl start foo
FATA[0000] 1 errors:
failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running createRuntime hook #0: exit status 1, stdout: , stderr: time="2024-08-23T14:07:24+09:00" level=fatal msg="failed to lock state dir: unable to unmarshall lifecycle data: unexpected end of JSON input": unknown
```

### Describe the results you received and expected

Received: error
Expected: starts

### What version of nerdctl are you using?

```console
$ sudo nerdctl version
Client:
 Version:	v2.0.0-rc.1
 OS/Arch:	linux/amd64
 Git commit:	778975fcaa57e365ee44dbad4a9d8d63180ae320
 buildctl:
  Version:	v0.0.0+unknown

Server:
 containerd:
  Version:	v2.0.0-rc.3
  GitCommit:	27de5fea738a38345aa1ac7569032261a6b1e562
 runc:
  Version:	1.2.0-rc.2
  GitCommit:	v1.2.0-rc.2-0-gf2d2ee5e
```

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

```console
$ sudo nerdctl info
Client:
 Namespace:	default
 Debug Mode:	false

Server:
 Server Version: v2.0.0-rc.3
 Storage Driver: overlayfs
 Logging Driver: json-file
  Cgroup Driver:  : systemd
  Cgroup Version: : 2
 Plugins:
  Log:     fluentd journald json-file syslog
  Storage: native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile:	builtin
  cgroupns
 Kernel Version:   6.8.0-40-generic
 Operating System: Ubuntu 24.04 LTS
 OSType:           linux
 Architecture:     x86_64
 CPUs:             4
 Total Memory:     15.57GiB
 Name:             suda-ws01
 ID:               5097992c-2934-44bd-a1cb-5ac20cf2013f
```

apostasie:
[...]

PR #3362, opened 2024-08-24
Hardening lifecycle-state-store, name-store, and oci-hooks

### LINKED ISSUE #3351: ` name "foo" is already used by ID ""`
Looks like the name store may accidentally get stuck in an improper state
```console
$ sudo nerdctl run -it --rm --name=foo alpine
FATA[0000] name "foo" is already used by ID ""

$ sudo nerdctl rm -f foo
ERRO[0000] 1 errors:
no such container: foo

$ sudo nerdctl run -it --rm --name=foo alpine 
FATA[0000] name "foo" is already used by ID ""
```

Probably a recent regression in the main branch.

Workaround: `sudo rm /var/lib/nerdctl/1935db59/names/default/foo`

### LINKED ISSUE #3357: When run fails on (bridge) networking setup, it breaks the namestore, leading to duplicate names
### Description

Yet another variant / breakout of #3352 

You need to edit onCteateRuntime so that it panics.

```
nerdctl rm -f $(nerdctl ps -aq)
nerdctl run --net bridge -d --name foo debian sleep Inf
systemctl --user stop containerd
systemctl --user start containerd
sleep 1
echo "restart foo"
nerdctl start foo
sleep 1
echo "run bar"
nerdctl run -d --name bar debian sleep Inf
nerdctl run -d --name bar debian sleep Inf
echo "result in duplication"
nerdctl ps -a
```

na

main

None

### Host information

_No response_

apostasie:
**First commit**

Although #3350 got closed as a fluke, the error message seen in the logs clearly shows that the lifecycle state file could end-up messed-up, triggering a failure.

This does two things:
- write atomically, so that we never end-up with half-written / hosed JSON in the file
- tolerate load errors: this file is not critical and purely informative and should soft fail with a warning instead

This will make the LifecycleState more robust and resistant to host shutdowns.

Note: we should consider a review of all our filesystem based stores and consider moving to atomic writing for them.

**Second commit**

Fix #3351 by making us resistant to the condition and allowing a container to acquire a "broken" name.

This obviously should never happen, but it does ^.

**Third commit**

Fix #3357 - does not fix the underlying issue, which is possibly upstream - but effectively mitigate it.
Since onPostStop is (wrongly?) called after an error in onCreateRuntime - while the container will NOT be deleted - we inhibit the normal cleanup flow in that case.

### DIFF
--- a/pkg/ocihook/state/state.go
+++ b/pkg/ocihook/state/state.go
@@ -24,6 +24,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/containerd/log"
+
 	"github.com/containerd/nerdctl/v2/pkg/lockutil"
 )
 
@@ -71,6 +73,8 @@ func (lf *LifecycleState) Load() error {
 	} else {
 		err = json.Unmarshal(data, lf)
 		if err != nil {
+			// Logging an error, as Load errors are generally ignored downstream
+			log.L.Error("unable to unmarshall lifecycle data")
 			return fmt.Errorf("unable to unmarshall lifecycle data: %w", err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


apostasie:
Failure is the usual  IPFS/compose.

pkg/ocihook/state/state.go
```diff
@@ -71,18 +74,26 @@ func (lf *LifecycleState) Load() error {
 	} else {
 		err = json.Unmarshal(data, lf)
 		if err != nil {
+			backout := filepath.Join(lf.stateDir, lifecycleFile, ".bak")
```
AkihiroSuda, line 77: Maybe
```suggestion
			backout := filepath.Join(lf.stateDir, lifecycleFile + ".bak")
```
Also, where is this backout created?

apostasie: Looks like I dropped a stash while rewriting commits.
Amending now.

apostasie: Apologies for this.
Fixed.

apostasie:
Grumble grumble.
Interactive rebase...
Should be ready in a second.

apostasie:
@AkihiroSuda CI is green. Mistake addressed.

AkihiroSuda (approved):
Thanks

Issue #3325, opened 2024-08-18
Remove dependency on `go.uber.org/mock`


AkihiroSuda:
https://github.com/search?q=repo%3Acontainerd%2Fnerdctl+go.uber.org%2Fmock&type=code

The tests that are using `go.uber.org/mock` should be rewritten to use plain structs and interfaces for simplicity.

fahedouch:
we can use mockgen to generate mock interface [implem](https://github.com/uber-go/mock?tab=readme-ov-file#running-mockgen) and put then in seperate package. No need to redo the mockgen work.

apostasie:
> we can use mockgen to generate mock interface [implem](https://github.com/uber-go/mock?tab=readme-ov-file#running-mockgen) and put then in seperate package. No need to redo the mockgen work.

We would still depend on `go.uber.org/mock` (during test runs) though, right?

PR #3364, opened 2024-08-26
Add a generic abstraction for filesystem based stores

### LINKED ISSUE #3365: Containers across namespaces boundary show up in /etc/hosts
### Description

```
nerdctl run -d --namespace one --name deb1 debian sleep Inf
nerdctl run -d --namespace two --name deb2 debian sleep Inf
nerdctl exec --namespace two deb2 cat /etc/hosts
```

This can lead to situations where entries in /etc/hosts are duplicated (when containers in different namspaces have the same name).

Of course this also breaks the promise of some isolation between ns.

na

na

main

None

### Host information

_No response_

apostasie:
Although our codebase makes use of flock when we manipulate the filesystem (normally...) , we have had a relatively significant number of issues in the past with our various "file stores". They typically fall in the following categories:
- unprotected concurrent access to the same resource
- filesystem details leaking upward into higher level abstractions where they are manipulated unsafely
- lock that does not really lock
- lock that should lock over a larger section of code
- sequential (individually locked) operations being done assuming that data cannot change in-between them
- partial writes to the filesystem

Issues arising from these are generally heinous to debug, as they usually manifest in a flaky way (as they may only happen in high-concurrency situations / under pressure), or tend to bubble-up in different places making seemingly unrelated operations fail in cryptic ways.

Truth is, it is not 100% trivial to achieve a concurrency-safe and atomic filesystem storage system, even a very simple one like what we do - operating system crash or restarts, or generally interrupted writes, for example, are often overlooked as something that has to be thought about, to boot.

While we have fixed a number of these issues, there is clearly more lingering, and we are bound to produce even more in the future short of having a stronger basis.
I do not think it is fair to assume that everyone will be very mindful and intentional with their locking, or that the filesystem is a good enough, "safe" abstraction...

Furthermore, and even when done right, these things are inherently repetitive (mutex / filesystem lock / path sanitization / atomic write) and we now have enough of these "filestores" to justify having a shared base:
- namestore
- volumestore
- lifecycledata
- hostsstore

**This PR first commit does introduce an interface and a reference (filesystem) implementation** of a reusable `Store` component that every "store" can leverage and that provides:
- concurrency safety (thread safe and safe across concurrent invocations of the binary)
- atomic operations (write will never leave incomplete data in a file)
- path components validation, per-platform
- a small set of simple methods that should cover most use-cases
- completely sealed underlying implementation that does not leak
- proper, detailed errors
- some level of locking enforcement (eg: will error if there is no lock)
- tests

Structs that leverage it are thus freed from having to deal with mutex / filelocks, or underlying implementations (eg: filesystem). They can rely on a simple API that is responsible for just that, and can focus solely on the abstraction they have to provide for their specific data.

Hopefully, if things go our way, we should never see again a bug report about hosed json files / broken systems on restart, or flakyness involving stores...

**The second commit is a mere cleanup of the (useless) mock for the volume store**

**The third commit is a refactor of our various stores** to leverage that:
- all four stores have been reviewed and their syntax harmonized
- better erroring
- code duplication removal
- all explicit locking mechanisms have been removed (to simplify the user life) (with the exception of the VolumeStore) and all stores now expose methods tailored for their use cases that lock on their own
- all use cases have been reviewed to make sure nothing is racy (exceptions with Composer, obviously)
- ensuring locking is done at the namespace level, or lower

Incidentally, the following known issues are getting fixed or partially fixed:
- fix #3365
- part of #3325

Finally, I do appreciate this is a far reaching PR, as it is touching code in the critical path of basically everything.
To a large extent, though, this is mere refactoring and literally no novelty.


cmd/nerdctl/namespace.go
```diff
@@ -119,13 +118,10 @@ func namespaceLsAction(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			log.L.Warn(err)
 		} else {
-			volEnts, err := volStore.List(false)
```
apostasie, line 122: List() is expensive.
Replaced with the much lighter Count.

pkg/cmd/compose/compose.go
```diff
@@ -77,15 +77,8 @@ func New(client *containerd.Client, globalOptions types.GlobalCommandOptions, op
...
-		_, volGetErr := volStore.Get(volName, false)
-		if volGetErr == nil {
-			return true, nil
-		} else if errors.Is(volGetErr, errdefs.ErrNotFound) {
-			return false, nil
-		}
-		return false, volGetErr
-	}
+	// FIXME: this is racy. See note in up_volume.go
+	options.VolumeExists = volStore.Exists
```
apostasie, line 81: There is now a `Exists` method, much cheaper than `Get`

pkg/cmd/container/remove.go
```diff
@@ -128,18 +128,10 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions
 		return err
 	}
 	// Get volume store
-	volStore, err := volume.Store(globalOptions.Namespace, globalOptions.DataRoot, globalOptions.Address)
+	volStore, err := volumestore.New(dataStore, globalOptions.Namespace)
 	if err != nil {
 		return err
 	}
-	// Note: technically, it is not strictly necessary to acquire an exclusive lock on the volume store here.
```
apostasie, line 135: Removing the lock in that case (see comment).

pkg/cmd/volume/prune.go
```diff
@@ -33,60 +35,50 @@ func Prune(ctx context.Context, client *containerd.Client, options types.VolumeP
 	if err != nil {
 		return err
 	}
-	err = volStore.Lock()
```
apostasie, line 36: Replacing all of that with a `volStore.Prune(func)` that does locking.

pkg/identifiers/validate_windows.go
```diff
@@ -1,41 +0,0 @@
-/*
```
apostasie, line 1: Moved inside the store.
These things are restrictions linked to the filesystem implementation, and not conceptually to restrictions we necessarily want to apply to identifiers.
Put otherwise: they are orthogonal concerns.

pkg/mountutil/mountutil_test.go
```diff
@@ -0,0 +1,37 @@
+/*
```
apostasie, line 1: This replace the whole uber go mock thing for mountutil.

apostasie:
This is ready for review and CI is green (save the Kube flaky test).
This will need to be rebased atop of #3362 that presumably should get merged first.

pkg/store/filestore.go
```diff
@@ -0,0 +1,415 @@
...
+
+	entries := []string{}
+	for _, dirEntry := range dirEntries {
+		entries = append(entries, dirEntry.Name())
+	}
+
+	return entries, nil
+}
+
+/*
```
apostasie, line 240: Will remove before merging.
This is not used.

pkg/store/store.go
```diff
@@ -0,0 +1,102 @@
...
+	Set(data []byte, key ...string) error
+	// Delete removes a key or a group
+	Delete(key ...string) error
+	// Location returns the absolute path to a certain resource
+	// Note that this technically "leaks" (filesystem) implementation details up.
+	// It is necessary though when we are going to pass these filepath to containerd for eg.
+	Location(key ...string) (string, error)
+
+	// GroupRename will rename a group (eg: dir)
+	// NOTE: not used right now. Implementation is also commented out, provided here pending use case identification.
```
apostasie, line 88: Will remove before merge. Unused.

pkg/cmd/container/remove.go
```diff
@@ -128,18 +129,10 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions
 		return err
 	}
 	// Get volume store
-	volStore, err := volume.Store(globalOptions.Namespace, globalOptions.DataRoot, globalOptions.Address)
+	volStore, err := volumestore.New(dataStore, globalOptions.Namespace)
```
apostasie, line 132: Calling volume.Store is useless in that case, as we already have dataStore.

pkg/cmd/container/rename.go
```diff
@@ -44,7 +44,7 @@ func Rename(ctx context.Context, client *containerd.Client, containerID, newCont
 	if err != nil {
 		return err
 	}
-	hostst, err := hostsstore.NewStore(dataStore)
+	hostst, err := hostsstore.New(dataStore, options.GOptions.Namespace)
```
apostasie, line 47: HostsStore signature got changed to be per-namespace - align with other stores, and guarantees better isolation between namespaces and less "global" locking.

pkg/cmd/container/run_mount.go
```diff
@@ -258,7 +258,7 @@ func generateMountOpts(ctx context.Context, client *containerd.Client, ensuredIm
 
 		log.G(ctx).Debugf("creating anonymous volume %q, for \"VOLUME %s\"",
 			anonVolName, imgVolRaw)
-		anonVol, err := volStore.Create(anonVolName, []string{})
+		anonVol, err := volStore.CreateWithoutLock(anonVolName, []string{})
```
apostasie, line 261: The Create method now provides locking.
CreateWithoutLock emphasizes that this method MUST be guardrailed inside a locked section.

pkg/containerutil/container_network_manager.go
```diff
@@ -401,7 +409,6 @@ func (m *hostNetworkManager) SetupNetworking(ctx context.Context, containerID st
 
 	// Prep the meta
 	hsMeta := hostsstore.Meta{
-		Namespace:  container.Labels[labels.Namespace],
```
apostasie, line 404: This property is useless (and was before IIRC).
The meta is being saved under a namespaced location in the store already.

pkg/containerutil/container_network_manager.go
```diff
@@ -503,11 +510,20 @@ func (m *hostNetworkManager) ContainerNetworkingOpts(_ context.Context, containe
 	resolvConfPath := filepath.Join(stateDir, "resolv.conf")
 	copyFileContent("/etc/resolv.conf", resolvConfPath)
 
-	etcHostsPath, err := hostsstore.AllocHostsFile(dataStore, m.globalOptions.Namespace, containerID)
+	hs, err := hostsstore.New(dataStore, m.globalOptions.Namespace)
```
apostasie, line 513: Write only once, instead of allocating first, then writing.

pkg/identifiers/validate.go
```diff
@@ -42,9 +42,5 @@ func ValidateDockerCompat(s string) error {
 		return fmt.Errorf("identifier %q must match pattern %q: %w", s, AllowedIdentfierChars, errdefs.ErrInvalidArgument)
 	}
 
-	if err := validatePlatformSpecific(s); err != nil {
```
apostasie, line 45: Platform validation of identifiers is a filesystem concern.
These are moved into the `store.Store`, as they are orthogonal concerns.

cmd/nerdctl/volume_remove_linux_test.go
```diff
@@ -39,8 +40,8 @@ func TestVolumeRemove(t *testing.T) {
 	base := testutil.NewBase(t)
 
 	inUse := errdefs.ErrFailedPrecondition.Error()
-	malformed := errdefs.ErrInvalidArgument.Error()
-	notFound := errdefs.ErrNotFound.Error()
+	malformed := store.ErrInvalidArgument.Error()
+	notFound := store.ErrNotFound.Error()
```
AkihiroSuda, line 44: ?

pkg/cmd/volume/rm.go
```diff
@@ -33,17 +33,10 @@ import (
 )
 
 func Remove(ctx context.Context, client *containerd.Client, volumes []string, options types.VolumeRemoveOptions) error {
```
apostasie, line 35: Thinking more about this, the entire section starting with client.Containers should be inside the locked segment.
Will amend.

cmd/nerdctl/volume_remove_linux_test.go
apostasie: Yeah, been back and forth on this.
I still think we have a general problem in the codebase with regard to errors management.
Since we generally do `fmt.Errorf` instead of using exported errors (`var ErrFoo = errors.New`), consuming code cannot test with `errors.Is`, making it hard to specialize behavior dependent on the type of error we get.

Some of my recent PR _do_ introduce better, finer grain and explicit error-ing.
Obviously, some of the errors should be shared across different sections of code, maybe into an `errs` subpackage.

Which brings the question of whether we want to try and reuse some (or all) of `containerd/errdefs` (for example `ErrNotFound`).

I have mixed feelings about whether we should do that or not, and I think I am leaning towards having our own independent `errs` package, as it will make things clearer that what nerdctl returns is nerdctl errors (even if down there, they happen to alias containerd errdef).

Thoughts?

apostasie:
Rebase broke some tests.
Investigating.

apostasie:
- rebased
- removed commented out extra stuff
- changed `volStore.Remove` semantics as discussed above to allow for locking of a function provided as a generator

CI is green.

cmd/nerdctl/volume_remove_linux_test.go
AkihiroSuda: I'm not sure about this. Easier to merge this PR if this part can be omitted.

apostasie: Ok. Omitted with latest push.

apostasie:
Rebased. Pending green CI.

pkg/store/filestore.go
AkihiroSuda: This should be removed now?

AkihiroSuda:
`GroupRename()` seems added (and commented out) in the first commit and then removed in the third commit.
Probably this should be just removed from the first commit

apostasie:
> `GroupRename()` seems added (and commented out) in the first commit and then removed in the third commit. Probably this should be just removed from the first commit

My bad. Thanks for catching it.
Last push corrects it.

(CI failing on Hub being 500)

pkg/store/store.go
```diff
@@ -0,0 +1,98 @@
...
+	// It is necessary though when we are going to pass these filepath to containerd for eg.
+	Location(key ...string) (string, error)
+
+	// GroupSize will return the combined size of all objects stored under the group (eg: dir)
+	GroupSize(key ...string) (int64, error)
+	// GroupEnsure ensures that a given group (eg: directory) exists
+	GroupEnsure(key ...string) error
+}
+
+// Manager describes a locking mechanism that can be used to encapsulate operations in a safe way
```
fahedouch, line 93:
```suggestion
// Locker describes a locking mechanism that can be used to encapsulate operations in a safe way
```

fahedouch:
big UP for this PR, I have to points to raise here:
- please avoid huge PR
-  we may need a doc for this abstraction (e.g big picture system design)

apostasie:
> big UP for this PR 👍 , I have two points to raise here:
> 
> * please avoid huge PR, it is really hard to review
> * we might need a document for this abstraction (e.g. big picture abstraction design)

Thanks a lot @fahedouch 
Really appreciate the sharp eye review
On my phone now but will address your comments momentarily

ack on splitting in more commits
Will definitely do in upcoming work

pkg/cmd/volume/prune.go
```diff
@@ -33,60 +35,50 @@ func Prune(ctx context.Context, client *containerd.Client, options types.VolumeP
...
+		if err != nil {
+			return nil, err
+		}
 
-	// Iterate through the known volumes, making sure we do not remove in-use volumes
-	// but capture as well anon volumes (if --all was passed)
-	for _, volume := range volumes {
-		if _, ok := usedVolumesList[volume.Name]; ok {
-			continue
+		usedVolumesList, err := usedVolumes(ctx, containers)
```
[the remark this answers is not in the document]
apostasie: Mmmm... I see.
Well, we do still need the `usedVolumes` list to display a warning to the user if they ask to rm an in-use volume.

Your point still stand though, we might want to cleanup things a bit more under `pkg/cmd` later on.

AkihiroSuda (approved):
Thanks

fahedouch (approved):
Thanks

PR #3366, opened 2024-08-26
Changing bufio.Scanner to bufio.Reader to support large message

CerberusQc:
From the Issue https://github.com/containerd/nerdctl/issues/3343

As stated in the bufio Documentation, when using the bufio.Scanner, if the buffer exceed the max size of the token, the Scanner fail silently with an I/O error.

In the case of a container started with nerdctl, the main process using stdout/stderr hang forever when sending text bigger than 64kb (bufio.MaxScanTokenSize) without any newline character. (Discovered from a JSON string dumped in the logs). bufio documentation state to switch to a Reader instead when you do not have control over the size of the message.

Small test that is failing with the current master but working with this fix.

@apostasie recommend to check the other place in the code for the bufio.Scanner usage but I am not really a GO expert

### DIFF
--- a/pkg/logging/logging.go
+++ b/pkg/logging/logging.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -170,16 +171,23 @@ func loggingProcessAdapter(ctx context.Context, driver Driver, dataStore string,
 	processLogFunc := func(reader io.Reader, dataChan chan string) {
 		defer wg.Done()
 		defer close(dataChan)
-		scanner := bufio.NewScanner(reader)
-		for scanner.Scan() {
-			if scanner.Err() != nil {
-				log.L.Errorf("failed to read log: %v", scanner.Err())
-				return
+		r := bufio.NewReader(reader)
+
+		var err error
+
+		for err == nil {
+			var s string
+			s, err = r.ReadString('\n')
+
+			if len(s) > 0 {
+				dataChan <- strings.TrimSuffix(s, "\n")
 			}
-			dataChan <- scanner.Text()
+
+			if err != nil && err != io.EOF {
+				log.L.WithError(err).Error("failed to read log")
+			}
 		}
 	}
-
 	go processLogFunc(stdoutR, stdout)
 	go processLogFunc(stderrR, stderr)
 	go func() {
--- a/pkg/logging/logging_test.go
+++ b/pkg/logging/logging_test.go
@@ -1,0 +1,115 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package logging
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"math/rand"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/containerd/containerd/v2/core/runtime/v2/logging"
+)
+
+type MockDriver struct {
+	processed      bool
+	receivedStdout []string
+	receivedStderr []string
+}
+
+func (m *MockDriver) Init(dataStore, ns, id string) error {
+	return nil
+}
+
+func (m *MockDriver) PreProcess(dataStore string, config *logging.Config) error {
+	return nil
+}
+
+func (m *MockDriver) Process(stdout <-chan string, stderr <-chan string) error {
+	for line := range stdout {
+		m.receivedStdout = append(m.receivedStdout, line)
+	}
+	for line := range stderr {
+		m.receivedStderr = append(m.receivedStderr, line)
+	}
+	m.processed = true
+	return nil
+}
+
+func (m *MockDriver) PostProcess() error {
+	return nil
+}
+
+func TestLoggingProcessAdapter(t *testing.T) {
+	// Will process a normal String to stdout and a bigger one to stderr
+	normalString := generateRandomString(1024)
+
+	// Generate 64KB of random text of bufio MaxScanTokenSize
+	// https://github.com/containerd/nerdctl/issues/3343
+	hugeString := generateRandomString(bufio.MaxScanTokenSize)
+
+	// Prepare mock driver and logging config
+	driver := &MockDriver{}
+	stdoutBuffer := bytes.NewBufferString(normalString)
+	stderrBuffer := bytes.NewBufferString(hugeString)
+	config := &logging.Config{
+		Stdout: stdoutBuffer,
+		Stderr: stderrBuffer,
+	}
+
+	// Execute the logging process adapter
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err := loggingProcessAdapter(ctx, driver, "testDataStore", config)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/logging/logging.go
```diff
@@ -170,16 +170,24 @@ func loggingProcessAdapter(ctx context.Context, driver Driver, dataStore string,
...
+		for {
+			s, err := scanner.ReadString('\n')
+			if err != nil {
+				if err == io.EOF {
+					// Send the last chunk if there's any data
+					if len(s) > 0 {
+						dataChan <- s
+					}
+				} else {
+					log.L.Errorf("failed to read log: %v", err)
```
apostasie, line 184: `log.L.WithError(err).Error("failed to read log")`

apostasie:
@CerberusQc you need to DCO your commit with `git -S` (and squash your commits here).

apostasie:
> @apostasie recommend to check the other place in the code for the bufio.Scanner usage but I am not really a GO expert

That does not have to happen right now. Maybe if you can open a new issue, in the line of: "follow-up to #3343: review uses of bufio.Scanner", so that someone can pick that up later?

pkg/logging/logging.go
```diff
@@ -170,16 +170,24 @@ func loggingProcessAdapter(ctx context.Context, driver Driver, dataStore string,
...
 		defer wg.Done()
 		defer close(dataChan)
-		scanner := bufio.NewScanner(reader)
-		for scanner.Scan() {
-			if scanner.Err() != nil {
-				log.L.Errorf("failed to read log: %v", scanner.Err())
+		scanner := bufio.NewReader(reader)
+
+		for {
+			s, err := scanner.ReadString('\n')
```
apostasie, line 176: My bet is that `bufio.Scanner.Scan` does NOT include LF, while `ReadString` does.
Pretty sure it will trip tests. If I am right, just strip a possibly trailing LF before sending?

apostasie: @CerberusQc read here if you do not know yet how to run specific integration tests locally:
https://github.com/containerd/nerdctl/blob/main/docs/testing.md

apostasie: @CerberusQc read here if you do not know yet how to run specific integration tests locally:
https://github.com/containerd/nerdctl/blob/main/docs/testing.md

CerberusQc:
> @CerberusQc you need to DCO your commit with `git -S` (and squash your commits here).

@apostasie I will need help with this one, I have created a GPG key, signed the last 4 commits, when I tried to sign the first commit, it created a merge, now I can't squash correctly. I have no idea of what I am doing at this point and I am really not familiar with the process, sorry about this

apostasie:
> > @CerberusQc you need to DCO your commit with `git -S` (and squash your commits here).
> 
> @apostasie I will need help with this one, I have created a GPG key, signed the last 4 commits, when I tried to sign the first commit, it created a merge, now I can't squash correctly. I have no idea of what I am doing at this point and I am really not familiar with the process, sorry about this

No problem.

So, first, the DCO is unrelated to actual key signing.
What you MUST do for the DCO when commiting is `git commit -s ...`
That will add the DCO to your commit message (https://wiki.linuxfoundation.org/dco) 
(apologies if I previously said `-S` - this one if for signing, which is nice, but not required)

About the rest - tis not you, tis git :D   

You have 5 commits here, just soft reset them (commits are rolled-back, but your changes are preserved locally):
```
git reset --soft HEAD~5
```

And now you can start "fresh": just `add` then `commit`
```
git add pkg/logging/logging.go pkg/logging/logging_test.go

# Some basic sanity checks before committing
make lint
make lint-imports

# Change to whatever message suits you
git commit -S -m "fix log issue with large messages (bufio.Scanner > bufio.Reader)"

# Now force-push
git push -f
```

Future stuff - if you do need more modifications, just amend instead of adding new commits on top:
```
git add whatever
git commit --amend
git push -f
```

Does that help?

CerberusQc:
@apostasie I think I have managed to make the basic squashing happen (feeling super noob here) I will check later for the unit tests to fix the stripping of newline

apostasie:
> @apostasie I think I have managed to make the basic squashing happen (feeling super noob here) I will check later for the unit tests to fix the stripping of newline

Failure is visible on integration tests.

To reproduce locally, make sure your updated nerdctl binary is in your path, then:
```
go test ./cmd/nerdctl/ -run TestLogsOutStreamsSeparated
```

(TestLogsOutStreamsSeparated is one of the failing integration test)

pkg/logging/logging.go
CerberusQc: wanted to keep it DRY but would need to process the TrimSuffix even if there is an error that is not `io.EOF`.
I am really open to idea on this one

apostasie:
> @apostasie thank you for guiding me through this, I am really learning today and yes, I prefer SSH signing, I will switch to this in the future.
> 
> I have run the whole integration suite in container without any fail, let's see how the pipeline goes

That looks promising.
Good stuff here, @CerberusQc.

pkg/logging/logging.go
apostasie: On that note:

"
ReadString reads until the first occurrence of delim in the input, returning a string containing the data up to and including the delimiter. If ReadString encounters an error before finding a delimiter, it returns the data read before the error and the error itself (often io.EOF). ReadString returns err != nil if and only if the returned data does not end in delim.
"
So:
- err == nil is the only case where you should trim
- you should send the data regardless of what error we get

I would maybe go for something like this (untested...):

```go
	if err == nil {
		s = strings.TrimSuffix(s, "\n")
	}else if err != nil && err != io.EOF {
		log.L.WithError(err).Error("failed to read log")
	}
	dataChan <- s
```

CerberusQc: We will need to exit the loop when an error occur, but 
```GO
	s, err := scanner.ReadString('\n')
	if err != nil && err != io.EOF {
		log.L.WithError(err).Error("failed to read log")
		return
	}
	dataChan <- strings.TrimSuffix(s, "\n")
```
should do the job since the 2 cases we want to send is when there is no error or when the error is not EOF right?

apostasie: I think we should send in all cases - there may be data, with or without error (and regardless of the error).

take what I say with salt here - not my forte

CerberusQc: how would we exit the for loop in that case? Maybe I am missing something with channel and defer.

apostasie: I would break.

Here is something that should be closer to something functioning:

```go
processLogFunc := func (reader io.Reader, dataChan chan string) {
	defer wg.Done()
	defer close(dataChan)
	scanner := bufio.NewReader(reader)

	for {
		s, err := scanner.ReadString('\n')
		if err == nil {
			s = strings.TrimSuffix(s, '\n')
		}
		dataChan <- s
		if err != nil && err != io.EOF {
			log.L.WithError(err).Error("failed to read log")
			break
		}
	}
}

```

Alternatively, stuff like that would work too to break the loop:

```
var err error
for ; err == nil; {

}
```

pkg/logging/logging.go
```diff
@@ -170,16 +171,20 @@ func loggingProcessAdapter(ctx context.Context, driver Driver, dataStore string,
...
+		var s string
+
+		for err == nil {
+			s, err = scanner.ReadString('\n')
+			if err != nil && err != io.EOF {
+				log.L.WithError(err).Error("failed to read log")
 			}
-			dataChan <- scanner.Text()
+
+			dataChan <- strings.TrimSuffix(s, "\n")
```
CerberusQc, line 185: TestLogsOutStreamsSeparated was failing if we would trim the suffix only when there is no error

CerberusQc:
@apostasie had some trouble making work the unit test and integration test on macOS but, looks good now, I decided to go with the second implementation plus a check on the length of the string, without it the unit tests does not pass for the TestLogsOutStreamsSeparated

apostasie:
LGTM

Tagging @AkihiroSuda @fahedouch for maintainers' review.

apostasie (approved):

AkihiroSuda (approved):
Thanks

PR #3367, opened 2024-08-26
Allow Canary to fail the build

apostasie:
Right now, canary integration test failures are ignored.
This was on-purpose initially, as I was expecting canary to be very unstable.

Although, the effect of `continue-on-error` is that these tasks are reported as green, and failures do go unnoticed.

This PR just remove that, making sure that we are appropriately noticing canary test failures.
If it turns out that canary is just failing too much, well, maybe we fix the issues - or we revert this change.

cc @AkihiroSuda if we can get that soon, that would be lovely!

### DIFF
--- a/.github/workflows/test-canary.yml
+++ b/.github/workflows/test-canary.yml
@@ -65,7 +65,6 @@ jobs:
       - name: "Run unit tests"
         run: go test -v ./pkg/...
       - name: "Run integration tests"
-        continue-on-error: true
         run: docker run -t --rm --privileged test-integration
 
   windows:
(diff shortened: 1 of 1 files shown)


apostasie:
cc @fahedouch :)

fahedouch:
I'm not sure we want to run this for every PR, it seems unstable and it may break CI PRs that target the stable version of deps. We may have independent workers who manage this periodically.

apostasie:
> I'm not sure we want to run this for every PR, it seems unstable and it may break CI PRs that target the stable version of deps. We may have independent workers who manage this periodically.

We can of course revisit this if it is a PITA.
Honestly, right now, I am more annoyed by flaky tests on the regular CI (TestIPFSComposeX being the worst offender).

A couple of additional notes:
- this does not target _unstable_ versions of dependencies - it only target the _latest release_ of dependencies (including pre-releases, that is true)
- though the idea of having a system that is independent of PRs is attractive, experience of this unfortunately showed that people tend to ignore these

Anyhow, let see how this goes?

PR #3369, opened 2024-08-26
fix: nerdctl stats on containers without a memory limit returns host memory limit

cezar-r:
Fix: https://github.com/runfinch/finch/issues/45

 - Fixed bug where `nerdctl stats` would show 16 exbibytes of memory limit when memory limit on container was not set.

### DIFF
--- a/pkg/statsutil/stats_linux.go
+++ b/pkg/statsutil/stats_linux.go
@@ -17,6 +17,10 @@
 package statsutil
 
 import (
+	"bufio"
+	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/vishvananda/netlink"
@@ -35,11 +39,10 @@ func calculateMemPercent(limit float64, usedNo float64) float64 {
 }
 
 func SetCgroupStatsFields(previousStats *ContainerStats, data *v1.Metrics, links []netlink.Link) (StatsEntry, error) {
-
 	cpuPercent := calculateCgroupCPUPercent(previousStats, data)
 	blkRead, blkWrite := calculateCgroupBlockIO(data)
 	mem := calculateCgroupMemUsage(data)
-	memLimit := float64(data.Memory.Usage.Limit)
+	memLimit := getCgroupMemLimit(float64(data.Memory.Usage.Limit))
 	memPercent := calculateMemPercent(memLimit, mem)
 	pidsStatsCurrent := data.Pids.Current
 	netRx, netTx := calculateCgroupNetwork(links)
@@ -59,11 +62,10 @@ func SetCgroupStatsFields(previousStats *ContainerStats, data *v1.Metrics, links
 }
 
 func SetCgroup2StatsFields(previousStats *ContainerStats, metrics *v2.Metrics, links []netlink.Link) (StatsEntry, error) {
-
 	cpuPercent := calculateCgroup2CPUPercent(previousStats, metrics)
 	blkRead, blkWrite := calculateCgroup2IO(metrics)
 	mem := calculateCgroup2MemUsage(metrics)
-	memLimit := float64(metrics.Memory.UsageLimit)
+	memLimit := getCgroupMemLimit(float64(metrics.Memory.UsageLimit))
 	memPercent := calculateMemPercent(memLimit, mem)
 	pidsStatsCurrent := metrics.Pids.Current
 	netRx, netTx := calculateCgroupNetwork(links)
@@ -80,6 +82,36 @@ func SetCgroup2StatsFields(previousStats *ContainerStats, metrics *v2.Metrics, l
 		PidsCurrent:      pidsStatsCurrent,
 	}, nil
 
+}
+
+func getCgroupMemLimit(memLimit float64) float64 {
+	if memLimit == float64(^uint64(0)) {
+		return getHostMemLimit()
+	}
+	return memLimit
+}
+
+func getHostMemLimit() float64 {
+	file, err := os.Open("/proc/meminfo")
+	if err != nil {
+		return float64(^uint64(0))
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if strings.HasPrefix(scanner.Text(), "MemTotal:") {
+			fields := strings.Fields(scanner.Text())
+			if len(fields) >= 2 {
+				memKb, err := strconv.ParseUint(fields[1], 10, 64)
+				if err == nil {
+					return float64(memKb * 1024) // kB to bytes
+				}
+			}
+			break
+		}
+	}
+	return float64(^uint64(0))
 }
 
 func calculateCgroupCPUPercent(previousStats *ContainerStats, metrics *v1.Metrics) float64 {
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -21,13 +21,13 @@ require (
...
 	github.com/containerd/go-cni v1.1.10
-	github.com/containerd/imgcrypt v1.2.0-rc1.0.20240709223013-f3769dc3e47f
+	github.com/containerd/imgcrypt v1.2.0-rc1.0.20240724013606-c091455481f6
 	github.com/containerd/log v0.1.0
 	github.com/containerd/nydus-snapshotter v0.14.1-0.20240806063146-8fa319bfe9c5
 	github.com/containerd/platforms v0.2.1
-	github.com/containerd/stargz-snapshotter v0.15.2-0.20240709063920-1dac5ef89319
-	github.com/containerd/stargz-snapshotter/estargz v0.15.2-0.20240709063920-1dac5ef89319
-	github.com/containerd/stargz-snapshotter/ipfs v0.15.2-0.20240709063920-1dac5ef89319
+	github.com/containerd/stargz-snapshotter v0.15.2-0.20240826180748-fbc3f6a1d4aa
```
Shubhranshu153, line 28: avoid updating hash for unrelated things.

pkg/statsutil/stats_linux.go
```diff
@@ -73,6 +72,27 @@ func SetCgroup2StatsFields(previousStats *ContainerStats, metrics *v2.Metrics, l
 
 }
 
+func getMemLimit(metrics *v1.Metrics) float64 {
```
Shubhranshu153, line 75: getCgroupMemLimit

pkg/statsutil/stats_linux.go
```diff
@@ -73,6 +72,27 @@ func SetCgroup2StatsFields(previousStats *ContainerStats, metrics *v2.Metrics, l
...
+	memLimit := float64(metrics.Memory.Usage.Limit)
+	if memLimit == float64(^uint64(0)) {
+		return float64(getHostMemoryLimit())
+	}
+	return memLimit
+}
+
+func getMemLimit2(metrics *v2.Metrics) float64 {
```
Shubhranshu153, line 83: getCgroup2MemLimit

pkg/statsutil/stats_linux.go
```diff
@@ -73,6 +72,27 @@ func SetCgroup2StatsFields(previousStats *ContainerStats, metrics *v2.Metrics, l
...
+	memLimit := float64(metrics.Memory.UsageLimit)
+	if memLimit == float64(^uint64(0)) {
+		return float64(getHostMemoryLimit())
+	}
+	return memLimit
+}
+
+func getHostMemoryLimit() uint64 {
+	virtualMemLimit, _ := mem.VirtualMemory()
```
Shubhranshu153, line 92: why virtual memory?

cezar-r: Followed their documentation - https://github.com/shirou/gopsutil?tab=readme-ov-file#usage 
I didn't catch any other relevant methods

go.mod
cezar-r: These changes occurred after running `go mod tidy`, which I needed to run after adding the `gopsutil` package. Let me see how I can work around this

pkg/statsutil/stats_linux.go
Shubhranshu153: yeah seems its called virtualmemory

go.mod
Shubhranshu153: probably is fine

apostasie: Tried it, and `go mod tidy` did not up these dependencies.
Maybe you did `go get -u` somehow?
Anyhow, I don't think we should touch these at this time.

go.mod
```diff
@@ -59,6 +59,7 @@ require (
 	github.com/pelletier/go-toml/v2 v2.2.3
 	github.com/rootless-containers/bypass4netns v0.4.1
 	github.com/rootless-containers/rootlesskit/v2 v2.3.1
+	github.com/shirou/gopsutil/v3 v3.24.5
```
apostasie, line 62: hey @cezar-r - is there a reason to not use /v4?

go.mod
cezar-r: fixed in latest push

go.mod
cezar-r: This was the recommended version that was prompted by Go and I was having some issues with v4 as well in `go.sum` but I can re-evaluate if needed @apostasie

apostasie: My concern is that v3 will drop out of support soon. Back in May: `v3 will not be updated except for high level security issues.`

cezar-r: good catch, updated it to v4 in latest push

go.mod
```diff
@@ -59,6 +59,7 @@ require (
 	github.com/pelletier/go-toml/v2 v2.2.3
 	github.com/rootless-containers/bypass4netns v0.4.1
 	github.com/rootless-containers/rootlesskit/v2 v2.3.1
+	github.com/shirou/gopsutil/v4 v4.24.7
```
AkihiroSuda, line 62: This seems to introduce a bunch of other deps that should not be necessary for fixing the issue

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/3369/files#r1735502693

go.mod
cezar-r: Thanks, will explore other options

cezar-r: > This seems to introduce a bunch of other deps that should not be necessary for fixing the issue

Apologies for the delay, the latest commit implements this feature without the need for other deps. 

I do want to note; it seems like there was a discussion upstream in the containerd cgroups package relating to this issue ([containerd/cgroups#265](https://github.com/containerd/cgroups/pull/265#issuecomment-1418250556)) that never got resolved. Is there alignment for the one of  4 options from that discussion? Should this fix even be implemented in this package or should it be upstream in containerd? 

Let me know your thoughts @AkihiroSuda

cezar-r:
> This seems to introduce a bunch of other deps that should not be necessary for fixing the issue

Apologies for the delay, the latest commit implements this feature without the need for other deps. 

I do want to note; it seems like there was a discussion upstream in the containerd cgroups package relating to this issue ([containerd/cgroups#265](https://github.com/containerd/cgroups/pull/265#issuecomment-1418250556)) that never got resolved. Is there alignment for the one of  4 options from that discussion? Should this fix even be implemented in this package or should it be upstream in containerd? 

Let me know your thoughts @AkihiroSuda

AkihiroSuda:
> Should this fix even be implemented in this package or should it be in containerd?

This package

Shubhranshu153:
@AkihiroSuda 

Following this discussion:
is the recommendation to have config value (via os environment or config.json irrespective) that users can configure and return the value and if not then return not configured/nil (option 4)

If not configured:
should we return the system max value, as that what docker seems to return and people trying to integrate with systems which were based on docker might want this to return max system limit of the current system. We probably can do it in finch under dockercompat flags, but feels like the right place for it is nerdctl. 


Thanks

Shubhranshu153:
Thanks @AkihiroSuda 
In that case we can return the max system limit of the current system: (this is from a 15.26 GB system)

```
% grep MemTotal /proc/meminfo
MemTotal:       16004064 kB
```

```
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT   MEM %     NET I/O     BLOCK I/O   PIDS
2b07ca5e9b88   goofy_khorana   0.00%     276KiB / 15.26GiB   0.00%     720B / 0B   0B / 0B     1
```

cezar-r:
Reopened the PR as it seems the implementation is aligned with what we need

AkihiroSuda (approved):
Thanks

PR #3374, opened 2024-08-28
Disable flaky #827 test

apostasie:
This test (which is btw testing that we are still broken) is unfortunately very flaky, to the point it is a PITA.

Disabling it for now - at least until we fix the actual issue this is meant to test.

@AkihiroSuda @djdongjin


djdongjin (commented):
Can we just add a `t.Skip("the reason to skip")` at the begining of the test? I think that's clear than commenting out code.

apostasie:
@djdongjin done

djdongjin (approved):
LGTM

AkihiroSuda (approved):
Thanks

PR #3379, opened 2024-08-29
Refactor cmd/nerdctl

apostasie:
This is in light of #3375 and #1680

The overall goal is to break down cmd/nerdctl.

These commits do:
- move some commonly used functions into `cmd/nerdctl/helpers`
  - these have been roughly split into distinct files that belong to different categories (testing, cobra, etc.)
  - see final note below on not attempting to clean these up further for now
- move to subpackages:
  - `cmd/nerdctl/completion`
  - `cmd/nerdctl/login`
  - `cmd/nerdctl/namespace`
  - `cmd/nerdctl/network`
  - `cmd/nerdctl/volume`
  - `cmd/nerdctl/image`
  - `cmd/nerdctl/builder`
  - `cmd/nerdctl/apparmor`
  - `cmd/nerdctl/internal`

This is purely refactoring, and there is no functional or logic change (see testing note at the bottom though).

Further note that this PR does not try to "cleanup" anything besides that. Specifically out of scope for now (could be done in a subsequent PR):
- name stuttering reduction
- possible cleanup of the "helpers" folder

Also, this PR is big enough now - what remains to be done (containers and compose) are big chunks on their own, and I would rather merge this first then tackle these two in the next PR.

This is on-top of #3384 that should be merged first.

PR #3384, opened 2024-08-31
Testing-rig fixes

apostasie:
In preparation for #3379

Couple of things:

1. by default, go test will parallelize sub-packages testing - once we move to `./cmd/nerdctl/...` this will break on us. To prevent this, pass along `-p 1` which will instruct go test to only test one sub-package at a time. Note that this is not the same as _parallelized tests_ - tests calling `t.Parallel` will still be run in parallel with others (and this is controlled by `-parallel`)
2. rootless tests did use a different stance altogether to test (first build a test binary, then call `go tool`). I do not know for sure why this was done (maybe @AkihiroSuda would know). My hunch is that this was motivated by the fact we ssh back into the container - which does lose both cwd and PATH. I did fix that in `test-integration-rootless.sh` instead, so that rootless can just use the familiar, normal `go test` routine

This is on-top of #3385 which should be merged first.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -21,17 +21,33 @@ Be sure to first `make && sudo make install`
 
 ```bash
 # Test all with nerdctl (rootless mode, if running go as a non-root user)
-go test ./cmd/nerdctl/...
+go test -p 1 ./cmd/nerdctl/...
 
 # Test all with nerdctl rootful
-go test -exec sudo ./cmd/nerdctl/...
+go test -p 1 -exec sudo ./cmd/nerdctl/...
 
 # Test all with docker
-go test ./cmd/nerdctl/... -args -test.target=docker
+go test -p 1 ./cmd/nerdctl/... -args -test.target=docker
 
 # Test just the tests(s) which names match TestVolume.*
-go test ./cmd/nerdctl/... -run "TestVolume.*"
+go test -p 1 ./cmd/nerdctl/... -run "TestVolume.*"
+# Or alternatively, just test the subpackage
+go test ./cmd/nerdctl/volume
 ```
+
+### About parallelization
+
+By default, when `go test ./foo/...` finds subpackages, it does create _a separate test binary
+per sub-package_, and execute them _in parallel_.
+This effectively will make distinct tests in different subpackages to be executed in
+parallel, regardless of whether they called `t.Parallel` or not.
+
+The `-p 1` flag does inhibit this behavior, and forces go to run each sub-package
+sequentially.
+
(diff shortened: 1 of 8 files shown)


apostasie:
@AkihiroSuda this is ready.
PTAL at your convenience.

CI is (mostly) green (failure is the usual `TestIPFSComposeUp/stargz`).

Should be merged after #3385.

docs/testing.md
```diff
@@ -21,16 +21,18 @@ Be sure to first `make && sudo make install`
 
 ```bash
 # Test all with nerdctl (rootless mode, if running go as a non-root user)
-go test ./cmd/nerdctl/...
+go test -p 1 ./cmd/nerdctl/...
```
AkihiroSuda, line 24: Thanks, I think this is the reason why we have failed to split `*_test.go`.

In the document could you explain what is `-p` and how it differs from `-parallel`?
The explanation should be copied to GHA yamls too.

AkihiroSuda: I also wonder if we can print a human-readable error when `-p 1` is not specified.

AkihiroSuda: Also wonder how this affects the run-time of the CI

apostasie: > Also wonder how this affects the run-time of the CI

Theoretically, this MAY impact our runtime.

Currently, we first run sequentially any test that is not parallel. Then we run all tests that can be parallelized together, with parallelization at GOMAXPROCS.

With the new model, the tests that can be parallelized are run together, but only inside the subpackage (and each subpackage is run sequentially).

But then, how many cores do we have on github actions? 4?

Back of the envelop:
- 15 subpackages
- about 1000 tests -> ~70 tests per package
- 160 calls to t.Parallel (but then, for loops, so, let's say 300) -> ~20 tests marked parallel

If we do indead have a small number of cores (<=4), then it is likely close to zero impact...

apostasie: > I also wonder if we can print a human-readable error when `-p 1` is not specified.

Maybe?

We could have the test initialization method to check the status of an flock and bail-out with an error if it is in use?
A bit heavy-handed, but then, it might work.

apostasie: > Thanks, I think this is the reason why we have failed to split `*_test.go`.
> 
> In the document could you explain what is `-p` and how it differs from `-parallel`? The explanation should be copied to GHA yamls too.

Yes.
This is not fully trivial knowledge.

Done with latest push.

AkihiroSuda (approved):
Thanks

PR #3388, opened 2024-08-31
Do not load build output if dest is specified

### LINKED ISSUE #3386: `nerdctl build --output "type=oci,dest=-" . > output.tar` does not output to stdout
### Description

`nerdctl build --output "type=tar,dest=-" . > output.tar` works as expected, the root of the image outputs to stdout, and is then redirected to `output.tar`. This is also the expected behavior of `type=oci`, but this is the actual behavior:

```
$ nerdctl build --output "type=oci,dest=-" . > output.tar
$ cat output.tar
unpacking overlayfs@sha256:3f163869fade0eb290d0a847f8c5fb32b75009d862c59558deb3b21472b8c732 (sha256:3f163869fade0eb290d0a847f8c5fb32b75009d862c59558deb3b21472b8c732)...
Loaded image: overlayfs@sha256:3f163869fade0eb290d0a847f8c5fb32b75009d862c59558deb3b21472b8c732
```

1. `mkdir output && nerdctl build --output "type=oci,dest=-" . | tar -C output -xvf - && ls output`

```
oci-layout
index.json
...
```

1.7.6

Lima

### Host information

`Linux lima-default 6.8.0-40-generic #40-Ubuntu SMP PREEMPT_DYNAMIC aarch64 aarch64 aarch64 GNU/Linux`

apostasie:
This is not a great PR.
It merely fix #3386 - but clearly this whole code should be rewritten IMHO (`strings.Contains` does not seem like the right approach to properly parse the option here).

A proper cleanup and handling of corner cases could/should be done in a follow-up PR, and add some tests, but at least this quick fix here does address a couple of issues:
- nerdctl wrongly exits 1 with "no image was built" when dest is specified
- inability to output oci images when dest is set

PTAL at your convenience.

Tagging @MayCXC

### DIFF
--- a/pkg/cmd/builder/build.go
+++ b/pkg/cmd/builder/build.go
@@ -228,7 +228,9 @@ func generateBuildctlArgs(ctx context.Context, client *containerd.Client, option
 			output = fmt.Sprintf("type=local,dest=%s", output)
 		}
 		if strings.Contains(output, "type=docker") || strings.Contains(output, "type=oci") {
-			needsLoading = true
+			if !strings.Contains(output, "dest=") {
+				needsLoading = true
+			}
 		}
 	}
 	if tags = strutil.DedupeStrSlice(options.Tag); len(tags) > 0 {


apostasie:
For reference:
https://docs.docker.com/reference/cli/docker/buildx/build/#docker

MayCXC:
nice, happy to see that exit 1 is fixed too 😎

AkihiroSuda (approved):
Thanks, maybe we should have a (unit) test for `needsLoading`

PR #3383, opened 2024-08-30
Refactor cmd/nerdctl, part 2

apostasie:
This is on top of #3379 and continuation of the work started there.

As the previous PR, this does not contain any logic change or fixes - this is purely refactoring.

This one deals with:
- container
- compose
- leftovers (inspect, system)

PR #3394, opened 2024-09-02
Fix TestDiff failure with docker

apostasie:
TestDiff started failing with target docker after moving `cmd/container` to a subpackage.

The actual reason for the failure is unclear to me, but the bottom-line is that it does not work with Parallel, which suggests it is interacting with another parallelized test.

As for why this is only showing up (more) now, the answer is simply the reduction in size of the parallelized test pool (since we only parallelize per sub-package now), which makes it much more likely that the interacting test is actually running in parallel with TestDiff.

Anyhow, this will unblock #3383 and can / should be merged now.

### DIFF
--- a/cmd/nerdctl/container_diff_linux_test.go
+++ b/cmd/nerdctl/container_diff_linux_test.go
@@ -23,7 +23,11 @@ import (
 )
 
 func TestDiff(t *testing.T) {
-	t.Parallel()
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks, you might be hitting some bug of Docker

apostasie:
> Thanks, you might be hitting some bug of Docker

Very likely, yes.

PR #3405, opened 2024-09-03
Draft design document for store

apostasie:
As requested by @fahedouch in #3364, this adds kind of a "design" document for the `store` implementation (still to be merged in 3364).

I am not sure this is the right tone, and I feel a bit bizarre explaining things that may (or may not) be obvious, so, please let me know your thoughts bluntly on whether any of this is helpful or not.

Furthermore, this PR creates a new `docs/dev` folder that is developer oriented (instead of user oriented) and moves the testing doc in there as well.


docs/dev/testing.md
AkihiroSuda, line 1: Maybe this should be a separate commit

docs/dev/store.md
```diff
@@ -0,0 +1,163 @@
+# Store
```
AkihiroSuda, line 1: "Store" needs explanation

apostasie: Amended that to more explicitly refer to `pkg/store`.

docs/dev/testing.md
apostasie: Done.

AkihiroSuda (approved):
Thanks

PR #3407, opened 2024-09-04
Lint for other platforms

### LINKED ISSUE #3382: lint broken on windows and freebsd
### Description

`GOOS=windows make lint` or `GOOS=freebsd make lint` are currently broken.

This is mostly about tests - but some minor stuff in main code as well.

We currently do not tests these on the CI.

Questions:
- should we care?
- if yes, do we want it enabled on the CI as well?

na

na

main

None

### Host information

_No response_

apostasie:
Fix #3382 and enforce linting checks for other platforms.

This is trivial changes to appease the linter gods.


cmd/nerdctl/container/container_cp_unix.go
```diff
@@ -1,3 +1,5 @@
+//go:build unix
```
AkihiroSuda: Isn't this implied by the file name suffix?

AkihiroSuda (approved):
Thanks

cmd/nerdctl/container/container_cp_unix.go
apostasie: I think only GOOS/GOARCH values are recognized as "magic" suffix implying a build tag.

IIRC they tried to introduce the same semantic for `_unix` but it backfired and broke a bunch of projects so they only added the `// go:build unix` directive but not the file matching.

PR #3411, opened 2024-09-04
Login refactor (small breakout): add support for explicitly setting --insecure-registry=false

apostasie:
This is the ground work for fixing #3046.

See commit message for details.

TL;DR:
`docker login --insecure-registry false localhost` should NOT treat localhost as insecure.

### DIFF
--- a/cmd/nerdctl/helpers/flagutil.go
+++ b/cmd/nerdctl/helpers/flagutil.go
@@ -22,6 +22,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
+	"github.com/containerd/nerdctl/v2/pkg/config"
 )
 
 func ProcessImageVerifyOptions(cmd *cobra.Command) (opt types.ImageVerifyOptions, err error) {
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -39,6 +39,8 @@ type Config struct {
 	HostsDir         []string `toml:"hosts_dir"`
 	Experimental     bool     `toml:"experimental"`
 	HostGatewayIP    string   `toml:"host_gateway_ip"`
+
+	ExplicitInsecureRegistry bool `toml:"-"`
```
AkihiroSuda, line 43: Not sure about this. The struct is dedicated for TOML properties.

apostasie: Yeah, I was thinking about that.

We could use a different struct instead (or just an extra param) - but then we would have to modify the signature of a lot of functions in the path of `pull`, `push`, `login`.

Let me know what you prefer.

apostasie: Ok. Alternative proposal (overloading globalconfig) with the latest push.
LMK your thoughts.

AkihiroSuda:
> Implementation will follow in a subsequent PR that will rewrite login.go.

The struct looks good, but I expect the implementation to be accompanied in the same PR

apostasie:
> > Implementation will follow in a subsequent PR that will rewrite login.go.
> 
> The struct looks good, but I expect the implementation to be accompanied in the same PR

The implementation does require rather large modifications to dockerconfigresolver and login code which are difficult to breakdown.

What about I close this PR then - and just land this as part of the (larger) PR to come?

I will try my best to have as many separate commits as possible to ease review.

Issue #3413, opened 2024-09-04
dockerConfigFile.GetCredentialsStore.Store likely faulty symlink behavior


apostasie:
### Description

When DOCKER_CONFIG points to a directory that does contain a relative, dangling `config.json` symlink, Docker credentials store will (apparently) wrongly resolve the link to the current working directory.

This can be reproduced with the following:

```
mkdir -p /tmp/foo
ln -s doesnotexist /tmp/foo/config.json
cd ~
DOCKER_CONFIG=/tmp/foo nerdctl login
cat ~/doesnotexist
ln -s /tmp/foo
```

This does suggest there is a bug in moby/docker somewhere where readlink is used to resolve against pwd instead of the parent dir (/tmp/foo).

In turn, if, for some reason, the file cannot be created in the current working directory (for example if it is readonly), this will error in a very baffling way with a very confusing message:
```
rename /tmp/TestBrokenCredentialsStore705218110/008/docker-config2430087685/config.json2911426040 doesnotexist: invalid cross-device link
 ```

First spotted in https://github.com/containerd/nerdctl/pull/3293#pullrequestreview-2231972713 although at the time I was unable to diagnose it.

While this is very likely a docker bug, we need to fix these tests to deal with that.

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
This ticket is both about making sure our tests work (fixed by the linked PR), but we could keep it around as "external" for now as well - or if somebody wants to, we might further diagnose it and report it to moby.

apostasie:
THIS

https://github.com/docker/cli/blob/master/cli/config/configfile/file.go#L172-L174

apostasie:
@thaJeztah you might want to fix this ^.

Readlink does not resolve the path for you, and by the time you write it, it will be relative to cwd.

You might consider using  instead `filepath.EvalSymlinks` (on the full path) - but then, this will error if the target does not exist, so, you have to account for that as well... maybe just `os.WriteFile("")` the file to ensure it exists... will work in _some_ cases... 

But then... you do MkdirAll *before* resolving the links (https://github.com/docker/cli/blob/master/cli/config/configfile/file.go#L144-L147), so... you will fail mkdir-ing where the link(s) actually resolve...

I certainly appreciate that an atomic write is the right thing to do, but in that case, you just cannot get away with it with just   readlink...

[...]

PR #3414, opened 2024-09-04
Fix test to workaround docker credstore symlink fault

apostasie:
Workaround #3413

As described in https://github.com/containerd/nerdctl/issues/3413 there seems to be a fault in docker credentials store that makes relative symlinks resolve to pwd instead of where they are. This in turn will break our test if the current working directory is on another read-only volume (typical with Lima). This changeset does Chdir to the parent temp directory so we can workaround that problem.

### DIFF
--- a/pkg/imgutil/dockerconfigresolver/credentialsstore_test.go
+++ b/pkg/imgutil/dockerconfigresolver/credentialsstore_test.go
@@ -168,27 +168,41 @@ func TestBrokenCredentialsStore(t *testing.T) {
 		},
 	}
 
-	t.Run("Broken Docker Config testing", func(t *testing.T) {
+	t.Run("Docker Config testing with a variety of filesystem situations", func(t *testing.T) {
+		// Do NOT parallelize this test, as it relies on Chdir, which would have side effects for other tests.
 		registryURL, err := Parse("registry")
 		if err != nil {
 			t.Fatal(err)
 		}
 
-		for _, tc := range testCases {
-			t.Run(tc.description, func(t *testing.T) {
+		for _, testCase := range testCases {
+			tc := testCase
+			t.Run(tc.description, func(tt *testing.T) {
+				// See https://github.com/containerd/nerdctl/issues/3413
+				var oldpwd string
 				directory := tc.setup()
-				cs, err := NewCredentialsStore(directory)
-				assert.ErrorIs(t, err, tc.errorNew)
+				oldpwd, err = os.Getwd()
+				assert.NilError(tt, err)
+				// Ignore the error, as the destination may not be a directory
+				_ = os.Chdir(directory)
+				tt.Cleanup(func() {
+					err = os.Chdir(oldpwd)
+					assert.NilError(tt, err)
+				})
+
+				var cs *CredentialsStore
+				cs, err = NewCredentialsStore(directory)
+				assert.ErrorIs(tt, err, tc.errorNew)
 				if err != nil {
 					return
 				}
 
 				var af *Credentials
 				af, err = cs.Retrieve(registryURL, true)
-				assert.ErrorIs(t, err, tc.errorRead)
+				assert.ErrorIs(tt, err, tc.errorRead)
 
 				err = cs.Store(registryURL, af)
-				assert.ErrorIs(t, err, tc.errorWrite)
+				assert.ErrorIs(tt, err, tc.errorWrite)
 			})
 		}
 	})


apostasie:
@AkihiroSuda can we get this in?

AkihiroSuda:
Can we just skip the test when the pwd doesn't satisfy the condition?

apostasie:
> Can we just skip the test when the cwd doesn't satisfy the condition?

I don’t think it is a good idea.
1. The docker SDK is faulty here and testing cwd for whatever conditions to verify that docker will not fail seems counter-productive when it is not supposed to do anything in cwd in the first place
2. Our testing rig is also problematic - we should really be in a temp dir by default instead

pkg/imgutil/dockerconfigresolver/credentialsstore_test.go
```diff
@@ -169,6 +169,7 @@ func TestBrokenCredentialsStore(t *testing.T) {
 	}
 
 	t.Run("Broken Docker Config testing", func(t *testing.T) {
```
fahedouch, line 171: there is also a healthy Docker configuration

fahedouch:
Hi @apostasie,

I don't see where the symbolic link is involved here!

apostasie:
Hey @fahedouch :)

If `config.json` is a relative symlink (inside the location pointed at by `$DOCKER_CONFIG`), the Docker SDK will wrongly resolve that against CWD (https://github.com/docker/cli/blob/master/cli/config/configfile/file.go#L171-L178).

The workaround here is just avoiding bad effects of that, by going `Chdir` into a temp location.

apostasie:
More details in the discussion with Sebastiaan in #3414

apostasie:
For a "real' fix: either Docker should take of that since the bug is theirs, or we come-up with a proper, non-buggy atomic write implementation and create the file first before querying the Docker cred store.

pkg/imgutil/dockerconfigresolver/credentialsstore_test.go
```diff
@@ -177,6 +178,12 @@ func TestBrokenCredentialsStore(t *testing.T) {
 		for _, tc := range testCases {
 			t.Run(tc.description, func(t *testing.T) {
 				directory := tc.setup()
+				// See https://github.com/containerd/nerdctl/issues/3413
+				oldpwd, err := os.Getwd()
```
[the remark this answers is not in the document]
apostasie: > this test is not parallelized.

Yes, and it should stay that way.

> rollback dir at the end of the test

Geez. My bad, and absolutely yes. Will fix.

pkg/imgutil/dockerconfigresolver/credentialsstore_test.go
apostasie: Sure. Suggesting we clean that up at a later point?

fahedouch: sure

apostasie:
@fahedouch `t.Cleanup` addressed.

Also changed the test group title after all.

PTAL at your convenience.

Thanks!

pkg/imgutil/dockerconfigresolver/credentialsstore_test.go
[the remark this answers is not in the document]
fahedouch: `oldpwd` should be outside the loop otherwise `os.Getwd()` of the second iteration is equal to  `directory`because `t.Cleanup` is  executed once by the end of the test.

pkg/imgutil/dockerconfigresolver/credentialsstore_test.go
```diff
@@ -177,6 +178,16 @@ func TestBrokenCredentialsStore(t *testing.T) {
...
+				assert.NilError(t, err)
+				// Ignore the error, as the destination may not be a directory
+				_ = os.Chdir(directory)
+				t.Cleanup(func() {
```
fahedouch, line 186: could we put `t.Cleanup` outside the loop ?

apostasie: Mmm... we do `Chdir` to whatever is returned by the specific _subtest_ `setup` routine, so, it should stay there, right?
Maybe I misunderstand your point?

fahedouch: `t.Cleanup` are executed in LIFO order  after the test return, the last one will override all of them

apostasie: Put otherwise. « Each subtest should leave things as they were when they exit ».

fahedouch: `t.Cleanup` funcs are called when the test and all its subtests complete; I think that having redundant `t.Cleanup` call for each subset test is useless.

what I suggest is:
````
oldpwd, err := os.Getwd()
assert.NilError(t, err)
directory := tc.setup()
_ = os.Chdir(directory)
t.Cleanup(func() {
	err = os.Chdir(oldpwd)
	assert.NilError(t, err)
})
for _, tc := range testCases {
			t.Run(tc.description, func(t *testing.T) {
				cs, err := NewCredentialsStore(directory)
				assert.ErrorIs(t, err, tc.errorNew)
				if err != nil {
					return
				}

				var af *Credentials
				af, err = cs.Retrieve(registryURL, true)
				assert.ErrorIs(t, err, tc.errorRead)

				err = cs.Store(registryURL, af)
				assert.ErrorIs(t, err, tc.errorWrite)

			})
		}
````

Comment (apostasie) [reply pkg/imgutil/dockerconfigresolver/credentialsstore_test.go:186]: How would `directory := tc.setup()` work outside of the loop that defines `tc`?

Comment (apostasie) [reply pkg/imgutil/dockerconfigresolver/credentialsstore_test.go:186]: I do not get where you are going with this.

The t.Cleanup calls are not "redundant". They are restoring things to what they were before the subtest started so that we do not pollute the next subtest (which is btw currently a recurring problem in our tests). 

This whole thing here is about avoiding the bad effects of a bug in docker, and work-arounding the fact that our testing rig is not segregating tests properly (we should be in a tempdir to start with).

Comment (fahedouch) [reply pkg/imgutil/dockerconfigresolver/credentialsstore_test.go:186]: > How would directory := tc.setup() work outside of the loop that defines tc?

Sorry my pseudo code was wrong, this is what I am suggesting 

```
	      oldpwd, err := os.Getwd()
	      assert.NilError(t, err)
             // Register cleanup to reset cwd after executing testsCases
	      t.Cleanup(func() {
			err = os.Chdir(oldpwd)
			assert.NilError(t, err)
	      })
	      for _, tc := range testCases {
			t.Run(tc.description, func(t *testing.T) {
				directory := tc.setup()
				_ = os.Chdir(directory)
				cs, err := NewCredentialsStore(directory)
				assert.ErrorIs(t, err, tc.errorNew)
				if err != nil {
					return
				}

				var af *Credentials
				af, err = cs.Retrieve(registryURL, true)
				assert.ErrorIs(t, err, tc.errorRead)

				err = cs.Store(registryURL, af)
				assert.ErrorIs(t, err, tc.errorWrite)

			})
		}
```
> The t.Cleanup calls are not "redundant". They are restoring things to what they were before the subtest started so that we do not pollute the next subtest (which is btw currently a recurring problem in our tests).


nope, the `t.Cleanup` func is only registered after each test subset but they are all executed after the parent test returns (please check the def of [Cleanup](https://pkg.go.dev/testing@go1.23.1#T.Cleanup))

Comment (apostasie) [reply pkg/imgutil/dockerconfigresolver/credentialsstore_test.go:186]: Thanks a lot @fahedouch 
Appreciate the clarification a lot.

This is definitely a very interesting (and important conversation).

So,

```
func TestTest(t *testing.T) {

	t.Cleanup(func() {
		fmt.Println("After main test")
		assert.Assert(t, false)
	})

	t.Run("subtest1", func(tt *testing.T) {
		fmt.Println("Inside subtest1")
		tt.Cleanup(func() {
			fmt.Println("After subtest1")
		})
	})

	t.Run("subtest2", func(tt *testing.T) {
		fmt.Println("Inside subtest2")
		tt.Cleanup(func() {
			fmt.Println("After subtest2")
		})
	})

	fmt.Println("About to end main test")
}
```

Will output (expectedly):
```
Inside subtest1
After subtest1
Inside subtest2
After subtest2
About to end main test
After main test
```

This above is what I intended (cleanup executed after each subtest) - and how my test is actually behaving... but... the catch is...

The fact that my code is shadowing `t` makes things extra confusing to read (I will update my PR to not do that and use `tt` instead).

apostasie: Also, apologies for the delay in answering here. I was out for a few days.
Cheers.

apostasie: @fahedouch latest push does clarify intent by no longer shadowing `t` (we really need better test tooling...).

Let me know if the intent is now clearer.

Thanks again!

fahedouch: thank you, it's clearer now!

fahedouch (approved):
Thanks

Issue #3265, opened 2024-07-31
[Design discussion] docker credentials and hosts.toml


apostasie:
### What is the problem you're trying to solve

Currently, we retrieve credentials from the traditional docker location.
Docker has no notion of "hosts/mirrors", and credentials are stored per host/port.

On the other hand, we support containerd hosts.toml (to some extent).

Currently, when we are faced with things like that:
```
server = "http://foo.com"

[host."http://localhost:5000"]
  capabilities = ["pull", "resolve"]
```

We will:
- first retrieve credentials associated with foo.com, if any
- then get the hosts list from hosts.toml
- try the host, REFUSE to send the credentials as the host does not match (this is good!) and fail

So, it is impossible to actually use hosts.toml with authentication.

The fundamental problem here is that we are trying to reconcile conceptually different worlds (the docker creds world and the hosts toml world).

My suggestion here would be the following:

1. `nerdctl login foo.com`
  - retrieve docker credentials for `foo.com`
  - do NOT try to contact any of the configured hosts - just contact `foo.com`
  - if there are any host configured for `foo.com`
     - IF contacting foo.com FAILED (airgap scenario for example), provide some meaningful explanation and suggest them that they should try login against these hosts (localhost:5000 in this example)
     - IF login in foo.com did SUCCEED, still output a gentle message to the user that foo.com is currently configured to resolve to `localhost:5000`, and that we are NOT authenticated against it (and if they want to do so, they should `nerctl login localhost:5000`)
2. `nerdctl pull/push foo.com/bla`
  - use the current flow (resolve hosts.toml, try the hosts in order)
  - for each host, do retrieve the credentials for that host

I appreciate this is a tricky change, with wide ranging impact, but it seems to me that currently, we simply do not have a correct solution for mirrors requiring authentication.

Maybe I am missing something here... please point it out if so :-)

Thanks a lot folks!

Taggin @fahedouch @AkihiroSuda and any of the other good folks here.

### Describe the solution you'd like

na

apostasie:
I now believe it is not possible to reconcile docker credentials and hosts.toml spec.
Specifically, a given `server` or `host` section may be used multiple times for different registry namespaces.
These hosts will get used with the appended `?ns=NAMESPACE` query parameter, which allows the server to understand for which namespace they are being queried.

Docker credentials only saves credentials per host+port, so, we cannot have multiple credentials sets per host (eg: along namespaces).

What I am thinking we can do:

Plan A. 

[...]

PR #3417, opened 2024-09-06
[Proposal] design document for authentication rework

apostasie:
Our current implementation of login suffers from structural issues manifest in a variety of ways (see https://github.com/containerd/nerdctl/issues/3072), and is vastly incomplete when it comes to supporting hosts.toml properly.

The core issue is that it predates these mechanisms, and has been architected with different requirements in mind that need to be re-evaluated today, especially in light of upcoming features like oauth device-flow.

First and foremost though, it lacks a lot of clarity about what is supposed to happen when, in a context where a lot of legacy features have complex interactions and effects:
- the insecure-registry flag
- the "default" behaviors for localhost
- the use of the docker credentials store to retrieve credentials 

... to a point that even seasoned contributors are confused as to whether we should allow a scheme to be passed in front of a registry url when we do login, or what exactly should happen in term of downgrading protocols or implied port and schemes.

Besides the bugs, reconciling these legacy features, the requirement to use docker credentials, and being able to actually use hosts.toml does require significant changes to our login and dockerconfigresolver code - aka: a full rewrite.

Some of these details are outlined in https://github.com/containerd/nerdctl/issues/3265 - though this is not the place for a proper reference - and some of the groundwork for this rewrite have been recently merged.

This document offers solutions, and describes precisely what happens when, and how things interact together. It is not a replacement for the hosts.toml slapstick spec, but rather a description of what we want to achieve with the upcoming rewrite.

While there is a lot of important details in there that should not be overlooked, the key parts of the proposal are the introduction of a new nerdctl login flag allowing to authenticate with endpoints, an experimental scheme for storing endpoint credentials, and clarification on the role and use of --registry-flag.

Given the extent of the changes to come - that will be very hard to break down into small pieces - I felt like it was better to start here and making sure we are aligned on the vision.

Cheers.

- A.


apostasie:
Oh project checks... I just hate you.

docs/dev/auth_design.md
```diff
@@ -0,0 +1,334 @@
...
+the docker hub registry which uses a fully qualified URL.
+
+Since docker does not support `hosts.toml` and since _endpoints_ are not
+the same thing as an implied registry host for a namespace, we store
+_endpoint_ credentials using a different schema.
+
+Docker will not recognize this schema, hence will not wrongly send these
+credentials when trying to log in into a known _endpoint_ as a registry.
+
+The schema is: `nerdctl-experimental://namespace.example:123/?endpoint=myserver.example:456`
```
AkihiroSuda, line 329: Where is this expected to be stored?

docs/dev/auth_design.md
```diff
@@ -0,0 +1,334 @@
...
+
+As clearly shown above, this is currently experimental, and is subject to change
```
AkihiroSuda, line 331: The document should draw a clear line on what is already (experimentally) implemented and what is being newly proposed

docs/dev/auth_design.md
apostasie: Inside the normal docker config location.

docs/dev/auth_design.md
apostasie: TBH part of the problem here is that it is unclear what we are supporting _right now_.

For example, we support bits and pieces of hosts.toml, but are failing on others.
Furthermore, our `insecure` resolution / downgrade is just baroque, and the --insecure-registry flag works in certain circumstances and not other...

I'll see what I can do to make things more clear.

docs/dev/auth_design.md
AkihiroSuda: Does Docker allow it?

apostasie: Yes, it will just do nothing with it.

docs/dev/auth_design.md
apostasie: Done.

apostasie:
@fahedouch would love your opinion here overall.

The take-away really is:
- spelling out loud stuff that is tribal knowledge right now, and making it clear what we do wrt scheme, port, localhost behavior
- proposing a way to deal with authentication when hosts.toml are involved (that is the new nerdctl flag and docker credentials scheme)

apostasie:
@AkihiroSuda once I get the testing/CI work done, is it fine with you that I move to implementation of this?

It might be simpler for people to discuss with actual code?

apostasie:
Closing for now.

PR #3434, opened 2024-09-18
Fix panic 'provided file is not a console'

### LINKED ISSUE #3433: Panic: `not a console`, follow-up to #3297
### Description

As outlined in #3297, there are more places that are likely to panic when we do not have a tty.

While the patch in #3300 fixed it for `exec` for a specific use case, clearly `run` can crash in the same circumstances.

Suggestion is to wrap the third party term console library so that it returns an error we can check instead of crashing.

na

na

main

None

### Host information

_No response_

apostasie:
See commit message.

Fixes #3433

This should put this panic issue to rest.

### DIFF
--- a/cmd/nerdctl/container/container_exec.go
+++ b/cmd/nerdctl/container/container_exec.go
@@ -18,9 +18,7 @@ package container
 
 import (
 	"errors"
-	"os"
 
-	"github.com/moby/term"
 	"github.com/spf13/cobra"
 
 	containerd "github.com/containerd/containerd/v2/client"
@@ -58,6 +56,7 @@ func NewExecCommand() *cobra.Command {
 }
 
 func processExecCommandOptions(cmd *cobra.Command) (types.ContainerExecOptions, error) {
+	// We do not check if we have a terminal here, as container.Exec calling console.Current will ensure that
 	globalOptions, err := helpers.ProcessRootCmdFlags(cmd)
 	if err != nil {
 		return types.ContainerExecOptions{}, err
@@ -85,13 +84,6 @@ func processExecCommandOptions(cmd *cobra.Command) (types.ContainerExecOptions, 
 	if flagT {
 		if flagD {
 			return types.ContainerExecOptions{}, errors.New("currently flag -t and -d cannot be specified together (FIXME)")
-		}
-	}
-
-	_, isTerminal := term.GetFdInfo(os.Stdin)
-	if !flagD {
-		if flagT && flagI && !isTerminal {
-			return types.ContainerExecOptions{}, errors.New("the input device is not a TTY")
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


apostasie:
Ping @AkihiroSuda - pending green CI.

apostasie:
Ahhhh windows, windows.

@sondavidb @jsturtevant any idea for me on this?

sondavidb:
Was trying to set up containerd on my local machine and it has been a lot more difficult than I anticipated 😓

Looks like the change you made had the integration pass but not the canary. Is this test known to be flaky? Looking through the changes it seems fine so I'm pretty stumped as well

apostasie:
@sondavidb I think `TestRunWithTtyAndDetached` is failing very often - but may succeed occasionally.

Might be a recent regression, and unrelated to this changeset, as my other PRs get the failure too.

Investigating...

apostasie:
@sondavidb maybe a side-effect of https://github.com/containerd/nerdctl/pull/3304 which did move that test around?

If there is some racyness involved, moving it in a different place may produce more adverse timing conditions?

Anyhow.

Ping @AkihiroSuda @jsturtevant as it seems to fail very often.

apostasie:
Test is broken on main.

Moving discussion to #3437

apostasie:
@AkihiroSuda besides what appears to be an unrelated regression / flakyness (#3437) this is ready.

pkg/consoleutil/consoleutil.go
```diff
@@ -18,8 +18,22 @@ package consoleutil
 
 import (
 	"context"
+	"os"
+
+	"github.com/containerd/console"
 )
 
+// Current is from https://github.com/containerd/console/blob/main/console.go#L68-L80
```
AkihiroSuda, line 26: s/main/<specific version>/

cmd/nerdctl/container/container_exec.go
```diff
@@ -88,13 +86,6 @@ func processExecCommandOptions(cmd *cobra.Command) (types.ContainerExecOptions,
 		}
 	}
 
-	_, isTerminal := term.GetFdInfo(os.Stdin)
-	if !flagD {
-		if flagT && flagI && !isTerminal {
-			return types.ContainerExecOptions{}, errors.New("the input device is not a TTY")
-		}
-	}
-
```
AkihiroSuda, line 97: Maybe add a comment line to explain why there is no need to check `isTerminal`

apostasie: Done

pkg/consoleutil/consoleutil.go
apostasie: Done.

AkihiroSuda (approved):
Thanks

PR #3441, opened 2024-09-20
Update gpu.md: add example command for the Nvidia driver is installed by the gpu-operator

zgfh:
Update gpu.md: add example command for the Nvidia driver is installed by the gpu-operator

fix: https://github.com/containerd/nerdctl/issues/2755

### DIFF
--- a/docs/gpu.md
+++ b/docs/gpu.md
@@ -82,4 +82,7 @@ echo "/run/nvidia/driver/usr/lib/x86_64-linux-gnu" > /etc/ld.so.conf.d/nvidia.co
 ldconfig
 ```
 
-And then, the `nerdctl run --gpus` can run successfully.
+And then, the `nerdctl run --gpus` can run successfully. eg:
+```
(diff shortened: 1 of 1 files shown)


yankay (approved):

docs/gpu.md
```diff
@@ -82,4 +82,7 @@ echo "/run/nvidia/driver/usr/lib/x86_64-linux-gnu" > /etc/ld.so.conf.d/nvidia.co
 ldconfig
 ```
 
-And then, the `nerdctl run --gpus` can run successfully.
+And then, the `nerdctl run --gpus` can run successfully. eg:
+```
+nerdctl run -it --rm --runtime=/usr/local/nvidia/toolkit/nvidia-container-runtime --gpus all nvidia/cuda:12.3.1-base-ubuntu20.04 nvidia-smi
```
AkihiroSuda, line 87: `--runtime` should not be needed on the "typical" setup?
(When `nvidia-container-runtime` is present in `$PATH`?)

zgfh: i have try ,this not work with no param runtime

image

AkihiroSuda: @ktock Can you confirm? Is this a bug?

ktock: @zgfh Can you confirm `nvidia-container-cli` is in PATH?

PR #3443, opened 2024-09-20
Enable other containers to join network namespace of the none network

Shubhranshu153:
## Description

There is a use case with pause containers, where other containers attaches to the pause container network. The pause container is launched with network none. This requires the pause container have a copy of hosts, hostname and resolv conf.
It also seems to share a net namespace, the containers must also share a user namespace.

The solution is to have a copy of the hosts/hostname and resolv.conf. In case of container network, add userns and netns both.

Want to confirm is this an acceptable solution and i can send out an PR for it.
Steps to reproduce the issue

    Create a pause container with network none.
    Create another container with --net container:

## Describe the results you received and expected

It would display errors with resolv.conf not found and once those configs are added would see an error with sys fs.
Expected result is to be able to connect to the network of pause container.
What version of nerdctl are you using?

1.7.5
Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None
Host information

lima vm (fedora image), but can be reproduced in any architecture.

### DIFF
--- a/cmd/nerdctl/container/container_run_network_linux_test.go
+++ b/cmd/nerdctl/container/container_run_network_linux_test.go
@@ -24,7 +24,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
> Want to confirm is this an acceptable solution 

Yes if it works with Docker.

Needs an integration test.

pkg/containerutil/container_network_manager.go
```diff
@@ -175,9 +175,64 @@ func (m *noneNetworkManager) InternalNetworkingOptionLabels(_ context.Context) (
...
+	content, err := os.ReadFile("/etc/hosts")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	etcHostsPath, err := hs.AllocHostsFile(containerID, content)
+	if err != nil {
+		return nil, nil, err
+	}
+	copyFileContent("/etc/hosts", etcHostsPath)
```
apostasie, line 206: /etc/hosts content has already been writen on line 202.

Shubhranshu153: ack

pkg/containerutil/container_network_manager.go
```diff
@@ -175,9 +175,64 @@ func (m *noneNetworkManager) InternalNetworkingOptionLabels(_ context.Context) (
...
+		return nil, nil, err
+	}
+
+	stateDir, err := ContainerStateDirPath(m.globalOptions.Namespace, dataStore, containerID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	resolvConfPath := filepath.Join(stateDir, "resolv.conf")
+	copyFileContent("/etc/resolv.conf", resolvConfPath)
```
apostasie, line 190: That would make containers using `none` to now use the host `resolv.conf`, right?
This does not seem right.

pkg/containerutil/container_network_manager.go
```diff
@@ -175,9 +175,64 @@ func (m *noneNetworkManager) InternalNetworkingOptionLabels(_ context.Context) (
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	content, err := os.ReadFile("/etc/hosts")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	etcHostsPath, err := hs.AllocHostsFile(containerID, content)
```
apostasie, line 202: Same comment as above. It does not seem like `none` should have the host `/etc/hosts` content...

pkg/containerutil/container_network_manager.go
Shubhranshu153: checked with docker and they have it for none, personally i dont think it should be part of none, but if it is not attaching to a none network namespace (custom use cases) seems to throw error.

apostasie: It does not look like they have it here...

`docker run --net none debian cat /etc/resolv.conf`

```
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.

nameserver 192.168.5.2
search .

# Based on host file: '/run/systemd/resolve/resolv.conf' (legacy)
# Overrides: []
```


vs.

`cat /etc/resolv.conf`

```
nameserver 127.0.0.53
options edns0 trust-ad
search .
```

Shubhranshu153: this is my output (docker 25)
docker run --network none alpine:latest cat /etc/resolv.conf 
<redacted>
nameserver 10.4.4.10
options timeout:1 attempts:2

```
 sudo docker run --network none alpine:latest cat /etc/hostname
9d25f3a65441

```

Shubhranshu153: to be fair i tested in another system i got same as yours too, so not sure which one is correct,  but for some system is taking from /run/systemd/resolv/resolv.conf?

apostasie: It likely does not matter much, since there will be no network anyhow. Concern here is more about information leakage.
I would suggest we either leave it empty or mimic the behavior of cniNetworkManager

Shubhranshu153: not sure if information leak is a concern in network none, if one can ssh from host they have access to /etc/resolv.conf of the host anyway. 

I like the idea of mimicing the behavior of cniNetworkManager, that seems to take into account adding host, dns  etc. 

while we are on the topic, if we use --network container:containerd id with 
```
		nameServers   = m.netOpts.DNSServers
		searchDomains = m.netOpts.DNSSearchDomains
		dnsOptions    = m.netOpts.DNSResolvConfOptions
```

dont we want to add them to the resolv.conf, similarly if --network host with these options, not sure they are getting added

```
[shubhum@lima-finch nerdctl]$ sudo nerdctl run  --network host --hostname default --dns test --dns-search testname alpine:latest cat /etc/resolv.conf
# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8).
# Do not edit.
#
# This file might be symlinked as /etc/resolv.conf. If you're looking at
# /etc/resolv.conf and seeing this text, you have followed the symlink.
#
# This is a dynamic resolv.conf file for connecting local clients to the
# internal DNS stub resolver of systemd-resolved. This file lists all
# configured search domains.
#
# Run "resolvectl status" to see details about the uplink DNS servers
# currently in use.
#
# Third party programs should typically not access this file directly, but only
# through the symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a
# different way, replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 127.0.0.53
options edns0 trust-ad
search ant.amazon.com amazon.com
```

Shubhranshu153: changed this configs to some generic default values

pkg/containerutil/container_network_manager.go
Shubhranshu153: changed this configs to some generic default values

apostasie:
@Shubhranshu153 will be slow to review - travelling right now. Will do asap.

pkg/containerutil/container_network_manager.go
```diff
@@ -150,6 +150,11 @@ type noneNetworkManager struct {
 	client        *containerd.Client
 }
 
+func WriteDefaultResolvConf(filePath string) error {
```
apostasie, line 153: I think we should use the existing `resolvconf.Build`.
It is not currently safe to use it (not atomic, nor locking), but it is better IMO than duplicating the problem.

pkg/containerutil/container_network_manager.go
```diff
@@ -179,9 +184,62 @@ func (m *noneNetworkManager) InternalNetworkingOptionLabels(_ context.Context) (
...
+		return nil, nil, err
+	}
+
+	stateDir, err := ContainerStateDirPath(m.globalOptions.Namespace, dataStore, containerID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	resolvConfPath := filepath.Join(stateDir, "resolv.conf")
+	if err := WriteDefaultResolvConf(resolvConfPath); err != nil {
```
apostasie, line 200: See note above.

apostasie (commented):
Left a note about `WriteDefaultResolveConf`.
Otherwise, these changes seem good and will bring us more in line with what docker does, so +1.

This needs tests - especially around your use case of another container using the none container net.

Shubhranshu153:
will add tests.

Shubhranshu153:
> Left a note about `WriteDefaultResolveConf`. Otherwise, these changes seem good and will bring us more in line with what docker does, so +1.
> 
> This needs tests - especially around your use case of another container using the none container net.

so overall behaviour wise it is like docker but the configs dont still match, which is probably fine as we dont expose host details if we add specifically the things we want to expose.

Shubhranshu153:
@apostasie added the tests, for the configs i have made it docker non compat because of the explanation above.
TAL. 
Thanks.

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -504,6 +505,37 @@ func TestSharedNetworkStack(t *testing.T) {
...
 }
 
+func TestSharedNetworkWithNone(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		t.Skip("--network=container:<container name|id> only supports linux now")
+	}
+	base := testutil.NewBase(t)
+
+	containerName := testutil.Identifier(t)
+	defer base.Cmd("rm", "-f", containerName).AssertOK()
```
apostasie, line 515: Generally speaking:
- favor t.Cleanup to defer
- do also run your cleanup routine before the test itself, so that local repeat runs do not fail if there are leftover from prior runs

Alternatively, you may want to use the new [test framework](https://github.com/containerd/nerdctl/blob/main/docs/testing/tools.md) that is designed to take care of these.

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -712,3 +748,42 @@ func TestRunContainerWithStaticIP6(t *testing.T) {
...
+
+	// If running on Linux, verify /etc/hostname is correctly set
+	if runtime.GOOS == "linux" {
+		containerHostName := "testcontainer"
+		cmd := base.Cmd("run", "--rm", "--net", "none", "--hostname", containerHostName, testutil.CommonImage, "cat", "/etc/hostname")
+		output := cmd.Run().Combined()
+		hostname := strings.TrimSpace(output)
+
+		if len(containerHostName) > 12 {
+			require.Equal(t, containerHostName[:12], hostname[:12])
```
apostasie, line 771: assert.Equal instead? Not sure why we would want testity

go.mod
```diff
@@ -61,6 +61,7 @@ require (
 	github.com/rootless-containers/rootlesskit/v2 v2.3.1
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.9.0
```
apostasie, line 64: See note above. I am not sure why we would want to add testify.

apostasie:
> @apostasie added the tests, for the configs i have made it docker non compat because of the explanation above. PTAL. Thanks.

Thanks @Shubhranshu153 

I left a few comments on the tests. I will give this a spin locally later today too, but otherwise looks good.

cmd/nerdctl/container/container_run_network_linux_test.go
Shubhranshu153: it is just prior to the run for the statement for running the continous container. (sleep infinity)

cmd/nerdctl/container/container_run_network_linux_test.go
Shubhranshu153: thanks will update it.

go.mod
Shubhranshu153: will remove it.

Shubhranshu153:
@apostasie PTAL
Not sure why FreeBSD is failing

Shubhranshu153:
@AkihiroSuda @apostasie
need to add some test but need some guidance here
In Docker you can add something like:
```
docker run --rm   --dns 8.8.8.8   --dns 8.8.4.4 --add-host somewhat.example.com:127.0.0.1   --dns-search example.com  --network none  -it ubuntu:latest
```
and you can get them added

```
dev-dsk-shubhum-2b-5b633f8c % docker run --rm   --dns 8.8.8.8   --dns 8.8.4.4 --add-host somewhat.example.com:127.0.0.1   --dns-search example.com  --network none  -it ubuntu:latest cat /etc/hosts
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
127.0.0.1       somewhat.example.com

(24-10-31 0:01:58) <0> [~]  
dev-dsk-shubhum-2b-5b633f8c % docker run --rm   --dns 8.8.8.8   --dns 8.8.4.4 --add-host somewhat.example.com:127.0.0.1   --dns-search example.com  --network none  -it ubuntu:latest cat /etc/resolv.conf 
search example.com
nameserver 8.8.8.8
nameserver 8.8.4.4
options timeout:1 attempts:2
```
So added these functionalities to none network, want some guidance if am missing functionality wise, i reused the code from host network as it seems it just used values from netopts. But i plan to refactor it to a single function (cleanup wise)

apostasie:
Will have a look as time allows.
Thanks @Shubhranshu153

Shubhranshu153:
@AkihiroSuda @apostasie 
now that nerdctl 2.0 is out want to check if i can get some feedback on the above proposal. 
Idea is to support the flags and behaviour as close as possible to docker.
Thank You

go.mod
```diff
@@ -124,6 +124,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/stefanberger/go-pkcs11uri v0.0.0-20230803200340-78284954bff6 // indirect
+	github.com/stretchr/objx v0.5.2 // indirect
```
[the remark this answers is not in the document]
Shubhranshu153: Need to run go mod tidy use the nerdctl test framework and fix a few things in code, just checking if my logic is correct to add the options DNS and host options for none network and probably need to be done for host network as well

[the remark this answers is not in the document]
apostasie: I believe you added it initially with https://github.com/containerd/nerdctl/pull/3443#pullrequestreview-2404666593

But really it should not be necessary.

Can you try removing the line from go.mod then doing go mod tidy again?

Will look at the rest later today.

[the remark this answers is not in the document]
Shubhranshu153: Thank you, the current version is more for making sure the direction is correct. I will be having another pr using the new test framework and updates for the host network config.

apostasie:
@Shubhranshu153 

I am a bit confused about the thing overall.

```
docker run -ti --net container:paused debian bash
docker: Error response from daemon: cannot join network of a non running container: c66b2c44df029075540e67ceb7ed1de63ee89f98baad49d823eeeab31dc1e08a.
ERRO[0000] error waiting for container: context canceled
```

So, Docker does not seem to allow starting a container using the network of a paused container - which does make sense.

apostasie:
And sorry for the late review.
Was busy elsewhere lately.

Shubhranshu153:
> @Shubhranshu153
> 
> I am a bit confused about the thing overall.
> 
> ```
> docker run -ti --net container:paused debian bash
> docker: Error response from daemon: cannot join network of a non running container: c66b2c44df029075540e67ceb7ed1de63ee89f98baad49d823eeeab31dc1e08a.
> ERRO[0000] error waiting for container: context canceled
> ```
> 
> So, Docker does not seem to allow starting a container using the network of a paused container - which does make sense.
> 
> We should still support using a started container with --net none as a network for another container though (doesn't seem to work currently).

The container i am trying to connect is a pause container (the container is in running state) which is basically sleep infinity container with a non network. From what I understand it is used to create a network namespace to which other containers can join and share.

apostasie:
Sorry for the confusion :-)

I fought by "pause container" you meant "created".

pkg/containerutil/container_network_manager.go
```diff
@@ -179,9 +248,75 @@ func (m *noneNetworkManager) InternalNetworkingOptionLabels(_ context.Context) (
 
 // ContainerNetworkingOpts Returns a slice of `oci.SpecOpts` and `containerd.NewContainerOpts` which represent
 // the network specs which need to be applied to the container with the given ID.
-func (m *noneNetworkManager) ContainerNetworkingOpts(_ context.Context, _ string) ([]oci.SpecOpts, []containerd.NewContainerOpts, error) {
+func (m *noneNetworkManager) ContainerNetworkingOpts(_ context.Context, containerID string) ([]oci.SpecOpts, []containerd.NewContainerOpts, error) {
```
apostasie, line 251: I am increasingly uncomfortable with the amount of duplication between `ContainerNetworkingOpts` implementations (and even more so about the differences in there). But then, refactoring that should probably be a separate endeavor.

Shubhranshu153: Not really as I was writing it I too saw that so I plan to refactor it but before that just wanted to confirm the direction so that the effort is not put in the wrong place

apostasie:
@Shubhranshu153 I think directionally that is ok.

Though, I would suggest really strong testing on this in making sure we are truly aligned with Docker:
- specifically test what the content of /etc/hosts, /etc/resolv.conf and /etc/hostname are for `none` containers and `container:foo` containers, to make sure we are fully aligned with docker

Also, I am wondering if hosts files are now going to updated for `none` network containers - I do not think that should happen, as:
- this has a (small?) performance angle
- part of the reason for running `none` containers is complete networking isolation from the other containers and host - so, let's make sure we do not break that promise by leaking networking details from other containers info in there

apostasie:
As I said in my comment, I am also concerned about the amount of duplication / differences between network managers (specifically ContainerNetworkingOpts).

We really need to clean that stuff up eventually and revise our approach there.

Shubhranshu153:
Yes testing needs to be rigorous not to break the promise of isolation as in that  case it will be sec vulnerability. 

Will send out a revision once I clean and have some robust testing in place.

Thanks

AkihiroSuda:
What's the current status of this?

Shubhranshu153:
@AkihiroSuda 
The PR requires a some refactor and additional testing for some of the cases. 
Will get out another revision shortly for another review.

Have been out for quite sometime due to family health reasons.

Thank You.

pkg/containerutil/container_network_manager.go
```diff
@@ -553,7 +742,7 @@ func (m *hostNetworkManager) ContainerNetworkingOpts(_ context.Context, containe
 	}
 
 	// `/etc/hostname` does not exist on FreeBSD
-	if runtime.GOOS == "linux" && m.netOpts.UTSNamespace != UtsNamespaceHost {
+	if runtime.GOOS == "linux" {
```
Shubhranshu153, line 745: uts namespace is already checked

Shubhranshu153:
@apostasie  
Updated the code and added the following
1. DNS options, hostname checks, UTS checks for network none, network host and shared network
2. Checked for docker compatibility for the above
3. Moved some of the test to new framework
4. Few changes in the network setup for managing docker incompatibility/bugs

Let me know if anything needs to be added. If incremental additional bug fixes/modification (not related to this PR)
 are required i can add it in a separate pr

Shubhranshu153:
@AkihiroSuda @apostasie 
Checking if someone can take a look at the pr. The pr contains:

1. DNS options, hostname checks, UTS checks for network none, network host and shared network
2. Checked for docker compatibility for the above
3. Moved the test to new framework
4. Few changes in the network setup for managing docker incompatibility/bugs
5. A minor fix for test consistency.

Some more refactor can be done for the Setup and cleanup scripts but will like to take it to a new pr. Let me know if you want   anything to be added.

Thank You.

Shubhranshu153:
@AkihiroSuda 
Hi Akihiro,
Checking if anyone might be available to review the pr. 
Thank You.

AkihiroSuda (approved):
Thanks

PR #3446, opened 2024-09-21
Prevent concurrency on commit and remove

### LINKED ISSUE #1391: Failed to execute nerdctl commit command at the same time
k8s 1.21.9 containerd 1.3.7

If two commit command are performed at the same time, one succeeds but the other one fails with a digest already exists error message.

Committed different pods created with the same base image, and assigned different image names and tag.

time="2022-08-29T16:40:14+09:00" level=warning msg="Image lacks label \"nerdctl/platform\", assuming the platform to be \"linux/amd64\""
time="2022-08-29T16:40:14+09:00" level=fatal msg="failed to export layer: snapshot \"sha256:738a919f97cf7edfe7a17bbc6058ab8a79539400e6f6c084ddd4c0a2829bbfb3-parent-view\": already exists"

### LINKED ISSUE #3440: Implement a locking mechanism for container operations that require persistent state

Concurrently running operations that require containers to remain "untouched" for a certain period of time are currently not working.

See for example:
https://github.com/containerd/nerdctl/issues/1391

So, clearly, `commit` is not safe. Probably `remove` as well - maybe others.

Suggestion here is to leverage the existing lifecycle store and start enforcing locking where need be.

na

### Additional context

_No response_

apostasie:
See commit message for some details.

- fix #1391
- fix #3440

### DIFF
--- a/cmd/nerdctl/container/container_remove_test.go
+++ b/cmd/nerdctl/container/container_remove_test.go
@@ -24,16 +24,18 @@ import (
 
 func TestRemoveContainer(t *testing.T) {
 	t.Parallel()
+
 	base := testutil.NewBase(t)
 	tID := testutil.Identifier(t)
 
 	// ignore error
 	base.Cmd("rm", tID, "-f").AssertOK()
 
-	base.Cmd("run", "-d", "--name", tID, testutil.CommonImage, "sleep", "infinity").AssertOK()
+	base.Cmd("run", "-d", "--name", tID, testutil.NginxAlpineImage).AssertOK()
 	defer base.Cmd("rm", tID, "-f").AssertOK()
 	base.Cmd("rm", tID).AssertFail()
 
-	base.Cmd("kill", tID).AssertOK()
+	// `kill` does return before the container actually stops
+	base.Cmd("stop", tID).AssertOK()
 	base.Cmd("rm", tID).AssertOK()
 }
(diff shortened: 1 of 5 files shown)


apostasie:
@AkihiroSuda green.

No tests for now.
I need time to think about and build a good set of helpers for concurrency testing.

cmd/nerdctl/container/container_remove_test.go
```diff
@@ -24,16 +24,18 @@ import (
...
 	t.Parallel()
+
 	base := testutil.NewBase(t)
 	tID := testutil.Identifier(t)
 
 	// ignore error
 	base.Cmd("rm", tID, "-f").AssertOK()
 
-	base.Cmd("run", "-d", "--name", tID, testutil.CommonImage, "sleep", "infinity").AssertOK()
+	base.Cmd("run", "-d", "--name", tID, testutil.NginxAlpineImage).AssertOK()
```
AkihiroSuda, line 34: ?

apostasie: The same test was duplicated for Windows and Linux (now merged).
The windows version was using `NginxAlpineImage` - the linux version was using `CommonImage`.
I picked the windows version.

AkihiroSuda (approved):
Thanks

apostasie:
@AkihiroSuda turns out this is breaking things when the container was not created by us.

Will submit a follow-up PR immediately to fix the regression.

PR #3449, opened 2024-09-21
Rename `containerd-rootless-setuptool.sh`

abitrolly:
`containerd-rootless-setuptool.sh` is now `containerd-rootless-systemd.sh`

To clarify what this tool does.

Closes #66.


fahedouch:
Hi @abitrolly, I don't think is a good idea to  suffix the script name by `systemd` as it setups multiple (buildkit, containerd..) tool using systemd. We may add some notes to doc.

abitrolly:
@fahedouch then it is probably `containerd` prefix that should be removed. Something like `setuptool-rootless-systemd.sh`.

AkihiroSuda:
For the sake of compatibility the script name shouldn’t be changed unless there is an extremely convincing reason.

Also, the script may potentially support non-systemd inits (if somebody works on it).

So I’m rejecting this PR.

Issue #3451, opened 2024-09-21
TestBuildIsShareableForCompatiblePlatform unclear what the intent is


apostasie:
### Description

`TestBuildIsShareableForCompatiblePlatform` relies on `AssertErrNotContains`, which apparently does not test stderr, but stdout.

Since there is never anything on stdout in that case, none of these asserts are doing anything.

Furthermore, the test on a different architecture will only pass if emulation is enabled, but fail otherwise.

Put otherwise, the test right now is doing two things:
- verify that we can build, which is already in another test
- verify that we can build on a different platform, which assumes emulation is installed

It is unclear to me what the test intent is.

Can we clarify?

Thanks.

Tagging @Shubhranshu153 

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
I will assume this was meant to test building on a different from the host but currently emulated platform, and rewrite the test accordingly.

Shubhranshu153:
i can fix it, AssertErrNotContains -> from the name i always assumed it checks stderr, probably a simple redirection of everything to stdout would fix it. 

the test checks if a cross build is happening  then a tarball is created. so if the test are running on amd then that portion would run and test but if its running on arm it would get skipped.

apostasie:
> i can fix it, AssertErrNotContains -> from the name i always assumed it checks stderr, probably a simple redirection of everything to stdout would fix it.

Probably this method should go. It does not really serve a purpose, since what we are checking here is that the build is successful (exitcode 0 should be sufficient).

> the test checks if a cross build is happening then a tarball is created. so if the test are running on amd then that portion would run and test but if its running on arm it would get skipped.

This is not what is happening right now.
If you are on _anything but_ armv7 (eg: arm64), the test will run, and _will fail_ if emulation is not installed (which is the default).
The test should first verify that we support building on something else than the current platform, then build for that (and we should remove the hardcoded `arm` check).

Shubhranshu153:
```
Probably this method should go. It does not really serve a purpose, since what we are checking here is that the build is successful (exitcode 0 should be sufficient)

```
 -thats not entirely correct, we are checking if the build was successful and export of the image is not a tarball for cases where we are not doing a cross platform build.

yeah agreed removing hardcoded stuff, not sure about a good way to test it out, let me check.

apostasie:
[...]

PR #3456, opened 2024-09-23
remove is shareable test

Shubhranshu153:
Based on discussions in https://github.com/containerd/nerdctl/issues/3451

### DIFF
--- a/cmd/nerdctl/builder/builder_build_test.go
+++ b/cmd/nerdctl/builder/builder_build_test.go
@@ -25,8 +25,6 @@ import (
 
 	"gotest.tools/v3/assert"
 
-	"github.com/containerd/platforms"
-
 	"github.com/containerd/nerdctl/v2/cmd/nerdctl/helpers"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 )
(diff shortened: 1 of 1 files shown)


Shubhranshu153:
tentatively created this pull request based on the conversation in https://github.com/containerd/nerdctl/issues/3451#issuecomment-2366912574

technically i dont see the risk as long as builds are successful and that is already tested. 
@apostasie

apostasie:
Thanks @Shubhranshu153 

As outlined in the ticket, I have no gripe against the existence of the test itself :-).

I am fine with removing it of course if it serves no purpose (haven't looked closely into unit tests around that).

Shubhranshu153:
The test is checking if the image built is shareable in case the target platform and the platform it is built on matches, Before the pr we used to tar and load it if the --platform option was provided even if the platform matches for the build system.

@AkihiroSuda  would like some suggestion weather to refactor this test to be more robust (and also fix the bug) or we want to remove it altogether. Unit test are there but the exact test is not there in E2E setup.

Shubhranshu153:
Checking back on this if we want to remove it or refactor it.
Thanks.

AkihiroSuda (approved):
Thanks

PR #3462, opened 2024-09-24
Do not print an error message on non-0 exec exit code

zouyee:
This was added with an earlier exec, and it is very confusing. the error had nothing to do with Containerd; it was the executable we ran inside the container that errored, and per task run convention we should set the Containerd exit code to the process's exit code and print no error.

ctr exec
```
 ctr --debug -n k8s.io task exec -t --exec-id nginx XXX bash -c  'bash ./faketime.sh 2024 9 22 22';echo $?
tomcat: stopped
tomcat: started
1
```

nerdctl exec
```
nerdctl -n k8s.io exec  xxx bash -c  'bash ./faketime.sh 2024 9 22 22';echo $?
tomcat: stopped
tomcat: started
FATA[0005] exec failed with exit code 1
1
```

### DIFF
--- a/pkg/cmd/container/exec.go
+++ b/pkg/cmd/container/exec.go
@@ -133,12 +133,9 @@ func execActionWithContainer(ctx context.Context, client *containerd.Client, con
 		return nil
 	}
 	status := <-statusC
-	code, _, err := status.Result()
+	_, _, err := status.Result()
 	if err != nil {
 		return err
-	}
-	if code != 0 {
-		return fmt.Errorf("exec failed with exit code %d", code)
 	}
 	return nil
 }


AkihiroSuda:
I believe this was purposely printed for consistency with Docker, but can't verify it on phone

apostasie:
FWIW:
```
docker exec -ti foo bash -c 'bash ./faketime.sh xxx'; echo $?
bash: ./faketime.sh: No such file or directory
127


nerdctl exec -ti bar bash -c 'bash ./faketime.sh xxx'; echo $?
bash: ./faketime.sh: No such file or directory
FATA[0000] exec failed with exit code 127
1
```

It looks like this may have been introduced in:
https://github.com/containerd/nerdctl/commit/fae1696fba90fb3b1d2bb9098ea102b6ca1e4859#diff-5878c73318981892177c543bc6b07cb0e70f9b3989b533f285ee51c0dc57bf05R202

Not clear what was the reason, as this is part of a refactor.

pkg/cmd/container/exec.go
```diff
@@ -133,13 +133,10 @@ func execActionWithContainer(ctx context.Context, client *containerd.Client, con
 		return nil
 	}
 	status := <-statusC
-	code, _, err := status.Result()
+	_, _, err = status.Result()
 	if err != nil {
 		return err
 	}
-	if code != 0 {
```
apostasie, line 140: @zouyee not displaying the message is one thing, but clearly we still need to error out.

apostasie (changes requested):

pkg/cmd/container/exec.go
zouyee: The executable we ran inside the container that errored, we need to print error ?

apostasie: We need to return an error from this method, otherwise nerdctl will exit 0.

zouyee: Why does exec function care about the results of execution inside the container?

AkihiroSuda: It should be just same as Docker

zouyee: nerdctl exec
```
nerdctl -n k8s.io exec  xxx bash -c  'bash ./faketime.sh xxx';echo $?
tomcat: stopped
tomcat: started
FATA[0005] exec failed with exit code 1
1
```

docker exec
```
docker exec xxx  bash -c  'bash ./faketime.sh xxx';echo $?
tomcat: stopped
tomcat: started
1
```

apostasie: Yes, we should do the same thing as docker.
With your patch right now, we no longer return exit 1 because you do not return a go error.

fahedouch: Agree with you @zouyee  you may write stderr to `/dev/null` as workaround

PR #3455, opened 2024-09-23
Test rework, part 2

apostasie:
This is a first follow-up to #3418, rewriting more tests to leverage the new tooling.

In almost all cases, this is "identical" with the previous state of affairs, as these changes are not meant to introduce new tests, and normally do not radically change the existing tests intention.

However, here or in upcoming follow-ups:
- "requirements" have been precisely reviewed
   - some tests that were previously not "private" are now properly isolated
   - some tests that were previously not running for Docker (or other previously overstated requirements) will now run
- all tests are now parallelized by default

I appreciate this is hard to review.
This is a full-on rewrite.
But then, we are talking _solely_ about tests here, with no code or logic change - and the proof is in the pudding: aka green CI.

I am splitting different subfolders tests in different commits.
As I am converting more tests, I will add more commits to this branch.

We should still merge this PR whenever convenient (pending green of course) - and I will just open more PRs - that should help keep this bite-size and avoid conflicting with main.

Issue #3437, opened 2024-09-18
[URGENT] Regression? TestRunWithTtyAndDetached is broken


apostasie:
### Description

See #3436 for a no-op PR that fails the test.

This has been failing a lot on all my recent PRs, but seems racy.

It is possibly a recent regression.

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Note: github is experiencing network issues rn, so.......

apostasie:
https://github.com/containerd/nerdctl/actions/runs/10950055624/job/30404542869?pr=3435#step:8:473

apostasie:
https://github.com/containerd/nerdctl/actions/runs/10965099031/job/30450151113?pr=3418#step:8:565

apostasie:
https://github.com/containerd/nerdctl/actions/runs/10968930365/job/30460923241?pr=3446#step:8:473

apostasie:
https://github.com/containerd/nerdctl/actions/runs/10974186024/job/30472417850?pr=3435#step:9:476

apostasie:
@AkihiroSuda this is a major pita.
I know it is just a test, but it is failing incredibly often which begs the question of whether the code is really working.

Tagging for milestone v2?

TinaMor:
@apostasie Have you started investigating this? I'll pick this up this week.

apostasie:
Hey @TinaMor 
Been merely poking the test w/o sleeps - no real work or new insight on my side (also I do not have a windows setup).

apostasie:
@TinaMor going to skip that test for the time being (#3541), until we figure out what's going on and can re-enable it.

PR #3435, opened 2024-09-18
Ensure all layers are here when tagging, committing, saving, or converting

apostasie:
See #3425 and linked issues for context.

- fix #3425
- fix #827 (commit)
- fix #2357 (save)
- fix #2327 (convert)
- fix #2506 (tag)
- fix #3026 (commit)

Unconfirmed (no test), but cannot reproduce anymore with the patch:
- fix #3420

Unconfirmed, but probably fixed as well:
- #2588 (untested)

Reviewers, most of this is test wrangling:
- enabling Kube test which was missing a few bits
- fixing broken image inspect tests that are now more likely to show leftovers

PR #3464, opened 2024-09-25
Test rework, part 3

apostasie:
On top of #3455 

Notably:
- first commit has a raft of evolutions to test tooling in preparation for more test rewrites
- second commit tentatively addresses flakyness of TestAttach
- third commit disables parallelism for Docker on prune and rmi tests

### DIFF
--- a/cmd/nerdctl/volume/volume_prune_linux_test.go
+++ b/cmd/nerdctl/volume/volume_prune_linux_test.go
@@ -56,54 +56,58 @@ func TestVolumePrune(t *testing.T) {
 	}
 
 	// This set must be marked as private, since we cannot prune without interacting with other tests.
-	testGroup := &test.Group{
-		{
-			Description: "prune anonymous only",
-			Require:     nerdtest.Private,
-			Command:     test.RunCommand("volume", "prune", "-f"),
-			Setup:       setup,
-			Cleanup:     cleanup,
-			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
-				return &test.Expected{
-					Output: test.All(
-						test.DoesNotContain(data.Get("anonIDBusy")),
-						test.Contains(data.Get("anonIDDangling")),
-						test.DoesNotContain(data.Get("namedBusy")),
-						test.DoesNotContain(data.Get("namedDangling")),
-						func(stdout string, info string, t *testing.T) {
-							helpers.Ensure("volume", "inspect", data.Get("anonIDBusy"))
-							helpers.Fail("volume", "inspect", data.Get("anonIDDangling"))
-							helpers.Ensure("volume", "inspect", data.Get("namedBusy"))
-							helpers.Ensure("volume", "inspect", data.Get("namedDangling"))
-						},
-					),
-				}
+	testCase := &test.Case{
+		Description: "Prune",
+		Require:     nerdtest.Private,
+		SubTests: []*test.Case{
+			{
+				Description: "prune anonymous only",
+				NoParallel:  true,
+				Command:     test.RunCommand("volume", "prune", "-f"),
+				Setup:       setup,
+				Cleanup:     cleanup,
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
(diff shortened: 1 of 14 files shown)


cmd/nerdctl/volume/volume_list_test.go
```diff
@@ -272,7 +272,7 @@ func TestVolumeLsFilter(t *testing.T) {
 				Description: "Retrieving name=volume1 and name=volume2",
 				// FIXME: https://github.com/containerd/nerdctl/issues/3452
 				// Nerdctl filter behavior is broken
-				Require: nerdtest.Docker,
+				Require: nerdtest.NerdctlNeedsFixing,
```
apostasie, line 275: This is the same requirement (alias-ed) - but we should use `NerdctlNeedsFixing` when a test is disabled for nerdctl because of a bug, to facilitate search.

cmd/nerdctl/volume/volume_prune_linux_test.go
```diff
@@ -56,54 +56,58 @@ func TestVolumePrune(t *testing.T) {
 	}
 
 	// This set must be marked as private, since we cannot prune without interacting with other tests.
-	testGroup := &test.Group{
```
apostasie, line 59: This pattern does not work here, as the `Group` will run in `Parallel` with other tests, breaking things for Docker.
It works with nerdctl because we are making the subtests pivate - for Docker, the subtests disable parallel, but since the parent stays parallel, it blows up.

Changing to a `Test` parent instead of a group, and disabling parallel on the subtests.

apostasie:
@AkihiroSuda let me rebase this so we get a clean view (and might amend something).
#3435 is more important to review either way (if you do have free cycles).

apostasie:
Hitting 429 with Docker Hub :-(

Otherwise, this is now rebased and ready.

AkihiroSuda:
Needs rebase

apostasie:
> Needs rebase

Will wait for #3483 to land as the CI is currently bust.

apostasie:
Rebased (w. CI bustage fix)

pkg/testutil/test/data.go
```diff
@@ -117,16 +116,10 @@ func (dt *data) getConfig() map[ConfigKey]ConfigValue {
 }
 
 func defaultIdentifierHashing(name string) string {
```
AkihiroSuda, line 118: Why change?

apostasie: I did hit a case where the message was over 76, which got shortened to a sha256.

Turned out distribution registry rejected it (as an image name), as there is an undocumented limitation rejecting valid image names that are `64-byte hexadecimal strings`...

apostasie: I will revert this to a readable form (the sha1 are hurting readability) - it takes a bit more work though to handle the len limitation while keeping readability and avoiding collisions.

apostasie:
Rebased.

CI is green (except the usual windows reg).

AkihiroSuda:
Do you plan to revert the hashing func?

https://github.com/containerd/nerdctl/pull/3464#discussion_r1783956620

apostasie:
> Do you plan to revert the hashing func?
> 
> [#3464 (comment)](https://github.com/containerd/nerdctl/pull/3464#discussion_r1783956620)

Yes - here is the next round version:
https://github.com/containerd/nerdctl/blob/5c240d780c7ea9e3d1ae14d7646def999c3801ca/pkg/testutil/test/data.go#L124-L141

This seems to bother you - if you prefer I can just backport it here right away - the thing is, this whole thing (rewrite test+tooling) is so huge that it has to be in incremental steps... I appreciate it does introduce a bit of churn, which is always unfortunate...

apostasie:
Making things simpler - just backported the new version of the hashing method.

Pending CI.

apostasie:
Windows failures are the usual TestRunWithTtyAndDetached aka #3437

AkihiroSuda (approved):
Thanks

Issue #3480, opened 2024-10-02
We should have a publish scripts for the testing images we need, and we should not depend on third party images


apostasie:
### Description

See #3479 for the last straw.

Our tests do depend on a number of images that we do not control - usually on Docker Hub.
There are good reasons to test Docker Hub - just not to use third party images.

Also, even for the images that do not change (hosted on ghcr.io), we do not have a way to repush them if we needed to.

Suggesting we create a nerdctl account on hub, and start pushing the images we need there.

@AkihiroSuda I can take care of this, but then I would need to transfer the account to you.
Or you take care of it directly instead.

Then I can come up with a script that would push everything we need and we get rid of the dependencies on `busybox` (etc.).

LMK.

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

na

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

AkihiroSuda:
We already have: https://github.com/stargz-containers/image-ci

apostasie:
> We already have: https://github.com/stargz-containers/image-ci ( Plain vanilla non-estargz images are mirrored too with `-org` suffix https://github.com/orgs/stargz-containers/packages/container/alpine/17932394?tag=3.15.3-org )

We still need images on the Hub to test hub behavior.

AkihiroSuda:
What behavior of hub?

apostasie:
> What behavior of hub?

Resolution of shortnames and authentication.
Generally, we should still make sure we work well with Hub (regardless of what we think of it).

I see your point though - we might be able to do what we need without making assumptions on the images content.

I'll see what I can do to yank out `busybox` out of problem tests and come back here eventually.

PR #3483, opened 2024-10-02
Fix broken TestImageInspectDifferentValidReferencesForTheSameImage

### LINKED ISSUE #3479: [URGENT] TestImageInspectDifferentValidReferencesForTheSameImage is broken on main
### Description

Test assumes busybox from Hub would not change semantics (eg: that `stable` and `latest` would always be the same image).
This changed a few hours ago.

All builds will be broken until we land a fix.

na

na

main

None

### Host information

@AkihiroSuda am on it.
Can you assign this to me?

apostasie:
Fix #3479

Reviewers: since the official `busybox` is a moving target, and has just shifted under our feet, I replaced it with a copy under `apostasie/busybox`.

This is a short term fix so that the CI does not stay bust - but I do not like the idea of having images in my personal account.

(Part of) the current intent of this test was apparently to test variants of image names only available on the hub.

To get rid of my images here, here are our options:
- trash the test, and reduce the scope to not use short names
- implement hosts.toml properly, so that we can start using ghcr with authentication as a mirror to Hub (in progress, but will take time)
- #3480

### DIFF
--- a/cmd/nerdctl/image/image_inspect_test.go
+++ b/cmd/nerdctl/image/image_inspect_test.go
@@ -18,7 +18,6 @@ package image
 
 import (
 	"encoding/json"
-	"runtime"
 	"strings"
 	"testing"
 
@@ -71,16 +70,12 @@ func TestImageInspectSimpleCases(t *testing.T) {
 func TestImageInspectDifferentValidReferencesForTheSameImage(t *testing.T) {
 	nerdtest.Setup()
 
-	platform := runtime.GOOS + "/" + runtime.GOARCH
-
 	tags := []string{
 		"",
 		":latest",
-		":stable",
 	}
 	names := []string{
 		"busybox",
-		"library/busybox",
 		"docker.io/library/busybox",
 		"registry-1.docker.io/library/busybox",
 	}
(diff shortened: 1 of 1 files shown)


apostasie:
cc @AkihiroSuda

Tagging @manugupt1 

Review at your convenience.
CI failures are a Github Actions fluke, and the usual IPFSCompose.

cmd/nerdctl/image/image_inspect_test.go
```diff
@@ -94,16 +90,16 @@ func TestImageInspectDifferentValidReferencesForTheSameImage(t *testing.T) {
 			nerdtest.Private,
 		),
 		Setup: func(data test.Data, helpers test.Helpers) {
-			helpers.Ensure("pull", "alpine", "--platform", platform)
-			helpers.Ensure("pull", "busybox", "--platform", platform)
-			helpers.Ensure("pull", "busybox:stable", "--platform", platform)
-			helpers.Ensure("pull", "registry-1.docker.io/library/busybox", "--platform", platform)
-			helpers.Ensure("pull", "registry-1.docker.io/library/busybox:stable", "--platform", platform)
```
AkihiroSuda, line 101: Can we just remove this line and then keep using the official image?

apostasie: Sure, we can.
That is not just this line though - stable is used throughout.

cmd/nerdctl/image/image_inspect_test.go
```diff
@@ -71,16 +70,12 @@ func TestImageInspectSimpleCases(t *testing.T) {
 func TestImageInspectDifferentValidReferencesForTheSameImage(t *testing.T) {
 	nerdtest.Setup()
 
-	platform := runtime.GOOS + "/" + runtime.GOARCH
```
apostasie, line 74: Earlier version of this test was using linux/amd64, because it was hardcoding the sha.
When I rewrote the test a few weeks ago, I removed the static sha and replaced by a lookup, allowing to use whatever platform - though I did not remove the platform mechanism itself for some reason.
It is unnecesary.

AkihiroSuda:
Could you update the commit message to explain why `:stable` is no longer testable?

apostasie:
CI is green-ish (Docker failure is unrelate and will get fixed in #3464).

apostasie:
> Could you update the commit message to explain why `:stable` is no longer testable?

Doing now.

AkihiroSuda (approved):
Thanks

PR #3484, opened 2024-10-02
Fix login persistence regression against Hub

apostasie:
Login no longer persists for Docker Hub, as a mistake in the code was not using the special case Docker Hub key to save in the store.

Incidentally, we should start testing more against Hub generally to avoid things like that.
That implies creating a test account over there, and issuing a read-only token to be added here.


apostasie:
@AkihiroSuda this one is ready.

AkihiroSuda (approved):
Thanks

PR #3293, opened 2024-08-11
Login & authentication rework, part 1: credentials management and URL parsing 

apostasie:
This is a breakout of #3249 and first part of fixing our credentials management / login issues (see 3249 for details about issues, including #3072 which is the reason a significant rewrite is warranted IMHO).

The focus is this PR is solely:
- providing a CredentialsStore abstraction that takes care of any magic / legacy identifiers
- handling all registry url parsing & scheme / url monkeying in one place (RegistryURL) 

This does:
- fix #3245 and a number of other oddities (now enshrined in tests)
- significantly increase testing (+~1000 lines is just tests), both unit testing for `CredentialsStore` and `RegistryURL` parsing, and integration testing, which should hopefully give us a much firmer understanding of what is supposed to happen when
- reduce code duplication
- greatly simplify handling of registry urls
- allows fixing further login issues as a follow-up, on better grounds
- allows for future implementation of namespaced hosts login

What this does NOT:
- does not try to solve #3265 (that will come later)
- does not do anything for hosts.toml resolution (ditto)
- does not add Harbor testing (ditto)

PR #3485, opened 2024-10-02
Fix Docker Hub push regression

apostasie:
Docker Hub push is currently broken (regression from #3293).

This fixes it.

Same as the other regression - we need to test against Hub with a dummy account and actual content for push and pull.


apostasie:
@AkihiroSuda this one is ready as well.

AkihiroSuda (approved):
Thanks

Issue #3086, opened 2024-06-13
`networks` related code is racy


apostasie:
### Description

For some reason, network tests are very racy on my rig.

Just taking `network_remove_linux_test`, I get several different conditions quite fast:

 * [ ] `task xyz not found: not found`
 * [ ] `reading /etc/cni/net.d/nerdctl-nerdctl-testnetworkremovebyid.conflist: open /etc/cni/net.d/nerdctl-nerdctl-testnetworkremovebyid.conflist: no such file or directory`
 * [ ] `failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: time=\"2024-06-13T03:36:47Z\" level=fatal msg=\"no such network: \\\"nerdctl-testnetworkprune\\\"\"\nFailed to write to log, write /var/lib/nerdctl/1935db59/containers/nerdctl-testnetworkprune/9c8747e53f2fa1cba99145d19f48648a85d63ccf5ae7167038a3536336c2d6aa/oci-hook.startContainer.log: file already closed: unknown`

From a cursory reading, it feels to me like number 1 is somewhere in `netutils` making assumptions about the availability of objects.

Number 2 is probably also in netutils - seems to me like racyness between checking that a network exist and a later operation that depends on reading the config.
Could be that for certain operation we do not use filelock (properly).

Number 3 is more worrisome.

### Steps to reproduce the issue

go test

### Describe the results you received and expected

Fail 1 out of 10 times with a variety of different reasons.

### What version of nerdctl are you using?

1.7.6

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Following on the refactor and ongoing test tooling rewrite, it is easier to repro:

```
while true; do go test  ./cmd/nerdctl/network/ -count 1; done
```


Will trigger: `time="2024-09-22T11:49:18-07:00" level=fatal msg="error reading /home/dmp.linux/.config/cni/net.d/nerdctl-test/nerdctl-testnetworkinspect_test_network_inspect.conflist: open /home/dmp.linux/.config/cni/net.d/nerdctl-test/nerdctl-testnetworkinspect_test_network_inspect.conflist: no such file or directory"` after a few tries.

apostasie:
```
--- FAIL: TestNetworkLsFilter (0.18s)
[...]

PR #3491, opened 2024-10-03
Make CreateNetwork safer wrt concurrency

apostasie:
As I am progressing through rewriting tests, the actual code issues are starting to show up more.
#3086 is definitely biting.

While we should consider rewriting CNIEnv on top of `Store` and ensure it is fully safe, this here will help a little bit in the interim.

Previous version had a lock, but somehow did enumerate the conf folder outside of it.


apostasie:
Specifically, I have hopes that this might fix the spurious `reading /etc/cni/net.d/XXXX.conflist: open /etc/cni/net.d/XXX.conflist: no such file or directory` that we are seeing regularly.

apostasie:
@AkihiroSuda CI is green. This is ready.

AkihiroSuda (approved):
Thanks

PR #3499, opened 2024-10-05
Remove images from Docker Hub from tests

### LINKED ISSUE #3498: Docker Hub: 429s keep breaking the builds
### Description

Docker Hub has stringent (non-sensical?) rate limitations on their service that no other registry provider impose.

As such, Hub will rapidly respond 429 - effectively breaking our tests.
Since this "temporary" ban apply for some time, it will further prevent any re-attempt.

Suggesting we either "pre-heat" all needed images at the start of our test suites, so that at least if pull are going to fail, we fail fast - or we remove all of these images from our test suite and use copies stored on ghcr.

https://github.com/containerd/nerdctl/actions/runs/11188098475/job/31106292438?pr=3497#step:7:524

429

main

None

### Host information

_No response_

apostasie:
Fix #3498

### DIFF
--- a/cmd/nerdctl/image/image_list_test.go
+++ b/cmd/nerdctl/image/image_list_test.go
@@ -95,12 +95,12 @@ LABEL version=0.1`, testutil.CommonImage)
 	base.Cmd("build", "-t", tempName, "-f", buildCtx+"/Dockerfile", buildCtx).AssertOK()
 	defer base.Cmd("rmi", tempName).AssertOK()
 
-	busyboxGlibc, busyboxUclibc := "busybox:glibc", "busybox:uclibc"
-	base.Cmd("pull", busyboxGlibc).AssertOK()
-	defer base.Cmd("rmi", busyboxGlibc).AssertOK()
-
-	base.Cmd("pull", busyboxUclibc).AssertOK()
-	defer base.Cmd("rmi", busyboxUclibc).AssertOK()
+	// This test is about testing local filtering of image names - as such, we do not need remote images at all
+	taggedOne, taggedTwo := "taggedimage:xfoox", "taggedimage:yzfooyz"
+	base.Cmd("tag", testutil.CommonImage, taggedOne).Run()
+	base.Cmd("tag", testutil.CommonImage, taggedTwo).Run()
+	defer base.Cmd("rmi", taggedOne).AssertOK()
+	defer base.Cmd("rmi", taggedTwo).AssertOK()
 
 	// before/since filters are not compatible with DOCKER_BUILDKIT=1? (but still compatible with DOCKER_BUILDKIT=0)
 	if base.Target == testutil.Nerdctl {
@@ -119,8 +119,8 @@ LABEL version=0.1`, testutil.CommonImage)
 	base.Cmd("images", "--filter", "label=foo=bar", "--filter", "label=version=0.2").AssertOutNotContains(tempName)
 	base.Cmd("images", "--filter", "label=version").AssertOutContains(tempName)
 	base.Cmd("images", "--filter", fmt.Sprintf("reference=%s*", tempName)).AssertOutContains(tempName)
-	base.Cmd("images", "--filter", "reference=busy*:*libc*").AssertOutContains("glibc")
-	base.Cmd("images", "--filter", "reference=busy*:*libc*").AssertOutContains("uclibc")
+	base.Cmd("images", "--filter", "reference=tag*:*foo*").AssertOutContains("xfoox")
+	base.Cmd("images", "--filter", "reference=tag*:*foo*").AssertOutContains("yzfooyz")
 }
 
 func TestImagesFilterDangling(t *testing.T) {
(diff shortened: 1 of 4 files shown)


apostasie:
Note that this here is highlighting another of our test issues: we do have a number of platform specific files, but we ALSO do mark certain tests as skipped if we are on windows.

Using the two mechanisms simultaneously is problematic, and leading to situations like this here, where we have to add non-sensical definitions.

Suggestion is to stop using platform specific files for tests entirely, and instead just use appropriate `Skip`.
- easier to know how many / which tests are being skipped on a certain platform
- easier to enable / disable a single test for a platform without having to move a bunch of files around

apostasie:
TestRunWithTtyAndDetached #3437 

Besides that ^, it is ready.

cmd/nerdctl/image/image_list_test.go
```diff
@@ -96,10 +96,9 @@ LABEL version=0.1`, testutil.CommonImage)
 	defer base.Cmd("rmi", tempName).AssertOK()
 
 	busyboxGlibc, busyboxUclibc := "busybox:glibc", "busybox:uclibc"
-	base.Cmd("pull", busyboxGlibc).AssertOK()
+	base.Cmd("tag", testutil.CommonImage, busyboxGlibc).Run()
+	base.Cmd("tag", testutil.CommonImage, busyboxUclibc).Run()
```
AkihiroSuda, line 100: Needs a comment line to explain why not pull real images

AkihiroSuda: I also wonder if we can avoid using fake "busybox" names

AkihiroSuda (changes requested):

cmd/nerdctl/image/image_list_test.go
apostasie: Yeah - I'll just use some other name instead - it doesn't really matter.

apostasie: > Needs a comment line to explain why not pull real images

Well, nothing is really "real" in our software world, right? :-) - jk :-)

I will write a comment.

fahedouch:
Hi @apostasie, there is also a rate limiting  on ghcr and everywhere!
A non-expensive solutions would be :
- to check if image exist in content storage before making a pull otherwise pull 
- cache image list in mem struct for each pull and make a mem check for each pull.

unfortunately there is no flag that does this check in the pull subcommand

apostasie:
> Hi @apostasie, there is also a rate limiting on ghcr and everywhere! A non-expensive solutions would be :
> 
> * to check if image exist in content storage before making a pull otherwise pull
> * cache image list in mem struct for each pull and make a mem check for each pull.
> 
> unfortunately there is no flag that does this check in the pull subcommand

At the start of tests, we could spin a local registry and push there the images we need.

Then we could leverage hosts.toml to use the local registry as cache during the tests.

apostasie:
@AkihiroSuda updated to remove the fake "busybox" name and added a comment.

apostasie:
`TestRunWithTtyAndDetached`

apostasie:
> Hi @apostasie, there is also a rate limiting on ghcr and everywhere! A non-expensive solutions would be :

I am sure there is - we just never hit it on the CI, or debugging locally...

cmd/nerdctl/image/image_list_test.go
```diff
@@ -95,12 +95,12 @@ LABEL version=0.1`, testutil.CommonImage)
...
-	defer base.Cmd("rmi", busyboxGlibc).AssertOK()
-
-	base.Cmd("pull", busyboxUclibc).AssertOK()
-	defer base.Cmd("rmi", busyboxUclibc).AssertOK()
+	// This test is about testing local filtering of image names - as such, we do not need remote images at all
+	taggedGlibc, taggedUclibc := "taggedimage:glibc", "taggedimage:uclibc"
+	base.Cmd("tag", testutil.CommonImage, taggedGlibc).Run()
+	base.Cmd("tag", testutil.CommonImage, taggedUclibc).Run()
+	defer base.Cmd("rmi", taggedGlibc).AssertOK()
+	defer base.Cmd("rmi", taggedUclibc).AssertOK()
```
AkihiroSuda, line 103: Why not just use simple meta syntax vars such as "foo" "bar"?
The implementation of libc doesn't really relate to the test.

apostasie: The existing test was using that - I just did not change it - and I don't think it matters one way or the other.
And yes, it is pretty much `vfoov` and `zzfoozz`

I am also not particularly interested in spending more energy than needed on these tests in their current form. They are neither particularly good nor complete enough, and I would rather focus on the rewrite.

Anyhow, I will update if you prefer.

apostasie: Done

apostasie: > The existing test was using that - I just did not change it - and I don't think it matters one way or the other. And yes, it is pretty much `vfoov` and `zzfoozz`
> 
> I am also not particularly interested in spending more energy than needed on these tests in their current form. They are neither particularly good nor complete enough, and I would rather focus on the rewrite.
> 
> Anyhow, I will update if you prefer.

Hope this didn't sound defensive.

I guess lately I have been working on tests / QA way too much for my own good :-).

Cheers.

apostasie:
Latest failures are the usual - `TestIPFSComposeUp` and `TestRunWithTtyAndDetached `

AkihiroSuda (approved):
Thanks

PR #3503, opened 2024-10-06
switch cosign registry from GCR to GHCR

bobcallaway:
This changes the Dockerfile to pull the cosign container image from GHCR instead of Google Cloud. This helps the Sigstore team manage their cloud spend (as GHCR is provided for free and Google Cloud Artifact Registry is not).

Note the container hash does not change and images are posted to both locations upon cosign's release process.


fahedouch (approved):
Thanks

Issue #3518, opened 2024-10-08
CNI plugins: firewall / iptables failure (seen on CI)


apostasie:
### Description

?

```
=== RUN   TestIssue2993/Issue_#2993_-_nerdctl_no_longer_leaks_containers_and_etchosts_directories_and_files_when_containers_are_removed.
=== PAUSE TestIssue2993/Issue_#2993_-_nerdctl_no_longer_leaks_containers_and_etchosts_directories_and_files_when_containers_are_removed.
=== CONT  TestIssue2993/Issue_#2993_-_nerdctl_no_longer_leaks_containers_and_etchosts_directories_and_files_when_containers_are_removed.
    container_create_linux_test.go:305: ======================== Pre-test cleanup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test --data-root /tmp/TestIssue2993Issue_#2993_-_nerdctl_no_longer_leaks_containers_and_etchosts_directories_and_files_when_containers_are_removed.39[502](https://github.com/containerd/nerdctl/actions/runs/11229452200/job/31215023669?pr=3517#step:6:503)85910/001 rm -f testissue2993-issue-2993-nerdctl-no-longer-leaks-containers-and-etc-e5720a82
    container_create_linux_test.go:305: ======================== Test setup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test run --data-root /tmp/TestIssue2993Issue_#2993_-_nerdctl_no_longer_leaks_containers_and_etchosts_directories_and_files_when_containers_are_removed.3950285910/001 --name testissue2993-issue-2993-nerdctl-no-longer-leaks-containers-and-etc-e5720a82 -d ghcr.io/stargz-containers/alpine:3.13-org sleep infinity
    command.go:112: assertion failed: expect.ExitCode is not result.ExitCode: Expected exit code: 0
        
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test run --data-root /tmp/TestIssue2993Issue_#2993_-_nerdctl_no_longer_leaks_containers_and_etchosts_directories_and_files_when_containers_are_removed.3950285910/001 --name testissue2993-issue-2993-nerdctl-no-longer-leaks-containers-and-etc-e5720a82 -d ghcr.io/stargz-containers/alpine:3.13-org sleep infinity
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
[...]

PR #3517, opened 2024-10-08
[WIP] Test rework permanent next-target

apostasie:
Note:
This PR is a dump of all next- things for test rework.
It will never be merged, and will be forked-out into smaller PR.

Github actions changes:
- cosmetic efforts on test display titles (more compact, more consistent)
- lint on all platforms instead of just linux
- removal of all nick/retries
- separation of flaky / non-flaky in different steps - non-flaky now run without any retry - flaky do run with gotestsum retry, but no longer nick/retries

Integration test entrypoint:
- now lives inside `./hack/test-integration.sh`
- additionally displays slow tests at the end of the run

Dockerfile changes:
- removed CMD - replaced in favor of a direct call to ./hack/test-integration.sh
- removed useless stages

Currently, test suites have been ran locally (no-flaky, on the host directly) 100 times without failure:
- `./cmd/nerdctl` < 1 sec per run
- `./cmd/nerdctl/completion`  < 1 sec per run
- `./cmd/nerdctl/issues` < 10 secs per run
- `./cmd/nerdctl/system` < 10 secs per run
- `./cmd/nerdctl/volume` < 10 secs per run
- `./cmd/nerdctl/network` < 10 secs per run
- `./cmd/nerdctl/image` < 30 secs per run
- `./cmd/nerdctl/ipfs`

WORK IN PROGRESS rn:
- `./cmd/nerdctl/builder`
- `./cmd/nerdctl/container`

NOT part of this PR:
- `./cmd/nerdctl/login`
- `./cmd/nerdctl/compose`

Issue #2908, opened 2024-04-03
nerdctl fails when running concurrently due to CNI errors: `CHAIN_USER_ADD failed (File exists): chain CNI-ISOLATION-STAGE-2`


aojea:
### Description

See in https://github.com/kubernetes-sigs/kind/issues/3533

```
Command Output: time="2024-04-01T08:34:37Z" level=fatal msg="failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: time=\"2024-04-01T08:34:34Z\" level=fatal msg=\"failed to call cni.Setup: plugin type=\\\"firewall\\\" failed (add): running [/usr/sbin/iptables -t filter -N CNI-ISOLATION-STAGE-2 --wait]: exit status 4: iptables v1.8.7 (nf_tables):  CHAIN_USER_ADD failed (File exists): chain CNI-ISOLATION-STAGE-2\\n\"\nFailed to write to log, write /var/lib/nerdctl/1935db59/containers/default/18e88fcb538d49417539810b[25](https://github.com/kubernetes-sigs/kind/actions/runs/8505888448/job/23295140055?pr=3563#step:8:26)67922886e120771be00f165b5d64cf41a381f5/oci-hook.createRuntime.log: file already closed: unknown"

Stack Trace: 
sigs.k8s.io/kind/pkg/errors.WithStack
	sigs.k8s.io/kind/pkg/errors/errors.go:59
sigs.k8s.io/kind/pkg/exec.(*LocalCmd).Run
	sigs.k8s.io/kind/pkg/exec/local.go:124
sigs.k8s.io/kind/pkg/cluster/internal/providers/nerdctl.createContainerWithWaitUntilSystemdReachesMultiUserSystem
	sigs.k8s.io/kind/pkg/cluster/internal/providers/nerdctl/provision.go:383
sigs.k8s.io/kind/pkg/cluster/internal/providers/nerdctl.planCreation.func3
	sigs.k8s.io/kind/pkg/cluster/internal/providers/nerdctl/provision.go:123
sigs.k8s.io/kind/pkg/errors.UntilErrorConcurrent.func1
	sigs.k8s.io/kind/pkg/errors/concurrent.go:30
runtime.goexit
	runtime/asm_amd64.s:1598
Error: Process completed with exit code 1.
```

### Steps to reproduce the issue

[...]

PR #3522, opened 2024-10-08
Enforce global lock in oci hooks

apostasie:
See
#3518
#2908

and probably others.

At this point, nerdctl simply cannot be used in any non trivial scenario involving CNI and concurrency.

Besides local context, this started cropping up as soon as I disabled retries on the CI and cranked up parallelism on tests.

The bottom-line MAY be that CNI plugins cannot be used concurrently at all - whether by design or not - this is unconfirmed though.

A "proper" fix is likely very involved - probably rewriting most / all of CNIEnv, or the cni plugins themselves. As floated in the tickets, it is not even clear if separate binaries + ocihooks is a tenable design at all.

Anyhow, this is the best I can think of right now that does not involve too much work: enforce a global lock in hooks. It is a hammer. It will hurt performance, likely. But then the choice is "does not work" vs. "works slowly".
As stated in the comment, it will probably not even cover all cases - but it is better than nothing and will make things a bit more acceptable.

This has been forked out of #3517 which is currently ground zero.

### DIFF
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -39,6 +39,7 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/bypass4netnsutil"
 	"github.com/containerd/nerdctl/v2/pkg/dnsutil/hostsstore"
 	"github.com/containerd/nerdctl/v2/pkg/labels"
+	"github.com/containerd/nerdctl/v2/pkg/lockutil"
 	"github.com/containerd/nerdctl/v2/pkg/namestore"
 	"github.com/containerd/nerdctl/v2/pkg/netutil"
 	"github.com/containerd/nerdctl/v2/pkg/netutil/nettype"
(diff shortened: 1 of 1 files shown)


apostasie:
(pending green)

apostasie:
Failure is the usual TestIPFSComposeUp/stargz

pkg/ocihook/ocihook.go
```diff
@@ -92,6 +93,27 @@ func Run(stdin io.Reader, stderr io.Writer, event, dataStore, cniPath, cniNetcon
...
+	// See
+	// https://github.com/containerd/nerdctl/issues/3518
+	// https://github.com/containerd/nerdctl/issues/2908
+	// and likely others
+	// Fixing these issues would require a lot of work, possibly even stopping using individual cni binaries altogether
+	// or at least being very mindful in what operation we call inside CNIEnv at what point, with filesystem locking.
+	// This below is a stopgap solution that just enforces a global lock
+	// Note this here is probably not enough, as concurrent CNI operations may happen outside of the scope of ocihooks
+	// through explicit calls to Remove, etc.
+	lockDir := filepath.Join(cniNetconfPath)
```
AkihiroSuda, line 106: Doesn't seem to be joined

apostasie: Geez. 🤦🏻‍♂️🐟
on my phone.
Will fix.

pkg/ocihook/ocihook.go
```diff
@@ -92,6 +93,26 @@ func Run(stdin io.Reader, stderr io.Writer, event, dataStore, cniPath, cniNetcon
...
+	// See
+	// https://github.com/containerd/nerdctl/issues/3518
+	// https://github.com/containerd/nerdctl/issues/2908
+	// and likely others
+	// Fixing these issues would require a lot of work, possibly even stopping using individual cni binaries altogether
+	// or at least being very mindful in what operation we call inside CNIEnv at what point, with filesystem locking.
+	// This below is a stopgap solution that just enforces a global lock
+	// Note this here is probably not enough, as concurrent CNI operations may happen outside of the scope of ocihooks
+	// through explicit calls to Remove, etc.
+	err = os.MkdirAll(cniNetconfPath, 0o777)
```
AkihiroSuda, line 106: Why 777?

apostasie: Good catch.
I don't think there is a reason, and it should work with 0o700

AkihiroSuda (approved):
Thanks

PR #3523, opened 2024-10-08
Lock errdefs minor versions for dependabot

apostasie:
errdefs recent minor updates are breaking compatibility, and since we depend on containerd, which depends on 0.1.0, we are not going to update that anytime soon.

Meanwhile, dependabot keeps trying: #3521 

Hopefully, that should now limit that?


apostasie:
Should not do anything on the CI, so, assuming green.

@AkihiroSuda opinion on this?

AkihiroSuda:
In #3521 I ran `ignore this minor version` command

PR #3529, opened 2024-10-12
Fix CNIEnv concurrency issue

apostasie:
This is very similar to #3522, and addresses another facet of the same underlying problem (CNIEnv not being safe to use concurrently).

Failure is typically showing-up with `images` in integration tests (on my laptop, shows up after 20+ runs).

This is not a good long-term solution (the right solution is to rewrite CNIEnv with concurrency in mind) - but it will alleviate the problem in a bunch of cases for now.


apostasie:
@AkihiroSuda ready to go.

AkihiroSuda (approved):
Thanks

Issue #3516, opened 2024-10-08
Multi-platform image code is broken


apostasie:
### Description

I was banging my head on the desk chasing #3513 and friends - but eventually, trying something simple on a fresh install:

```
# Use whatever platform is not your native platform
nerdctl pull --platform linux/amd64  debian
nerdctl tag debian foo
nerdctl push foo
```

This fails on 1.7.6 (and main, obviously)

```
FATA[0000] failed to create a tmp single-platform image "docker.io/library/foo:latest-tmp-reduced-platform": content digest sha256:4466c6813e1d9a85fbac707afe419e13ce9a823f7041316bab860d8c0a50f91d: not found
```

This seems to be pervasive across the codebase (and certainly part of the problem we have been chasing with the many incarnations of `content digest not found`).

I am wondering if the multi-platform image concept/code overall is considered experimental at this point - or if there are known gotchas?

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Actually, in light of #3512 - looks like most image operations are just broken (including just `images --filter`) when manipulating multi-platform images.

:(

apostasie:
@AkihiroSuda is there a document somewhere that lays-out the UX for images command when an image is not the native arch? or when it is a multi-platform image?

apostasie:
The specific repro from this ticket has been fixed by the recent series of `content digest not found` PRs.

There are still issues with multi-platform code, but we should close this and possibly open new ones with appropriate repros.

Issue #3513, opened 2024-10-07
[MAJOR] content digest not found


apostasie:
This is meant to collect more information about the issue known as "content digest not found".

While #3435 (and follow-up for IPFS) did address the most obvious cases, we obviously have more.

Here is the currently know list:
- #3509
- #3510
- #3512

The gist of it appears to be, at this point:
- seems like containerd `converter.Convert` (calling on imageservice.Create) may return an image with missing content in some circumstances
- seems like any call to `content.ReadBlob` (specifically as seen inside imgutil) may fail and should be gated by an operation to first force fetch the blob in question

At this point, above is speculative, although it seems consistent with the debugging info in the tickets.

apostasie:
Note that these bugs are elusive.
You have to run the test suite for the affected parts for 50+ times before you hit it.

apostasie:
Apparently happens as well on **run**:

This does reinforce the notion that there is a problem in containerd (possibly https://github.com/containerd/containerd/issues/8973 )

https://github.com/containerd/nerdctl/actions/runs/11397513713/job/31713086655?pr=3535#step:7:538

```
Command:  /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 run --pull always -d --name testremove-remove-image-with-running-container-without-f-8c22042c ghcr.io/stargz-containers/alpine:3.13-org sleep infinity
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   ghcr.io/stargz-containers/alpine:3.13-org:                                     resolved       |++++++++++++++++++++++++++++++++++++++| 
        index-sha256:ec14c7992a97fc11425907e908340c6c3d6ff602f5f13d899e6b7027c9b4133a: done           |++++++++++++++++++++++++++++++++++++++| 
        elapsed: 0.1 s                                                                 total:  1.6 Ki (15.7 KiB/s)                                      
        ghcr.io/stargz-containers/alpine:3.13-org:                                        resolved       |++++++++++++++++++++++++++++++++++++++| 
        index-sha256:ec14c7992a97fc11425907e908340c6c3d6ff602f5f13d899e6b7027c9b4133a:    done           |++++++++++++++++++++++++++++++++++++++| 
        manifest-sha256:e103c1b4bf019dc290bcc7aca538dc2bf7a9d0fc836e186f5fa34945c5168310: done           |++++++++++++++++++++++++++++++++++++++| 
[...]

Issue #3512, opened 2024-10-07
content digest not found issue is impacting `nerdctl images --filter`


apostasie:
### Description

```
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	40.188s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	33.235s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	41.313s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	38.859s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	42.244s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	38.793s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	40.892s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	41.808s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	42.223s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	41.142s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	35.809s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	38.968s
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/image	43.957s
test target: "nerdctl"
Waiting for port to become available... 80
Waiting for port to become available... 80
Waiting for port to become available... 80
Waiting for port to become available... 80
Waiting for port to become available... 80
Waiting for port to become available... 80
--- FAIL: TestImagesFilter (2.10s)
    image_list_test.go:271: ======================== Pre-test cleanup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test builder prune --all --force
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test rmi taggedimage:one-fragment-one
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test rmi taggedimage:two-fragment-two
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test rmi testimagesfilter-2a4e0faa
    image_list_test.go:271: ======================== Test setup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test pull ghcr.io/stargz-containers/alpine:3.13-org
[...]

PR #3530, opened 2024-10-12
Alleviate image --filter missing digest issue

apostasie:
See #3512 #3513 #3516 for some context.

In a shell, filtering requires config to be there.
With multi-arch / non-native arch, retrieving the config might fail for specific images, which will fail the entire operation.

Turning this into a soft failure to alleviate the pain.

### DIFF
--- a/pkg/imgutil/filtering.go
+++ b/pkg/imgutil/filtering.go
@@ -28,6 +28,7 @@ import (
 
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/containerd/containerd/v2/core/images"
+	"github.com/containerd/log"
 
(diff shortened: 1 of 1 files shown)


apostasie:
@AkihiroSuda ready to go. Will definitely squash some of the flakyness.

Pending CI.

apostasie:
What about we just disable this `TestRunWithTtyAndDetached ` test?
At least until the windows folk look into it.
It is just failing almost constantly - even with retries.

AkihiroSuda:
> What about we just disable this `TestRunWithTtyAndDetached ` test? At least until the windows folk look into it. It is just failing almost constantly - even with retries.

I think we should just skip it

AkihiroSuda (approved):
Thanks

Issue #3536, opened 2024-10-14
Add support for running container from OCI archive without requiring separate load command


austinvazquez:
### What is the problem you're trying to solve

I have a use case which is moving from podman to nerdctl/containerd which currently runs container image from OCI archive without requiring load as a separate step.

### Describe the solution you'd like

I would like to provide a similar experience:

```bash
$ nerdctl run oci-archive:///path/to/oci/archive /bin/bash
```

austinvazquez:
Thanks everyone for the reviews and feedback on this one.

PR #3537, opened 2024-10-14
Add support for running container from OCI archive

austinvazquez:
### Issue
Proposal for https://github.com/containerd/nerdctl/issues/3536

### Description
This change adds the ability to create and run containers from OCI archive without performing separate load step.

### Testing
As this behavior is not provided by Docker, this change adds a nerdctl-only test for this behavior.

### DIFF
--- a/cmd/nerdctl/container/container_create_linux_test.go
+++ b/cmd/nerdctl/container/container_create_linux_test.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/containerd/containerd/v2/defaults"
 
+	"github.com/containerd/nerdctl/v2/cmd/nerdctl/helpers"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nettestutil"
@@ -305,4 +306,35 @@ func TestIssue2993(t *testing.T) {
 	}
 
 	testCase.Run(t)
+}
+
+func TestCreateFromOCIArchive(t *testing.T) {
+	testutil.RequiresBuild(t)
+	testutil.RegisterBuildCacheCleanup(t)
+
+	// Docker does not support creating containers from OCI archive.
+	testutil.DockerIncompatible(t)
+
+	base := testutil.NewBase(t)
+	imageName := testutil.Identifier(t)
+	containerName := testutil.Identifier(t)
+
+	teardown := func() {
+		base.Cmd("rm", "-f", containerName).Run()
+		base.Cmd("rmi", "-f", imageName).Run()
+	}
+	defer teardown()
+	teardown()
+
+	const sentinel = "test-nerdctl-create-from-oci-archive"
+	dockerfile := fmt.Sprintf(`FROM %s
+	CMD ["echo", "%s"]`, testutil.CommonImage, sentinel)
+
+	buildCtx := helpers.CreateBuildContext(t, dockerfile)
+	tag := fmt.Sprintf("%s:latest", imageName)
+	tarPath := fmt.Sprintf("%s/%s.tar", buildCtx, imageName)
+
+	base.Cmd("build", "--tag", tag, fmt.Sprintf("--output=type=oci,dest=%s", tarPath), buildCtx).AssertOK()
+	base.Cmd("create", "--rm", "--name", containerName, fmt.Sprintf("oci-archive://%s", tarPath)).AssertOK()
+	base.Cmd("start", "--attach", containerName).AssertOutContains("test-nerdctl-create-from-oci-archive")
 }
--- a/cmd/nerdctl/container/container_run_test.go
+++ b/cmd/nerdctl/container/container_run_test.go
@@ -658,3 +658,31 @@ func TestRunQuiet(t *testing.T) {
 
 	assert.Assert(t, wasQuiet(result.Combined(), sentinel), "Found %s in container run output", sentinel)
 }
+
+func TestRunFromOCIArchive(t *testing.T) {
+	testutil.RequiresBuild(t)
+	testutil.RegisterBuildCacheCleanup(t)
+
+	// Docker does not support running container images from OCI archive.
+	testutil.DockerIncompatible(t)
+
+	base := testutil.NewBase(t)
+	imageName := testutil.Identifier(t)
+
+	teardown := func() {
+		base.Cmd("rmi", "-f", imageName).Run()
+	}
+	defer teardown()
+	teardown()
+
+	const sentinel = "test-nerdctl-run-from-oci-archive"
+	dockerfile := fmt.Sprintf(`FROM %s
+	CMD ["echo", "%s"]`, testutil.CommonImage, sentinel)
+
+	buildCtx := helpers.CreateBuildContext(t, dockerfile)
+	tag := fmt.Sprintf("%s:latest", imageName)
+	tarPath := fmt.Sprintf("%s/%s.tar", buildCtx, imageName)
+
+	base.Cmd("build", "--tag", tag, fmt.Sprintf("--output=type=oci,dest=%s", tarPath), buildCtx).AssertOK()
+	base.Cmd("run", "--rm", fmt.Sprintf("oci-archive://%s", tarPath)).AssertOutContainsAll(fmt.Sprintf("Loaded image: %s", tag), sentinel)
+}
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -134,6 +134,7 @@ Run a command in a new container.
 Usage: `nerdctl run [OPTIONS] IMAGE [COMMAND] [ARG...]`
 
 :nerd_face: `ipfs://` prefix can be used for `IMAGE` to pull it from IPFS. See [`ipfs.md`](./ipfs.md) for details.
+:nerd_face: `oci-archive://` prefix can be used for `IMAGE` to specify a local file system path to an OCI formatted tarball.
 
 Basic flags:
 
@@ -423,6 +424,7 @@ Create a new container.
 Usage: `nerdctl create [OPTIONS] IMAGE [COMMAND] [ARG...]`
 
 :nerd_face: `ipfs://` prefix can be used for `IMAGE` to pull it from IPFS. See [`ipfs.md`](./ipfs.md) for details.
+:nerd_face: `oci-archive://` prefix can be used for `IMAGE` to specify a local file system path to an OCI formatted tarball.
 
 The `nerdctl create` command similar to `nerdctl run -d` except the container is never started. You can then use the `nerdctl start <container_id>` command to start the container at any point.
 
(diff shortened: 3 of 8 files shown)


austinvazquez:
If maintainers would prefer, I can break https://github.com/containerd/nerdctl/pull/3537/commits/79e4e24fbaaf94a13461bb48d9710a9f0c55505d into its own PR. The change was trivial at first before I added docs changes and integration test. LMK your thoughts.

cmd/nerdctl/container/container_create.go
```diff
@@ -409,12 +409,17 @@ func processContainerCreateOptions(cmd *cobra.Command) (types.ContainerCreateOpt
 	if err != nil {
 		return opt, err
 	}
+	quiet, err := cmd.Flags().GetBool("quiet")
```
AkihiroSuda, line 412: Should be a separate PR

AkihiroSuda: Does Docker or Podman have this flag?

Otherwise —pull-quiet might be less confusing? idk

cmd/nerdctl/container/container_run.go
```diff
@@ -361,6 +363,33 @@ func runAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if imageRef := args[0]; strings.HasPrefix(imageRef, "oci-archive://") {
```
AkihiroSuda, line 366: This has to be supported in `nerdctl quiet` too (See ipfs:// code)

cmd/nerdctl/container/container_run.go
```diff
@@ -361,6 +363,33 @@ func runAction(cmd *cobra.Command, args []string) error {
...
+			AllPlatforms: false,
+			Quiet:        createOpt.ImagePullOpt.Quiet,
+		})
+		if err != nil {
+			return err
+		}
+
+		// Running multi-image OCI archives is not yet supported.
+		if len(images) != 1 {
+			return errors.New("multiple images in OCI archive found")
```
AkihiroSuda, line 387: What about just defaulting to the first one? (With a warning)

cmd/nerdctl/container/container_run_test.go
```diff
@@ -631,3 +631,29 @@ func TestRunAttachFlag(t *testing.T) {
 		})
 	}
 }
+
+func TestRunFromOCIArchive(t *testing.T) {
+	testutil.RequiresBuild(t)
+	testutil.RegisterBuildCacheCleanup(t)
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 638: Needs a comment

docs/command-reference.md
AkihiroSuda, line 1: oci-archive://  needs socks too

cmd/nerdctl/container/container_create.go
austinvazquez: 👍🏼 Opened https://github.com/containerd/nerdctl/pull/3551
Docker has --quiet across load, run, and create.

austinvazquez:
Going to let this CI run finish before rebasing for https://github.com/containerd/nerdctl/pull/3551

docs/command-reference.md
austinvazquez: Addressed in https://github.com/containerd/nerdctl/pull/3537/commits/d669f3d0af9f235ce61d744c3e21b3495a9b5e69. Added documentation for both create and run.

chews93319 (approved):

pkg/cmd/container/create.go
```diff
@@ -123,6 +124,37 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
 	}
 	opts = append(opts, platformOpts...)
 
+	if imageRef := args[0]; strings.HasPrefix(imageRef, "oci-archive://") {
```
AkihiroSuda, line 127: Protocol prefixes should be parsed in the same place as ipfs://

https://github.com/containerd/nerdctl/blob/f240e58d2e5922642128fee245d933a6550ddb46/pkg/referenceutil/referenceutil.go#L91

pkg/referenceutil/referenceutil_test.go
```diff
@@ -290,6 +290,14 @@ func TestReferenceUtil(t *testing.T) {
 			Tag:          "latest",
 			ExplicitTag:  "",
 		},
+		"oci-archive:///tmp/build/saved-image.tar": {
+			Protocol:     "oci-archive",
+			Path:         "/tmp/build/saved-image.tar",
+			Error:        "",
+			String:       "/tmp/build/saved-image.tar",
+			Suggested:    "oci-archive-/tmp/-abcde",
```
apostasie, line 298: You likely need to do something for the suggested name of oci-archive://.

This suggested name is not valid for a container name (it is used when you do not specify a name for your container).

Maybe something in the line of:
```
func (ir *ImageReference) SuggestContainerName(suffix string) string {
	name := "untitled"
	if ir.Protocol != "" && ir.Domain == "" {
		if ir.Protocol == OCIArchive {
			name = string(ir.Protocol) + "-" + path.Base(ir.Path)[:shortIDLength]
		}else{
			name = string(ir.Protocol) + "-" + ir.String()[:shortIDLength]
		}
	} else if ir.Path != "" {
		name = path.Base(ir.Path)
	}
	return name + "-" + suffix[:5]
}

```

... although that is likely not sufficient, as filesystem might allow characters that are not safe to use as image identifiers.

apostasie: Maybe we keep it simple and only suggest ‘oci-archive-container-id’ so that we do not have to deal with local path sanitization.

pkg/cmd/container/create.go
```diff
@@ -123,6 +124,37 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
...
+			return nil, nil, err
+		} else if len(images) == 0 {
+			// This is a regression and should not occur.
+			return nil, nil, errors.New("OCI archive did not contain any images")
+		}
+
+		image := images[0].Name
+		// Multi-image archive provided, default to first image found.
+		if len(images) != 1 {
+			log.L.Warnf("multi-image OCI archive provided, defaulting to image %s...", image)
```
ktock, line 152:
```suggestion
			log.L.Warnf("multiple images are found for the platform, defaulting to image %s...", image)
```

pkg/imgutil/load/load.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/containerd/nerdctl/v2/pkg/imgutil"
+	"github.com/containerd/nerdctl/v2/pkg/platformutil"
+)
+
+// FromArchive will load and unpack the images from the provided tar archive specified in image load options.
+func FromArchive(ctx context.Context, client *containerd.Client, options types.ImageLoadOptions) ([]images.Image, error) {
+	if options.Input != "" {
+		f, err := os.Open(options.Input)
+		if err != nil {
+			return []images.Image{}, err
```
ktock, line 43: why not `return nil, err`?

pkg/referenceutil/referenceutil_test.go
austinvazquez: Thanks @apostasie. Let me take another look here. We load the image from the archive before and reset the image ref to the name in the image manifest before calling parse again on that reference. I shouldn't have just made this test work though. Maybe it would be better to have another type for oci-archive image references which doesn't support suggested container name as it doesn't make sense.

pkg/imgutil/load/load.go
austinvazquez: I feel like I picked this up early in my Go learning (maybe a project I worked did this) but yeah looks like nil is recommended to avoid memory alloc.

pkg/referenceutil/referenceutil_test.go
apostasie: I’ll have a look around as well on how this is used. 
thanks @austinvazquez !

pkg/imgutil/load/load.go
austinvazquez: Addressed in https://github.com/containerd/nerdctl/pull/3537/commits/d8ab512e735c20ab922198c584b8cf69ba192793

pkg/cmd/container/create.go
austinvazquez: Applied in https://github.com/containerd/nerdctl/pull/3537/commits/932bfaf0a599fbda3f19377cc827092c371318bd

pkg/referenceutil/referenceutil_test.go
austinvazquez: Should be addressed with https://github.com/containerd/nerdctl/pull/3537/commits/403499804af29636fe6168a430ee2a0ca687fc79. @apostasie, I took another look and changed how I was looking at the problem. When parsing an raw image reference which is a path for an OCI archive, we should fail and let the caller (`pkg/create`) know the reference is a path to an image archive and needs to be loaded before (re)parsing. This should resolve the quirkiness of having a image reference object which was not really a valid image reference.

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -306,3 +307,34 @@ func TestIssue2993(t *testing.T) {
...
+
+	// Docker does not support creating containers from OCI archive.
+	testutil.DockerIncompatible(t)
+
+	base := testutil.NewBase(t)
+	imageName := testutil.Identifier(t)
+	containerName := testutil.Identifier(t)
+
+	teardown := func() {
+		base.Cmd("rm", containerName).Run()
```
apostasie, line 323: `-f` for both rm and rmi would be better in teardown.

cmd/nerdctl/container/container_run_test.go
```diff
@@ -658,3 +658,31 @@ func TestRunQuiet(t *testing.T) {
...
+	testutil.RegisterBuildCacheCleanup(t)
+
+	// Docker does not support running container images from OCI archive.
+	testutil.DockerIncompatible(t)
+
+	base := testutil.NewBase(t)
+	imageName := testutil.Identifier(t)
+
+	teardown := func() {
+		base.Cmd("rmi", imageName).Run()
```
apostasie, line 673: -f

pkg/referenceutil/referenceutil_test.go
```diff
@@ -273,6 +273,10 @@ func TestReferenceUtil(t *testing.T) {
 			Tag:          "latest",
 			ExplicitTag:  "",
 		},
+		"oci-archive:///tmp/build/saved-image.tar": {
+			Protocol: "oci-archive",
```
apostasie, line 277: Protocol is not necessary here, since the test stops as it errors (and since your latest version does not set the protocol)

pkg/referenceutil/referenceutil_test.go
apostasie: Yeah, reading through - it is a bit weird indeed, because, yep, it is not really an image ref, it is a filesystem location... - I think it is ok as you put it now.

apostasie (approved):

apostasie:
@austinvazquez thanks! - I left a few nits comments.

AkihiroSuda (approved):
Thanks

Issue #3531, opened 2024-10-12
Suggesting we rewrite (parts of) imgutil


apostasie:
### What is the problem you're trying to solve

In light of #3516 it is clear there are issues with code that likely pre-dates the introduction of multi-architecture.

There is also a certain amount of duplication - a lot of what `pkg/cmd/image/list.go` does should really be provided by imgutil.

Also, imgutil is making assumptions that the current native platform version of the image is necessarily available - or that other platforms variants simply do not matter (especially wrt labels).

Finally, the imagewalker (see #3502) might be part of the problem - along with the added abstraction layers (`EnsuredImage`) on top / mixed with the already crowded and confusing containerd API (Image aka `provides the model for how containerd views container images.` vs. Image aka `describes an image used by containers`).

### Describe the solution you'd like

Suggesting someone motivated starts from scratch:
- review all the places where we are currently manipulating Images
- list what they need
- come-up with a multi-platform-aware abstraction for "NerdImage" to cover all of these use-cases (maybe that is an evolution of `EnsuredImage`) and seal the implementation away so that we stop shooting ourselves in the foot with containerd types
- re-evaluate all helpers inside imgutil against use cases - simplify them, make them solidly multi-platform aware
- remove most of the code from `pkg/cmd/image/list.go` and other places that do access low-level details directly
- come-up with serious test - we have close to 0 multi-platform tests for images

PR #3542, opened 2024-10-14
`referenceutil` cleanup

apostasie:
This is a first step at enhancing and fixing issues in imgutil (#3531) - focusing specifically on `referenceutil`.

We currently use a mix of `nerdctl/referenceutil` and `github.com/distribution/reference` as far as I can tell without much consistency.

We obviously introduced our own "helpers" in `referenceutil` (primarily to accommodate for ipfs images), sometimes with identical names to the upstream API:
- ParseDockerRef
- ParseAny
- ParseAnyReference
- ParseIPFSRefWithScheme

... in addition to the existing:
- ParseNormalizedNamed
- ParseDockerRef
- ParseAnyReference
- etc

This abundance of methods that are doing _almost_ exactly the same thing - with subtle differences - seems hard to justify - be it on our side, or on the distribution side.
The distribution ref code was designed ten years ago now (fall 2014), with different notions about images names, backward compatibility requirements with the old python registry (prior to the go implem), and quite frankly, a questionable design (the whole "cast it to `XYZ` if you want to access properties"), and a number of embarrassing bugs (specifically around tags+digests), that IIRC have been creeping far out (into buildkit), causing hard-to-figure-out caching issues.
It seems like `reference` has been barely touched since then. 

At this point, it is doubtful that anyone could say _top of the head_ what are the differences between all of these or what are the ramifications if you use one rather than the other.

While this is challenging already for long-timers (_I certainly do need to read the source multiple times every time I touch these_), it is just plain bad for newcomers.

Add to that normalization concerns, the familiarization mechanisms that work only post normalization. This is just not a good API.

Finally, we seem to parse these references multiples times during the same flow as we pass strings around - in a context where `distribution/reference` is notoriously hungry for memory.

This PR is suggesting that we stop using `distributionref` entirely (eg: except inside referenceutil obviously), and replace all of our stuff with a new, _single method_ - `Parse` - and _one_ struct.

The first commit is the implementation of the new referenceutil + tests.
The second commit adapts the codebase - this is not too aggressive of a refactor, and we might want to further clean this up to reduce duplicate parsing - but that would require changing methods signatures, which this PR tries to avoid.

Let me know your thoughts - especially if I am missing something here that would not be in the tests, and that would justify specialized parsing in certain cases.

### DIFF
--- a/cmd/nerdctl/image/image_list.go
+++ b/cmd/nerdctl/image/image_list.go
@@ -70,49 +70,47 @@ Properties:
 	return imagesCommand
 }
 
-func processImageListOptions(cmd *cobra.Command, args []string) (types.ImageListOptions, error) {
+func processImageListOptions(cmd *cobra.Command, args []string) (*types.ImageListOptions, error) {
 	globalOptions, err := helpers.ProcessRootCmdFlags(cmd)
 	if err != nil {
-		return types.ImageListOptions{}, err
+		return nil, err
 	}
 	var filters []string
-
 	if len(args) > 0 {
-		canonicalRef, err := referenceutil.ParseAny(args[0])
+		parsedReference, err := referenceutil.Parse(args[0])
 		if err != nil {
-			return types.ImageListOptions{}, err
+			return nil, err
 		}
-		filters = append(filters, fmt.Sprintf("name==%s", canonicalRef.String()))
-		filters = append(filters, fmt.Sprintf("name==%s", args[0]))
+		filters = []string{fmt.Sprintf("name==%s", parsedReference)}
 	}
 	quiet, err := cmd.Flags().GetBool("quiet")
 	if err != nil {
-		return types.ImageListOptions{}, err
+		return nil, err
 	}
 	noTrunc, err := cmd.Flags().GetBool("no-trunc")
 	if err != nil {
-		return types.ImageListOptions{}, err
+		return nil, err
 	}
 	format, err := cmd.Flags().GetString("format")
 	if err != nil {
-		return types.ImageListOptions{}, err
+		return nil, err
 	}
 	var inputFilters []string
 	if cmd.Flags().Changed("filter") {
 		inputFilters, err = cmd.Flags().GetStringSlice("filter")
 		if err != nil {
-			return types.ImageListOptions{}, err
+			return nil, err
 		}
 	}
 	digests, err := cmd.Flags().GetBool("digests")
 	if err != nil {
-		return types.ImageListOptions{}, err
+		return nil, err
 	}
 	names, err := cmd.Flags().GetBool("names")
 	if err != nil {
-		return types.ImageListOptions{}, err
+		return nil, err
 	}
-	return types.ImageListOptions{
+	return &types.ImageListOptions{
 		GOptions:         globalOptions,
 		Quiet:            quiet,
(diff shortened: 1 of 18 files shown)


pkg/cmd/builder/build.go
```diff
@@ -235,19 +235,19 @@ func generateBuildctlArgs(ctx context.Context, client *containerd.Client, option
 	}
 	if tags = strutil.DedupeStrSlice(options.Tag); len(tags) > 0 {
 		ref := tags[0]
-		named, err := distributionref.ParseNormalizedNamed(ref)
+		parsedReference, err := referenceutil.Parse(ref)
```
apostasie, line 238: Harmonizing variables names from `named`, `canonicalRef`, etc, to just the same `parsedReference`.

pkg/cmd/image/inspect.go
```diff
@@ -39,22 +37,16 @@ import (
 
 func inspectIdentifier(ctx context.Context, client *containerd.Client, identifier string) ([]images.Image, string, string, error) {
 	// Figure out what we have here - digest, tag, name
-	parsedIdentifier, err := referenceutil.ParseAnyReference(identifier)
+	parsedReference, err := referenceutil.Parse(identifier)
 	if err != nil {
-		return nil, "", "", fmt.Errorf("invalid identifier %s: %w", identifier, err)
```
apostasie, line 44: Useless. The underlying implementation does return that already.

cmd/nerdctl/image/image_list.go
```diff
@@ -70,49 +70,47 @@ Properties:
 	return imagesCommand
 }
 
-func processImageListOptions(cmd *cobra.Command, args []string) (types.ImageListOptions, error) {
+func processImageListOptions(cmd *cobra.Command, args []string) (*types.ImageListOptions, error) {
```
apostasie, line 73: Change the signature here so that we can return nil instead of `types.ImageListOptions{}`.

cmd/nerdctl/image/image_list.go
```diff
@@ -70,49 +70,47 @@ Properties:
...
-
 	if len(args) > 0 {
-		canonicalRef, err := referenceutil.ParseAny(args[0])
+		parsedReference, err := referenceutil.Parse(args[0])
 		if err != nil {
-			return types.ImageListOptions{}, err
+			return nil, err
 		}
-		filters = append(filters, fmt.Sprintf("name==%s", canonicalRef.String()))
-		filters = append(filters, fmt.Sprintf("name==%s", args[0]))
```
apostasie, line 86: This does not seem to serve a purpose.
Am I missing something here?

pkg/imgutil/imgutil.go
```diff
@@ -61,7 +61,7 @@ type PullMode = string
 // GetExistingImage returns the specified image if exists in containerd. Return errdefs.NotFound() if not exists.
 func GetExistingImage(ctx context.Context, client *containerd.Client, snapshotter, rawRef string, platform ocispec.Platform) (*EnsuredImage, error) {
 	var res *EnsuredImage
-	imagewalker := &imagewalker.ImageWalker{
+	imgwalker := &imagewalker.ImageWalker{
```
apostasie, line 64: Not shadowing import.

ktock (approved):

PR #3543, opened 2024-10-15
Global compose lock

apostasie:
Our compose implementation is inherently not safe to use concurrently.

The root issue is that we do shell out multiple times to nerdctl to manipulate a bunch of resources, while assuming a state of the world that has possibly changed meanwhile.
Because of that, we cannot leverage API locking mechanisms (eg: as implemented in VolStore, etc) across multiple distincts invocations of nerdctl.

This PR suggests that we put in place a global filesystem lock for compose.

This is not pretty - and it will definitely prevent running multiple `nerdctl compose` commands in parallel (they will be queued).
Also, this is not going to solve _all_ concurrency / racy issues with compose - as normal nerdctl operations may still run in parallel with a compose invocation.

But at least, it will prevent some of the most egregious problems - until we fix the core issues. 

Let me know your thoughts.

### DIFF
--- a/pkg/cmd/compose/compose.go
+++ b/pkg/cmd/compose/compose.go
@@ -35,14 +35,31 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/composer/serviceparser"
 	"github.com/containerd/nerdctl/v2/pkg/imgutil"
 	"github.com/containerd/nerdctl/v2/pkg/ipfs"
+	"github.com/containerd/nerdctl/v2/pkg/lockutil"
 	"github.com/containerd/nerdctl/v2/pkg/netutil"
 	"github.com/containerd/nerdctl/v2/pkg/referenceutil"
 	"github.com/containerd/nerdctl/v2/pkg/signutil"
 	"github.com/containerd/nerdctl/v2/pkg/strutil"
 )
 
+//nolint:unused
+var locked *os.File
+
 // New returns a new *composer.Composer.
 func New(client *containerd.Client, globalOptions types.GlobalCommandOptions, options composer.Options, stdout, stderr io.Writer) (*composer.Composer, error) {
+	// Compose right now cannot be made safe to use concurrently, as we shell out to nerdctl for multiple operations,
+	// preventing us from using the lock mechanisms from the API.
+	// This here imposes a global lock, effectively preventing multiple compose commands from being run in parallel and
+	// preventing some of the problems with concurrent execution.
+	// This should be removed once we have better, in-depth solutions to make this concurrency safe.
+	// Note that we do not close the lock explicitly. Instead, the lock will get released when the `locked` global
+	// variable will get collected and the file descriptor closed (eg: when the binary exits).
+	var err error
+	locked, err = lockutil.Lock(globalOptions.DataRoot)
+	if err != nil {
+		return nil, err
+	}
+
 	cniEnv, err := netutil.NewCNIEnv(globalOptions.CNIPath, globalOptions.CNINetConfPath, netutil.WithNamespace(globalOptions.Namespace), netutil.WithDefaultNetwork())
 	if err != nil {
 		return nil, err


pkg/cmd/compose/compose.go
```diff
@@ -35,14 +35,29 @@ import (
...
+	// preventing us from using the lock mechanisms from the API.
+	// This here imposes a global lock, effectively preventing multiple compose commands from being run in parallel and
+	// preventing some of the problems with concurrent execution.
+	// This should be removed once we have better, in-depth solutions to make this concurrency safe.
+	var err error
+	locked, err = lockutil.Lock(globalOptions.DataRoot)
+	if err != nil {
+		return nil, err
+	}
+
```
AkihiroSuda, line 60: Where is unlock?

pkg/cmd/compose/compose.go
```diff
@@ -35,14 +35,29 @@ import (
...
 	"github.com/containerd/nerdctl/v2/pkg/ipfs"
+	"github.com/containerd/nerdctl/v2/pkg/lockutil"
 	"github.com/containerd/nerdctl/v2/pkg/netutil"
 	"github.com/containerd/nerdctl/v2/pkg/referenceutil"
 	"github.com/containerd/nerdctl/v2/pkg/signutil"
 	"github.com/containerd/nerdctl/v2/pkg/strutil"
 )
 
+//nolint:unused
+var locked *os.File
```
AkihiroSuda, line 46: Why global var?

pkg/cmd/compose/compose.go
apostasie: When `nerdctl compose` exits, it will release the lock - we do not release explicitly.

The reason I did no do it explicitly is because (I think) we would have to have that code inside *all* compose commands, which would make for a large/more intrusive patch.
This makes here for a much simpler patch.

pkg/cmd/compose/compose.go
apostasie: Otherwise the lock will get released once we exit the scope of `New()`.
We could add the lock to the Composer struct, but I am not sure that would bring much value?

pkg/cmd/compose/compose.go
AkihiroSuda: Could you add code comments to explain that

pkg/cmd/compose/compose.go
AkihiroSuda: Could you add code comments?

Also, how is the file lock associated with the lifetime of the Go variable? I don’t think *os.File has a destructor method to unlock the file?

pkg/cmd/compose/compose.go
AkihiroSuda: Also wondering if we can just add *Composer.Close

pkg/cmd/compose/compose.go
apostasie: > Could you add code comments?

Sure.

> Also, how is the file lock associated with the lifetime of the Go variable? I don’t think *os.File has a destructor method to unlock the file?

I was surprised too.

My (vague) explanation is that garbage collection does close the file descriptor - and yes, flock() will release the lock when the fd gets closed.

pkg/cmd/compose/compose.go
apostasie: > Also wondering if we can just add `*Composer.Close()`.
> The caller of `compose.New` should have `defer c.Close()`

Sure.

Is it worth the effort and extra code though?

AkihiroSuda:
Do you plan to update this for Unlock()”

apostasie:
> Do you plan to update this for Unlock()?

I would rather not - unless you think it would really be better? Just lmk.

As far as I can think about this, the meaning of this lock is to stay for the entire life of the command execution. Unlocking it manually does not seem to serve much purpose. This here overall is a ugly hack to alleviate a hard to fix problem - but should not stay in the long term.

AkihiroSuda (approved):
Sure, let’s merge

PR #3544, opened 2024-10-15
[fix] fix image filter with nonexist image

### LINKED ISSUE #3511: `images --filter=non-existent` inconsistent (docker vs. nerdctl)
### Description

```
docker images --filter since=non-exists-image; echo $?
Error response from daemon: No such image: non-exists-image:latest
1
```

```
nerdctl images --filter since=non; echo $?
0
```

This recently got fixed because of a panic - in #3469 - though the tests being currently disabled for Docker made this go under the radar.

Tagging @xyz-li in case you are interested in revisiting. 

na

na

main

None

### Host information

_No response_

xyz-li:
Fix #3511

This commit does not cover this situation `nerdctl images -f since=IMAGE_ID`

### DIFF
--- a/pkg/imgutil/filtering.go
+++ b/pkg/imgutil/filtering.go
@@ -145,18 +145,28 @@ func FilterByCreatedAt(ctx context.Context, client *containerd.Client, before []
 			maxTime = time.Now()
 		)
 
+		fetchImageNames := func(names []string) string {
+			parsedNames := make([]string, 0, len(names))
+			for _, name := range names {
+				parsedNames = append(parsedNames, strings.TrimPrefix(name, "name=="))
+			}
+			return strings.Join(parsedNames, ",")
+		}
+
 		imageStore := client.ImageService()
 		if len(before) > 0 {
 			beforeImages, err := imageStore.List(ctx, before...)
 			if err != nil {
 				return []images.Image{}, err
 			}
-			if len(beforeImages) > 0 {
-				maxTime = beforeImages[0].CreatedAt
-				for _, image := range beforeImages {
-					if image.CreatedAt.After(maxTime) {
-						maxTime = image.CreatedAt
-					}
+			if len(beforeImages) == 0 {
+				//nolint:stylecheck
+				return []images.Image{}, fmt.Errorf("No such image: %s", fetchImageNames(before))
+			}
+			maxTime = beforeImages[0].CreatedAt
+			for _, image := range beforeImages {
+				if image.CreatedAt.After(maxTime) {
+					maxTime = image.CreatedAt
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/imgutil/filtering.go
```diff
@@ -145,18 +145,34 @@ func FilterByCreatedAt(ctx context.Context, client *containerd.Client, before []
 			maxTime = time.Now()
 		)
 
+		parseImageName := func(names []string) string {
+			var name string
+			if len(names) > 1 {
+				name = names[1]
+			} else {
+				name = names[0]
+			}
```
djdongjin, line 154: can you explain why we use `names[1]` (not 0, not >1) if len(names) > 1?

xyz-li: If you run command `nerdctl images -f since=nginx`. Then the first item is `docker.io/library/nginx:latest`, and the second item is `nginx`

https://github.com/containerd/nerdctl/blob/427f1cb6697af2a5c8d17550fb58d9f09aaa0a9e/pkg/imgutil/filtering.go#L96-L97

djdongjin: how about lets just use `strings.Join(names, ",")`? given this is what we're actually doing.

pkg/imgutil/filtering.go
```diff
@@ -145,18 +145,34 @@ func FilterByCreatedAt(ctx context.Context, client *containerd.Client, before []
...
 			}
-			if len(beforeImages) > 0 {
-				maxTime = beforeImages[0].CreatedAt
-				for _, image := range beforeImages {
-					if image.CreatedAt.After(maxTime) {
-						maxTime = image.CreatedAt
-					}
+			if len(beforeImages) == 0 {
+				//nolint:stylecheck
+				return []images.Image{}, fmt.Errorf("No such image: %s", parseImageName(before))
```
djdongjin, line 170: why do we need `nolint` here?

xyz-li: "No such image: aaaa:latest" is the output of docker. If don't add this `nolint`, then golint-ci won't pass, or we can change the output.

pkg/imgutil/filtering.go
xyz-li: `FATA[0000] No such image: docker.io/library/aaaa:latest,aaaa:latest`. I don't know if this is better or not?

pkg/imgutil/filtering.go
djdongjin: >  If don't add this nolint, then golint-ci won't pass

what's the error message from golint-ci? (Or you mean `"No such image: aaaa:latest" is the output of docker.` is the error message from golint-ci?)

xyz-li:
```
 Running [/home/runner/golangci-lint-1.61.0-linux-amd64/golangci-lint run  --verbose] in [/home/runner/work/nerdctl/nerdctl] ...
  Error: pkg/imgutil/filtering.go:163:30: ST1005: error strings should not be capitalized (stylecheck)
  				return []images.Image{}, fmt.Errorf("No such image: %s", fetchImageNames(before))
  				                         ^
  Error: pkg/imgutil/filtering.go:179:30: ST1005: error strings should not be capitalized (stylecheck)
  				return []images.Image{}, fmt.Errorf("No such image: %s", fetchImageNames(since))
  	
```

djdongjin (approved):

PR #3551, opened 2024-10-15
Add image load and container run quiet mode

austinvazquez:
### Issue
Split from https://github.com/containerd/nerdctl/pull/3537

### Description
This change adds `-q|--quiet` flag on `nerdctl load`, `nerdctl container create`, and `nerdctl container run` commands to suppress image pull and load output.

### Testing
1. Manual tested behavior aligned with Docker 25.
```bash
$ docker load -q -i alpine-latest.tar 
Loaded image: alpine:latest
$ _output/nerdctl load -q -i alpine-latest.tar 
Loaded image: alpine:latest
$ 
```
```bash
$ docker run --rm --quiet alpine:latest echo "test run quiet"
test run quiet
$ _output/nerdctl run --rm --quiet alpine:latest echo "test run quiet"
test run quiet
$ 
```
2. Added integration tests

### DIFF
--- a/cmd/nerdctl/container/container_create.go
+++ b/cmd/nerdctl/container/container_create.go
@@ -409,12 +409,17 @@ func processContainerCreateOptions(cmd *cobra.Command) (types.ContainerCreateOpt
 	if err != nil {
 		return opt, err
 	}
+	quiet, err := cmd.Flags().GetBool("quiet")
+	if err != nil {
+		return opt, err
+	}
 	opt.ImagePullOpt = types.ImagePullOptions{
 		GOptions:      opt.GOptions,
 		VerifyOptions: imageVerifyOpt,
 		IPFSAddress:   opt.IPFSAddress,
 		Stdout:        opt.Stdout,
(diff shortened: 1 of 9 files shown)


AkihiroSuda (approved):
Thanks

AkihiroSuda:
> Not necessarily in this PR, but I am wondering if `--quiet` should be a global flag - given the fact we do need to pull missing content on many operations (convert, tag, commit, push) ([containerd/containerd#8973](https://github.com/containerd/containerd/issues/8973))

Probably this is safe, but Docker didn't do this

apostasie:
> > Not necessarily in this PR, but I am wondering if `--quiet` should be a global flag - given the fact we do need to pull missing content on many operations (convert, tag, commit, push) ([containerd/containerd#8973](https://github.com/containerd/containerd/issues/8973))
> 
> Probably this is safe, but Docker didn't do this

I guess containerd lazy pulling / remote snapshotters does not apply for Docker, which is why they do not have this class of problem.

PR #3553, opened 2024-10-16
Fix over-eager windows path restrictions

apostasie:
Current regexp will match anything _starting_ with forbidden patterns.
This changes properly reduces that to paths that _exactly_ match the forbidden patterns.

Discovered while working on enabling unit-tests for windows.


djdongjin (approved):
LGTM, thanks

AkihiroSuda (approved):
Thanks

PR #3555, opened 2024-10-16
Builder tests

apostasie:
This PR focused solely on rewriting builder tests with the new test framework.

Any other changes (in testutil) are removal of no-longer used helpers, and making sure we can compile on other platforms as we make test files move away from _platform.go patterns.

Besides 1-1 rewrite, a few test changes are worth noting:
- (re)introducing a non-native platform build test
- making some tests previously ran only on linux to also run on windows
- also note that we no longer perform the `builder prune` op that was previously in place. The key upside is that tests can now be ran in parallel instead of sequentially, with no downside as far as I can tell (if there is any, we can always --no-cache where it matters)

This PR has been reduced as much as feasible to not contain any other changes.

I know it is still big (1k lines)... - lmk if you think there is a good way to break this up.


apostasie:
Failure is #3556

apostasie:
Re-run failures are:
- TestLogsWithInheritedFlags for Docker - already spotted as flaky with Docker before - is the test broken or is docker broken?
- TestAttach: despite a lot of efforts to tweak delays, this keeps failing again

apostasie:
Failures are fixed by #3563

pkg/testutil/testutil.go
```diff
@@ -859,21 +859,3 @@ func KubectlHelper(base *Base, args ...string) *Cmd {
 		Base: base,
 	}
 }
-
-// SetupDockerContainerBuilder creates a Docker builder using the docker-container driver
-// and adds cleanup steps to test cleanup. The builder name is returned as output.
-//
-// If not docker, this function returns an empty string as the builder name.
-func SetupDockerContainerBuilder(t *testing.T) string {
```
apostasie, line 867: This is used only once and can be simplified in the test in-place.

pkg/testutil/testutil_windows.go
```diff
@@ -43,7 +43,9 @@ const (
 	NginxAlpineImage            = "registry.k8s.io/e2e-test-images/nginx:1.14-2"
 	NginxAlpineIndexHTMLSnippet = "<title>Welcome to nginx!</title>"
 
-	GolangImage = "fixme-test-using-this-image-is-disabled-on-windows"
+	GolangImage  = "fixme-test-using-this-image-is-disabled-on-windows"
```
apostasie, line 46: As we remove more of platform-specific test files and replace with platform-dependent `t.Skip`, we need a few placeholders so that we compile.

The point of this (as exposed ealier), is to make platform-specific skip/unskip easier than it is today.

ktock (approved):

apostasie:
Thanks a lot @ktock

PR #3557, opened 2024-10-16
Add delay to tentatively fix flakyness with Docker

apostasie:


cmd/nerdctl/container/container_logs_test.go
```diff
@@ -95,6 +95,9 @@ func TestLogsWithInheritedFlags(t *testing.T) {
 	base.Cmd("run", "-d", "--name", containerName, testutil.CommonImage,
 		"sh", "-euxc", "echo foo; echo bar").AssertOK()
 
+	// It appears this test flakes out with Docker seeing only "foo\n"
+	// Tentatively adding a pause in case this is just slow
+	time.Sleep(time.Second)
```
AkihiroSuda, line 100: Seems a bug in Docker?

AkihiroSuda (approved):
Thanks

cmd/nerdctl/container/container_logs_test.go
apostasie: Maybe - if that does not help, I ll just mark it skipped for docker

apostasie:
Similar failure happened in `TestStopWithStopSignal`.
Documenting with #3564.

There may indeed be some bug in docker logs.

Issue #3558, opened 2024-10-16
Disable TestAttach


apostasie:
### Description

Despite many attempts at tweaking timeout, TestAttach keeps on failing.

I do have a rewritten version that seems to flake less, but there is still a lot of work on containers tests before I can send it.

Suggesting we skip it for the time being for quality of life.

### Steps to reproduce the issue

Run tests.

### Describe the results you received and expected

```
=== RUN   TestAttach
=== PAUSE TestAttach
=== CONT  TestAttach
    container_attach_linux_test.go:58: assertion failed: 
        Command:  /usr/bin/unbuffer -p /usr/local/bin/nerdctl --namespace=nerdctl-test run -it --name nerdctl-testattach ghcr.io/stargz-containers/alpine:3.13-org
        ExitCode: 0
        Stdout:   
        Stderr:   
        
        Failures:
        Expected stdout to contain "read detach keys"
--- FAIL: TestAttach (7.03s)
```

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #3561, opened 2024-10-17
Add pseudo tty support for tests

apostasie:
Right now we rely on unbuffer.

This is problematic for a couple of reasons:
- unbuffer merges stdout and stderr together
- it is a binary dependency on the host
- in light of #3558 it makes debugging more complicated by having another (blackbox) piece at play

Suggesting we just remove unbuffer.

This PR implements basic pty support for tests, and removes unbuffer from a first test.

### DIFF
--- a/pkg/testutil/test/command.go
+++ b/pkg/testutil/test/command.go
@@ -45,6 +45,7 @@ type GenericCommand struct {
 	envBlackList []string
 	stdin        io.Reader
 	async        bool
+	pty          bool
 	timeout      time.Duration
 	workingDir   string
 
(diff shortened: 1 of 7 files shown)


pkg/testutil/test/command.go
```diff
@@ -90,6 +96,16 @@ func (gc *GenericCommand) Run(expect *Expected) {
...
 		env = iCmdCmd.Env
+
+		if gc.pty {
+			pty, tty, _ := Open()
+			iCmdCmd.Stdin = tty
+			iCmdCmd.Stdout = tty
+			iCmdCmd.Stderr = tty
+			defer pty.Close()
+			defer tty.Close()
+		}
```
AkihiroSuda, line 107: The pty variable is just closed without using it?

apostasie: It is just the handle of /dev/ptmx - only purpose is to get a new /dev/pts/x

AkihiroSuda (approved):
Thanks

PR #3562, opened 2024-10-17
Image completion and tests

apostasie:
There does not seem to be a reason for image completion specialization per-platform, nor for limiting completion tests to linux.

This came out while noticing some completion tests for builder that are not linux specific.


apostasie:
@AkihiroSuda can you poke the CI? (`Unable to connect to archive.ubuntu.com`)

AkihiroSuda (approved):
Thanks

PR #3535, opened 2024-10-14
CI adjustments

apostasie:
~~NOTE: this is pending #3572, #3573, #3574, #3575~~

This is the next part of the test rework effort.

### The most important change here

Is that we will now have two separate steps for every integration test run:
- without retry (this is new) - running all tests that are deemed safe ONCE, and hard failing if one of them fails
- with retry (this is the previous behavior) - running any _other_ test

Note that the above split does NOT apply to ipv6, nor kube, tests. These are considered safe.

Right now, the balance is about 286 (rootful, no retry) vs. 500 (with retry).
Hopefully, as we fix and migrate more tests, and address bugs in nerdctl, it will shift.

Note that running legacy tests without retries is currently very painful and does surface a lot of conditions that have been previously unseen / ignored for a long time. As a lot of bugs have been recently fixed with regards to concurrency, underlying issues are also starting to crop up, and these very likely run deep / are involved to fix.
These are chiefly:
- #3570
- #3516
- #3556
- #3513

This is the reason why we are - conservatively - leaving the pre-existing, legacy tests in the "with retries" bucket.

There is now a helper script, `test-integration.sh`.

When ran without any argument, if will run both steps in order:

If you want to run ONLY the first step (aka "safe tests"), call:
`test-integration.sh -test.only-flaky=false`

If you want to run ONLY the second step, call:
`test-integration.sh -test.only-flaky=true`

That works as well with gotestsum, or go test, evidently:

Note though that the default in that case is `-test.only-flaky=false`:
```
go test ./cmd/nerdctl/image
go test ./cmd/nerdctl/image -test.only-flaky
# etc...
```

### Other CI changes
- cosmetic efforts on test display titles (more compact, more consistent)
- removal of all nick/retries
- minor tweaks to the Dockerfile to adapt for new script

PR #3563, opened 2024-10-17
Change compose lock to use DataStore()

apostasie:
In some rare (?) circumstances, DataRoot might not exist yet (eg: fresh install, first command being run being a `compose` command).

Changing the lock location to rely on `DataStore()` instead, which does provide existence guarantee (and is also a bit more specific).

Failure was seen on unmerged PRs CI runs when parallelization triggered an execution order that produced the situation above.


apostasie:

apostasie:
@djdongjin if you are still around, I can use this to fix CI bustage on upcoming PRs (specifically #3535)

AkihiroSuda (approved):
Thanks

PR #3550, opened 2024-10-15
Part 5: tests

apostasie:
Broken out of #3535

PR #3564, opened 2024-10-17
Chores

apostasie:
- duplicated title test fixed
- marking / unmarking tests as flaky based on recent runs
- spurious comments
- (useless) leftovers from previous iteration in referenceutils_tests

This contains some changes broken out of #3550 (while it is being reviewed, we can still get these in and I will rebase)

### DIFF
--- a/cmd/nerdctl/image/image_list_test.go
+++ b/cmd/nerdctl/image/image_list_test.go
@@ -18,6 +18,7 @@ package image
 
 import (
 	"fmt"
+	"runtime"
 	"slices"
 	"strings"
 	"testing"
@@ -117,6 +118,13 @@ func TestImages(t *testing.T) {
 		},
 	}
 
+	if runtime.GOOS == "windows" {
+		testCase.Require = test.Require(
+			testCase.Require,
+			nerdtest.IsFlaky("https://github.com/containerd/nerdctl/issues/3524"),
+		)
+	}
+
 	testCase.Run(t)
 }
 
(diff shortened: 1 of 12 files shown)


cmd/nerdctl/image/image_list_test.go
```diff
@@ -117,6 +118,13 @@ func TestImages(t *testing.T) {
 		},
 	}
 
+	if runtime.GOOS == "windows" {
+		testCase.Require = test.Require(
+			testCase.Require,
+			nerdtest.IsFlaky("https://github.com/containerd/nerdctl/issues/3524"),
+		)
```
djdongjin, line 125: can you explain this a bit?

apostasie: Sure.

On windows, this may fail (randomly) - detailed in the linked ticket.

`nerdtest.IsFlaky("issue link")` is part of the new test framework (https://github.com/containerd/nerdctl/blob/main/docs/testing/tools.md#requirements)

It marks a test as flaky, which makes it run only if `-test.only-flaky` is provided to the integration test.
Very soon, with PR #3535 we will have separate steps to run non-flaky (no retries, hard failure) and flaky tests (retries, hard fail only if all retries fail).

Now, this test also has a requirement of `not(docker)`: https://github.com/containerd/nerdctl/blob/ce8e077d9548277ea05e543565cb7d33b7e57e65/cmd/nerdctl/image/image_list_test.go#L40
and I do not want that to get overridden here (otherwise, if ran on windows against target docker, the test would not be skipped).

`test.Require(reqs...)` is a helper that just bundles together a set of requirements (the previous one `testCase.Require`, and the conditional one `IsFlaky`).

Lmk if that does not clarify or if your question was about something I missed?

Thanks @djdongjin

djdongjin (approved):

AkihiroSuda (approved):
Thanks

PR #3565, opened 2024-10-17
Fix semantic of Fail to not care about exit code

apostasie:
This is a small change broken out of #3550 which is pending review (will rebase once this is in).

`Fail()` is not expected to verify the exitcode, just that the task fail.
This is specifically important as Docker often returns 127 instead of 1 in certain conditions.


djdongjin (approved):
LGTM, thanks

PR #3566, opened 2024-10-17
Fix and enable CI unit testing for windows

apostasie:
Reopening erroneously closed previous PR.

This one no longer depends on other pending CI changes and can be merged now.

Take-away:
- couple of tests marked as not compatible
- one test that is blatantly disagreeing with code has been adapted
- error output specialization
- filepath.Join
- enforcing windows-unit on CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,8 +40,19 @@ jobs:
           make lint-imports
 
   test-unit:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 20
+    timeout-minutes: 5
+    name: unit | ${{ matrix.goos }}
+    runs-on: "${{ matrix.os }}"
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      matrix:
+        include:
+          - os: windows-2022
+            goos: windows
+          - os: ubuntu-24.04
+            goos: linux
     steps:
       - uses: actions/checkout@v4.2.1
         with:
@@ -51,6 +62,17 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
           check-latest: true
           cache: true
+      - if: ${{ matrix.goos=='windows' }}
+        uses: actions/checkout@v4.2.1
+        with:
+          repository: containerd/containerd
+          ref: v1.7.23
+          path: containerd
+          fetch-depth: 1
+      - if: ${{ matrix.goos=='windows' }}
+        name: "Set up CNI"
+        working-directory: containerd
+        run: GOPATH=$(go env GOPATH) script/setup/install-cni-windows
       - name: "Run unit tests"
         run: go test -v ./pkg/...
 
--- a/pkg/cmd/builder/build_test.go
+++ b/pkg/cmd/builder/build_test.go
@@ -17,7 +17,10 @@
 package builder
 
 import (
+	"fmt"
+	"path/filepath"
 	"reflect"
+	"runtime"
 	"testing"
 
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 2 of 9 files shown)


apostasie:
@djdongjin if you are around - this was previously dependent / blocked by other under-review PRs (eg: part5), but I rebased that differently, and could be merged now.

.github/workflows/test.yml
```diff
@@ -51,6 +62,17 @@ jobs:
...
+        uses: actions/checkout@v4.2.1
+        with:
+          repository: containerd/containerd
+          ref: v1.7.23
+          path: containerd
+          fetch-depth: 1
+      - if: ${{ matrix.goos=='windows' }}
+        name: "Set up CNI"
+        working-directory: containerd
+        run: GOPATH=$(go env GOPATH) script/setup/install-cni-windows
```
djdongjin, line 75: why only windows requires this?

apostasie: Magic trick. :-)

apostasie: Here is my tentative explanation: somewhere, unit tests does go through a code path that depends on `generateNetworkConfig`, which is going to check for a binary on the system for the corresponding plugin-type name (`GetPluginType`).

For windows, it is looking for "nat", which is only provided by the cni plugins.

For linux, it is looking for "bridge"... which is present on most systems and have nothing to do with CNI - hence we get away with running the unit tests.

I do not think the unit tests themselves actually depend on running the binary (obviously...), but yep, that is making the case that we should look into that code.

djdongjin (approved):

pkg/cmd/builder/build_test.go
```diff
@@ -213,6 +214,10 @@ func TestParseBuildctlArgsForOCILayout(t *testing.T) {
 		},
 	}
 
+	if runtime.GOOS == "windows" {
+		tests[1].expectedErr = "open D:\\tmp\\oci-layout\\index.json: The system cannot find the path specified."
```
AkihiroSuda, line 218: D:\\tmp shouldn't be hardcoded?

apostasie: Interesting.
Yeah, I guess it might vary from one install to the other?

I really wish I had a windows install around.

Let me look into this real quick.

apostasie: Ok, that should do the trick.

@AkihiroSuda PTAL.

pkg/cmd/builder/build_test.go
```diff
@@ -213,6 +216,15 @@ func TestParseBuildctlArgsForOCILayout(t *testing.T) {
 		},
 	}
 
+	if runtime.GOOS == "windows" {
+		abspath, err := filepath.Abs("/tmp/oci-layout")
```
AkihiroSuda, line 220: Shouldn't this be os.Getenv("TEMP")

apostasie: The path `/tmp/oci-layout` was / is hardcoded in the oci query part of the test:

https://github.com/containerd/nerdctl/blob/ca76611c84c34e606c8e4ef19047808573193906/pkg/cmd/builder/build_test.go#L206-L213

We could change it in both the test declaration and the assert - to GetTemp / Getenv (or whatever) else - but I am not sure it would serve much purpose: it seems this test is solely testing the parsing part of `parseBuildContextFromOCILayout` - which does exactly that (`filepath.Abs`) after stripping the scheme and will immediately error since there is no oci data in that location.

https://github.com/containerd/nerdctl/blob/ca76611c84c34e606c8e4ef19047808573193906/pkg/cmd/builder/build.go#L562

The path could very much be `/foo/bar` for all intent and purpose, and I presume who wrote the test did just put `/tmp/oci-layout` because that was the first thing they had in mind.

Lmk your thoughts.

apostasie: To be completely honest, I am not sure this test is bringing much (or any) value, but then, it is here - so...

PR #3573, opened 2024-10-18
Fix ensurecontainerstarted logic and better debug

apostasie:
Breakout of #3535 

This fixes logic issues in hasty porting of EnsureContainerStarted, and enhances debugability.

### DIFF
--- a/pkg/testutil/test/command.go
+++ b/pkg/testutil/test/command.go
@@ -28,6 +28,9 @@ import (
 	"gotest.tools/v3/icmd"
 )
 
+const ExitCodeGenericFail = -1
+const ExitCodeNoCheck = -2
+
 // GenericCommand is a concrete Command implementation
 type GenericCommand struct {
 	Config  Config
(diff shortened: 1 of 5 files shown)


pkg/testutil/test/command.go
```diff
@@ -117,7 +117,10 @@ func (gc *GenericCommand) Run(expect *Expected) {
 		// Build the debug string - additionally attach the env (which iCmd does not do)
 		debug := result.String() + "Env:\n" + strings.Join(env, "\n")
 		// ExitCode goes first
-		if expect.ExitCode == -1 {
+		if expect.ExitCode == -2 { //nolint:revive
+			// -2 means we do not care at all about exit code
+		} else if expect.ExitCode == -1 {
+			// -1 means any error
```
AkihiroSuda, line 123: These magic numbers should be defined as `const`

apostasie: Yep, good point.

Done.

AkihiroSuda (approved):
Thanks

PR #3574, opened 2024-10-18
Workaround and document weird docker issue

apostasie:
Seen while debugging #3535 

Not quite sure what is going on, but for now the workaround will do.

Tagging @thaJeztah in case there would be interest in investigating more.

### DIFF
--- a/cmd/nerdctl/container/container_commit_test.go
+++ b/cmd/nerdctl/container/container_commit_test.go
@@ -36,6 +36,12 @@ func TestCommit(t *testing.T) {
 				helpers.Anyhow("rmi", "-f", data.Identifier())
 			},
 			Setup: func(data test.Data, helpers test.Helpers) {
+				// FIXME: short of pulling first, docker will fail to start the container.
+				// Debugging shows container exited immediately. Nothing in the container logs. Not much in journalctl.
+				// It is not clear what is happening.
+				if nerdtest.IsDocker() {
+					helpers.Ensure("pull", testutil.CommonImage)
+				}
 				helpers.Ensure("run", "-d", "--name", data.Identifier(), testutil.CommonImage, "sleep", "infinity")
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/container/container_commit_test.go
```diff
@@ -36,6 +36,12 @@ func TestCommit(t *testing.T) {
 				helpers.Anyhow("rmi", "-f", data.Identifier())
 			},
 			Setup: func(data test.Data, helpers test.Helpers) {
+				// FIXME: short of pulling first, docker will fail to start the container.
+				// Debugging shows container exited immediately. Nothing in the container logs. Not much in journalctl.
```
AkihiroSuda, line 40: Can you repro the issue without using the nerdctl test suite?

AkihiroSuda: Should be reported to https://github.com/moby/moby/issues

apostasie: It does happen on #3535, but I cannot repro locally for now.

I believe it was previously hidden with retries.

I will report on moby if I manage to get a reproducer outside of the CI context.

AkihiroSuda (approved):

cmd/nerdctl/container/container_commit_test.go
thaJeztah: No idea; haven't seen that issue. With the default pull policy ("missing"), the CLI should trigger a pull when the API endpoint returns a 404; https://github.com/docker/cli/blob/8a7c5ae68f42061fa3085736b91a9e831e0eab07/cli/command/container/create.go#L272-L289

apostasie: I believe the pull was successful on run.
Totally spitballing: this might be a race issue with docker starting and starting a container?

Issue #3556, opened 2024-10-16
CNIEnv concurrency issues


apostasie:
### Description

Although #3491 and #3522 have fixed a lot of cases where CNI would fail because of concurrent access, there are still cases where this happens.

Here, on container create - but very likely everywhere else we manipulate CNIEnv.

We can continue playing wack-a-mole on this and fix every occurrence piece-meal, though it seems like rewriting CNIEnv in a safe way would be a better approach at this point.

The fundamental problems are:
- we rely on CNI implementation
    - not safe wrt concurrency, as it is walking dirs without a lock
    - writes are not atomic, leaving systems in broken / inconsistent states
- we do lock in some places, but not everywhere, as this is an afterthought and not part of the design
- we have complicated code, with private methods calling public ones, further complicated enforcing locking
- we unnecessarily walk the directory repeatedly during the same flow
- some of the logic currently in pkg/cmd should really be part of the methods of CNIEnv

### Steps to reproduce the issue

```
FAIL: cmd/nerdctl/network TestNetworkCreate/with_MTU (0.17s)
    network_create_linux_test.go:108: ======================== Pre-test cleanup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test network rm testnetworkcreate-with-mtu-1b256b01
    network_create_linux_test.go:108: ======================== Test setup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test network create testnetworkcreate-with-mtu-1b256b01 --driver bridge --opt com.docker.network.driver.mtu=9216
    network_create_linux_test.go:108: ======================== Test Run ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test run --rm --net testnetworkcreate-with-mtu-1b256b01 ghcr.io/stargz-containers/alpine:3.13-org ifconfig eth0
    command.go:112: assertion failed: expect.ExitCode is not result.ExitCode: Expected exit code: 0
        
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test run --rm --net testnetworkcreate-with-mtu-1b256b01 ghcr.io/stargz-containers/alpine:3.13-org ifconfig eth0
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
[...]

PR #3583, opened 2024-10-21
GHA cache image build dependencies

apostasie:
Folllow-on to discussion in #3580

What this does:
- enable the use of github actions cache for docker build of the integration image
- in the Dockerfile, separate third-party dependency build stage and nerdctl stage
- in github workflows, separate out the step that builds these dependencies (read-write to cache) from the steps that prepare the final integration image (read-only from cache)

The result is:
- with a cold cache, instead of building everything 9 times (4 rootful, 4 rootless, ipv6) , we only build the dependencies stage 4 times
- the dependencies stage is then cached, and _subsequent runs on the same branch_ will not rebuild it (unless the dependencies versions have changed, or the Dockerfile has been altered)
- note that we SHOULD also get the cache from the main branch as well, so, once it is merged and the cache is hot, new PRs should benefit from the cache as well on their first run (eg: not build)

The key benefit here is about reducing the network traffic required to produce out test images (hence reducing the opportunity for failure due to third-party server hiccups). It is similar in intent to #3580 .

Incidentally, we will also get a small speed boost for the overall run.
Obviously, GHA cache is not "free": it takes time to retrieve and time to store - so, part of the speed gains from not-building are negated by the cache r/w.

Nevertheless, this looks promising for increased reliability (and reduced transactions with docker hub / debian / ubuntu).

This PR also has a couple of minor changes to the workflow file (reduced timeouts, cosmetic comments, along with bumping the size of the arm64 instance as previously discussed). If preferable, I can split these out.

Further refactoring / changes to the Dockerfile could bring more stuff in the dependencies stage.
This PR has been conservative on that front and staid with the minimal possible changes to the Dockerfile, so that we can decide separately if we want a more in-depth restructuring of it or not.

Finally note that GHA cache is rather limited (10G), and going over the limit will prune prior entries indiscriminately - we might want to keep an eye on that and check that this proposed implementation here stays under the limit to fully benefit from it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,47 @@ env:
   LONG_TIMEOUT: 60
 
 jobs:
+  # This job builds the dependency target of the test docker image for all supported architectures and cache it in GHA
+  build-dependencies:
+    timeout-minutes: 10
+    name: dependencies |  ${{ matrix.containerd }} | ${{ matrix.arch }}
+    runs-on: "${{ matrix.runner }}"
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runner: ubuntu-24.04
+            containerd: v1.6.36
+            arch: amd64
+          - runner: ubuntu-24.04
+            containerd: v1.7.23
+            arch: amd64
+          - runner: ubuntu-24.04
+            containerd: v2.0.0-rc.5
+            arch: amd64
+          - runner: arm64-8core-32gb
+            containerd: v2.0.0-rc.5
+            arch: arm64
+    env:
+      CONTAINERD_VERSION: "${{ matrix.containerd }}"
+      ARCH: "${{ matrix.arch }}"
+    steps:
+      - uses: actions/checkout@v4.2.1
+        with:
+          fetch-depth: 1
+      - name: "Expose GitHub Runtime variables for gha"
+        uses: crazy-max/ghaction-github-runtime@v3
+      - name: "Build dependencies for the integration test environment image"
+        run: |
+          docker buildx create --name with-gha --use
+          docker buildx build \
+            --output=type=docker \
+            --cache-to type=gha,mode=max,scope=${ARCH}-${CONTAINERD_VERSION} \
+            --cache-from type=gha,scope=${ARCH}-${CONTAINERD_VERSION} \
+            --target build-dependencies --build-arg CONTAINERD_VERSION=${CONTAINERD_VERSION} .
+
   test-unit:
+    # FIXME:
     # Supposed to work: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-returning-a-json-data-type
     # Apparently does not
     # timeout-minutes: ${{ fromJSON(env.SHORT_TIMEOUT) }}
@@ -56,7 +96,8 @@ jobs:
         run: make test-unit
 
   test-integration:
-    timeout-minutes: 60
+    needs: build-dependencies
+    timeout-minutes: 30
     name: rootful | ${{ matrix.containerd }} | ${{ matrix.runner }}
     runs-on: "${{ matrix.runner }}"
     strategy:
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -69,6 +69,7 @@ RUN xx-apt-get update -qq && xx-apt-get install -qq --no-install-recommends \
     libbtrfs-dev \
     libseccomp-dev \
     pkg-config
+RUN git config --global advice.detachedHead false
```
apostasie, line 72: Silence useless / noisy git messages.

Dockerfile
```diff
@@ -117,25 +118,19 @@ RUN xx-go --wrap && \
...
-COPY . /go/src/github.com/containerd/nerdctl
-WORKDIR /go/src/github.com/containerd/nerdctl
+RUN git config --global advice.detachedHead false
 
 FROM build-base AS build-minimal
 RUN BINDIR=/out/bin make binaries install
 # We do not set CMD to `go test` here, because it requires systemd
 
-FROM build-base AS build-full
+FROM build-base AS build-deps
```
apostasie, line 127: This is the part that splits out third-party dependencies from nerdctl.

apostasie:
@AkihiroSuda @djdongjin this should _not_ go in before #3535, but this looks really promising in term of minimizing transactions with Hub / third party services.

LMK your thoughts overall and I will polish it and rebase on top of ^ after it merges

Dockerfile
```diff
@@ -117,25 +118,19 @@ RUN xx-go --wrap && \
...
 
 FROM build-base AS build-minimal
 RUN BINDIR=/out/bin make binaries install
 # We do not set CMD to `go test` here, because it requires systemd
 
-FROM build-base AS build-full
+FROM build-base AS build-dependencies
 ARG TARGETARCH
 ENV GOARCH=${TARGETARCH}
-RUN BINDIR=/out/bin make binaries install
```
apostasie, line 130: Move down nerdctl, which is the part that changes (almost) all the time.

Dockerfile
```diff
@@ -181,13 +176,6 @@ RUN git clone https://github.com/containerd/imgcrypt.git /go/src/github.com/cont
   git checkout "${IMGCRYPT_VERSION}" && \
   CGO_ENABLED=0 make && DESTDIR=/out make install && \
   echo "- imgcrypt: ${IMGCRYPT_VERSION}" >> /out/share/doc/nerdctl-full/README.md
-ARG ROOTLESSKIT_VERSION
```
apostasie, line 184: Move down rootless kit, as we support two versions.

Dockerfile
```diff
@@ -237,6 +232,14 @@ RUN echo "" >> /out/share/doc/nerdctl-full/README.md && \
   mv /tmp/SHA256SUMS /out/share/doc/nerdctl-full/SHA256SUMS && \
   chown -R 0:0 /out
 
+FROM build-dependencies AS build-full
```
apostasie, line 235: Restore nerdctl from above.

.github/workflows/test.yml
```diff
@@ -56,7 +99,7 @@ jobs:
         run: make test-unit
 
   test-integration:
-    timeout-minutes: 60
+    timeout-minutes: 30
```
apostasie, line 102: Lower timeouts.

.github/workflows/test.yml
```diff
@@ -262,11 +344,6 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
           cache: true
           check-latest: true
-      - name: "Print docker info"
```
apostasie, line 265: Debugging remnants.

.github/workflows/test.yml
```diff
@@ -324,13 +401,14 @@ jobs:
...
   test-integration-freebsd:
-    timeout-minutes: 60
+    timeout-minutes: 30
     name: FreeBSD
     # ubuntu-24.04 lacks the vagrant package
     runs-on: ubuntu-22.04
-
     steps:
       - uses: actions/checkout@v4.2.2
+        with:
```
apostasie, line 410: Harmonize

apostasie:

Would like to this is in action for a few PRs to better evaluate how the cache behaves.

.github/workflows/test.yml
```diff
@@ -127,13 +189,18 @@ jobs:
...
           sudo sysctl -w net.ipv6.conf.all.forwarding=1
           sudo sysctl -w net.ipv4.ip_forward=1
-      - name: Enable IPv6 for Docker
+      - name: "Expose GitHub Runtime variables for gha"
+        uses: crazy-max/ghaction-github-runtime@v3
+      - name: Enable IPv6 for Docker, and configure docker to use containerd for gha
         run: |
           sudo mkdir -p /etc/docker
-          echo '{"ipv6": true, "fixed-cidr-v6": "2001:db8:1::/64", "experimental": true, "ip6tables": true}' | sudo tee /etc/docker/daemon.json
+          echo '{"features": {"containerd-snapshotter": true}, "ipv6": true, "fixed-cidr-v6": "2001:db8:1::/64", "experimental": true, "ip6tables": true}' | sudo tee /etc/docker/daemon.json
```
AkihiroSuda, line 197: `containerd-snapshotter` isn't necessary if you use `docker buildx` ? (probably with `docker buildx create` to create a standalone non-moby buildkitd instance)

apostasie: Interesting.
That would certainly be better.
Let me look into that later today.

apostasie: Done. Thanks for the tips.

.github/workflows/test.yml
```diff
@@ -15,7 +15,49 @@ env:
...
+    env:
+      CONTAINERD_VERSION: "${{ matrix.containerd }}"
+      ARCH: "${{ matrix.arch }}"
+    steps:
+      - uses: actions/checkout@v4.2.1
+        with:
+          fetch-depth: 1
+      - name: "Expose GitHub Runtime variables for gha"
+        uses: crazy-max/ghaction-github-runtime@v3
+      - name: "Enable containerd to be able to use gha cache"
```
AkihiroSuda, line 48: Not containerd

apostasie: Right. Thanks.
Removed the separate step altogether with the latest push.

AkihiroSuda (approved):
Thanks

apostasie:
@AkihiroSuda failure is likely #3556
Can you poke it?

.github/workflows/test.yml
```diff
@@ -106,19 +159,22 @@ jobs:
...
-        # ubuntu-20.04: cgroup v1, ubuntu-22.04 and later: cgroup v2
         include:
           - ubuntu: 24.04
             containerd: v2.0.0-rc.5
+            arch: amd64
     env:
-      UBUNTU_VERSION: "${{ matrix.ubuntu }}"
       CONTAINERD_VERSION: "${{ matrix.containerd }}"
+      ARCH: "${{ matrix.arch }}"
+      RUNNER: "ubuntu-${{ matrix.ubuntu }}"
```
ktock, line 176: Is this used in the test scripts?

ktock (approved):
LGTM, sorry for slow review.

.github/workflows/test.yml
apostasie: No, you are right, it is useless.
I was using it previously as part of the cache-key, but have since stopped.

Will remove.
Thanks!

apostasie:
> LGTM, sorry for slow review.

No problem @ktock.
Thanks!

ktock (approved):

apostasie:
While working on something else, I am quickly getting 429s from gha cache, and jobs failing / timeouting because of throttled requests.

I am not optimistic about this overall.
GHA cache does not feel like a usable / scalable product - bad API, lilliputian limitations.

Suggesting we give it a couple of weeks and consider reverting this / tweak it / going back to other ideas (eg: proxy caching).

@AkihiroSuda @ktock

PR #3586, opened 2024-10-22
Update go dependencies

apostasie:
This is updating go dependencies on main.

Main change is containernetworking/plugins (to v1.6) and a few bugfix bumps on indirect deps.

Note that this does (temporarily, hopefully) pin hcsshim to the last working version that does not bring back in containerd v1.7.


go.mod
```diff
@@ -7,10 +7,19 @@ go 1.22.0
...
 
+// FIXME:
+// hcsshim is pinned until the transitive dependency to containerd v1.7 is resolved
+// https://github.com/microsoft/hcsshim/pull/2293/files
+replace github.com/Microsoft/hcsshim => github.com/Microsoft/hcsshim v0.12.7
+
+// FIXME:
+// plugins are pinned until they restore support for golang 1.22
+replace github.com/containernetworking/plugins => github.com/containernetworking/plugins v1.5.1
+
```
AkihiroSuda, line 18: Why do we need these replaces

go.mod
```diff
@@ -116,29 +125,29 @@ require (
...
+	go.opentelemetry.io/otel/trace v1.31.0 // indirect
+	golang.org/x/exp v0.0.0-20241009180824-f66d83c29e7c // indirect
 	golang.org/x/mod v0.21.0 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 // indirect
-	google.golang.org/grpc v1.67.0 // indirect
-	google.golang.org/protobuf v1.34.2 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20241021214115-324edc3d5d38 // indirect
+	google.golang.org/grpc v1.67.1 // indirect
+	google.golang.org/protobuf v1.35.1 // indirect
 	lukechampine.com/blake3 v1.3.0 // indirect
```
AkihiroSuda, line 151: Usually we do not update indirect deps unless there is a specific reason

go.mod
apostasie: To pin the dependencies to that specific version.

AkihiroSuda: Then can we just leave go.mod as-is?

apostasie: Sure, we can.

The suggestion here is about being more friendly and explicit about why these two deps cannot move right now.

Right now, there is probably 2, or maybe 3 people who know why.
For anyone else, trying to upgrade dependencies will fail in confusing ways (now and in the future), making it less approachable to touch go.mod for contributors.

go.mod
apostasie: Fair.

PR #3589, opened 2024-10-22
Remove no-op description from top-level tests

apostasie:
`Description` was mandatory previously with the early version of the new test toolkit.
Top-level tests do relax that (since it is a no-op), and these serve no purpose now.


AkihiroSuda (approved):
Thanks

PR #3591, opened 2024-10-22
Add support for writing to ptys

apostasie:
This is the next step in removing `unbuffer` from our testing, adding the ability to write to stdin with ptys.

### DIFF
--- a/cmd/nerdctl/container/container_attach_linux_test.go
+++ b/cmd/nerdctl/container/container_attach_linux_test.go
@@ -17,9 +17,11 @@
 package container
 
 import (
-	"bytes"
+	"errors"
+	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"gotest.tools/v3/assert"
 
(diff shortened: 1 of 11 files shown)


apostasie:
Failure is unrelated.

pkg/testutil/test/command.go
```diff
@@ -105,8 +107,18 @@ func (gc *GenericCommand) Run(expect *Expected) {
 			iCmdCmd.Stdin = tty
 			iCmdCmd.Stdout = tty
 			iCmdCmd.Stderr = tty
-			defer pty.Close()
-			defer tty.Close()
+
+			for _, writer := range gc.ptyWriters {
```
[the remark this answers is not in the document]
apostasie: Yes indeed.
Will update.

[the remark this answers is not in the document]
apostasie: Reverting to draft for the time being - being abroad I may not have time to look into this immediately.

Thanks @AkihiroSuda .

pkg/testutil/test/command.go
```diff
@@ -105,8 +107,18 @@ func (gc *GenericCommand) Run(expect *Expected) {
...
+				go func() {
+					err := writer(pty)
+					assert.NilError(gc.t, err)
+				}()
+			}
+
+			defer func() {
```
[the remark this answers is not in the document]
fahedouch: this may help when debugging issues ;)

pkg/testutil/test/command.go
```diff
@@ -105,8 +107,18 @@ func (gc *GenericCommand) Run(expect *Expected) {
 			iCmdCmd.Stdin = tty
 			iCmdCmd.Stdout = tty
 			iCmdCmd.Stderr = tty
-			defer pty.Close()
-			defer tty.Close()
```
fahedouch, line 109: please don't move `defer tty.Close()`

apostasie: Why?

pkg/testutil/test/command.go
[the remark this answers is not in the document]
apostasie: Thanks for the review @fahedouch 

this is not ready though - I left it half assed a little while back, but will get it proper soon!

pkg/testutil/test/command.go
fahedouch: not mandatory but you can just defer closing the tty before the goroutines loop

apostasie:
Ok, back on this...

Refresher: the motivation for this overall is to get rid of unbuffer, which has issues:
- forces a third-party binary dependency on the host
- merges together stdout and stderr (it's a pty...)
- closes streams in unexpected ways, forcing us to introduce arbitrary waits
- since it is not go, we cannot debug it as part of the normal testing process

The proposed solution introduces a pure go implementation leveraging ptmx for test tooling.
Right now, it is linux only.

The second commit uses the above to rewrite `detach` tests, removing unbuffer and focusing on avoiding flakyness and racy behaviors.
It also reverts #3558 since the new implementation addresses the underlying issue, and re-enables all tests for Docker as well.

The third separate commit fixes #3881 - as I tripped it while working on this.

Note that icmd is problematic (see #3883) and we might have to get rid of it eventually.
Right now, there is a very ugly hack in place, reading from the master and stuffing that instead of `icmd.Stdout()` in some cases.
This part of the code starts to smell really funny and might need some cleanup, but for now I think we should give this whole thing a spin and see how it goes before considering cleanup.

Also note that this PR removes entirely `DelayOnceReader`, which is harmful: its only motivation was to workaround unbuffer stream closing timing, but in the process it was hiding other sources of flakyness, specifically related to the way we proxy stdin to read detach keys which may race the container reading prior stdin content.

Finally, the rewritten tests should see a decent speed boost.

If this all checks-out with you folks, next PR will rewrite other tests using unbuffer so that we can close the chapter on it.

CI is pending, so there might be a few more minor adjustments (will mark "ready" when it is green), but otherwise, this is ready for review.

@AkihiroSuda @fahedouch

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -28,133 +30,185 @@ import (
...
-	t.Skip("This test is very unstable and currently skipped. See https://github.com/containerd/nerdctl/issues/3558")
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
 
-	skipAttachForDocker(t)
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "--rm", "-it", "--name", data.Identifier(), testutil.CommonImage)
+		cmd.WithPseudoTTY(func(f *os.File) error {
+			_, err := f.Write([]byte{16, 17})
```
AkihiroSuda, line 58: What are these magic numbers?

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -28,133 +30,185 @@ import (
...
-		))),
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		// Run interactively and detach
+		cmd := helpers.Command("attach", "--detach-keys=ctrl-a,ctrl-b", data.Identifier())
+		cmd.WithPseudoTTY(func(f *os.File) error {
+			_, _ = f.WriteString("echo mark${NON}mark\n")
+			// Interestingly, and unlike with run, on attach, docker (like nerdctl) ALSO needs a pause so that the
+			// container can read stdin before we detach
+			time.Sleep(time.Second)
+			_, err := f.Write([]byte{1, 2})
```
AkihiroSuda, line 141: The comment about the magic numbers was lost

cmd/nerdctl/container/container_attach_linux_test.go
apostasie: ascii codepoints for detach keys

apostasie: ctrl+p and ctrl+q

cmd/nerdctl/container/container_attach_linux_test.go
apostasie: ascii code points seem obvious - but then I don't mind either way. Will re-add it.

cmd/nerdctl/container/container_attach_linux_test.go
apostasie: Done.

cmd/nerdctl/container/container_attach_linux_test.go
apostasie: Done.

AkihiroSuda (approved):
Thanks

apostasie:
> sorry for my lack of responsiveness, thank you for your efforts.

No problem at all @fahedouch 
The fun is all mine - I get to hack on cool stuff and share it :-)

PR #3610, opened 2024-10-29
Rewrite container remove test

apostasie:


cmd/nerdctl/container/container_remove_test.go
```diff
@@ -20,22 +20,30 @@ import (
...
+		helpers.Ensure("run", "-d", "--name", data.Identifier(), testutil.CommonImage, "sleep", "inf")
+	}
 
-	// ignore error
-	base.Cmd("rm", tID, "-f").AssertOK()
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
 
-	base.Cmd("run", "-d", "--name", tID, testutil.NginxAlpineImage).AssertOK()
```
AkihiroSuda, line 34: Why was this using nginx 🤷‍♂️

cmd/nerdctl/container/container_remove_test.go
```diff
@@ -20,22 +20,30 @@ import (
...
+		helpers.Fail("rm", data.Identifier())
 
-	// `kill` does return before the container actually stops
-	base.Cmd("stop", tID).AssertOK()
-	base.Cmd("rm", tID).AssertOK()
+		// FIXME: should (re-)evaluate this
+		// `kill` seems to return before the container actually stops
+		helpers.Ensure("stop", data.Identifier())
+
+		return helpers.Command("rm", data.Identifier())
```
AkihiroSuda, line 45: Not specific to this PR, but you do not need to call `Identifier()` multiple times

AkihiroSuda (approved):
Thanks

cmd/nerdctl/container/container_remove_test.go
apostasie: In that case, no.
`containerID := data.Identifier()` will do (inside the `Command` function).

cmd/nerdctl/container/container_remove_test.go
apostasie: Some idiot added it earlier: https://github.com/containerd/nerdctl/commit/2d2979242d250e21f567eac2c4d9ad125ad5c900

😆

More seriously, it just got ported over from a windows test https://github.com/containerd/nerdctl/commit/2d2979242d250e21f567eac2c4d9ad125ad5c900#diff-a15a0a8784b576a40769421789d705b44bcc372734a97dd8aa7d6c581574fb87L34

Which was initially introduced with:
https://github.com/containerd/nerdctl/commit/c7a5bc3ad146dc15d6bd2e00cee397c84537daa1

My guess is this: on windows, we only have busybox or nginx.
For some reason, the busybox we have (from k8) is stuck in the past and does not support `sleep inf`, so, people rely on nginx instead when they want something running in the background?

apostasie: Now that I think more about it... on windows, this test is now not testing the same thing as on linux...

apostasie: Let's ignore the problem for now.

I need to do a full round of `sleep infinity` cleanup anyhow.

PR #3611, opened 2024-10-29
Rewrite container prune test

apostasie:


AkihiroSuda (approved):
Thanks


Not specific to this PR, but you do not need to call `data.Identifier` multiple times with the same argument

PR #3617, opened 2024-10-30
Update busybox image (support sleep inf)

apostasie:
1.28 does not support inf(inity) as an argument for sleep - which is something we are doing a lot in tests.

This came up while debugging docker failures, due to a different test tagging busybox as alpine.

This PR updates to the latest available containerd busybox image.


AkihiroSuda (approved):
Thanks

PR #3618, opened 2024-10-30
Fix test leaking tagged Alpine image

apostasie:
This test is/was leaking busybox tagged as alpine.

Further tests that would call `run alpine` would then not pull, since the image is already here.

But then - see #3617 - if said test was to `run -d sleep inf` - it would fail silently, because the version of busybox we are using has an old `sleep` that does not support `inf`, causing cascading errors into unrelated commands.


AkihiroSuda (approved):
Thanks

Issue #3623, opened 2024-10-31
zstdchunked error deleting temp layers


GrigoryEvko:
### Description

The issues seems to be related to the code imported from stargz-snapshotter, because it persists across both nerdctl and stargz-snapshotter packages. (reported here https://github.com/containerd/stargz-snapshotter/issues/1842)

Using both packages (nerdctl with `nerdctl image convert --zstdchunked --oci src target`) results in similar errors:

```
[...]

PR #3626, opened 2024-10-31
Regression: fix convert unable to find image

apostasie:
Convert using non fully qualified images will fail. Likely a regression with the refutil cleanup.

Spotted by @GrigoryEvko on unrelated #3623


apostasie:
@AkihiroSuda must have for 2.0 (pending green)

AkihiroSuda (approved):
Thanks

PR #3624, opened 2024-10-31
Enable more tests to run on windows

apostasie:
On top of #3631 and #3632.

This PR enables tests on Windows, or annotate them to indicate why they can't run there.

The following have been reviewed:
   - root
   - volume
   - system
   - network
   - namespace
   - issues
   - ipfs
   - internal
   - inspect
   - apparmor
   - completion

The following need to be reviewed more thoroughly:
   - image - needs a windows Registry for push and pull, might come with #3636
   - login - needs a windows Registry, might come with #3636
   - builder
   - compose
   - container

PR #3631, opened 2024-11-01
Add Infinity constant to enable windows tests with old busybox

apostasie:
Despite having recently updated busybox to a more recent version that does support `sleep inf` (#3617), it seems like the windows version of it that we depend on still does not.

(if you do have a windows rig available and can investigate what is going on here that would be lovely)

Making more tests available for windows would thus require that we replace all calls to sleep inf or sleep infinity with sleep ARBITRARY_NUMBER.

This is not so good as it would be confusing if the duration is meaningful or not, and as different tests will likely start using different arbitrary numbers.

So this PR suggests we use a constant instead, `Infinity`, hiding the fact that we (for now) make this `3600`.


apostasie:
Geez.
Git ate my commits 😭.
I want my money back!

AkihiroSuda:
> Geez. Git ate my commits 😭. I want my money back!

FYI `git reflog` might be useful to recover lost commits

AkihiroSuda:
Merged via #3624

PR #3579, opened 2024-10-20
[WIP] Testing

apostasie:
To be broken out in separate PRs.

Using right now for quick validation on the CI.

PR #3636, opened 2024-11-02
[Experimental] windows registry image

apostasie:


apostasie:
@AkihiroSuda @djdongjin @fahedouch et al: would like team opinion on this.

Problems:
1. we currently do not have a registry image for windows, so, we cannot test anything related to login, push or pull on that platform 
2. the busybox image for windows we are currently using is *ancient* (we fetch it from an old kube e2e repo, with no idea how they produced it). Specifically, it predates busybox supporting `sleep Infinity` (discussed earlier somewhere else)

This PR is doing two things:
1. build a Windows Registry image, so that we can unlock image ops testing on windows
2. build a *recent* busybox image for windows that we do understand, control and can update

PR #3579 (WIP) uses both of these test images as a sanity check.

Before I put more work in this, I would like to know:

a. is this a direction we want to pursue?
b. should we have this stuff here, and push the images when the CI runs on main, from inside `ghcr-image-build-and-publish.yml`?

Thanks.

apostasie:
TODO/missing:
- need curl inside the busybox image

AkihiroSuda:
> a. is this a direction we want to pursue?

👍 

> b. should we have this stuff here, and push the images when the CI runs on main, from inside ghcr-image-build-and-publish.yml?

👍 

But this CI job does not need to run for every main commit.
Probably `workflow_dispatch` is enough?

apostasie:
Thanks for the feedback.

> But this CI job does not need to run for every main commit. Probably `workflow_dispatch` is enough?

For sure.

Let me get this in shape and will ping again when it is ready for review.

apostasie:
Windows nerds, questions for you  (tagging the usual suspects @sondavidb  and @jsturtevant )

1. It seems that mcr.microsoft.com/windows/nanoserver:ltsc2022 by default runs as `ContainerUser`, and not `ContainerAdministrator`.

Either this is a change compared to previous versions of nanoserver, or the windows busybox image we had previously was forcing `USER ContainerAdministrator`.

Currently,  I am going for compatibility, meaning I am forcing admin in https://github.com/containerd/nerdctl/pull/3636/files#diff-0a09737c4052666fccb706d792f24be6f876a77dcf12eeb7b9ccaacc249a5bf1R138

Is this the right approach? Or should we instead go with nanoserver default and adapt the tests instead?


2. What should be the entrypoint and default command for our busybox image?
Right now, I am going with:
https://github.com/containerd/nerdctl/pull/3636/files#diff-0a09737c4052666fccb706d792f24be6f876a77dcf12eeb7b9ccaacc249a5bf1R139-R140

Is this the right thing to do?

apostasie:
Closing for now.

PR #3639, opened 2024-11-03
Remove useless TestIssue108 test

apostasie:
This is a test that was created for an issue in dependency `urfave/cli`.

Nerdctl has since moved to cobra cli instead.

Suggesting we remove the test as not making sense anymore.

### DIFF
--- a/cmd/nerdctl/issues/main_linux_test.go
+++ b/cmd/nerdctl/issues/main_linux_test.go
@@ -20,41 +20,8 @@ import (
 	"testing"
 
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
-	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
-	"github.com/containerd/nerdctl/v2/pkg/testutil/test"
 )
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
> This is a test that was created for an issue in dependency `urfave/cli`.
> 
> Nerdctl has since moved to cobra cli instead.
> 
> Suggesting we remove the test as not making sense anymore.

Why no sense? We still have to make sure that cobra doesn’t have the same issue

apostasie:
> > This is a test that was created for an issue in dependency `urfave/cli`.
> > Nerdctl has since moved to cobra cli instead.
> > Suggesting we remove the test as not making sense anymore.
> 
> Why no sense? We still have to make sure that cobra doesn’t have the same issue

IMHO, it seems extremely unlikely that two unrelated libraries would develop the same bug, especially now with the level of adoption that cobra has seen.

That being said, I am not opinionated on this.
We can keep this if you think it is worth it.
Just lmk.

apostasie:
Closing this - while this test does not really serve a purpose IMO, it is not harmful either.

Issue #3625, opened 2024-10-31
Configuring custom IP address for nerdctl default bridge


swagatbora90:
### What is the problem you're trying to solve

I would like to set a custom subnet range for the nerdctl bridge IP which is currently set to 10.4.0.0/24. In docker, this can be done by setting the `bip` option in daemon config, however I could not find a similar option in nerdctl. A quick and hacky way to get around this is to manually update the bridge cni conf file to set the subnet and gateway, however I would like to discuss if this can be added as a option in nerdctl toml. 

Also, unlike docker, nerdctl does not allow removing the default bridge network, so we will need some additional handling to detect the change of bridge ip (in case the option gets added after the bridge is already created), and trigger a update workflow. I have a reference implementation [here](https://github.com/swagatbora90/nerdctl/commit/3749bc19c6ef56c603414ad1ed99f798543bff6b) here. 

Is this something the community would consider to add as a new feature? If so I can go ahead and open a PR. 

### Describe the solution you'd like

Introduce a new option in nerdctl.toml
```
| `bridge_ip`         | `--bridge-ip`                       | `NERDCTL_BRIDGE_IP`      | IP address for the default nerdctl bridge network 
```

That configures the default 'bridge' network with the specified ip address. If the bridge network is already created, adding this new option will update the existing 'bridge' IP provided there are no running containers attached to that network.

apostasie:
Hey @swagatbora90 

Curious about your use-case here.

Couldn't you create your own bridge network instead?
```
nerdctl network create mybridge --subnet 10.0.10.0/24
nerdctl network inspect mybridge
```

swagatbora90:
Hi @apostasie,

My usecase is using finch with nerdctl. finch relies on the default bridge network that nerdctl creates. So containers launched without any network `--network` get attached to this default network.  

Creating a new bridge network does not work, since the default network has special treatment in nerdctl which we would like to preserve, such as protection again network prune commands.

apostasie:
Thanks @swagatbora90 
Makes sense.

PR #3640, opened 2024-11-04
feat: adds IP address configuration for default bridge network

swagatbora90:
Adds an option in nerdctl.toml to configure the the default bridge network IP

Fixes: https://github.com/containerd/nerdctl/issues/3625

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -45,6 +45,7 @@ experimental   = true
 | `hosts_dir`         | `--hosts-dir`                      |                           | `certs.d` directory                                                                                                                                              | Since 0.16.0     |
 | `experimental`      | `--experimental`                   | `NERDCTL_EXPERIMENTAL`    | Enable  [experimental features](experimental.md)                                                                                                                 | Since 0.22.3     |
 | `host_gateway_ip`   | `--host-gateway-ip`                | `NERDCTL_HOST_GATEWAY_IP` | IP address that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP address of the host. It has no effect without setting --add-host | Since 1.3.0      |
+| `bridge_ip`         | `--bridge-ip`                      | `NERDCTL_BRIDGE_IP`       | IP address for the default nerdctl bridge network, e.g., 10.1.100.1/24                                                                                           | Since 2.1.0      |
 
 The properties are parsed in the following precedence:
 1. CLI flag
--- a/cmd/nerdctl/helpers/flagutil.go
+++ b/cmd/nerdctl/helpers/flagutil.go
@@ -99,6 +99,10 @@ func ProcessRootCmdFlags(cmd *cobra.Command) (types.GlobalCommandOptions, error)
 	if err != nil {
 		return types.GlobalCommandOptions{}, err
 	}
+	bridgeIP, err := cmd.Flags().GetString("bridge-ip")
+	if err != nil {
+		return types.GlobalCommandOptions{}, err
+	}
 	return types.GlobalCommandOptions{
 		Debug:            debug,
 		DebugFull:        debugFull,
@@ -113,6 +117,7 @@ func ProcessRootCmdFlags(cmd *cobra.Command) (types.GlobalCommandOptions, error)
 		HostsDir:         hostsDir,
 		Experimental:     experimental,
 		HostGatewayIP:    hostGatewayIP,
+		BridgeIP:         bridgeIP,
 	}, nil
 }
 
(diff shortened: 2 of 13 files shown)


docs/config.md
```diff
@@ -45,6 +45,7 @@ experimental   = true
 | `hosts_dir`         | `--hosts-dir`                      |                           | `certs.d` directory                                                                                                                                              | Since 0.16.0     |
 | `experimental`      | `--experimental`                   | `NERDCTL_EXPERIMENTAL`    | Enable  [experimental features](experimental.md)                                                                                                                 | Since 0.22.3     |
 | `host_gateway_ip`   | `--host-gateway-ip`                | `NERDCTL_HOST_GATEWAY_IP` | IP address that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP address of the host. It has no effect without setting --add-host | Since 1.3.0      |
+| `bridge_ip`         | `--bridge-ip`                       | `NERDCTL_BRIDGE_IP`      | IP address for the default nerdctl bridge network              | Since 1.7.8 |
```
AkihiroSuda, line 48:
```suggestion
| `bridge_subnet`         | `--bridge-subnet`                       | `NERDCTL_BRIDGE_SUBNET`      | IP subnet for the default nerdctl bridge network, e.g., 10.1.100.0/24             | Since 2.1.0 |
```

pkg/netutil/netutil_unix.go
```diff
@@ -43,6 +43,7 @@ import (
 const (
 	DefaultNetworkName = "bridge"
 	DefaultCIDR        = "10.4.0.0/24"
+	DefaultGatewayIP   = "10.4.0.1"
```
AkihiroSuda, line 46: This is redundant, as it can be computed from DefaultCIDR

docs/config.md
swagatbora90: The current implementation actually follows docker convention of `bip` which requires a specific IP and a CIDR. Any reason to prefer  subnet instead of IP? I can make the change to use subnet instead and let nerdctl choose the first IP of the subnet as gateway.

AkihiroSuda: I thought specifying CIDR might be more consistent with CNI, but it seems also fine to follow `dockerd --bip=<IP>` convention.

No strong opinion from me.

apostasie:
Docker CI failure: #3652

swagatbora90:
> Just for note, I'm not sure if `--bridge-ip` is equal with `bip` in Docker daemon or not. If two flag is equal, I think we need change the network create behavior at the same time.

Can you please elaborate? I don't think network create has a `--bridge-ip ` option. You provide `--subnet`, `--ip-range` and `--gateway`. `bip` is defined in the daemon level.

swagatbora90:
> This option in docker will have an implicit behavior, it will change the network start range. For example, for the the same command docker network create abc. The abc IP range will be 172.18.0.0/24(for example) and will be 10.11.1.0/24 if we set the bip to 10.11.0.0/24

I could not reproduce this behavior in my local testing:

```
% sudo cat /etc/docker/daemon.json
{
  "bip": "240.10.0.1/24",
  "data-root": "/local/docker",
  "dns": ["10.4.4.10"]
}

dev-dsk-sbora-2c-4ee7e306 % docker network inspect bridge | jq '.[0].IPAM.Config'
[
  {
    "Subnet": "240.10.0.0/24",
    "Gateway": "240.10.0.1"
  }
]

% docker network create -d bridge test-net
10e34d98783d9562f27f4248f970745d53357e9fd2d0c0488422258c2243a586

% docker network inspect test-net | jq '.[0].IPAM.Config'
[
  {
    "Subnet": "172.18.0.0/16",
    "Gateway": "172.18.0.1"
  }
]
```

For new bridge network (without a subnet range specified), docker still picks the `172.18/16` range. I found these ranges are defined here https://github.com/moby/moby/blob/master/libnetwork/ipamutils/utils.go#L12-L25, so this is still from a predefined range of non-conflicting network ranges and not implicitly set from `bip`

Zheaoli:
> ```
> }
> ]
> ```
> 
> 
>     
>   
> 
> For new bridge network (without a subnet range specified), docker still picks the `172.18/16` range. I found these ranges are defined here https://github.com/moby/moby/blob/master/libnetwork/ipamutils/utils.go#L12-L25, so this is still from a predefined range of non-conflicting network ranges and not implicitly set from `bip`.
> 
> 
> Docker actually provides a daemon level option `default-address-pools` to specify custom network ranges for user defined networks. But that can be a follow up feature if needed, since have the option to pass `--subnet` with network create to do exactly that.

You are right. I mix bip and `default-address-pools` together

Zheaoli (approved):
LGTM on CI green.

Maybe you just need update the code in main branch and rerun the CI

Zheaoli (approved):
LGTM, All CI green

AkihiroSuda (approved):

AkihiroSuda (approved):
Thanks

PR #3665, opened 2024-11-12
refactor: Use a more direct and less error-prone return value

pinglanlu:
Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone.

Just like  https://github.com/containerd/nerdctl/blob/8b814ca7fe29cb505a02a3d85ba22860e63d15bf/pkg/netutil/subnet/subnet.go#L76   and  https://github.com/containerd/nerdctl/blob/8b814ca7fe29cb505a02a3d85ba22860e63d15bf/pkg/netutil/subnet/subnet.go#L92


Zheaoli (approved):
LGTM, but need update with main branch after #3666 has been merged to rerun the CI

Issue #3661, opened 2024-11-08
Persist Logs on container restart


coderbirju:
### What is the problem you're trying to solve

In nerdctl, logs are not persisted when containers are restarted or recreated. This behavior is observed in the following scenarios:

- Using` nerdctl run` followed by `nerdctl start -a`
- Using `nerdctl restart`
- Using `nerdctl create` followed by `nerdctl start` and `nerdctl restart`
In all these cases, only the most recent logs are available after a container restart or recreation.

### Describe the solution you'd like

 nerdctl should log persist logs similar to Docker and Podman. 

Specifically:
Logs should be preserved across container restarts.

When using `nerdctl start -a` on a previously run container, both old and new logs should be accessible.
The `nerdctl restart` command should not replace previous logs but append new ones.
Containers created with `nerdctl create` and then started should maintain their logs across multiple starts.

coderbirju:
So it seems that [go-logrotate](https://github.com/fahedouch/go-logrotate/tree/6a8beddaea39b2b9c77109d7fa2fe92053c063e5) has a bug when `MaxBytes` is set to -1 and rotates the logs for every re-invocation. creating another log file with a `.1` suffix.

tushar5526:
Hi 👋 I would be happy to help if this issue is still up for grab!

tushar5526:
> Using nerdctl restart

Logs persist in this case. I tried the following with an nginx container. 

```
nerdctl create nginx
nerdctl start  <nginx-container>
nerdctl logs -f <nginx-container>

# in a different terminal, exec a bash shell in the nginx container
curl http://localhost:80/random-endpoint

# stop and start the container 
```
I can still see the logs persisting between restarts in this case.

PR #3670, opened 2024-11-13
fix: default json-file log size to 100MB

coderbirju:
#3661 
This commit makes the json-file logger to default logrotate maxBytes configuration of 100MB.

It is a temporary fix to a bug addressed by this [logrotate PR](https://github.com/fahedouch/go-logrotate/pull/50).

Setting the default values in the config passed to logrotate fixes the issue temporarily but having unlimited size for container logs will be available when the issue is addressed upstream. 

Testing done :

Verified that the logging mechanism works and appends the logs to the same log file and doesn't create a new log file.

### DIFF
--- a/pkg/logging/json_logger.go
+++ b/pkg/logging/json_logger.go
@@ -88,10 +88,11 @@ func (jsonLogger *JSONLogger) PreProcess(dataStore string, config *logging.Confi
 	l := &logrotate.Logger{
 		Filename: jsonFilePath,
 	}
-	//maxSize Defaults to unlimited.
-	var capVal int64
-	capVal = -1
+	// MaxBytes is the maximum size in bytes of the log file before it gets
+	// rotated. If not set, it defaults to 100 MiB.
+	// see: https://github.com/fahedouch/go-logrotate/blob/6a8beddaea39b2b9c77109d7fa2fe92053c063e5/logrotate.go#L500
 	if capacity, ok := jsonLogger.Opts[MaxSize]; ok {
+		var capVal int64
 		var err error
 		capVal, err = units.FromHumanSize(capacity)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/logging/json_logger.go
```diff
@@ -88,10 +88,9 @@ func (jsonLogger *JSONLogger) PreProcess(dataStore string, config *logging.Confi
 	l := &logrotate.Logger{
 		Filename: jsonFilePath,
 	}
-	//maxSize Defaults to unlimited.
-	var capVal int64
-	capVal = -1
+	//maxSize Defaults to 100MB.
```
AkihiroSuda, line 91: Where is this set to 100MiB?

How was this value chosen?

coderbirju: 100MB is the default value that go-logrotate specifies for each log file. 
[defaultMaxsize  = 100](https://github.com/fahedouch/go-logrotate/blob/6a8beddaea39b2b9c77109d7fa2fe92053c063e5/logrotate.go#L500)

AkihiroSuda: Could you add that context to the code comment?

djdongjin (approved):
LGTM, thanks

pkg/logging/json_logger.go
```diff
@@ -88,10 +88,11 @@ func (jsonLogger *JSONLogger) PreProcess(dataStore string, config *logging.Confi
 	l := &logrotate.Logger{
 		Filename: jsonFilePath,
 	}
-	//maxSize Defaults to unlimited.
-	var capVal int64
-	capVal = -1
+	// MaxBytes is the maximum size in bytes of the log file before it gets
+	// rotated. If not set, it defaults to 100 MiB.
+	// see: https://github.com/fahedouch/go-logrotate/blob/main/logrotate.go#L500
```
AkihiroSuda, line 93: "main" has to be replaced with an exact version or a commit hash for permanence

coderbirju:
Could this fix be backported to nerdctl 1.7.x, considering containerd 2.0 is not yet LTS. @AkihiroSuda @djdongjin

AkihiroSuda (approved):
Thanks, feel free to open cherry-pick PR for v1.7

AkihiroSuda:
> LTS

For nerdctl we do not have any LTS (yet)

PR #3673, opened 2024-11-13
fix nerdctl ps slow on  heavy IO system by using goroutine

ningmingxiao:
formatter.ContainerStatus will use some time. If plenty of calls are coming, this will use much time.
before    
```
time nerdctl_old -n k8s.io  ps -a
real    2m27.206s
user    0m0.336s
sys     0m0.218s
```
after this commit   
```
time nerdctl_new -n k8s.io  ps -a
real    0m3.333s
user    0m0.145s
sys     0m0.074s
```
@AkihiroSuda

### DIFF
--- a/pkg/cmd/container/list.go
+++ b/pkg/cmd/container/list.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	containerd "github.com/containerd/containerd/v2/client"
@@ -40,11 +41,11 @@ import (
 
 // List prints containers according to `options`.
 func List(ctx context.Context, client *containerd.Client, options types.ContainerListOptions) ([]ListItem, error) {
-	containers, err := filterContainers(ctx, client, options.Filters, options.LastN, options.All)
+	containers, cMap, err := filterContainers(ctx, client, options.Filters, options.LastN, options.All)
 	if err != nil {
 		return nil, err
 	}
-	return prepareContainers(ctx, client, containers, options)
+	return prepareContainers(ctx, client, containers, cMap, options)
 }
 
 // filterContainers returns containers matching the filters.
@@ -53,14 +54,14 @@ func List(ctx context.Context, client *containerd.Client, options types.Containe
 //   - all means showing all containers (default shows just running).
 //   - lastN means only showing n last created containers (includes all states). Non-positive values are ignored.
 //     In other words, if lastN is positive, all will be set to true.
-func filterContainers(ctx context.Context, client *containerd.Client, filters []string, lastN int, all bool) ([]containerd.Container, error) {
+func filterContainers(ctx context.Context, client *containerd.Client, filters []string, lastN int, all bool) ([]containerd.Container, map[string]string, error) {
 	containers, err := client.Containers(ctx)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	filterCtx, err := foldContainerFilters(ctx, containers, filters)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	containers = filterCtx.MatchesFilters(ctx)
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/container/list.go
```diff
@@ -112,58 +104,101 @@ func (x *ListItem) Label(s string) string {
...
-			containerSize, err := getContainerSize(ctx, snapshotter, info.SnapshotKey)
-			if err != nil {
-				return nil, err
-			}
-			li.Size = containerSize
+	}
+	return result, nil
+}
+
+func combineErrors(errs []error) error {
```
djdongjin, line 191: you can use `errors.Join(errs...)`.

pkg/cmd/container/list.go
```diff
@@ -112,58 +104,101 @@ func (x *ListItem) Label(s string) string {
...
+				id = id[:12]
 			}
-			return nil, err
+			cStatus := formatter.ContainerStatus(ctx, c)
+			// only show Up status container
+			if !all {
+				if !strings.HasPrefix(cStatus, "Up") {
+					return
+				}
+			}
```
djdongjin, line 142: =>

```golang
if !all &&!strings.HasPrefix(cStatus, "Up") {
  return
}
```

djdongjin: also, what's the improvement if we mvoe this logic into `filterContainers`?

Now the filter is done in 2 separate functions, and `filterContainers` has to return a `all` to indiciate if it's a `all` filter or not, which is a little confusing.

ningmingxiao: old version the bottleneck is  formatter.ContainerStatus, and call it twice.So delete one is necessary.

ningmingxiao: or use goroutine just in filterContainers func.

pkg/cmd/container/list.go
```diff
@@ -53,14 +54,14 @@ func List(ctx context.Context, client *containerd.Client, options types.Containe
 //   - all means showing all containers (default shows just running).
 //   - lastN means only showing n last created containers (includes all states). Non-positive values are ignored.
 //     In other words, if lastN is positive, all will be set to true.
-func filterContainers(ctx context.Context, client *containerd.Client, filters []string, lastN int, all bool) ([]containerd.Container, error) {
+func filterContainers(ctx context.Context, client *containerd.Client, filters []string, lastN int, all bool) ([]containerd.Container, map[string]string, error) {
```
djdongjin, line 57: can you add a comment explaining that, `formatter.ContainerStatus(ctx, c)` is time consuming so we do it in goroutines and return the result as a map.

Also can you update the PR description with the latest benchmark result from this implementation.

ningmingxiao: done, thank you.

djdongjin (approved):
LGTM, thanks

pkg/cmd/container/list.go
```diff
@@ -112,7 +130,7 @@ func (x *ListItem) Label(s string) string {
 	return x.LabelsMap[s]
 }
 
-func prepareContainers(ctx context.Context, client *containerd.Client, containers []containerd.Container, options types.ContainerListOptions) ([]ListItem, error) {
+func prepareContainers(ctx context.Context, client *containerd.Client, containers []containerd.Container, cMap map[string]string, options types.ContainerListOptions) ([]ListItem, error) {
```
AkihiroSuda, line 133: `cMap` needs a comment to explain its keys and values.

ningmingxiao: done

AkihiroSuda: Doesn't seem done

pkg/cmd/container/list.go
```diff
@@ -77,18 +78,36 @@ func filterContainers(ctx context.Context, client *containerd.Client, filters []
 		}
 	}
 
+	var wg sync.WaitGroup
+	cMap := make(map[string]string)
+	var mu sync.Mutex
+	// formatter.ContainerStatus(ctx, c) is time consuming so we do it in goroutines and return the container's id with status as a map.
+	// prepareContainers func will use this map to avoid call formatter.ContainerStatus again.
+	for _, c := range containers {
+		wg.Add(1)
```
fahedouch, line 87:
```suggestion
                c := c
		wg.Add(1)
```

the above line code simplifies the code and avoid using a mutex, right ?

pkg/cmd/container/list.go
```diff
@@ -77,18 +78,36 @@ func filterContainers(ctx context.Context, client *containerd.Client, filters []
...
+		go func(ctx context.Context, c containerd.Container) {
+			defer wg.Done()
+			cStatus := formatter.ContainerStatus(ctx, c)
+			mu.Lock()
+			if c.ID() != "" {
```
fahedouch, line 92: is it possible to have an empty id ?

if yes we my move this check even before creating the goroutine

pkg/cmd/container/list.go
```diff
@@ -77,18 +78,36 @@ func filterContainers(ctx context.Context, client *containerd.Client, filters []
 		}
 	}
 
+	var wg sync.WaitGroup
+	cMap := make(map[string]string)
```
fahedouch, line 82:
```suggestion
	statusPerContainer := make(map[string]string)
```

pkg/cmd/container/list.go
```diff
@@ -136,6 +155,13 @@ func prepareContainers(ctx context.Context, client *containerd.Client, container
 		if options.Truncate && len(id) > 12 {
 			id = id[:12]
 		}
+		var status string
+		s, ok := cMap[c.ID()]
```
fahedouch, line 159:
```suggestion
		if s, ok := cMap[c.ID()]; ok {
		     status = s
		}
```

pkg/cmd/container/list.go
fahedouch: @ningmingxiao WDYT ?

djdongjin: > avoid using a mutex

I don't think it will avoid the mutex. Current write to a map still need the mutex.

fahedouch: indeed map are note safe for [concurrent](https://go.dev/doc/faq#atomic_maps) write. thanks @djdongjin  for your pointer ;)

fahedouch (approved):
Thanks

PR #3683, opened 2024-11-19
Pass appropriate --device arguments from compose to run

ryfow:
This was passing a stringified version of the DeviceMapping struct instead of a source:target:permissions string.


AkihiroSuda (approved):
Thanks

Issue #3637, opened 2024-11-02
support default Dynamic IPv6 subnet allocation and ip6tables


ghost:
### What is the problem you're trying to solve

nerdctl doesn't listen on ipv6 by default

### Describe the solution you'd like

https://docs.docker.com/engine/daemon/ipv6/

### Additional context

https://github.com/containerd/nerdctl/pull/127

Zheaoli:
For now, the nerdctl has already support IPV6, I'm not sure what behavior is you want. Would you mind provide more detail about this?

PR #3686, opened 2024-11-21
Enable IPv6 on compose networks where enable_ipv6 is true

darktohka:
There is a compose feature called enable_ipv6 that can be set on a network.

When this is currently set on a compose network, it has no effect. Compose does let us know that the EnableIPv6 flag is ignored, though.

However, `nerdctl network create` does support IPv6, it can be set using the `--ipv6` flag. When the `--ipv6` flag is set, IPv6 IPAM is added just like IPv4 IPAM to the network. However, when `--ipv6` is not set, the IPv6 IPAM is ignored and only the IPv4 settings are added.

We should add support for `enable_ipv6`. This PR aims just to do that.

Please note that this is not enough for compose to support IPv6... see issue https://github.com/containerd/nerdctl/issues/3637. I'm not yet familiar enough with the codebase to add such a feature. But this should allow manual configuration.

Added two tests:
- one test to test enable_ipv6 true, the IPv6 subnet should stay in the IPAM configuration after creation
- one test to test enable_ipv6 false, the IPv6 subnet is filtered out of the IPAM configuration after creation and only IPv4 remains


darktohka:
Seems like the second test: "test to test enable_ipv6 false, the IPv6 subnet is filtered out of the IPAM configuration after creation and only IPv4 remains" fails on the Docker integration test.

Should we remove this test and only test if enable_ipv6: true works fine?

fahedouch:
> Seems like the second test: "test to test enable_ipv6 false, the IPv6 subnet is filtered out of the IPAM configuration after creation and only IPv4 remains" fails on the Docker integration test.
> 
> Should we remove this test and only test if enable_ipv6: true works fine?

I think we need to figure out why nerdctl runtime do not consider `enable_ipv6` false, it may point to an underlying problem.

PR #2839, opened 2024-02-23
fix: add network cleanup for kill and stop cmd

alegrey91:
Fix #2699

This PR provides code to cleanup CNI network setup when `kill` and `stop` command are typed.
Additionally, the `onStartContainer` hook was added in order to restore the CNI network setup when `start` command is typed (as suggested in the issue by @yankay).

## How to test

In order to test this PR, follow these commands:

```
# compile the binary with the latest changes
make

# create the first container with the built nerdctl version
sudo _output/nerdctl run --name ngx1 -d -p 9999:80 nginx:alpine

# check iptables rules have been setup
sudo iptables-save | grep 9999

# kill the first container previosly created
sudo _output/nerdctl kill ngx1

# (now the iptables rules should disappeared
# create the second container bound on the same port
sudo _output/nerdctl run --name ngx2 -d -p 9999:80 nginx:alpine

# check iptables rules have been setup
sudo iptables-save | grep 9999

# access the second container forward
curl 127.0.0.1:9999
```

## Credits

Thanks to @yankay and @89luca89 for the help

PR #3687, opened 2024-11-21
feat: add force-recreate and no-recreate for compose up command

pendo324:
Enable two missing docker compose up options: --force-recreate and --no-recreate

Also made it so that only logs from the latest run are displayed in the user's terminal (otherwise, all logs will be displayed, which could be confusing to users). This is the same behavior as the docker CLI today.

One issue which I noticed is that my testing container would keep getting a new IP address in its IPAM bridge network. I think as of #2839, the iptables rules should be cleaned up when a container is stopped, but its not happening for me for some reason.

### DIFF
--- a/cmd/nerdctl/compose/compose_up.go
+++ b/cmd/nerdctl/compose/compose_up.go
@@ -47,6 +47,8 @@ func newComposeUpCommand() *cobra.Command {
 	composeUpCommand.Flags().Bool("ipfs", false, "Allow pulling base images from IPFS during build")
 	composeUpCommand.Flags().Bool("quiet-pull", false, "Pull without printing progress information")
 	composeUpCommand.Flags().Bool("remove-orphans", false, "Remove containers for services not defined in the Compose file.")
+	composeUpCommand.Flags().Bool("force-recreate", false, "Recreate containers even if their configuration and image haven't changed.")
+	composeUpCommand.Flags().Bool("no-recreate", false, "Don't recreate containers if they exist, conflict with --force-recreate.")
 	composeUpCommand.Flags().StringArray("scale", []string{}, "Scale SERVICE to NUM instances. Overrides the `scale` setting in the Compose file if present.")
 	return composeUpCommand
 }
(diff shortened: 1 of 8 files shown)


cmd/nerdctl/compose/compose_up.go
```diff
@@ -102,6 +104,17 @@ func composeUpAction(cmd *cobra.Command, services []string) error {
 	if err != nil {
 		return err
 	}
+	forceRecreate, err := cmd.Flags().GetBool("force-recreate")
+	if err != nil {
+		return err
+	}
+	noRecreate, err := cmd.Flags().GetBool("no-recreate")
+	if err != nil {
+		return nil
```
austinvazquez, line 113: return error here?

cmd/nerdctl/compose/compose_up.go
```diff
@@ -102,6 +104,17 @@ func composeUpAction(cmd *cobra.Command, services []string) error {
 	if err != nil {
 		return err
 	}
+	forceRecreate, err := cmd.Flags().GetBool("force-recreate")
```
austinvazquez, line 107: Need to set these values on up options below.

cmd/nerdctl/compose/compose_up.go
austinvazquez, line 1: I think an integration test might be good to add. It will be a little effort but if you can have a test that does compose up on a container that creates a sentinel file on a volume mount and dies if the sentinel was not found. For --force-recreate the container should comeback and for --no-recreate the container should stay stopped.

cmd/nerdctl/compose/compose_up.go
pendo324: I copied this from `compose_create.go`, I think they should probably both return the errs, good catch

pkg/composer/container.go
```diff
@@ -63,3 +63,24 @@ func (c *Composer) containerExists(ctx context.Context, name, service string) (b
...
+		if err != nil {
+			return "", err
+		}
+		if name == containerLabels[labels.Name] {
+			// container exists
+			return container.ID(), nil
+		}
+	}
+	// container doesn't exist
+	return "", nil
```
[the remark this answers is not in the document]
pendo324: I think empty string + nil error is a good enough indication of a non-existent container and setting err should be saved for anything unexpected. WDTY?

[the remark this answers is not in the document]
austinvazquez: Yep sounds good. Thanks for taking another look.

AkihiroSuda (commented):
Could you update docs/command-reference.md too?

AkihiroSuda:
ping @pendo324

AkihiroSuda (approved):
Thanks

PR #3693, opened 2024-11-26
fix redundant error checks and comment format

ErikJiang:
1. removed redundant if ...; err != nil checks in:
* `pkg/lockutil/lockutil_unix.go`
* `pkg/containerutil/container_network_manager.go`

2. corrected the comment format for the exported variable in `pkg/mountutil/volumestore/volumestore.go`


Zheaoli (approved):
LGTM

Thanks

Issue #3704, opened 2024-11-30
/etc/resolv.conf is not mounted with the correct permissions when the host has a umask 0077


apostasie:
### Description

This is quite confounding.

```
sudo nerdctl run --rm -ti debian sh -c -- "apt-get update"
```

Will fail
```
Err:1 http://deb.debian.org/debian bookworm InRelease
  Temporary failure resolving 'deb.debian.org'
Err:2 http://deb.debian.org/debian bookworm-updates InRelease
  Temporary failure resolving 'deb.debian.org'
Err:3 http://deb.debian.org/debian-security bookworm-security InRelease
  Temporary failure resolving 'deb.debian.org'
```

(same with ubuntu)

BUT

```
sudo nerdctl run --rm -ti alpine sh -c -- "apk update; apk add curl"
```

Works just fine.

Furthermore:
- using `net host` works just fine - this definitely has to do with bridge
- using a different nameserver (`sudo nerdctl run --dns 1.1.1.1 --rm -ti debian bash`) does NOT fix the problem
- using docker with the same images, on the same host, _works just fine_

Since this is working with alpine, my intuition is to blame glibc.

@AkihiroSuda does this problem sound familiar in any way?

Any pointer on how to debug this?

### Describe the results you received and expected

na

### What version of nerdctl are you using?

Host is:
```
apo@amaterasu:~/post $ uname -a
Linux amaterasu 6.6.51+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.51-1+rpt3 (2024-10-08) aarch64 GNU/Linux
apo@amaterasu:~/post $ cat /etc/issue
Debian GNU/Linux 12 \n \l
```

```
apo@amaterasu:~/post $ sudo nerdctl version
WARN[0000] unable to determine buildctl version: exec: "buildctl": executable file not found in $PATH
Client:
 Version:	v2.0.0
 OS/Arch:	linux/arm64
 Git commit:	ef588dafa080e3dbc9c061ff3802affb66aef291
 buildctl:
  Version:

Server:
 containerd:
  Version:	1.7.23
  GitCommit:	57f17b0a6295a39009d861b89e3b3b87b005ca27
 runc:
  Version:	1.1.14
  GitCommit:	v1.1.14-0-g2c9f560
```

```
apo@amaterasu:~/post $ sudo nerdctl info
Client:
 Namespace:	default
 Debug Mode:	false

Server:
 Server Version: 1.7.23
[...]

PR #3708, opened 2024-12-02
Fix permissions for resolv.conf and hosts

apostasie:
WriteFile permissions are before umask is applied. For people using aggressive umasks (0077), /etc/resolv.conf will end-up unreadable for non root processes.

Fix https://github.com/containerd/nerdctl/issues/3704

### DIFF
--- a/pkg/dnsutil/hostsstore/hostsstore.go
+++ b/pkg/dnsutil/hostsstore/hostsstore.go
@@ -115,6 +115,14 @@ func (x *hostsStore) Acquire(meta Meta) (err error) {
 			return errors.Join(store.ErrSystemFailure, err)
 		}
 
+		// os.WriteFile relies on syscall.Open. Unless there are ACLs, the effective mode of the file will be matched
+		// against the current process umask.
+		// See https://www.man7.org/linux/man-pages/man2/open.2.html for details.
+		// Since we must make sure that these files are world readable, explicitly chmod them here.
+		if err = os.Chmod(loc, 0o644); err != nil {
+			err = errors.Join(store.ErrSystemFailure, err)
+		}
+
 		var content []byte
 		content, err = json.Marshal(meta)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


apostasie:
Note: we should yank out every and all filesystem direct access away.

The Store interface did a lot of that already, and did solve atomicity and concurrency, but this is yet another proof that letting people use the filesystem directly is almost always a mistake and we should provide safer abstractions instead.

pkg/dnsutil/hostsstore/hostsstore.go
```diff
@@ -114,6 +114,9 @@ func (x *hostsStore) Acquire(meta Meta) (err error) {
 		if err = os.WriteFile(loc, []byte{}, 0o644); err != nil {
 			return errors.Join(store.ErrSystemFailure, err)
 		}
+		if err = os.Chmod(loc, 0o644); err != nil {
```
AkihiroSuda, line 117: Could you add a comment to explain why `0o644` in the arg of `WriteFile` doesn't take an expected effect?

pkg/dnsutil/hostsstore/hostsstore.go
```diff
@@ -175,6 +178,9 @@ func (x *hostsStore) AllocHostsFile(id string, content []byte) (location string,
 		if err != nil {
 			err = errors.Join(store.ErrSystemFailure, err)
 		}
+		if err = os.Chmod(loc, 0o644); err != nil {
```
AkihiroSuda, line 181: same as above

pkg/resolvconf/resolvconf.go
```diff
@@ -317,7 +317,12 @@ func Build(path string, dns, dnsSearch, dnsOptions []string) (*File, error) {
 		return nil, err
 	}
 
-	return &File{Content: content.Bytes(), Hash: hash}, os.WriteFile(path, content.Bytes(), 0644)
+	err = os.WriteFile(path, content.Bytes(), 0o644)
+	if err != nil {
+		return nil, err
+	}
+
+	return &File{Content: content.Bytes(), Hash: hash}, os.Chmod(path, 0o644)
```
AkihiroSuda, line 325: same as above

pkg/dnsutil/hostsstore/hostsstore.go
apostasie: Done

pkg/dnsutil/hostsstore/hostsstore.go
```diff
@@ -114,6 +114,10 @@ func (x *hostsStore) Acquire(meta Meta) (err error) {
 		if err = os.WriteFile(loc, []byte{}, 0o644); err != nil {
 			return errors.Join(store.ErrSystemFailure, err)
 		}
+		// See https://www.man7.org/linux/man-pages/man2/open.2.html
```
AkihiroSuda, line 117: This doesn't explain how the umask rule differs across Go's `os.WriteFile` and `os.Chmod`

apostasie: No chmod implementation (system or language) would ever honor change the provided mode to account for umask - that would not make sense.

As for WriteFile, it calls `syscall.Open` underneath - and Open behavior wrt umask is covered in the man page.

Anyhow, changed the comment.

AkihiroSuda (approved):
Thanks

PR #3723, opened 2024-12-03
Cleanup namespace validation

apostasie:
Currently, namespace name is only validated in certain circumstances (eg: if we are going to access a container statedir). This may lead to inconsistencies, where we would allow forbidden namespaces in certain scenarios but properly restrict them in other scenarios.

Also, the current set of rules for validation are inadequate, and fail to cover a range of conditions that *will* fail on the filesystem.

So, suggesting:
- we always validate namespace, once, during argument parsing 
- we use the known good validator we have in place already in the store (ValidatePathComponent) instead of the current validator

This is introducing changes. Specifically, it will restrict previously allowed namespaces:
- namespaces that are longer than 255 characters
- namespaces containing protected OS/filesystem patterns (specifically: on windows, `con[0-9]`, etc)

And on the other hand, relax restrictions on certain patterns that were previously forbidden for no apparent reason (tilde, etc).

Most importantly though, note that containerd will *anyhow* identifier-validate the namespace name whenever it touches it... with the much more restrictive`^[A-Za-z0-9]+(?:[._-](?:[A-Za-z0-9]+))*$` - so, this is merely about cleaning things up on our side.

Let me know what you think.


pkg/nsutil/nsutil.go
```diff
@@ -1,47 +0,0 @@
...
-		return fmt.Errorf("namespace name cannot be empty")
-	}
-
-	// Slash and '$' for POSIX and backslash and '%' for Windows.
-	pathSeparators := "/\\%$"
-	if strings.ContainsAny(nsName, pathSeparators) {
-		return fmt.Errorf("namespace name cannot contain any special characters (%q): %s", pathSeparators, nsName)
-	}
-
-	specialAliases := []string{".", "..", "~"}
```
AkihiroSuda, line 39: This check should be retained

pkg/nsutil/nsutil_test.go
djdongjin, line 1: We should move the tests to `pkg/store/filestore_test.go` unless they're all covered by existing cases there.

apostasie: Fair.
(Pending CI) #3725 does test filesystem restrictions thoroughly now.

cmd/nerdctl/main.go
```diff
@@ -239,6 +240,9 @@ Config file ($NERDCTL_TOML): %s
 				return fmt.Errorf("invalid cgroup-manager %q (supported values: \"systemd\", \"cgroupfs\", \"none\")", cgroupManager)
 			}
 		}
+		if err = store.ValidatePathComponent(globalOptions.Namespace); err != nil {
```
djdongjin, line 243: nit: 

can we add a comment stating something like namespace should be a single valid path component.
 
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#path-segment-names

djdongjin (approved):
LGTM thanks

djdongjin (approved):

cmd/nerdctl/main.go
apostasie: Sure. Let me do that real quick.

apostasie:
Thanks @djdongjin 

Latest push addresses your comment, and has been rebased to account for newly added unit tests.

apostasie:
`TestComposeUp` failure seems unrelated.

AkihiroSuda (approved):
Thanks

PR #3726, opened 2024-12-04
Increase timeouts on CI

apostasie:
Either GH is busier or they downgraded the base instances we are using.

One way or the other:
- windows unit test now hit the 5 minutes mark
- building dependencies on non arm machine hit over 10 minutes mark
- ipv6 tests are hitting the 10 minutes mark

@AkihiroSuda  @djdongjin at your convenience.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ env:
 jobs:
   # This job builds the dependency target of the test docker image for all supported architectures and cache it in GHA
   build-dependencies:
-    timeout-minutes: 10
+    timeout-minutes: 15
     name: dependencies |  ${{ matrix.containerd }} | ${{ matrix.arch }}
     runs-on: "${{ matrix.runner }}"
     strategy:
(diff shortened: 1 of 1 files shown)


djdongjin (approved):
LGTM, thanks

apostasie:
Thanks @djdongjin 

Just increased ipv6 test as well.

Not sure if GH slowness is transient or not, but either way we need a bit of leeway and I did set these timeouts too aggressive in the first place.

apostasie:
Ok... something is very wrong with GH right now...

https://github.com/containerd/nerdctl/actions/runs/12150865305

apostasie:
@djdongjin there definitely was some transient issue with github causing a bunch of timeouts.

I still think relaxing the timeouts a bit on these is warranted and will not hurt anyhow (windows normally clocks in at 4min+, quite close to the current 5 mins limit), so, let's merge if you are ok with that.

fahedouch:
is the execution engine (e.g. Docker) that runs these jobs using the maximum CPU/memory available on the machine ? we will reach a limit with the timeout :/

apostasie:
> is the execution engine (e.g. Docker) that runs these jobs using the maximum CPU/memory available on the machine ? we will reach a limit with the timeout :/

Not sure. I will have a look around.

About that, I have mixed feelings about the fact that we are running tests inside a docker container overall:
- this does introduce a layer of complexity (and possible additional bugs - especially on networking, but maybe also snapshotter?)
- this also alters the type of situations we normally see on a host, by restricting us to this "simplified" docker env

I know we do that for a reason (provide environment consistency and reproducibility), but it comes at a cost ^.
Maybe we could have `canary` run on the host instead of inside the container (would obviously require a lot of work to set the env though, so...).

apostasie:
@fahedouch 

https://github.com/containerd/nerdctl/pull/3726/checks#step:3:87

docker info reports 4 CPU 16GB.

Is there somewhere else dockerd could constrain containers resources that would not show in docker info?

PR #3728, opened 2024-12-04
Chore: import cleanup

apostasie:
Suggesting we remove spurious prefixes for imports that make reading code more confusing than necessary:
- remove `gocni`, and use the expected `cni`
- remove `gocontext`, and use the expected `context`


fahedouch (approved):
Thanks

AkihiroSuda (approved):
Thanks

PR #3729, opened 2024-12-04
Chore: spurious helpers test cleanup

apostasie:
Remove dependency on misplaced helper CreateBuildContext  for next-gen tests (use data.TempDir(), assert in the test).

### DIFF
--- a/cmd/nerdctl/image/image_list_test.go
+++ b/cmd/nerdctl/image/image_list_test.go
@@ -19,6 +19,8 @@ package image
 import (
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"runtime"
 	"slices"
 	"strings"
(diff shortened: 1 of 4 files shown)


cmd/nerdctl/image/image_list_test.go
```diff
@@ -144,7 +145,9 @@ LABEL foo=bar
 LABEL version=0.1
 RUN echo "actually creating a layer so that docker sets the createdAt time"
 `, testutil.CommonImage)
-			buildCtx := testhelpers.CreateBuildContext(t, dockerfile)
+			buildCtx := data.TempDir()
+			err := os.WriteFile(filepath.Join(buildCtx, "Dockerfile"), []byte(dockerfile), 0o600)
+			assert.NilError(helpers.T(), err)
```
djdongjin, line 150: why we cannot use the helper directly?

apostasie: (I say "copy it over" because a bunch of legacy tests are still dependent on the existing helper and would not be able to use the new one)

apostasie:
Thanks @djdongjin 

Appreciate the review.

PR #3736, opened 2024-12-06
fix some obvious error

ningmingxiao:
If the requested key doesn’t exist, we get the value type’s default value
https://go.dev/blog/maps

mybe not a bug.


ningmingxiao:
If the requested key doesn’t exist, we get the value type’s default value

Issue #3719, opened 2024-12-02
[CI]: failure nerdctl - TestRemove/Remove_image_with_running_container_-_with_-f : images appears to be racy / fail in certain circumstances


apostasie:
### Description

Seen on the CI:

```
=== Failed
=== FAIL: cmd/nerdctl/image TestRemove/Remove_image_with_running_container_-_with_-f (0.93s)
    image_remove_test.go:306: ======================== Pre-test cleanup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 rm -f testremove-remove-image-with-running-container-with-f-e9c3b26f
    command.go:112: /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 rmi -f 
    image_remove_test.go:306: ======================== Test setup ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 run --pull always -d --name testremove-remove-image-with-running-container-with-f-e9c3b26f ghcr.io/stargz-containers/alpine:3.13-org sleep 3600
    command.go:112: /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 image inspect ghcr.io/stargz-containers/alpine:3.13-org
    image_remove_test.go:306: ======================== Test Run ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 rmi -f ghcr.io/stargz-containers/alpine:3.13-org
    command.go:112: /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 images
    command.go:112: assertion failed: expression is false: strings.Contains(stdout, compare): Output does not contain: "<none>"
        Command:  /usr/local/bin/nerdctl --namespace=testremove-private-a0b107b8 images
        ExitCode: 0
        Stdout:   REPOSITORY    TAG    IMAGE ID    CREATED    PLATFORM    SIZE    BLOB SIZE
        
        Stderr:   time="2024-12-02T22:04:46Z" level=warning msg="content digest sha256:ec14c7992a97fc11425907e908340c6c3d6ff602f5f13d899e6b7027c9b4133a: not found"
        
        Env:
        HOSTNAME=cd6946756de5
        SYSTEMD_EXEC_PID=47
        container=docker
        HOME=/root
        LANG=C.UTF-8
        INVOCATION_ID=c7ba162bd0dd405697721433a973325e
        TERM=xterm
        SHLVL=3
[...]

PR #3739, opened 2024-12-07
Small Ci adjustments & cleanups

apostasie:
Minor CI cleanup prepping for landing of custom test images for windows in a subsequent PR.

- suggesting we stop using explicit version for well-known actions, and use major versions instead
- removing parameters that are actually default values

### DIFF
--- a/.github/workflows/ghcr-image-build-and-publish.yml
+++ b/.github/workflows/ghcr-image-build-and-publish.yml
@@ -31,19 +31,19 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4.2.2
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf  # v3.2.0
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@c47758b77c9736f4b2ef4073d4d51994fabfe349  # v3.7.1
 
       # Login against a Docker registry except on PR
       # https://github.com/docker/login-action
       - name: Log into registry ${{ env.REGISTRY }}
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v3.3.0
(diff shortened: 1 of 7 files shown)


.github/workflows/ghcr-image-build-and-publish.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4.2.2
+        uses: actions/checkout@v4
```
AkihiroSuda, line 34: OpenSSF, CNCF, etc. rather suggest us to pin the specific tag with the commit hash
- https://github.com/lima-vm/lima/pull/2738

apostasie: I don't think this makes practical sense for GitHub owned actions (though it would for third party actions).

Then again, very likely none of our dependencies are safe, even against third-party.

Even containerd itself half-assed the job:
https://github.com/containerd/containerd/blob/main/.github/actions/install-go/action.yml#L13

One way or the other, this changeset here does not impact security or change that situation (4.2.2 is the same amount of risk as 4).

Happy to change them all to sha if this is desirable - though this can come as a different PR.

Let me know your thoughts.

apostasie: Ok, let's get this moving.
Using sha everywhere then.

PTAL at your convenience.

.github/workflows/test.yml
```diff
@@ -40,11 +40,11 @@ jobs:
       CONTAINERD_VERSION: "${{ matrix.containerd }}"
       ARCH: "${{ matrix.arch }}"
     steps:
-      - uses: actions/checkout@v4.2.2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
AkihiroSuda, line 43: Linter doesn't like the number of the whitespaces
```
43:73 [comments] too few spaces before comment
```

apostasie: Damn.
These things are so picky :D

apostasie: Should be fixed. yamllinter is happy locally.

apostasie:
Not sure what is going on with #3719 (and variants).
Maybe these tests are somehow broken.

Will look into it as it becomes more aggravating.

AkihiroSuda (approved):
Thanks

Issue #3702, opened 2024-11-29
images list API, avoid multiple prints of duplicate images.


fengwei0328:
### What is the problem you're trying to solve

Due to the `CRI` interface, whether pulling images or loading images, the `CRI` interface writes data to the meta.db again. 
The storage methods between` CRI` and `containerd` are inconsistent:
1 The `imageStore` on the `CRI` side performs `deduplication`, 
2 The `imageStore` on the ctr side reads directly from the database. 
This leads to the presence of duplicate entries when viewing images in the `k8s.io` namespace, which affects the user experience when there are a large number of images.

cri:
```
$ crictl images       
IMAGE                      TAG                 IMAGE ID            SIZE
docker.io/library/centos   7                   eeb6ee3f44bd0       76.1MB
```
nerdctl:
```$ nerdctl -n k8s.io images
REPOSITORY    TAG       IMAGE ID        CREATED           PLATFORM       SIZE         BLOB SIZE
centos        7         be65f488b776    59 minutes ago    linux/amd64    211.5 MiB    72.6 MiB
centos        <none>    be65f488b776    59 minutes ago    linux/amd64    211.5 MiB    72.6 MiB
<none>        <none>    be65f488b776    59 minutes ago    linux/amd64    211.5 MiB    72.6 MiB

```
ctr:
```$ ctr -n k8s.io i ls
REF                                                                                              TYPE                                                      DIGEST                                                                  SIZE     PLATFORMS                                                       LABELS                          
docker.io/library/centos:7                                                                       application/vnd.docker.distribution.manifest.list.v2+json sha256:be65f488b7764ad3638f236b7b515b3678369a5124c47b8d32916d6487418ea4 72.6 MiB linux/386,linux/amd64,linux/arm/v7,linux/arm64/v8,linux/ppc64le io.cri-containerd.image=managed 
docker.io/library/centos@sha256:be65f488b7764ad3638f236b7b515b3678369a5124c47b8d32916d6487418ea4 application/vnd.docker.distribution.manifest.list.v2+json sha256:be65f488b7764ad3638f236b7b515b3678369a5124c47b8d32916d6487418ea4 72.6 MiB linux/386,linux/amd64,linux/arm/v7,linux/arm64/v8,linux/ppc64le io.cri-containerd.image=managed 
[...]

PR #3744, opened 2024-12-10
Fix duplicate image entries in k8s.io namespaces

fengwei0328:
[fix3702](https://github.com/containerd/nerdctl/issues/3702)

### DIFF
--- a/cmd/nerdctl/image/image_list_test.go
+++ b/cmd/nerdctl/image/image_list_test.go
@@ -317,3 +317,70 @@ CMD ["echo", "nerdctl-build-notag-string"]
 
 	testCase.Run(t)
 }
+
+func TestImagesKubeWithKubeHideDupe(t *testing.T) {
+	nerdtest.Setup()
+
+	testCase := &test.Case{
+		Require: test.Require(
+			nerdtest.OnlyKubernetes,
+		),
+		Setup: func(data test.Data, helpers test.Helpers) {
+			helpers.Ensure("pull", "--quiet", testutil.BusyboxImage)
+		},
+		SubTests: []*test.Case{
+			{
+				Description: "The same imageID will not print no-repo:tag in k8s.io with kube-hide-dupe",
+				Command:     test.Command("--kube-hide-dupe", "images"),
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						Output: func(stdout string, info string, t *testing.T) {
+							var imageID string
+							var skipLine int
+							lines := strings.Split(strings.TrimSpace(stdout), "\n")
+							header := "REPOSITORY\tTAG\tIMAGE ID\tCREATED\tPLATFORM\tSIZE\tBLOB SIZE"
+							if nerdtest.IsDocker() {
+								header = "REPOSITORY\tTAG\tIMAGE ID\tCREATED\tSIZE"
+							}
+							tab := tabutil.NewReader(header)
+							err := tab.ParseHeader(lines[0])
+							assert.NilError(t, err, info)
+							found := true
+							for i, line := range lines[1:] {
+								repo, _ := tab.ReadRow(line, "REPOSITORY")
+								tag, _ := tab.ReadRow(line, "TAG")
+								if repo+":"+tag == testutil.BusyboxImage {
+									skipLine = i
+									imageID, _ = tab.ReadRow(line, "IMAGE ID")
+									break
+								}
+							}
+							for i, line := range lines[1:] {
+								if i == skipLine {
+									continue
+								}
+								id, _ := tab.ReadRow(line, "IMAGE ID")
+								if id == imageID {
+									found = false
+									break
+								}
+							}
+							assert.Assert(t, found, info)
+						},
+					}
+				},
+			},
+			{
+				Description: "the same imageId will print no-repo:tag in k8s.io without kube-hide-dupe",
+				Command:     test.Command("images"),
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						Output: test.Contains("<none>"),
+					}
+				},
+			},
+		},
+	}
+
+	testCase.Run(t)
+}
--- a/cmd/nerdctl/image/image_remove_test.go
+++ b/cmd/nerdctl/image/image_remove_test.go
@@ -351,3 +351,173 @@ func TestIssue3016(t *testing.T) {
 
 	testCase.Run(t)
 }
+
+func TestRemoveKubeWithKubeHideDupe(t *testing.T) {
+	var numTags, numNoTags int
+	testCase := nerdtest.Setup()
+	testCase.NoParallel = true
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("--kube-hide-dupe", "rmi", "-f", testutil.BusyboxImage)
+	}
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		numTags = len(strings.Split(strings.TrimSpace(helpers.Capture("--kube-hide-dupe", "images")), "\n"))
+		numNoTags = len(strings.Split(strings.TrimSpace(helpers.Capture("images")), "\n"))
+	}
+	testCase.Require = test.Require(
+		nerdtest.OnlyKubernetes,
+	)
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "After removing the tag without kube-hide-dupe, repodigest is shown as <none>",
+			NoParallel:  true,
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("pull", testutil.BusyboxImage)
+			},
+			Command: test.Command("rmi", "-f", testutil.BusyboxImage),
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: 0,
+					Errors:   []error{},
+					Output: func(stdout string, info string, t *testing.T) {
+						helpers.Command("--kube-hide-dupe", "images").Run(&test.Expected{
+							Output: func(stdout string, info string, t *testing.T) {
+								lines := strings.Split(strings.TrimSpace(stdout), "\n")
+								assert.Assert(t, len(lines) == numTags+1, info)
+							},
+						})
+						helpers.Command("images").Run(&test.Expected{
+							Output: func(stdout string, info string, t *testing.T) {
+								lines := strings.Split(strings.TrimSpace(stdout), "\n")
+								assert.Assert(t, len(lines) == numNoTags+1, info)
+							},
+						})
+					},
+				}
+			},
+		},
+		{
+			Description: "If there are other tags, the Repodigest will not be deleted",
+			NoParallel:  true,
+			Cleanup: func(data test.Data, helpers test.Helpers) {
+				helpers.Anyhow("--kube-hide-dupe", "rmi", data.Identifier())
+			},
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("pull", testutil.BusyboxImage)
+				helpers.Ensure("tag", testutil.BusyboxImage, data.Identifier())
+			},
+			Command: test.Command("--kube-hide-dupe", "rmi", testutil.BusyboxImage),
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: 0,
+					Errors:   []error{},
+					Output: func(stdout string, info string, t *testing.T) {
+						helpers.Command("--kube-hide-dupe", "images").Run(&test.Expected{
+							Output: func(stdout string, info string, t *testing.T) {
+								lines := strings.Split(strings.TrimSpace(stdout), "\n")
+								assert.Assert(t, len(lines) == numTags+1, info)
+							},
+						})
+						helpers.Command("images").Run(&test.Expected{
+							Output: func(stdout string, info string, t *testing.T) {
+								lines := strings.Split(strings.TrimSpace(stdout), "\n")
+								assert.Assert(t, len(lines) == numNoTags+2, info)
+							},
+						})
+					},
+				}
+			},
+		},
+		{
+			Description: "After deleting all repo:tag entries, all repodigests will be cleaned up",
+			NoParallel:  true,
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("pull", testutil.BusyboxImage)
+				helpers.Ensure("tag", testutil.BusyboxImage, data.Identifier())
+			},
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				helpers.Ensure("--kube-hide-dupe", "rmi", "-f", testutil.BusyboxImage)
+				return helpers.Command("--kube-hide-dupe", "rmi", "-f", data.Identifier())
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					Output: func(stdout string, info string, t *testing.T) {
+						helpers.Command("--kube-hide-dupe", "images").Run(&test.Expected{
+							Output: func(stdout string, info string, t *testing.T) {
+								lines := strings.Split(strings.TrimSpace(stdout), "\n")
+								assert.Assert(t, len(lines) == numTags, info)
+							},
+						})
+						helpers.Command("images").Run(&test.Expected{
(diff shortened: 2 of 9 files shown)


AkihiroSuda (changes requested):
This should be opt-in via a config entry like `--kube-hide-dupe`:
https://github.com/containerd/nerdctl/blob/main/docs/config.md

fengwei0328:
> This should be opt-in via a config entry like `--kube-hide-dupe`: https://github.com/containerd/nerdctl/blob/main/docs/config.md

Okay, I'll try to implement it. Initially, I wanted to implement it separately, but this is a preliminary version for the sake of testing and discussion.

pkg/idutil/imagewalker/imagewalker.go
```diff
@@ -39,7 +40,7 @@ type Found struct {
 	NameMatchIndex int    // Image index with a name matching the argument for `nerdctl rmi`.
 }
 
-type OnFound func(ctx context.Context, found Found) error
+type OnFound func(ctx context.Context, found Found) (error, bool)
```
AkihiroSuda, line 43: Needs a comment to explain the returned bool

AkihiroSuda: Also, why not return bool before error?

pkg/cmd/image/list.go
```diff
@@ -128,6 +129,21 @@ type imagePrintable struct {
 
 func printImages(ctx context.Context, client *containerd.Client, imageList []images.Image, options *types.ImageListOptions) error {
 	w := options.Stdout
+	var ImageList []images.Image
+	if options.GOptions.Namespace != "k8s.io" || options.All {
```
AkihiroSuda, line 133: Needs a comment to clarify how the behavior differs for k8s.io

AkihiroSuda: Needs a test too

fengwei0328: Ok, I'll add instructions and tests to `--kube-hide-dupe` to the feature refinement

pkg/cmd/image/list.go
```diff
@@ -128,6 +129,35 @@ type imagePrintable struct {
...
+		the same imageId underk8s.io is showing multiple results: repo:tag, repo:digest, configID.
+		We expect to display only repo:tag, consistent with other namespaces and CRI.
+		e.g.
+		nerdctl -n k8s.io images
+		REPOSITORY    TAG       IMAGE ID        CREATED        PLATFORM       SIZE         BLOB SIZE
+		centos        7         be65f488b776    3 hours ago    linux/amd64    211.5 MiB    72.6 MiB
+		centos        <none>    be65f488b776    3 hours ago    linux/amd64    211.5 MiB    72.6 MiB
+		<none>        <none>    be65f488b776    3 hours ago    linux/amd64    211.5 MiB    72.6 MiB
+		expect:
+		nerdctl --kube_hide_dupe -n k8s.io images
```
AkihiroSuda, line 143: s/_/-/

pkg/cmd/image/list.go
```diff
@@ -128,6 +129,35 @@ type imagePrintable struct {
 
 func printImages(ctx context.Context, client *containerd.Client, imageList []images.Image, options *types.ImageListOptions) error {
 	w := options.Stdout
+	var ImageList []images.Image
+	/*
+		the same imageId underk8s.io is showing multiple results: repo:tag, repo:digest, configID.
```
AkihiroSuda, line 134: typo: underk8s.io

pkg/config/config.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	Experimental     bool     `toml:"experimental"`
 	HostGatewayIP    string   `toml:"host_gateway_ip"`
 	BridgeIP         string   `toml:"bridge_ip, omitempty"`
+	KubeHideDupe     bool     `toml:"kube_hide_dupe"`
```
AkihiroSuda, line 43: docs/config.md has to be updated too

docs/config.md
```diff
@@ -46,6 +46,7 @@ experimental   = true
 | `experimental`      | `--experimental`                   | `NERDCTL_EXPERIMENTAL`    | Enable  [experimental features](experimental.md)                                                                                                                 | Since 0.22.3     |
 | `host_gateway_ip`   | `--host-gateway-ip`                | `NERDCTL_HOST_GATEWAY_IP` | IP address that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP address of the host. It has no effect without setting --add-host | Since 1.3.0      |
 | `bridge_ip`         | `--bridge-ip`                      | `NERDCTL_BRIDGE_IP`       | IP address for the default nerdctl bridge network, e.g., 10.1.100.1/24                                                                                           | Since 2.0.1      |
+| `kube-hide-dupe`    | `--kube-hide-dupe`                 |                           | Deduplicate images for Kubernetes with namespace k8s.io, no more redundant <none> ones are displayed    | Since 2.0.2      |
```
AkihiroSuda, line 49: 2.0.3

cmd/nerdctl/image/image_remove_test.go
```diff
@@ -351,3 +351,110 @@ func TestIssue3016(t *testing.T) {
 
 	testCase.Run(t)
 }
+
+func TestRemoveKubeWithKubeHideDupe(t *testing.T) {
```
AkihiroSuda, line 355: @apostasie Could you take a look?
Is this correct?
The test doesn't seem to involve with Kube

apostasie: Looking.

apostasie: @AkihiroSuda 

This is correct.

Repro locally:

```
# This creates the kind / kube cluster and configure what needs be
./hack/build-integration-kubernetes.sh
# Jump into the control plane to gain access to the created cluster
nerdctl exec -ti nerdctl-test-control-plane bash
```

This ^ is exactly what the CI is doing

Now that you are in the control plane, use nerdctl against the kind cluster:
```
nerdctl --namespace k8s.io images
nerdctl --namespace k8s.io pull busybox
nerdctl --namespace k8s.io images
```

^ now, `2919d0172f75` (in my case, on arm) appears as busybox and none/none

cmd/nerdctl/image/image_list_test.go
```diff
@@ -317,3 +317,39 @@ CMD ["echo", "nerdctl-build-notag-string"]
...
+func TestImagesKubeWithKubeHideDupe(t *testing.T) {
+	nerdtest.Setup()
+
+	testCase := &test.Case{
+		Require: test.Require(
+			nerdtest.OnlyKubernetes,
+			test.Not(nerdtest.Docker),
+		),
+		Setup: func(data test.Data, helpers test.Helpers) {
+			helpers.Ensure("pull", "--quiet", testutil.NginxAlpineImage)
```
apostasie, line 330: @fengwei0328 can you use `BusyboxImage` instead?
It is considerably smaller to pull

cmd/nerdctl/image/image_remove_test.go
```diff
@@ -351,3 +351,110 @@ func TestIssue3016(t *testing.T) {
...
+	var numTags, numNoTags int
+	testCase := nerdtest.Setup()
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("--kube-hide-dupe", "rmi", "-f", testutil.NginxAlpineImage)
```
apostasie, line 359: Same as above.

cmd/nerdctl/image/image_remove_test.go
```diff
@@ -351,3 +351,110 @@ func TestIssue3016(t *testing.T) {
...
+	testCase.Require = test.Require(
+		nerdtest.OnlyKubernetes,
+		test.Not(nerdtest.Docker),
+	)
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "After removing the tag without kube-hide-dupe, repodigest is shown as <none>",
+			NoParallel:  true,
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("pull", testutil.NginxAlpineImage)
```
apostasie, line 372: Same

cmd/nerdctl/image/image_remove_test.go
```diff
@@ -351,3 +351,110 @@ func TestIssue3016(t *testing.T) {
...
+			},
+			Command: test.Command("rmi", "-f", testutil.NginxAlpineImage),
```
apostasie, line 374: Same

cmd/nerdctl/image/image_list_test.go
```diff
@@ -317,3 +317,39 @@ CMD ["echo", "nerdctl-build-notag-string"]
...
 	testCase.Run(t)
 }
+
+func TestImagesKubeWithKubeHideDupe(t *testing.T) {
+	nerdtest.Setup()
+
+	testCase := &test.Case{
+		Require: test.Require(
+			nerdtest.OnlyKubernetes,
+			test.Not(nerdtest.Docker),
```
apostasie, line 327: You don't need to `test.Not(nerdtest.Docker)`.
Because docker is never ran inside a Kube context, `nerdtest.OnlyKubernetes` will always be false.

(btw, I am not even sure how to use Docker against containerd in kube, but that is a different, albeit interesting, question)

pkg/cmd/image/list.go
```diff
@@ -128,6 +130,46 @@ type imagePrintable struct {
...
+		expect:
+		nerdctl --kube-hide-dupe -n k8s.io images
+		REPOSITORY    TAG       IMAGE ID        CREATED        PLATFORM       SIZE         BLOB SIZE
+		centos        7         be65f488b776    3 hours ago    linux/amd64    211.5 MiB    72.6 MiB
+	*/
+	if options.GOptions.KubeHideDupe && options.GOptions.Namespace == "k8s.io" {
+		imageDigest := make(map[digest.Digest]bool)
+		var ImageNoTag []images.Image
+		for _, img := range imageList {
+			parsed, err := reference.ParseAnyReference(img.Name)
```
apostasie, line 152: You should use nerdctl `referenceutil.Parse` instead.

pkg/cmd/image/list.go
```diff
@@ -128,6 +130,46 @@ type imagePrintable struct {
 
 func printImages(ctx context.Context, client *containerd.Client, imageList []images.Image, options *types.ImageListOptions) error {
 	w := options.Stdout
+	var ImageList []images.Image
```
apostasie, line 133: That is not a good name.

It should not start with a cap letter, and is very confusing to look at while the arg is `imageList`.

Suggesting instead `finaleImageList`? or anything else you like?

pkg/cmd/image/list.go
apostasie: Mmm... do you actually _need_ distribution reference here for some reason?

pkg/idutil/imagewalker/imagewalker.go
```diff
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/distribution/reference"
```
apostasie, line 25: Same as above.
Unless there is a reason to use `distribution/reference`, one should use nerdctl referenceutils instead.

apostasie (changes requested):
Thanks a lot @fengwei0328 

I left a few comments.

cmd/nerdctl/image/image_remove_test.go
apostasie: > This is correct.

What I mean is: the test is actually doing what it says, and does interact with Kube, as nerdctl is interacting with the kind cluster and the k8s.io namespace

pkg/cmd/image/list.go
```diff
@@ -128,6 +130,46 @@ type imagePrintable struct {
...
+			}
+			if _, ok := parsed.(reference.Tagged); !ok {
+				ImageNoTag = append(ImageNoTag, img)
+				continue
+			}
+			ImageList = append(ImageList, img)
+			imageDigest[img.Target.Digest] = true
+		}
+		//Ensure that dangling images without a repo:tag are displayed correctly.
+		for _, ima := range ImageNoTag {
```
fengwei0328, line 164: https://github.com/containerd/nerdctl/pull/3744#discussion_r1894316679
Here, I need to obtain the interface corresponding to the image name, and then use interface.type to distinguish whether it's a standard tag( reference.Tagged) or a non-standard tag().
I simply expected standard implementation, but `referenceutil.Parse` works too.

pkg/idutil/imagewalker/imagewalker.go
```diff
@@ -98,6 +118,101 @@ func (w *ImageWalker) Walk(ctx context.Context, req string) (int, error) {
...
+	nameMatchIndex := -1
+
+	//Distinguish between tag and non-tag
+	for _, img := range images {
+		ref := img.Name
+		parsed, err := reference.ParseAnyReference(ref)
+		if err != nil {
+			continue
+		}
+		if _, ok := parsed.(reference.Tagged); ok {
```
fengwei0328, line 166: https://github.com/containerd/nerdctl/pull/3744#discussion_r1894316679 @apostasie
Here, I need to obtain the interface corresponding to the image name, and then use interface.type to distinguish whether it's a standard tag( reference.Tagged) or a non-standard tag().
I simply expected standard implementation, but referenceutil.Parse works too.

apostasie: @fengwei0328 nerdctl does support additional image reference names that are not supported by the standard distribution library (ipfs and oci-archive custom schemes, ipfs ids).

That is the reason we should use nerdctl Parse instead.

apostasie: Thanks a lot for the changes @fengwei0328 
Let me do another review real quick.

apostasie:
@fengwei0328 

Maybe we could have an extra test for the following case?

```
nerdctl pull busybox
nerdctl tag busybox foo
id="$(nerdctl images busybox -q)"
```

Then
```
nerdctl rmi -f "$id"
```

Versus
```
nerdctl --kube-hide-dupe rmi -f "$id"
```

And confirm that in both cases, all images with that id are correctly removed?

And the same test, this time with `id=$(nerdctl images debian -q --no-trunc)`

wdyt?

cmd/nerdctl/image/image_remove_test.go
```diff
@@ -351,3 +351,109 @@ func TestIssue3016(t *testing.T) {
...
 	testCase.Run(t)
 }
+
+func TestRemoveKubeWithKubeHideDupe(t *testing.T) {
+	var numTags, numNoTags int
+	testCase := nerdtest.Setup()
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("--kube-hide-dupe", "rmi", "-f", testutil.BusyboxImage)
+		numTags = len(strings.Split(strings.TrimSpace(helpers.Capture("--kube-hide-dupe", "images")), "\n"))
+		numNoTags = len(strings.Split(strings.TrimSpace(helpers.Capture("images")), "\n"))
```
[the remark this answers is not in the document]
fengwei0328: This is to avoid that there is already a `busybox` under the namespace, which will count more, resulting in a situation where it is not good to count pull new-busybox later

fengwei0328:
> Maybe we could have an extra test for the following case?

Ok, that's also I do manually every time I'm locally

apostasie:
> debian

Sorry, I definitely meant `busybox`, not `debian`.

The point is to verify that force-deleting by short id, and digest, is still working as expected, and will delete images with multiple tags.

fengwei0328:
> short id, and digest

short id and digest tests added

fengwei0328:
> > debian
> 
> Sorry, I definitely meant `busybox`, not `debian`.
> 
> The point is to verify that force-deleting by short id, and digest, is still working as expected, and will delete images with multiple tags.

short id and digest tests added

cmd/nerdctl/image/image_remove_test.go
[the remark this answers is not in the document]
apostasie: You should mark the `testCase` as `NoParallel` (like you did already for the subtests). Otherwise, there is a chance that other, concurrent tests may still race your `Cleanup` (in between your calls to `rmi` and `images`).

Then you should move everything from `Cleanup` to `Setup`.

apostasie (approved):

apostasie:
Thanks a lot @fengwei0328 

One last comment on making sure the root test is NoParallel, LGTM otherwise.

Then, to maintainers.

fengwei0328:
> Thanks a lot @fengwei0328
> 
> One last comment on making sure the root test is NoParallel, LGTM otherwise.
> 
> Then, to maintainers.

Thanks a lot @apostasie

AkihiroSuda (approved):
Thanks

PR #3760, opened 2024-12-13
CI: go install action

apostasie:
Currently, we carry GO_VERSION around in a bunch of files, along with calls to `actions/setup-go`, which also comes with it pinned version.

This here suggests we centralize all of that inside a new, in-tree github action that also bundles in the logic for canary and simplify pulling go for the consumer:
- default strategy is to get the explicitly pinned go version (currently 1.23.4)
- `strategy latest-stable` will retrieve the latest patch for the main go version (eg: 1.23.x)
- `strategy old-stable` will retrieve the latest patch for the previous go version (eg: 1.22.x)
- `strategy canary` will retrieve the latest RC
 
That should clean-up our CI files a tad, and allow for much simpler / centralized maintenance, and also better control over usage of cache.

PR #3762, opened 2024-12-13
CI: go lint action

### LINKED ISSUE #3761: CI: golangci is broken with go canary (1.24RC)
### Description

The core of the issue is that golangci is unable to work properly with a go version that is higher than the one it has been built with, resulting in hogging memory / CPU.

This can be reproduced locally on a laptop:
- install go 1.24RC
- download binary `curl -sSfLO https://github.com/golangci/golangci-lint/releases/download/v1.62.2/golangci-lint-1.62.2-darwin-arm64.tar.gz`

Then:

```
dmp@Synchronicity:~/Projects/go/nerd/nerdctl$ go version
go version go1.24rc1 darwin/arm64

dmp@Synchronicity:~/Projects/go/nerd/nerdctl$ time GOOS=linux /Users/dmp/tmp/golangci-lint-1.62.2-darwin-arm64/golangci-lint run
^C^C

real	2m10.275s
user	5m25.968s
sys	3m23.894s
```

vs.

```
dmp@Synchronicity:~/Projects/go/nerd/nerdctl$ go version
go version go1.23.4 darwin/arm64

dmp@Synchronicity:~/Projects/go/nerd/nerdctl$ time GOOS=linux /Users/dmp/tmp/golangci-lint-1.62.2-darwin-arm64/golangci-lint run

real	0m27.558s
user	1m38.488s
sys	0m47.445s
```

I have no idea what kind of voodoo golangci is doing, but clearly their binary installs are not usable for our purpose.

Solutions are either: the golangci github action offers to build / install instead of downloading pre-built (and that would work with go1.24), or we figure out how to install manually and remove the action, or we stop linting for canary (would be sad...).

It is unclear to me ATM if there will be unforeseen issues with 1.24 wrt golangci - from a glimpse, it seems that it is just fine and maps are just sugar.

_No response_

na

main

None

### Host information

_No response_

apostasie:
Fix #3761 

After consideration, suggesting we get rid of the action entirely, as it does not seem to be able to accommodate our usage:
- install mode binary does not work for canary
- install mode goinstall precludes using GOOS, which we also need (and since their is no way to breakdown install and run steps with the action...)
- therefore, since installation has to be handled separately anyhow, we would use the action solely to run the command `golangci-lint run --verbose` - which is ludicrous, and already provided by the makefile (and we should encourage the use of the same commands on the CI we expect people to run locally)

Besides dropping the golangci action, this PR folds the imports ordering (previously separate task) into the lint-go action as well.

Note that one of the effects of this is that we loose the ability to cache lint results.
This is actually a good thing, as our github cache usage is currently a source of concern - and the cache does not provide any value for us (the linting process runs in seconds).

Finally, this is on top of #3760 which should go first.


AkihiroSuda:
> Finally, this is on top of https://github.com/containerd/nerdctl/pull/3760 which should go first.


I feel #3760 is too much complicated

PR #3763, opened 2024-12-15
Darwin build

slonopotamus:
Closes #732


apostasie:
Interesting!

Thanks @slonopotamus 

Is it functional right now? eg: if I share the right folders from a linux VM on a mac, is it "working" enough?

apostasie:
@slonopotamus would love to see at least some of that to get in.
Would you still have time and motivation for this?

slonopotamus:
Whoops, forgot about this. Will see what can be done...

SamuelMarks:
Testing now. Seems to build fine. Had to install buildkit and do some configuration so that `containerd` would run at a path it has access to.

Currently at this issue:
```
% sudo _output/nerdctl --address '/opt/containerd/run/containerd.sock' version
Password:
WARN[0000] unable to determine runc version              error="exec: \"runc\": executable file not found in $PATH"
Client:
 Version:	v2.0.3-213-g3c2889e7
 OS/Arch:	darwin/arm64
 Git commit:	3c2889e71ec12431a7d93c2a954674bd12658dc3
 buildctl:
  Version:	0.20.1
  GitCommit:	de56a3c5056341667b5bad71f414ece70b50724f

Server:
 containerd:
  Version:	v1.3.2-vmw
  GitCommit:	
 runc:
  Version:
```

Looks like runc doesn't support macOS: https://github.com/opencontainers/runc/issues/1733

Will work on that repo I guess and try to port it?

slonopotamus:
runc can't and won't run on macOS, it is fully tied to Linux.

SamuelMarks:
Yeah I realised that pretty quickly…


So how is this meant to work then?

slonopotamus:
Well, there is a nonzero amount of other runtime shims that _could_ run natively on macOS (though they might not as of today):

* [containerd-shim-spin](https://github.com/spinframework/containerd-shim-spin)
* [runu](https://github.com/ukontainer/runu)
* [containerd-shim-wasm](https://docs.rs/containerd-shim-wasm/latest/containerd_shim_wasm/)
* [rund](https://github.com/darwin-containers/rund)

Note that things these shims run are not a conventional "containers" with Linux userspace in them, but something different. However, `nerdctl` should not care about this. Its job is to call proper APIs of containerd and pass it parameters that tell containerd to do what user requests.

And by the way, `nerdctl` already has some kind of FreeBSD support. FreeBSD also cannot run `runc` (for the same reasons as macOS cannot), however there exists [runj](https://github.com/samuelkarp/runj) shim that can run FreeBSD jails.

AkihiroSuda:
Needs rebase

.github/workflows/lint.yml
```diff
@@ -27,6 +27,8 @@ jobs:
             goos: linux
           - os: ubuntu-24.04
             goos: freebsd
+          - os: ubuntu-24.04
```
AkihiroSuda, line 30: Why ubuntu?

AkihiroSuda (approved):
Thanks

.github/workflows/lint.yml
slonopotamus: It just selects a runner. I always was under impression than Linux runner are faster and more available on GitHub Actions. Also, Ubuntu runner is used for Windows lint in this file.

apostasie: We use ubuntu for windows because something is botched wrt Makefile path when on windows...

Will move it to macos-15 - the point is not so much speed here, but verifying that the tooling itself actually works on the target platform.

Making things a bit more clear here: https://github.com/containerd/nerdctl/pull/4129/commits/9c83660b2c9f479fd6379df1f4402365c8ca4576

But yeah, it's a detail..

PR #3792, opened 2024-12-27
Fix image inspect exit code on image not found error

austinvazquez:
### Issue
Fixes regression inadvertently introduced with https://github.com/containerd/nerdctl/pull/3017 for image inspect command exit code when image does not exist.

### Description
This change updates the image inspect utility function to serialize an array of encountered errors. This aligns with the behavior of `nerdctl inspect`.

### Testing
Adds simple test case for image inspect on images that do not exist.
Required updates to test cases which assumed exit code 0 for image inspects on images that do not exist.

### DIFF
--- a/pkg/cmd/image/inspect.go
+++ b/pkg/cmd/image/inspect.go
@@ -18,6 +18,7 @@ package image
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"regexp"
 	"strings"
@@ -97,6 +98,7 @@ func Inspect(ctx context.Context, client *containerd.Client, identifiers []strin
 	defer cancel()
 
 	// Will hold the final answers
+	var errs []error
 	var entries []interface{}
 
 	snapshotter := containerdutil.SnapshotService(client, options.GOptions.Snapshotter)
@@ -104,7 +106,7 @@ func Inspect(ctx context.Context, client *containerd.Client, identifiers []strin
 	for _, identifier := range identifiers {
 		candidateImageList, requestedName, requestedTag, err := inspectIdentifier(ctx, client, identifier)
 		if err != nil {
-			log.G(ctx).WithError(err).WithField("identifier", identifier).Error("failure calling inspect")
+			errs = append(errs, fmt.Errorf("%w: %s", err, identifier))
 			continue
 		}
 
@@ -185,6 +187,8 @@ func Inspect(ctx context.Context, client *containerd.Client, identifiers []strin
 			// Store our image
 			// foundImages[validatedDigest] = validatedImage
 			entries = append(entries, validatedImage)
+		} else {
+			errs = append(errs, fmt.Errorf("no such image: %s", identifier))
 		}
 	}
 
@@ -193,6 +197,10 @@ func Inspect(ctx context.Context, client *containerd.Client, identifiers []strin
 		if formatErr := formatter.FormatSlice(options.Format, options.Stdout, entries); formatErr != nil {
 			log.G(ctx).Error(formatErr)
 		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("%d errors:\n%w", len(errs), errors.Join(errs...))
 	}
 
 	return nil
--- a/cmd/nerdctl/image/image_inspect_test.go
+++ b/cmd/nerdctl/image/image_inspect_test.go
@@ -18,6 +18,8 @@ package image
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"runtime"
 	"strings"
 	"testing"
@@ -60,6 +62,14 @@ func TestImageInspectSimpleCases(t *testing.T) {
 				Description: "typedFormat support (.ID)",
 				Command:     test.Command("image", "inspect", testutil.CommonImage, "--format", "{{.ID}}"),
 				Expected:    test.Expects(0, nil, nil),
+			},
+			{
+				Description: "Error for image not found",
+				Command:     test.Command("image", "inspect", "dne:latest", "dne2:latest"),
+				Expected: test.Expects(1, []error{
+					errors.New("no such image: dne:latest"),
+					errors.New("no such image: dne2:latest"),
+				}, nil),
 			},
 		},
 	}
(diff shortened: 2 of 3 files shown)


cmd/nerdctl/image/image_inspect.go
```diff
@@ -95,7 +97,13 @@ func imageInspectAction(cmd *cobra.Command, args []string) error {
 	}
 	defer cancel()
 
-	return image.Inspect(ctx, client, args, options)
+	err = image.Inspect(ctx, client, args, options)
+	if uw, ok := err.(interface{ Unwrap() []error }); ok {
+		errs := uw.Unwrap()
+		err = fmt.Errorf("%d errors: %v", len(errs), errs)
+	}
```
djdongjin, line 104: why do we need to unwrap here?

djdongjin (commented):
The fix LGTM, but I'm not sure why we need the errors unwrap.

cmd/nerdctl/image/image_inspect.go
austinvazquez: Thanks @djdongjin, I just thought it would be good to align with the format used by `nerdctl inspect` command.
https://github.com/containerd/nerdctl/blob/6f34536582f9c7211883432580e732c1ac3a49f0/cmd/nerdctl/inspect/inspect.go#L164

```bash
austin@nimbus:~/workspace/nerdctl$ mysudo nerdctl inspect dne:latest dne2:latest
FATA[0000] 2 errors: [no such object dne:latest no such object dne2:latest]
```

Open to suggestions here. Docker is having format of one error per line, but thought perhaps to stay consistent with nerdctl inspect instead.

```
austin@nimbus:~/workspace/nerdctl$ docker inspect dne:latest dne2:latest
[]
Error: No such object: dne:latest
Error: No such object: dne2:latest
austin@nimbus:~/workspace/nerdctl$ docker image inspect dne:latest dne2:latest
[]
Error response from daemon: No such image: dne:latest
Error response from daemon: No such image: dne2:latest
```

djdongjin: I feel one error per line will be visiually better? Also some other nerdctl command also does one error per line:

```
☁  ~   ➜ sudo nerdctl rmi sdaf sdfasdf
FATA[0000] 2 errors:
no such image: sdaf
no such image: sdfasdf
```

AkihiroSuda:
What is "dnf" here?
Doesn't seem related to Fedora's `dnf`

cmd/nerdctl/image/image_inspect.go
austinvazquez: SGTM, pushing new revision with suggestion.

austinvazquez:
@AkihiroSuda, oops sorry about that. I had shorthanded "did not find" not meaning to confuse dnf package manager. Lastest commit should have more clear wording.

pkg/cmd/image/inspect.go
```diff
@@ -195,5 +198,9 @@ func Inspect(ctx context.Context, client *containerd.Client, identifiers []strin
 		}
 	}
 
+	if len(errs) > 0 {
+		return fmt.Errorf("%d errors:\n%s", len(errs), strings.Join(errs, "\n"))
+	}
```
djdongjin, line 203: I think you can still use `var errs []error` to store all the errors, and here do

```golang
return fmt.Errorf("%d errors:\n%w", len(errs), errors.Join(errs...))
```

When you print a `errors.Join`ed error, it will automatically print one error per line.

pkg/cmd/image/inspect.go
```diff
@@ -97,14 +97,15 @@ func Inspect(ctx context.Context, client *containerd.Client, identifiers []strin
...
+	var errs []string
 	var entries []interface{}
 
 	snapshotter := containerdutil.SnapshotService(client, options.GOptions.Snapshotter)
 	// We have to query per provided identifier, as we need to post-process results for the case name + digest
 	for _, identifier := range identifiers {
 		candidateImageList, requestedName, requestedTag, err := inspectIdentifier(ctx, client, identifier)
 		if err != nil {
-			log.G(ctx).WithError(err).WithField("identifier", identifier).Error("failure calling inspect")
+			errs = append(errs, fmt.Sprintf("invalid reference format: %s", identifier))
```
djdongjin, line 108: Is `invalid reference format` the only possible error here? Or we should return the (wrapped) original error? e.g., something like `fmt.Errorf("invalid reference %s: %w", identifier, err)`

djdongjin (commented):
Left some minor comments

pkg/cmd/image/inspect.go
austinvazquez: It looks like other errors from `inspectIdentifier` are errors returned from containerd image service. i.e. https://github.com/austinvazquez/nerdctl/blob/58fb7731a84ac3f260d971f8822aee8329bf1bbf/pkg/cmd/image/inspect.go#L62

So wrapping is useful in that case, but no in the case the error is invalid reference format. Perhaps it is best to revert back to log statement? I originally brought this in as error to align more with Docker behavior, but maybe it isn't worth losing the information here when error talking to containerd.
```
austin@nimbus:~/workspace/nerdctl$ mysudo nerdctl image inspect ∞∞∞∞∞∞∞∞∞∞ dne:latest
FATA[0000] 2 errors:
invalid reference format: ∞∞∞∞∞∞∞∞∞∞: invalid reference format
no such image: dne:latest 
austin@nimbus:~/workspace/nerdctl$ 
```

djdongjin: How about let's do `fmt.Errorf("%w: %s", err, identifier)` 🤔?

The output will be something like:

```
invalid reference format: ∞∞∞∞∞∞∞∞∞∞
no such image: dne:latest 
containerd image service failed: containerd_err: dne:latest2
```

djdongjin (approved):
LGTM on CI green, thanks

pkg/cmd/image/inspect.go
austinvazquez: Thanks for the suggestion btw; that solution was way simpler.

apostasie:
Hey @austinvazquez 

Looks like this is introducing a regression with `--mode=native`.

When in mode native, https://github.com/austinvazquez/nerdctl/blob/aa5eb28ade74afcd6ff1cd29ae3b937d5fd62652/pkg/cmd/image/inspect.go#L127-L129 you never get a validateImage, so...

https://github.com/austinvazquez/nerdctl/blob/aa5eb28ade74afcd6ff1cd29ae3b937d5fd62652/pkg/cmd/image/inspect.go#L190-L192

I am not quite sure how the CI did not catch this. Will have a look at that.

Would you have time to revisit this here to fix the regression?

To reproduce, just `nerdctl image inspect --mode native debian; echo $?` should show this.

PR #3793, opened 2024-12-29
Refactor: Encapsulate net opts loading in internalLabels

motinsa:
### Context
The `create.go` file contained a TODO to formalize the way network 
options are loaded into `internalLabels`. The previous implementation 
manually assigned each field, which made it harder to maintain and extend.

### Solution
- Introduced a new method `LoadNetOpts` in the `internalLabels` struct.
- Encapsulated the logic for loading network options, improving 
  maintainability and extensibility.

### Impact
This change is backward-compatible and does not affect existing functionality.

Signed-off-by: Pedro Aibar <pedroaibarmarco@gmail.com>

### DIFF
--- a/pkg/cmd/container/create.go
+++ b/pkg/cmd/container/create.go
@@ -248,7 +248,7 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
 	envs = append(envs, "HOSTNAME="+netLabelOpts.Hostname)
 	opts = append(opts, oci.WithEnv(envs))
 
-	internalLabels.LoadNetOpts(netLabelOpts)
+	internalLabels.loadNetOpts(netLabelOpts)
 
 	// NOTE: OCI hooks are currently not supported on Windows so we skip setting them altogether.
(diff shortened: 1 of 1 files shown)


pkg/cmd/container/create.go
```diff
@@ -732,6 +726,16 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
 	return containerd.WithAdditionalContainerLabels(m), nil
 }
 
+// LoadNetOpts loads network options into InternalLabels.
+func (il *internalLabels) LoadNetOpts(opts types.NetworkOptions) {
```
yankay, line 730: `LoadNetOpts` is a public function. Could you please clarify why it is necessary for this function to be public rather than private?"

motinsa: Making LoadNetOpts public was unintentional. I've already updated the function to be private to align with its intended use case.

fahedouch (approved):
Thanks

PR #3797, opened 2025-01-07
update CNI plugins (1.6.2)

djdongjin:
https://github.com/containernetworking/plugins/releases/tag/v1.6.2

> (administrative note: release v1.6.1 was somehow double-created in GitHub; v1.6.2 is identical but fixes the duplication.)

Now 1.6.1 only has s390v artifacts and fails our CI:

https://github.com/containernetworking/plugins/releases/tag/v1.6.1

https://github.com/containerd/nerdctl/actions/runs/12641602765/job/35224329459?pr=3796

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@
 # Basic deps
 ARG CONTAINERD_VERSION=v2.0.1
 ARG RUNC_VERSION=v1.2.3
-ARG CNI_PLUGINS_VERSION=v1.6.1
+ARG CNI_PLUGINS_VERSION=v1.6.2
 
 # Extra deps: Build
 ARG BUILDKIT_VERSION=v0.18.2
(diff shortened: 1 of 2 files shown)


yankay (approved):

Dockerfile
```diff
@@ -57,18 +57,18 @@ FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-bookworm AS build-base-debia
 COPY --from=xx / /
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -qq && apt-get install -qq --no-install-recommends \
-    git \
-    dpkg-dev
+  git \
```
AkihiroSuda, line 60: ?

Dockerfile.d/SHA256SUMS.d/cni-plugins-v1.6.2
AkihiroSuda, line 1: [cni-plugins-v1.6.1](https://github.com/djdongjin/nerdctl/blob/b0e78eeeca2ff0ca8d35457d409075657ef2c597/Dockerfile.d/SHA256SUMS.d/cni-plugins-v1.6.1) can be removed

Dockerfile
djdongjin: ah probably editor auto format 🤦, let me change this back (and remove the 1.6.1 as well in the other comment), sry!

djdongjin: @AkihiroSuda actually this file has mixed 2-space and 4-space indent. After this all indents are 2-space. Maybe we should keep this format change? 

4 spaces:
https://github.com/containerd/nerdctl/blob/1220ce7ec2701d485a9b1beeea63dae3da134fb5/Dockerfile#L65-L71

2 spaces:
https://github.com/containerd/nerdctl/blob/1220ce7ec2701d485a9b1beeea63dae3da134fb5/Dockerfile#L79-L83

AkihiroSuda: Either is fine to me.

Probably we should have some linter to enforce either of one.

djdongjin: Sounds good, I will keep this format change for now then for consistency. Let me see if there are some linter that can run in CI.

PR #3807, opened 2025-01-10
Update containerd (1.7.25) in CI

austinvazquez:
This change updates containerd used in CI to 1.7.25
Release notes: https://github.com/containerd/containerd/releases/tag/v1.7.25
Full diff: https://github.com/containerd/containerd/compare/v1.7.24...v1.7.25

PR #3810, opened 2025-01-11
Bump up github.com/containerd/go-cni to 1.1.12 and fix CI

djdongjin:
CI has been failing on `test / FreeBSD (pull_request)` https://github.com/containerd/nerdctl/pull/3807, https://github.com/containerd/nerdctl/pull/3809:

> Notice it's looking for version `v0` which doesn't exist. I think it might because the vagrant version used in github action.
> Moving vagrant step to ubuntu 24.04 (similar to containerd's setup) does fix it.

```
==> default: Adding box 'generic/freebsd14' (v0) for provider: libvirt
    default: Downloading: https://vagrantcloud.com/generic/freebsd14

The box failed to unpackage properly. Please verify that the box
file you're trying to add is not corrupted and that enough disk space
is available and then try again.
The output from attempting to unpackage (if any):

bsdtar: Error opening archive: Unrecognized archive format
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -392,8 +392,7 @@ jobs:
   test-integration-freebsd:
     timeout-minutes: 30
     name: FreeBSD
-    # ubuntu-24.04 lacks the vagrant package
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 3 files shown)


fahedouch (approved):

AkihiroSuda (approved):
Thanks

PR #3818, opened 2025-01-14
reset spec if update returned error

zzzzzzzzzy9:
reset spec if update returned error

### DIFF
--- a/cmd/nerdctl/container/container_update_linux_test.go
+++ b/cmd/nerdctl/container/container_update_linux_test.go
@@ -1,0 +1,33 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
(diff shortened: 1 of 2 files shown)


zzzzzzzzzy9:
@AkihiroSuda could you please review this pr? thanks.

zzzzzzzzzy9:
1、update memory and gave a wrong args to restart.
```bash
nerdctl -n k8s.io update --memory=450000000 --restart=123 ef66
```
2、This will got an error like this:
```bash
FATA[0000] 1 errors:
unsupported restart policy "123", supported policies are: ["no" "always" "on-failure" "unless-stopped"]
```
3、inspect we can see the memory is 450000000 now. In fact, the memory is not being updated.

AkihiroSuda:
Can we have an integration test?

zzzzzzzzzy9:
Ok, I will do it.

zzzzzzzzzy9:
Done. @AkihiroSuda

cmd/nerdctl/container/container_update_linux_test.go
```diff
@@ -0,0 +1,33 @@
...
+package container
+
+import (
+	"testing"
+
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+)
+
+func TestUpdateContainer(t *testing.T) {
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 26: Why?

AkihiroSuda (approved):
Thanks, I'm merging this, but wondering if we can omit `--mode=native` and run some `(docker|nerdctl) exec` command instead

zzzzzzzzzy9:
We can't use exec because the update fails and the actual memory is not modified. But the original situation is that even though it has not been modified, `inspect --mode=native` shows that it has been modified. So exec can't see this problem.

Issue #3806, opened 2025-01-10
nerdctl save dismisses multiple tags for same image


MathiasRudolphSec:
### Description

When using `nerdctl save` for multiple tags that reverence the same image, only a single tag will be retained in the saved dump.

### Steps to reproduce the issue

1. `nerdctl pull quay.io/kiwigrid/k8s-sidecar:1.28.0`
2. `nerdctl pull kiwigrid/k8s-sidecar:1.28.0`
3. `nerdctl save quay.io/kiwigrid/k8s-sidecar:1.28.0 kiwigrid/k8s-sidecar:1.28.0 | pigz > images.tar.gz`

### Result
`nerdctl load -i images.tar.gz` produces:
```
unpacking quay.io/kiwigrid/k8s-sidecar:1.28.0 (sha256:4166a019eeafd1f0fef4d867dc5f224f18d84ec8681dbb31f3ca258ecf07bcf2)...
Loaded image: quay.io/kiwigrid/k8s-sidecar:1.28.0
```

### Expected Result
When I use docker to do the same:
`docker save quay.io/kiwigrid/k8s-sidecar:1.28.0 kiwigrid/k8s-sidecar:1.28.0 | gzip > docker-save.tar.gz`
the same `nerdctl load` produces:
```
unpacking quay.io/kiwigrid/k8s-sidecar:1.28.0 (sha256:40a8c7a9ef0598a807b908dc458a8057698b1f92a3c0c2fe4929a0d3700e5dc8)...
Loaded image: quay.io/kiwigrid/k8s-sidecar:1.28.0
unpacking docker.io/kiwigrid/k8s-sidecar:1.28.0 (sha256:40a8c7a9ef0598a807b908dc458a8057698b1f92a3c0c2fe4929a0d3700e5dc8)...
Loaded image: kiwigrid/k8s-sidecar:1.28.0
```
which is the expected behavior.

### What version of nerdctl are you using?

Manually installed from gitlab releases.
```bash
> sudo nerdctl version
Client:
 Version:       v2.0.2
 OS/Arch:       linux/amd64
 Git commit:    1220ce7ec2701d485a9b1beeea63dae3da134fb5
 buildctl:
  Version:      v0.15.2
  GitCommit:    9e14164a1099d3e41b58fc879cbdd6f2b2edb04e

Server:
 containerd:
  Version:      v2.0.1
  GitCommit:    88aa2f531d6c2922003cc7929e51daf1c14caa0a
 runc:
[...]

PR #3822, opened 2025-01-16
fix: save multiple images for the same image id to a tar archive

haytok:
Suppose we try to save multiple container images with the same image ID but different image names into a tar archive using the nerdctl save command.

When we then try to load container images from this tar archive using the nerdctl load command, not all container images will be loaded.

This behavior is reported and the details are described in the following:

- https://github.com/containerd/nerdctl/issues/3806

Therefore, this PR resolves this issue.

### DIFF
--- a/cmd/nerdctl/image/image_save_test.go
+++ b/cmd/nerdctl/image/image_save_test.go
@@ -138,3 +138,60 @@ func TestSave(t *testing.T) {
 
 	testCase.Run(t)
 }
+
+// TestSaveMultipleImagesWithSameIDAndLoad tests https://github.com/containerd/nerdctl/issues/3806
+func TestSaveMultipleImagesWithSameIDAndLoad(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/image/image_save_test.go
```diff
@@ -138,3 +138,60 @@ func TestSave(t *testing.T) {
 
 	testCase.Run(t)
 }
+
+// TestIssue3806 tests https://github.com/containerd/nerdctl/issues/3806
+func TestIssue3806(t *testing.T) {
```
AkihiroSuda, line 143: Can we have a more descriptive test name?

haytok: Thanks for commnets ! I've updated.

haytok:
`Run integration tests (flaky)` has failed ...
If it is flaky, could you run CI again?

fahedouch (approved):
Thanks

PR #3824, opened 2025-01-17
Correct `compose logs` command reference

rerpha:
I think this was copy-pasted from `compose up` so wasn't quite right.


AkihiroSuda (approved):
Thanks

Issue #3839, opened 2025-01-24
journald logging and tests


apostasie:
### Description

If the user is _not_ a member of group `adm` (or `systemd-journal`), `journalctl` will refuse to show logs, and bail out with exit code 1.

With nerdctl, we surface journalctl message, as-is: `No journal files were opened due to insufficient permissions`

That part is kinda ok.

But we still exit 0.

Now, in that scenario, all tests involving journald will fail.

I would suggest the following:
- `nerdctl logs foo` should exit 1 as well if journalctl does exit 1
- skip the tests if we lack permissions

Note that docker rootless also suck at it, and just return nothing, exit 0.

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Note: this is the default situation with lima.

PR #3841, opened 2025-01-24
Fix journald logger exit and wait behavior

### LINKED ISSUE #3840: nerdctl logs with log-driver journald does not `wait` on journalctl
### Description

`nerdctl logs -f foo`

will launch `journalctl SYSLOG_IDENTIFIER=44c10dfcf2c6 --output=cat -f` and just returns.

We should `cmd.Wait()`.

_No response_

na

main

None

### Host information

_No response_

apostasie:
This should fix #3840 and part of #3839

### DIFF
--- a/pkg/logging/journald_logger.go
+++ b/pkg/logging/journald_logger.go
@@ -173,11 +173,20 @@ func FetchLogs(stdout, stderr io.Writer, journalctlArgs []string, stopChannel ch
 	}
 
 	// Setup killing goroutine:
+	killed := false
 	go func() {
 		<-stopChannel
+		killed = true
 		log.L.Debugf("killing journalctl logs process with PID: %#v", cmd.Process.Pid)
 		cmd.Process.Kill()
 	}()
+
+	err = cmd.Wait()
+	if exitError, ok := err.(*exec.ExitError); ok {
+		if !killed && exitError.ExitCode() != 0 {
+			return fmt.Errorf("journalctl command exited with non-zero exit code (%d): %w", exitError.ExitCode(), exitError)
+		}
+	}
 
 	return nil
 }


pkg/logging/journald_logger.go
```diff
@@ -173,12 +173,21 @@ func FetchLogs(stdout, stderr io.Writer, journalctlArgs []string, stopChannel ch
...
 		<-stopChannel
+		killed = true
 		log.L.Debugf("killing journalctl logs process with PID: %#v", cmd.Process.Pid)
 		cmd.Process.Kill()
 	}()
 
+	err = cmd.Wait()
+	if exitError, ok := err.(*exec.ExitError); ok {
+		if !killed && exitError.ExitCode() != 0 {
+			return errors.New("journalctl command exited with non-zero exit code")
```
djdongjin, line 187: suggest:

`return fmt.Errorf("journalctl command exited with non-zero exit code (%d): %w", exitError.ExitCode(), exitError)`

apostasie: Thanks. Will do.

apostasie: Done.

djdongjin (approved):
LGTM, thanks

pkg/logging/journald_logger.go
```diff
@@ -173,12 +173,21 @@ func FetchLogs(stdout, stderr io.Writer, journalctlArgs []string, stopChannel ch
 	}
 
 	// Setup killing goroutine:
+	killed := false
 	go func() {
 		<-stopChannel
+		killed = true
```
AkihiroSuda, line 179: This should be set after killing the process?

AkihiroSuda: Also, this one should have a mutex

apostasie: > This should be set after killing the process?

But then, wouldn't `Wait()` return before `killed` is set?

apostasie: > Also, this one should have a mutex

I am not sure I understand why. `Wait()` will not return until the process is killed, right?

AkihiroSuda (approved):
Thanks

PR #3842, opened 2025-01-24
Fix test time to use UTC

apostasie:
(jsonfile does log UTC)


djdongjin (approved):
Did this cause this test to be flaky?

apostasie:
> Did this cause this test to be flaky?

It causes the test to fail dependent on the machine settings (for example on a local lima instance here).

AkihiroSuda (approved):
Thanks

PR #3843, opened 2025-01-24
Harden journald test

### LINKED ISSUE #3839: journald logging and tests

apostasie:
Fix #3839 - the other part of it - the first part is being fixed in #3841

### DIFF
--- a/cmd/nerdctl/container/container_logs_test.go
+++ b/cmd/nerdctl/container/container_logs_test.go
@@ -18,12 +18,14 @@ package container
 
 import (
 	"fmt"
+	"os/exec"
 	"runtime"
 	"strings"
 	"testing"
 	"time"
 
 	"gotest.tools/v3/assert"
+	"gotest.tools/v3/icmd"
 
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 )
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/container/container_logs_test.go
```diff
@@ -109,8 +111,14 @@ func TestLogsWithInheritedFlags(t *testing.T) {
 }
 
 func TestLogsOfJournaldDriver(t *testing.T) {
-	t.Parallel()
 	testutil.RequireExecutable(t, "journalctl")
+	journalctl, _ := exec.LookPath("journalctl")
+	res := icmd.RunCmd(icmd.Command(journalctl, "-xe"))
+	if res.ExitCode != 0 {
```
AkihiroSuda, line 117: Can we print stderr here?

apostasie: Sure. Will do.

apostasie: Done.

AkihiroSuda (approved):
Thanks

PR #3847, opened 2025-01-24
CI: test portability fix: ensure tini-custom is here

apostasie:


cmd/nerdctl/container/container_run_linux_test.go
```diff
@@ -277,6 +277,7 @@ func TestRunUlimit(t *testing.T) {
 func TestRunWithInit(t *testing.T) {
 	t.Parallel()
 	testutil.DockerIncompatible(t)
+	testutil.RequireExecutable(t, "tini-custom")
```
fahedouch, line 280: Why checking for `tini-custom` outside container ?

apostasie: > Why checking for `tini-custom` outside container ?

So that the test suite can pass (regardless of surrounding context).

apostasie: The point being: I don't think the test suite should only run, or pass, inside the container. Conditions inside are very specific, and give us a false sense of whether things are working or not outside.
Furthermore, for local development, it is just much faster to just work directly on the host.
So, the test suite should be able to run anywhere, with tests expressing requirements and skipping appropriately if not met.

fahedouch: But `tini-custom` is executed as parent process inside the container pid ns not in the host!

apostasie: tini-custom binary still has to exist on the "host" (https://github.com/containerd/nerdctl/blob/main/Dockerfile#L307)

apostasie: eg: `--init-binary` has to point to an _existing binary_ from where you are calling it.

fahedouch (approved):
Thanks

PR #3850, opened 2025-01-25
Cleanup XDGRuntime

apostasie:
From reading code, it seems to me like:

1. `XDGRuntimeDir()` should check the value of `ROOTLESSKIT_PARENT_EUID` (like `ParentEUID` does)
2. `getXDGRuntimeDir()` in bypass4netns is essentially duplicating `XDGRuntimeDir()`
3. `getRuntimeVariableDataDir` and `CNIRuntimeDir` should not ignore errors, and they seem to have faulty logic (relying on `Geteuid` regardless of conditions), or calling `rootlessutil.ParentEUID` which will redo stuff done already in `XDGRuntimeDir()`

PR addresses these.

### DIFF
--- a/pkg/buildkitutil/buildkitutil_linux.go
+++ b/pkg/buildkitutil/buildkitutil_linux.go
@@ -18,13 +18,12 @@ package buildkitutil
 
 import (
 	"fmt"
-
-	"github.com/containerd/log"
+	"os"
 
 	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
 )
 
-func getRuntimeVariableDataDir() string {
+func getRuntimeVariableDataDir() (string, error) {
 	// Per Linux Foundation "Filesystem Hierarchy Standard" version 3.0 section 3.15.
 	// Under version 2.3, this was "/var/run".
 	run := "/run"
@@ -32,9 +31,11 @@ func getRuntimeVariableDataDir() string {
 		var err error
 		run, err = rootlessutil.XDGRuntimeDir()
 		if err != nil {
-			log.L.Warn(err)
-			run = fmt.Sprintf("/run/user/%d", rootlessutil.ParentEUID())
+			if rootlessutil.IsRootlessChild() {
+				return "", err
+			}
+			run = fmt.Sprintf("/run/user/%d", os.Geteuid())
 		}
 	}
-	return run
+	return run, nil
 }
--- a/pkg/rootlessutil/xdg_linux.go
+++ b/pkg/rootlessutil/xdg_linux.go
@@ -18,8 +18,10 @@ package rootlessutil
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 )
 
 func XDGRuntimeDir() (string, error) {
@@ -28,8 +30,11 @@ func XDGRuntimeDir() (string, error) {
 	}
 	// Fall back to "/run/user/<euid>".
 	// Note that We cannot rely on os.Geteuid() because we might be inside UserNS.
-	if euid := os.Getenv("ROOTLESSKIT_PARENT_EUID"); euid != "" {
-		return "/run/user/" + euid, nil
+	if parentEuid, ok := os.LookupEnv("ROOTLESSKIT_PARENT_EUID"); ok {
+		if _, err := strconv.Atoi(parentEuid); err != nil {
+			return "", fmt.Errorf("invalid ROOTLESSKIT_PARENT_EUID environment variable value %q: %w", parentEuid, err)
+		}
+		return "/run/user/" + parentEuid, nil
 	}
 	return "", errors.New("environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/")
 }
--- a/pkg/buildkitutil/buildkitutil_freebsd.go
+++ b/pkg/buildkitutil/buildkitutil_freebsd.go
@@ -16,7 +16,7 @@
 
 package buildkitutil
 
-func getRuntimeVariableDataDir() string {
+func getRuntimeVariableDataDir() (string, error) {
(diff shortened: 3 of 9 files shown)


apostasie:
Was wondering if we should even incorporate into `XDGRuntimeDir` the extra logic
```
			if rootlessutil.IsRootlessChild() {
				return "", err
			}
			run = fmt.Sprintf("/run/user/%d", os.Geteuid())
```

but (for now) elected for pure refactoring with no change of behavior.

apostasie:
Not clear why the failure (some of it is definitely the usual stargz / layer not found), but the rest is weird (eg: platform timeout).

If we can poke it in case it is a fluke...

pkg/rootlessutil/xdg_linux.go
```diff
@@ -28,10 +29,11 @@ func XDGRuntimeDir() (string, error) {
 	}
 	// Fall back to "/run/user/<euid>".
 	// Note that We cannot rely on os.Geteuid() because we might be inside UserNS.
-	if euid := os.Getenv("ROOTLESSKIT_PARENT_EUID"); euid != "" {
-		return "/run/user/" + euid, nil
+	euid, err := strconv.Atoi(os.Getenv("ROOTLESSKIT_PARENT_EUID"))
+	if err != nil {
+		return "", errors.New("environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/")
```
AkihiroSuda, line 34: The new code is harder to read, as the error returned here is about `XDG_RUNTIME_DIR` while the original `err` is about `ROOTLESSKIT_PARENT_EUID`

apostasie: I don't like any of that code either.

We can add more fine grained errors. Just not sure this is worth the effort.
Lmk what you would prefer.

AkihiroSuda: The old code is easier to understand to me

pkg/rootlessutil/xdg_linux.go
```diff
@@ -28,10 +29,11 @@ func XDGRuntimeDir() (string, error) {
 	}
 	// Fall back to "/run/user/<euid>".
 	// Note that We cannot rely on os.Geteuid() because we might be inside UserNS.
-	if euid := os.Getenv("ROOTLESSKIT_PARENT_EUID"); euid != "" {
-		return "/run/user/" + euid, nil
```
AkihiroSuda, line 32: If you want to validate the format of `$ROOTLESSKIT_PARENT_EUID`, a validation code can be added here
```go
if _, err := strconv.Atoi(euid); err != nil {
  return fmt.Errorf("invalid $ROOTLESSKIT_PARENT_EUID value: %q: %w", euid, err)
}
```

pkg/rootlessutil/xdg_linux.go
apostasie: Updated.

apostasie:
Failures are a bunch of timeouts. Looks like the import cache manifest from github cache are extremely slow during the initial build phase:
- normal import ~ 1 second : https://github.com/containerd/nerdctl/actions/runs/13463005401/job/37622717754?pr=3850#step:4:90
- abnormal imports ~ 5 minutes+: https://github.com/containerd/nerdctl/actions/runs/13463005401/job/37622717494?pr=3850#step:4:72

Wether we should increase the _overall timeout_ over 30 minutes or not is something we can talk about - at least these type of events would not fail the builds - but the underlying issue seem to be github action cache.

apostasie:
Rebasing to force CI re-run.

pkg/rootlessutil/xdg_linux.go
```diff
@@ -28,8 +29,11 @@ func XDGRuntimeDir() (string, error) {
...
+	if parentEuid, ok := os.LookupEnv("ROOTLESSKIT_PARENT_EUID"); ok {
+		if _, err := strconv.Atoi(parentEuid); err != nil {
+			return "", errors.New("invalid ROOTLESSKIT_PARENT_EUID environment variable")
```
AkihiroSuda, line 34:
```suggestion
			return "", fmt.Errorf("invalid ROOTLESSKIT_PARENT_EUID environment variable value %q: %w", parentEuid, err)
```

pkg/buildkitutil/buildkitutil_linux.go
```diff
@@ -18,23 +18,24 @@ package buildkitutil
...
 	run := "/run"
 	if rootlessutil.IsRootless() {
 		var err error
 		run, err = rootlessutil.XDGRuntimeDir()
 		if err != nil {
-			log.L.Warn(err)
-			run = fmt.Sprintf("/run/user/%d", rootlessutil.ParentEUID())
+			if rootlessutil.IsRootlessChild() {
+				return "", err
+			}
```
AkihiroSuda, line 36: Warn should be still printed here?

pkg/rootlessutil/xdg_linux.go
apostasie: Sure.

pkg/buildkitutil/buildkitutil_linux.go
apostasie: I don't think it is necessary, since err will be bubbled up and printed out as an error anyhow.

apostasie: eg: if we leave the warn, we will just print the error twice (once as a warn, and then as an error)

pkg/rootlessutil/xdg_linux.go
apostasie: Done.

AkihiroSuda (approved):
Thanks

apostasie:
CI failure looks like #3556

I am not sure why we are suddenly hitting that one a lot more.

Did we recently change something wrt cni?

PR #3853, opened 2025-01-28
Prevent reference filter from failing on dangling images

### LINKED ISSUE #3852: `nerdctl images --filter` is broken with dangling untagged images
### Description

```
nerdctl pull busybox
nerdctl run --name foo -d busybox sleep inf
nerdctl rmi -f busybox
nerdctl images
nerdctl images --filter reference="test*"
```

`FATA[0000] unable to parse image name: : while filtering by reference because of invalid reference format`

The issue is in the way we handle dangling images (`<none>`).

Went back to august to see if it was a recent regression and seems to be alreadt there, if I am not mistaken.

Will do more tests with older versions.

_No response_

na

main

None

### Host information

_No response_

apostasie:
Fix #3852 

It seems likely that the same problem exists in other places.

Solutions:
a. review every code location where we manipulate the result of `imagesservice.List` and account for ":"
b. modify referenceutil to account for it

Thoughts?

~~Also 🍿 on the CI to see if that was a change in containerd 2.0.~~

### DIFF
--- a/pkg/imgutil/filtering.go
+++ b/pkg/imgutil/filtering.go
@@ -325,6 +325,11 @@ func matchesAllLabels(imageCfgLabels map[string]string, filterLabels map[string]
 func matchesReferences(image images.Image, referencePatterns []string) (bool, error) {
 	var matches int
 
+	// Containerd returns ":" for dangling untagged images - see https://github.com/containerd/nerdctl/issues/3852
+	if image.Name == ":" {
+		return false, nil
+	}
+
 	parsedReference, err := referenceutil.Parse(image.Name)
 	if err != nil {
 		return false, err


apostasie:
Also if somebody wants to add a test for that, that would be nice.
I don't have much energy for that right now - just bug-fixing drive-by.

apostasie:
Note: this is definitely affecting rmi as well.

Confusing message reads: `Stderr:   time="2025-01-27T16:21:20-08:00" level=fatal msg="1 errors:\nimage \":\": already exists"`

pkg/imgutil/filtering.go
```diff
@@ -325,6 +325,11 @@ func matchesAllLabels(imageCfgLabels map[string]string, filterLabels map[string]
 func matchesReferences(image images.Image, referencePatterns []string) (bool, error) {
 	var matches int
 
+	// Containerd returns ":" for dangling untagged images
+	if image.Name == ":" {
+		return false, nil
+	}
+
```
fahedouch, line 332: Should we consider fixing the dangling untagged image representation from `:` to `<none>` on the containerd side for better clarity ?

pkg/imgutil/filtering.go
```diff
@@ -325,6 +325,11 @@ func matchesAllLabels(imageCfgLabels map[string]string, filterLabels map[string]
 func matchesReferences(image images.Image, referencePatterns []string) (bool, error) {
 	var matches int
 
+	// Containerd returns ":" for dangling untagged images
```
fahedouch, line 328: Please add the issue link as this may evolve

pkg/imgutil/filtering.go
apostasie: Maybe. Problem is that `<none>` is not a valid image name, so, this has to be carefully done and figuring out everywhere we are manipulating these.

pkg/imgutil/filtering.go
apostasie: Sure.

apostasie: Done.

pkg/imgutil/filtering.go
apostasie: Of course ":" is not a valid image name either...

fahedouch: at least we can stick with docker naming convention which is `<none>`.
let me open an issue ticket on containerd to gather their input."

fahedouch (approved):
Thanks

PR #3860, opened 2025-01-31
Fix apparmor test not running

### LINKED ISSUE #3859: TestRunApparmor does not seem to ever run on the CI
### Description

Rootful:
https://github.com/containerd/nerdctl/actions/runs/13000786746/job/36258878338?pr=3853#step:8:430

Rootless:
https://github.com/containerd/nerdctl/actions/runs/13000786746/job/36258878879?pr=3853#step:9:613

In both cases:

```
=== SKIP: cmd/nerdctl/container TestRunApparmor (0.00s)
    container_run_security_linux_test.go:176: needs to be able to apply "nerdctl-default" profile
```

I don't know why, and this does not make much sense.

Maybe for rootless - although https://github.com/containerd/nerdctl/blob/main/Dockerfile.d/test-integration-rootless.sh#L21 needs a serious review then - as it seems bizarre that a script named "-rootless" would do certain things when `uid = 0`.

For rootful though, we should certainly be able to load the profile and apply it.

This is related to #3858 : our current testing rig is hard to reason about, and more importantly, is hiding issues and non-running tests.

1.
2.
3.

na

main

None

### Host information

_No response_

apostasie:
Fix #3859 

This PR addresses two underlying issues:
- our test rig did not mount securityfs
- the containerd/pkg hostsupport method is not usable here, as it was specifically designed to not work inside a container

### DIFF
--- a/Dockerfile.d/test-integration-rootless.sh
+++ b/Dockerfile.d/test-integration-rootless.sh
@@ -16,6 +16,10 @@
 
 set -eux -o pipefail
 if [[ "$(id -u)" = "0" ]]; then
+  # Ensure securityfs is mounted for apparmor to work
+  if ! mountpoint -q /sys/kernel/security; then
+    mount -tsecurityfs securityfs /sys/kernel/security
+  fi
 	if [ -e /sys/kernel/security/apparmor/profiles ]; then
 		# Load the "nerdctl-default" profile for TestRunApparmor
 		nerdctl apparmor load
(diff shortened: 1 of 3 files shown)


Dockerfile.d/test-integration-rootless.sh
```diff
@@ -16,6 +16,8 @@
 
 set -eux -o pipefail
 if [[ "$(id -u)" = "0" ]]; then
+  # Ensure securityfs is mounted for apparmor to work
+  mount -tsecurityfs securityfs /sys/kernel/security
```
fahedouch, line 20:
```suggestion
if ! mountpoint -q /sys/kernel/security; then
  mount -tsecurityfs securityfs /sys/kernel/security
fi
```

apostasie: Good one. Thanks. Will amend.

apostasie: Done.

fahedouch (approved):
LGTM. Thanks :)

Issue #3851, opened 2025-01-27
Epic: nerdctl namespace issues


apostasie:
### Description

- both `namespace update` and `namespace inspect` with non-existing namespaces (with possibly invalid names) do not error, and return results (test for: " ", "_", "∞", "") 
- if multiple identifiers are passed to `inspect`, it will return a `len(array) > 1` of bogus results
- `namespace update` should enforce the presence of a `--label`, as updating without any does not do anything
- `namespace list` should support a format argument
- `namespace inspect`, `namespace remove` and `namespace update` should have `completion` on namespaces names
- `namespace inspect` format support is questionnable: by default, inspect outputs valid json, but with --format json, it will output a stream of json entries - table and wide are not supported
- `namespace inspect` should at least have the same information available on `namespace list` (inc. number of volumes, containers, labels, etc)

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Should be a good set of "good first issues", and it should be relatively straightforward to fix these.

opjt:
I'm new to this project, but I'd like to give this issue a try. If possible, please assign it to me

nwneisen:
I'm looking through issues for something to start getting into the code.

It looks like 1, 5, and 6 have been fixed?

opjt:
@nwneisen 
1 has been fixed.

PR #3862, opened 2025-02-04
Namespace Improvements

Minterl:
Addresses come concerns in #3851.
The decision not to enforce name validation in namespace.Remove is intentional as it is likely namespaces do exists that violate the validation rules. Not being able to easily delete them would present an issue on its own.


apostasie:
Thanks for this @Minterl 

There is a few issues with your PR - suggesting you run `make lint` to figure it out so we can get that out of the way.

Also, for identifiers validation, `identifiers.ValidateDockerCompat` should be the way to go I think.

Happy to review more in-depth once these two are out of the way.

pkg/cmd/namespace/common.go
```diff
@@ -39,3 +43,19 @@ func labelArgs(labelStrings []string) map[string]string {
 
 	return labels
 }
+
+// Returns an error if name is invalid.
+func validateNamespaceName(name string) error {
```
TinaMor, line 48: @Minterl Could we add test cases for this?

pkg/cmd/namespace/common.go
```diff
@@ -39,3 +43,19 @@ func labelArgs(labelStrings []string) map[string]string {
...
+    for _, c := range(name) {
+        if (
```
TinaMor, line 50: hyphen is allowed but not checked in the if statement.

Unexpected Behavior: "abc-123" (hyphen is neither allowed nor rejected explicitly, meaning it gets rejected)

apostasie:
Hey @Minterl 

Are you interested in carrying this through the finish line?

Issue #3854, opened 2025-01-28
shasum fails on ./share/doc/nerdctl-full/README.md


kkumlien:
### Description

Checksum fails on README file, not a deal-breaker but should be fixed:

```shell
$ mkdir tmp && cd tmp
$ curl -LOs https://github.com/containerd/nerdctl/releases/download/v2.0.3/nerdctl-full-2.0.3-linux-amd64.tar.gz
$ curl -LOs https://github.com/containerd/nerdctl/releases/download/v2.0.3/SHA256SUMS
$ shasum -c SHA256SUMS 2>&1 | grep OK
nerdctl-full-2.0.3-linux-amd64.tar.gz: OK
$ tar Cxzf . nerdctl-full-2.0.3-linux-amd64.tar.gz
$ cp -i share/doc/nerdctl-full/SHA256SUMS .
cp: overwrite './SHA256SUMS'? y
$ shasum -c SHA256SUMS 2>&1 | grep -v OK
./share/doc/nerdctl-full/README.md: FAILED
shasum: WARNING: 1 computed checksum did NOT match
``` 

### Steps to reproduce the issue

1. download nerdctl-full-2.0.3-linux-amd64.tar.gz
2. unpack it
3. run checksum for  ./share/doc/nerdctl-full/README.md

### Describe the results you received and expected

I received "shasum: WARNING: 1 computed checksum did NOT match" and expected all checksums to match.

### What version of nerdctl are you using?

```
Client:
 Version:	v2.0.3
 OS/Arch:	linux/amd64
 Git commit:	ce61fc403417afc22134627449a87a5ed299ac11
 buildctl:
  Version:	v0.19.0
  GitCommit:	3637d1b15a13fc3cdd0c16fcf3be0845ae68f53d
``` 

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

```
$ hostnamectl
...
Operating System: Ubuntu 24.04.1 LTS              
          Kernel: Linux 6.8.0-51-generic
    Architecture: x86-64
...
```

PR #3870, opened 2025-02-07
Fix nerdctl-full shasum fails by move the sha256sum to build-full stage

yankay:
fix: #3854   by move the `sha256sum` to build-full stage.
And the issue is imported by #3583


AkihiroSuda (approved):
Thanks

PR #3887, opened 2025-02-15
Ci minor cleanup

apostasie:
Since we officially only support containerd 2.0, 1.7 and 1.6, and since we do not have test rig to test older containerd versions, the requirement tests for containerd < 1.6 does not seem useful anymore.

Second commit is also getting rid of old/depreacted/unused test helpers (also misplaced under cmd).


manugupt1 (approved):

AkihiroSuda (approved):
Thanks

Issue #3876, opened 2025-02-12
Regression: image inspect --mode native broken


apostasie:
### Description

See https://github.com/containerd/nerdctl/pull/3792#issuecomment-2629221450

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #3888, opened 2025-02-15
Fix FATA error when inspecting images in native mode

weiyuhang2011:
## Description
This PR fixes an issue where a FATA error is always displayed when inspecting images in native mode.

## Steps to reproduce the issue
```
> nerdctl image inspect --mode native abc
<abc image info>
FATA[0000] 1 errors:
no such image: abc
```

Fix: #3876 
Ref: https://github.com/containerd/nerdctl/pull/3792#issuecomment-2629221450

### DIFF
--- a/cmd/nerdctl/image/image_inspect.go
+++ b/cmd/nerdctl/image/image_inspect.go
@@ -17,6 +17,8 @@
 package image
 
 import (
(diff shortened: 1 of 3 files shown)


fahedouch (approved):
Thanks

PR #3890, opened 2025-02-16
CI: test framework, part II

apostasie:
~~This is on top of #3889 (on top of #3591) - should be merged in order.~~

This specific PR does cleanup and reorganize the `testutil/test` package.

Although this PR looks "large", it is not functionally changing anything (and is only touching tests and testing tools anyhow).

The key take-aways here are:
- move the new test module to its final destination (`mod/tigron`)
- split overcrowded module into more intuitive sub-packages:
  - `expect` (expect.Contains, expect.Match, expect.ExitCode)
  - `require` (require.Arm64, etc)
  - `test` (test.Case, etc)
- renamed some methods to more intuitive names (eg: `test.Require` becomes `require.All`)

A `tigron` is a tigon in french :) (the offspring of a tiger and a lion)

The next PR after that will focus on linting code with more stringent standards.


apostasie:
CI failure is #3914 - to be addressed by #3915

AkihiroSuda (approved):
Thanks

PR #3892, opened 2025-02-17
add signal option to container stop command

coderbirju:
This PR adds `signal` flag to the `nerdctl container stop` command

### DIFF
--- a/cmd/nerdctl/container/container_stop.go
+++ b/cmd/nerdctl/container/container_stop.go
@@ -41,6 +41,7 @@ func NewStopCommand() *cobra.Command {
 		SilenceErrors:     true,
 	}
 	stopCommand.Flags().IntP("time", "t", 10, "Seconds to wait before sending a SIGKILL")
+	stopCommand.Flags().StringP("signal", "s", "SIGTERM", "Signal to send to the container")
 	return stopCommand
 }
 
@@ -58,11 +59,20 @@ func processContainerStopOptions(cmd *cobra.Command) (types.ContainerStopOptions
 		t := time.Duration(timeValue) * time.Second
 		timeout = &t
 	}
+	var signal string
+	if cmd.Flags().Changed("signal") {
+		signalValue, err := cmd.Flags().GetString("signal")
+		if err != nil {
+			return types.ContainerStopOptions{}, err
+		}
+		signal = signalValue
+	}
 	return types.ContainerStopOptions{
 		Stdout:   cmd.OutOrStdout(),
 		Stderr:   cmd.ErrOrStderr(),
 		GOptions: globalOptions,
 		Timeout:  timeout,
+		Signal:   signal,
 	}, nil
 }
 
--- a/cmd/nerdctl/container/container_stop_linux_test.go
+++ b/cmd/nerdctl/container/container_stop_linux_test.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/coreos/go-iptables/iptables"
 	"gotest.tools/v3/assert"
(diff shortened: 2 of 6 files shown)


pkg/api/types/container_types.go
```diff
@@ -275,6 +275,9 @@ type ContainerStopOptions struct {
 	// Timeout specifies how long to wait after sending a SIGTERM and before sending a SIGKILL.
 	// If it's nil, the default is 10 seconds.
 	Timeout *time.Duration
+
+	// singal to send to the container, before sending sigkill
```
djdongjin, line 279: nit: sigkill -> `SIGKILL`

pkg/cmd/container/restart.go
```diff
@@ -35,7 +35,7 @@ func Restart(ctx context.Context, client *containerd.Client, containers []string
 			if found.MatchCount > 1 {
 				return fmt.Errorf("multiple IDs found with provided prefix: %s", found.Req)
 			}
-			if err := containerutil.Stop(ctx, found.Container, options.Timeout); err != nil {
+			if err := containerutil.Stop(ctx, found.Container, options.Timeout, "SIGTERM"); err != nil {
```
djdongjin, line 38: do you plan to add the same flag for `restart`? https://docs.docker.com/reference/cli/docker/container/restart/#signal

(not required, and not necessarily done in the same PR)

djdongjin (commented):
LGTM overall. Can we add a test for this? 

One possible test is let a container sleep and stop it with both `timeout` and `SIGKILL` signal, then the container should be stopped before the `timeout`. But feel free to use other test that can test new signal is take effect.

pkg/api/types/container_types.go
swagatbora90: nit typo: singal -> signal

pkg/cmd/container/restart.go
coderbirju: Yes, that will be a follow up PR

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -159,3 +160,39 @@ func TestStopCreated(t *testing.T) {
...
+	// Start a container that sleeps forever
+	base.Cmd("run", "-d", "--name", testContainerName, testutil.CommonImage, "sleep", "Inf").AssertOK()
+
+	// Stop the container with a 30-second timeout and SIGKILL
+	start := time.Now()
+	base.Cmd("stop", "--time=30", "--signal", "SIGKILL", testContainerName).AssertOK()
+	elapsed := time.Since(start)
+
+	// The container should be stopped almost immediately, well before the 30-second timeout
+	assert.Assert(t, elapsed < 30*time.Second, "Container wasn't stopped immediately with SIGKILL")
```
djdongjin, line 179: Let's use a shorter timeout (say 5 sec) just in case so it won't block CI too long even when the test fails.

djdongjin (approved):
LGTM overall, just one minor comment, thanks

swagatbora90 (approved):

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -81,13 +82,20 @@ func TestStopWithStopSignal(t *testing.T) {
...
-while [ $quit -ne 1 ]; do
-    printf 'wait quit'
+trap 'echo "SIGQUIT received"; quit=1' QUIT
+echo "Trap set"
+while true; do
+    if [ $quit -eq 1 ]; then
+        echo "Quitting loop"
+        break
+    fi
+    echo "In loop"
```
djdongjin, line 94: hi @coderbirju why is this change needed?

djdongjin: probably ok since it's just test

coderbirju: Does not change test logic, added a bit more logs and better loop break mechanism

cmd/nerdctl/container/container_stop.go
```diff
@@ -41,6 +41,7 @@ func NewStopCommand() *cobra.Command {
 		SilenceErrors:     true,
 	}
 	stopCommand.Flags().IntP("time", "t", 10, "Seconds to wait before sending a SIGKILL")
+	stopCommand.Flags().StringP("signal", "s", "SIGTERM", "Signal to send to the container")
```
AkihiroSuda, line 44: The docs needs to be updated too https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-stop

cmd/nerdctl/container/container_stop_linux_test.go
apostasie: This is now racy: https://github.com/containerd/nerdctl/actions/runs/13462973052/job/37622397545#step:8:2221

The additional echo is probably exacerbating the condition.
 Suggesting this consumes containers logs and wait to see "trap set" before sending the signal.

apostasie: Moving conversation to #3908

This is impacting a lot of recent builds on the Docker test part.

Issue #3897, opened 2025-02-19
Support bind propagation options in Compose volume long syntax


dancavallaro:
### What is the problem you're trying to solve

I have a use case that requires [configuring propagation options](https://docs.docker.com/engine/storage/bind-mounts/#configure-bind-propagation) on a bind mount defined in a Compose file, but `nerdctl compose` does not currently support bind propagation options.

For example, using this Compose file:

```yaml
services:
  nginx:
    container_name: nginx
    image: nginx:alpine
    volumes:
      - type: bind
        source: ./mnt
        target: /mnt
        read_only: true
        bind:
          propagation: shared
```

then running `sudo nerdctl compose up -d` results in this warning, and the propagation option being ignored:

```
WARN[0000] Ignoring: volume: Bind: [Propagation]
```

`nerdctl run` already supports this, it's just a matter of supporting the additional propagation option when parsing the compose file to generate the `nerdctl run` command. 

### Describe the solution you'd like

I would like `nerdctl compose` to parse propagation options and pass them to the underlying `nerdctl run` command. I've already implemented this with a unit test (plus manual testing): https://github.com/dancavallaro/nerdctl/commit/e764188f054143d6e2c5e8c371c8d91984fb4390. 

Any objection to accepting this contribution if I open a PR? Are there any changes you'd like to see before I do that?

AkihiroSuda:
> Any objection to accepting this contribution if I open a PR?

Nope, a contribution is welcome 👍

dancavallaro:
Thanks -- just opened https://github.com/containerd/nerdctl/pull/3903.

PR #3903, opened 2025-02-20
Support bind propagation options in Compose volume long syntax

dancavallaro:
### Description

Fixes https://github.com/containerd/nerdctl/issues/3897, adding support for bind propagation options in [Compose volume long syntax](https://docs.docker.com/reference/compose-file/services/#long-syntax-5).

### Testing

Added a new unit test exercising the Compose volume long syntax, including bind propagation mode. I also did some manual testing but nothing noteworthy beyond what the unit test already covers.


AkihiroSuda (approved):
Thanks

PR #3909, opened 2025-02-21
Remove transient sysinfo package now that moby has moved away from containerd v1 

apostasie:
During the migration to containerd v2, we forked sysinfo as docker was still depending on containerd v1 at that time.

With their latest 28 release, docker has now moved on, and we can get rid of the fork.

This PR removes the fork and bump docker dep to the required version.

There are also a few minor changes required to make it work:
- move to `runtime.NumCPU` as the sysinfo indirection has been removed upstream
- remove use of deprecated fields


apostasie:
Will unblock #3906

apostasie:
CI failure is #3513 

@AkihiroSuda if you can poke the CI, this is ready-to-go otherwise.

AkihiroSuda (approved):
Thanks

apostasie:
CI failure is #3556

PR #3912, opened 2025-02-22
[PRIORITY] fix: prevent acquire from blanking out hosts file

### LINKED ISSUE #3907: nerdctl run --net=host in rootful mode not working properly (not using `/etc/hosts` file from the host, since v2.0)
### Description

When using nerdctl run with --net=host in rootful mode, it is not using /etc/hosts file from the host, it is overwriting it with the default values. Seeing this behavior only with v2.0.x version of nerdctl.

1.
2.
3.

When using nerdctl run with --net=host in rootful mode , it should use the /etc/hosts file from the host, and don't overwrite with a default values.

nerdctl version : v2.0.3, but seeing the same behavior with v2.0.x
Containerd: 1.7.25

None

### Host information

_No response_

apostasie:
Fix #3907 

See comments for details.

ping @AkihiroSuda

### DIFF
--- a/pkg/dnsutil/hostsstore/hostsstore.go
+++ b/pkg/dnsutil/hostsstore/hostsstore.go
@@ -112,16 +112,21 @@ func (x *hostsStore) Acquire(meta Meta) (err error) {
 			return err
 		}
 
-		if err = os.WriteFile(loc, []byte{}, 0o644); err != nil {
-			return errors.Join(store.ErrSystemFailure, err)
-		}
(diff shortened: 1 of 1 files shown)


apostasie:
We should scrutinize this seriously.

I am not entirely convinced we are doing the right thing here wrt the content of the hosts file.

apostasie:
Marking it [PRIORITY] since this fixes a 2.0 regression.

AkihiroSuda (approved):
Thanks

PR #3917, opened 2025-02-23
Refactor: suggesting we remove the apparmor ls command

apostasie:
- this command very much duplicates what you get from aa-status
- most of its output is just unrelated to nerdctl in any way
- nerdctl does not allow loading any other profile but the default one

-> this command seems useless and unrelated to nerdctl.

Suggesting we just remove it as it does not seem to serve a purpose.

LMK if I am missing something or if there is an actual use-case for it that I did not understand.


AkihiroSuda:
I still prefer to keep `apparmor ls`, as it seems odd to solely have `apparmor inspect` without a command to list the profiles.

Anyway, removal of a non-experimental command may happen only when bumping up the major version, basically.

PR #3918, opened 2025-02-24
CI: rewrite signal tests

### LINKED ISSUE #3908: CI: tests sending signals to be caught by in-container shell scripts are racy
### Description

See recent regression with TestStopWithStopSignal in: https://github.com/containerd/nerdctl/actions/runs/13462973052/job/37622397545
and comment in
https://github.com/containerd/nerdctl/pull/3892#discussion_r1966067544

The fundamental problem seems to be that sending the signal may race the shell script setting up the trap.

Overall, we should consume the container logs first and check that the trap is fully set (greping for "script ready" or some other way for the script to signal the trap is set) before sending the signal.

This applies to tests relying on an in-container trap. Currently:
- `TestRunSigProxy` <- this one is currently somewhat safe as it forces wait 3 seconds (though arbitrary length waits are fragile and not a proper solution)
- `TestStopWithStopSignal`
- probably `TestRestartWithSignal `

1.
2.
3.

na

main

None

### Host information

_No response_

apostasie:
Fix #3908 

Currently, tests involving signals have issues:
- bash sig proxy code is duplicated
- tests are flaky, as they may not wait for the container to start or the trap to be set - and/or use arbitrary sleep waits 
- a couple of these tests only test half of what they should (if the shell script exits when trapping the signal, we fail to test that the container should exit or restart on its own, and not through the script exiting its loop)

This PR does:
- expose the ability to send a signal to the tested command (in test tooling)
- rewrite the SigProxy helper with the new test framework and expanded abilities
- rewrite the tests themselves, fixing the aforementioned shortcomings

Hopefully these tests should stop flaking out (assuming `log` will be reliable in these cases).

Note that this PR likely conflicts with other in-flight PR impacting test tooling (specifically #3889).
Depending on merge order, I will rebase of course.

### DIFF
--- a/cmd/nerdctl/container/container_restart_linux_test.go
+++ b/cmd/nerdctl/container/container_restart_linux_test.go
@@ -18,6 +18,7 @@ package container
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -26,6 +27,7 @@ import (
 
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/test"
 )
 
 func TestRestart(t *testing.T) {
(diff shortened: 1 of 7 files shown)


apostasie:
Rebased.

AkihiroSuda (approved):
Thanks

PR #3919, opened 2025-02-24
CI: remove unbuffer from container tests

apostasie:
This PR rewrites a pair of container tests that were previously using unbuffer, using the new `WithPseudoTTY` test tooling facility.


AkihiroSuda:
This should be removed too?

https://github.com/containerd/nerdctl/blob/46be3a7cad83c4787295047afd4e915b0c421107/Dockerfile#L279

apostasie:
> This should be removed too?
> 
> https://github.com/containerd/nerdctl/blob/46be3a7cad83c4787295047afd4e915b0c421107/Dockerfile#L279

We still have unbuffer in compose tests for now.

AkihiroSuda (approved):
Thanks

PR #3930, opened 2025-02-25
CI: arm64-8core-32gb -> ubuntu-24.04-arm

AkihiroSuda:
GHA now provides ARM runners for free

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
           - runner: ubuntu-24.04
             containerd: v2.0.2
             arch: amd64
-          - runner: arm64-8core-32gb
+          - runner: ubuntu-24.04-arm
             containerd: v2.0.2
             arch: arm64
     env:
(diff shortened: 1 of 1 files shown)


fahedouch (approved):
🚀

apostasie:
New instance is likely much smaller than what we had (timeouting here).

Given the current state of the runners, suggesting we bump the timeout (#3931) until I figure out a way to reduce run time.

fahedouch:
> New instance is likely much smaller than what we had (timeouting here).
> 
> Given the current state of the runners, suggesting we bump the timeout (#3931) until I figure out a way to reduce run time.

`ubuntu-24.04-arm` should perform better than `ubuntu-24.04-amd64` which is successful in the last pipeline 🤔

apostasie:
> > New instance is likely much smaller than what we had (timeouting here).
> > Given the current state of the runners, suggesting we bump the timeout (#3931) until I figure out a way to reduce run time.
> 
> `ubuntu-24.04-arm` should perform better than `ubuntu-24.04-amd64` which is successful in the last pipeline 🤔

`ubuntu-24.04-arm` had a significantly slower dependencies-prep step (+ 3 minutes - possibly because of invalidated github cache), which was enough to bring it over the total 30 minutes timeout.

I am refactoring the Dockerfile right now to deal with the new world order of 2025, where (amongst other things), github instances are smaller and where github cache sucks even more than in 2024 :p

apostasie (approved):

ktock (approved):

PR #3934, opened 2025-02-26
Enforce pkg isolation as part of golangci

apostasie:
Remove standalone hack file and leverage depguard instead.


.golangci.yml
```diff
@@ -135,6 +135,19 @@ linters-settings:
       - unnamedResult
       - unnecessaryBlock
 
+  depguard:
+    rules:
+      # pkg files must not depend on cobra nor anything in cmd
+      pkg:
+        files:
+#          - "$all"
```
AkihiroSuda, line 143: This comment can be removed?

.golangci.yml
```diff
@@ -135,6 +135,19 @@ linters-settings:
...
+          - "**/pkg/**/*.go"
+        deny:
+          - pkg: "github.com/spf13"
```
AkihiroSuda, line 146: This is over-approximation. spf13 provides non-CLI packages too.

.golangci.yml
apostasie: Yep... missed that.

.golangci.yml
apostasie: Addressed.

AkihiroSuda (approved):
Thanks

djdongjin (approved):
LGTM, thanks

PR #3940, opened 2025-02-26
Adding document analyzing CI/dockerfile

apostasie:
Title says all.

Sharing audit notes in a new document (we can also just slap this into an issue if it is better - I don't mind either way).

I am rewriting the Dockerfile now and will provide an updated audit to see if we have measurable improvements.

### DIFF
--- a/docs/dev/auditing_dockerfile.md
+++ b/docs/dev/auditing_dockerfile.md
@@ -1,0 +1,321 @@
+# Auditing dockerfile
+
+Because of the nature of GitHub cache, and the time it takes to build the dockerfile for testing, it is desirable
+to be able to audit what is going on there.
+
+This document provides a few pointers on how to do that, and some results as of 2025-02-26 (run inside lima, nerdctl main,
+on a macbook pro M1).
+
+## Intercept network traffic
+
+### On macOS
+
+Use Charles:
+- start SSL proxying
+- enable SOCKS proxy
+- export the root certificate
+
+### On linux
+
+Left as an exercise to the reader.
+
+### If using lima
+
+- restart your lima instance with `HTTP_PROXY=http://X.Y.Z.W:8888 HTTPS_PROXY=socks5://X.Y.Z.W:8888 limactl start instance` - where XYZW
+is the local ip of the Charles proxy (non-localhost)
+
+### On the host where you are running containerd
+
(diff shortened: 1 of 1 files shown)


docs/dev/auditing_dockerfile.md
```diff
@@ -0,0 +1,222 @@
...
+COPY charles-ssl-proxying-certificate.crt /usr/local/share/ca-certificates/
+RUN update-ca-certificates
+```
+
+Then replace any later "FROM" with our modified bases:
+```
+golang:${GO_VERSION}-bookworm => hack-build-base-debian
+golang:${GO_VERSION}-alpine => hack-build-base
+ubuntu:${UBUNTU_VERSION} => hack-base
+```
```
AkihiroSuda, line 58: ( I once had a BuildKit PR to allow injecting a hook to update CAs, but it wasn't accepted 😞 <https://github.com/moby/buildkit/pull/4669> Still thinking about an alternative... )

apostasie: Yeah.
I always end-up doing some ridiculous monkeying for that kind of stuff (same for overloading apt config, netrc, etc). Deviating the base like here, or using secrets (secrets is neat, but forces you to have additional --mount for every RUN, which is highly unpractical).

I just think the Dockerfile as a front-end is not aging too well... things like Dagger or Earthly are kinda touching on this overall question (eg: of a better frontend for developer that leverages the buildkit DAG).

But then, Dockerfile is everywhere now, so... something "new" comes up with a price...

docs/dev/auditing_dockerfile.md
```diff
@@ -0,0 +1,222 @@
+# Auditing dockerfile
+
+Because of the nature of GitHub cache, and the time it takes to build the dockerfile for testing, it is desirable
+to be able to audit what is going on there.
+
+This document provides a few pointers on how to do that, and some results as of 2025/02/26 (run inside lima, nerdctl main,
```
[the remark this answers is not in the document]
apostasie: Good one 🤓

apostasie:
@AkihiroSuda if you are fine with it, let's merge this.

I would like to have a baseline merged in for the ongoing CI build work.

AkihiroSuda (approved):
Thanks

Issue #3717, opened 2024-12-02
[CI]: failure nerdctl - TestCreate/logs appears to be flaky


apostasie:
### Description

Regularly seeing:

```
=== RUN   TestCreate/logs
    container_create_test.go:68: ======================== Pre-test cleanup ========================
    container_create_test.go:68: ======================== Test setup ========================
    container_create_test.go:68: ======================== Test Run ========================
    command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test logs testcreate-container-13f956b8
    command.go:112: assertion failed: expression is false: strings.Contains(stdout, compare): Output does not contain: "foo"
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test logs testcreate-container-13f956b8
        ExitCode: 0
        Stdout:   
        Stderr:   
        Env:
        HOSTNAME=7a31bdc569d5
        MEMORY_PRESSURE_WRITE=c29tZSAyMDAwMDAgMjAwMDAwMAA=
        SYSTEMD_EXEC_PID=80
        container=docker
        HOME=/root
        LANG=C.UTF-8
        MEMORY_PRESSURE_WATCH=/sys/fs/cgroup/system.slice/docker-entrypoint.service/memory.pressure
        INVOCATION_ID=b617c8bc19b646fab[564](https://github.com/containerd/nerdctl/actions/runs/12128694305/job/33815645478?pr=3716#step:7:565)5ab9fa1d7038
        TERM=xterm
        USER=root
        SHLVL=3
        CGO_ENABLED=0
        _=/usr/local/bin/gotestsum
        PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
        ***
        DOCKER_CONFIG=/tmp/TestCreatelogs930266936/001
        NERDCTL_TOML=/tmp/TestCreatelogs930266936/001/nerdctl.toml
    container_create_test.go:68: ======================== Processing subtests ========================
    case.go:164: ======================== Post-test cleanup ========================
--- FAIL: TestCreate/logs (0.02s)
```

It is unclear if the test is somehow broken, or if we have a deeper issue with logs.

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Possibly a recent regression.
Haven't seen this before.

[...]

Issue #3448, opened 2024-09-21
Issues and requirements for next test-tooling updates


apostasie:
As we rewrite tests, missing pieces and bugs in the tooling will appear.
This is to collect them:
- `mode private` propagation is problematic, as each subtest is in its own isolated namespace - ideally, the subtest would by default be in the parent namespace
- async commands are clunky to use right now - implemented foreground and background, but it would be better to just have a setting on the command that would somehow switch from RunCmd to Start+WaitCmd - design TBD

PR #3941, opened 2025-02-26
feat: add support for userns

Shubhranshu153:
Adds support for running containers with custom user namespace mappings through
the --userns flag in 'run' and 'create' commands. 

Key Features:
- Parse and utilize UID/GID mappings from /etc/subuid and /etc/subgid
- Support multi-entry mappings for user namespace configurations
- Integrate with containerd's remapping capabilities

Technical Details:
- Depends containerd.WithUserNSRemapperLabels for namespace configuration. So containerd mainline is required to compile. It can be run on containerd 1.7 if snapshotter support remap-ids.
- For native snapshotter overlay snapshotter with remapping support is added in containerd 2.x (>2.0.x)

Dependencies:
- Requires containerd with UserNSRemapperLabels support
- Compatible with overlay snapshotter in containerd mainline

Testing:
- Added unit tests for mapping configurations
- Added integration tests for container isolation

This enhancement improves container isolation by providing
flexible user namespace mapping capabilities.

Size: XL

### DIFF
--- a/cmd/nerdctl/builder/builder_build.go
+++ b/cmd/nerdctl/builder/builder_build.go
@@ -25,6 +25,8 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/containerd/log"
+
 	"github.com/containerd/nerdctl/v2/cmd/nerdctl/completion"
 	"github.com/containerd/nerdctl/v2/cmd/nerdctl/helpers"
 	"github.com/containerd/nerdctl/v2/pkg/api/types"
@@ -207,6 +209,13 @@ func processBuildCommandFlag(cmd *cobra.Command, args []string) (types.BuilderBu
 	extendedBuildCtx, err := cmd.Flags().GetStringArray("build-context")
 	if err != nil {
 		return types.BuilderBuildOptions{}, err
+	}
+
+	usernsRemap, err := cmd.Flags().GetString("userns-remap")
+	if err != nil {
+		return types.BuilderBuildOptions{}, err
+	} else if usernsRemap != "" {
+		log.L.Warn("userns remap is not supported with nerdctl build. dropping the config.")
 	}
 
 	return types.BuilderBuildOptions{
--- a/cmd/nerdctl/container/container_create.go
+++ b/cmd/nerdctl/container/container_create.go
@@ -464,6 +464,30 @@ func createOptions(cmd *cobra.Command) (types.ContainerCreateOptions, error) {
 	}
 	// #endregion
 
+	// #region for UserNS
+	opt.UserNS, err = cmd.Flags().GetString("userns-remap")
+	if err != nil {
+		return opt, err
+	}
+
+	userns, err := cmd.Flags().GetString("userns")
+	if err != nil {
+		return opt, err
+	}
+
+	if userns == "host" {
+		opt.UserNS = ""
+	} else if userns != "" {
+		return opt, fmt.Errorf("invalid user mode")
+	}
+
+	if opt.Privileged && opt.UserNS != "" {
+		//userns-remap is not supported with privileged flag.
+		// Ref: https://docs.docker.com/engine/security/userns-remap/
+		return opt, fmt.Errorf("privileged flag cannot be used with userns-remap")
+	}
+	// #endregion
+
 	return opt, nil
 }
 
--- a/cmd/nerdctl/container/container_create_linux_test.go
+++ b/cmd/nerdctl/container/container_create_linux_test.go
@@ -19,15 +19,19 @@ package container
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"syscall"
 	"testing"
 
 	"github.com/opencontainers/go-digest"
 	"gotest.tools/v3/assert"
 
 	"github.com/containerd/containerd/v2/defaults"
+	"github.com/containerd/nerdctl/mod/tigron/expect"
 	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
 
@@ -324,4 +328,185 @@ func TestCreateFromOCIArchive(t *testing.T) {
 	base.Cmd("build", "--tag", tag, fmt.Sprintf("--output=type=oci,dest=%s", tarPath), buildCtx).AssertOK()
 	base.Cmd("create", "--rm", "--name", containerName, fmt.Sprintf("oci-archive://%s", tarPath)).AssertOK()
 	base.Cmd("start", "--attach", containerName).AssertOutContains("test-nerdctl-create-from-oci-archive")
+}
+
+func TestUsernsMappingCreateCmd(t *testing.T) {
+	nerdtest.Setup()
+
+	testCase := &test.Case{
+		Require: require.All(
+			nerdtest.AllowModifyUserns,
+			nerdtest.RemapIDs,
+			require.Not(nerdtest.Docker)),
+		NoParallel: true,
+		Setup: func(data test.Data, helpers test.Helpers) {
+			data.Labels().Set("validUserns", "nerdctltestuser")
+			data.Labels().Set("expectedHostUID", "123456789")
+			data.Labels().Set("invalidUserns", "invaliduser")
+		},
+		SubTests: []*test.Case{
+			{
+				Description: "Test container create with valid Userns",
+				NoParallel:  true, // Changes system config so running in non parallel mode
+				Setup: func(data test.Data, helpers test.Helpers) {
+					err := appendUsernsConfig(data.Labels().Get("validUserns"), data.Labels().Get("expectedHostUID"), helpers)
+					assert.NilError(t, err, "Failed to append Userns config")
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					removeUsernsConfig(t, data.Labels().Get("validUserns"), helpers)
+					helpers.Anyhow("rm", "-f", data.Identifier())
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					helpers.Ensure("create", "--tty", "--userns-remap", data.Labels().Get("validUserns"), "--name", data.Identifier(), testutil.NginxAlpineImage)
+					return helpers.Command("start", data.Identifier())
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						ExitCode: 0,
+						Output: func(stdout string, info string, t *testing.T) {
+							actualHostUID, err := getContainerHostUID(helpers, data.Identifier())
+							assert.NilError(t, err, "Failed to get container host UID")
+							assert.Assert(t, actualHostUID == data.Labels().Get("expectedHostUID"), info)
+						},
+					}
+				},
+			},
+			{
+				Description: "Test container create failure with valid Userns and privileged flag",
+				NoParallel:  true, // Changes system config so running in non parallel mode
+				Setup: func(data test.Data, helpers test.Helpers) {
+					err := appendUsernsConfig(data.Labels().Get("validUserns"), data.Labels().Get("expectedHostUID"), helpers)
+					assert.NilError(t, err, "Failed to append Userns config")
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					removeUsernsConfig(t, data.Labels().Get("validUserns"), helpers)
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("create", "--tty", "--privileged", "--userns-remap", data.Labels().Get("validUserns"), "--name", data.Identifier(), testutil.NginxAlpineImage)
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						ExitCode: 1,
+					}
+				},
+			},
+			{
+				Description: "Test container create with invalid Userns",
+				NoParallel:  true, // Changes system config so running in non parallel mode
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					helpers.Anyhow("rm", "-f", data.Identifier())
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("create", "--tty", "--userns-remap", data.Labels().Get("invalidUserns"), "--name", data.Identifier(), testutil.NginxAlpineImage)
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						ExitCode: 1,
+					}
+				},
+			},
+		},
+	}
+	testCase.Run(t)
+}
+
+func getContainerHostUID(helpers test.Helpers, containerName string) (string, error) {
+	result := helpers.Capture("inspect", "--format", "{{.State.Pid}}", containerName)
+	pidStr := strings.TrimSpace(result)
+	pid, err := strconv.Atoi(pidStr)
+	if err != nil {
+		return "", fmt.Errorf("invalid PID: %v", err)
+	}
+
+	stat, err := os.Stat(fmt.Sprintf("/proc/%d", pid))
+	if err != nil {
+		return "", fmt.Errorf("failed to stat process: %v", err)
+	}
+
+	uid := int(stat.Sys().(*syscall.Stat_t).Uid)
+	return strconv.Itoa(uid), nil
+}
+
+func appendUsernsConfig(userns string, hostUID string, helpers test.Helpers) error {
+	addUser(userns, hostUID, helpers)
+	entry := fmt.Sprintf("%s:%s:65536\n", userns, hostUID)
+	tempDir := helpers.T().TempDir()
+	files := []string{"subuid", "subgid"}
+	for _, file := range files {
+
+		fileBak := filepath.Join(tempDir, file)
+		defer os.Remove(fileBak)
+		d, err := os.Create(fileBak)
+		if err != nil {
+			return fmt.Errorf("failed to create %s: %w", fileBak, err)
+		}
+
+		s, err := os.Open(filepath.Join("/etc", file))
+		if err != nil {
+			return fmt.Errorf("failed to open %s: %w", file, err)
+		}
+		defer s.Close()
+
+		_, err = io.Copy(d, s)
+		if err != nil {
+			return fmt.Errorf("failed to copy %s to %s: %w", file, fileBak, err)
+		}
+
+		f, err := os.OpenFile(fmt.Sprintf("/etc/%s", file), os.O_APPEND|os.O_WRONLY, 0644)
+		if err != nil {
+			return fmt.Errorf("failed to open %s: %w", file, err)
+		}
+		defer f.Close()
+
+		if _, err := f.WriteString(entry); err != nil {
+			return fmt.Errorf("failed to write to %s: %w", file, err)
+		}
+	}
+	return nil
+}
+
+func addUser(username string, hostID string, helpers test.Helpers) {
+	helpers.Custom("groupadd", "-g", hostID, username).Run(&test.Expected{
+		ExitCode: 0})
+	helpers.Custom("useradd", "-u", hostID, "-g", hostID, "-s", "/bin/false", username).Run(&test.Expected{
+		ExitCode: 0})
+}
+
+func removeUsernsConfig(t *testing.T, userns string, helpers test.Helpers) {
+	delUser(userns, helpers)
+	delGroup(userns, helpers)
+	tempDir := helpers.T().TempDir()
+	files := []string{"subuid", "subgid"}
+	for _, file := range files {
+		fileBak := filepath.Join(tempDir, file)
+		s, err := os.Open(fileBak)
+		if err != nil {
+			t.Logf("failed to open %s, Error: %s", fileBak, err)
+			continue
+		}
+		defer s.Close()
+
+		d, err := os.Open(filepath.Join("/etc/%s", file))
+		if err != nil {
+			t.Logf("failed to open %s, Error: %s", file, err)
+			continue
+
+		}
+		defer d.Close()
+
+		_, err = io.Copy(d, s)
+		if err != nil {
+			t.Logf("failed to restore. Copy %s to %s failed, Error %s", fileBak, file, err)
+			continue
+		}
+
+	}
+}
+
+func delUser(username string, helpers test.Helpers) {
+	helpers.Custom("userdel", username).Run(&test.Expected{ExitCode: expect.ExitCodeNoCheck})
+}
+
+func delGroup(groupname string, helpers test.Helpers) {
+	helpers.Custom("groupdel", groupname).Run(&test.Expected{ExitCode: expect.ExitCodeNoCheck})
 }
--- a/cmd/nerdctl/container/container_run.go
+++ b/cmd/nerdctl/container/container_run.go
@@ -288,7 +288,6 @@ func setCreateFlags(cmd *cobra.Command) {
 	// #endregion
 
 	cmd.Flags().String("ipfs-address", "", "multiaddr of IPFS API (default uses $IPFS_PATH env variable if defined or local directory ~/.ipfs)")
-
 	cmd.Flags().String("isolation", "default", "Specify isolation technology for container. On Linux the only valid value is default. Windows options are host, process and hyperv with process isolation as the default")
 	cmd.RegisterFlagCompletionFunc("isolation", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		if runtime.GOOS == "windows" {
@@ -296,6 +295,7 @@ func setCreateFlags(cmd *cobra.Command) {
 		}
 		return []string{"default"}, cobra.ShellCompDirectiveNoFileComp
 	})
+	cmd.Flags().String("userns", "", "Specify host to disable userns-remap")
 
 }
 
--- a/cmd/nerdctl/container/container_run_user_linux_test.go
+++ b/cmd/nerdctl/container/container_run_user_linux_test.go
@@ -17,9 +17,16 @@
 package container
 
 import (
+	"fmt"
 	"testing"
 
+	"gotest.tools/v3/assert"
+
+	"github.com/containerd/nerdctl/mod/tigron/require"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 )
 
 func TestRunUserGID(t *testing.T) {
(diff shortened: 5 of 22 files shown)


cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -324,3 +328,160 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+		defer f.Close()
+
+		if _, err := f.WriteString(entry); err != nil {
+			return fmt.Errorf("failed to write to %s: %w", file, err)
+		}
+	}
+	return nil
+}
+
+func addUser(username string, hostId string) error {
```
AkihiroSuda, line 443: This is scary to run by default. Needs to have an opt-in flag like `-test.allow-add-user`

Similar:
https://github.com/containerd/nerdctl/blob/c12aaa339989698d89d8b6824f0a9094db4fc5fe/pkg/testutil/testutil.go#L557

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -324,3 +328,160 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+	files := []string{"/etc/subuid", "/etc/subgid"}
+	for _, file := range files {
+		content, err := os.ReadFile(file)
+		if err != nil {
+			t.Logf("Failed to read %s: %v", file, err)
+			continue
+		}
+
+		newContent := strings.ReplaceAll(string(content), entry, "")
+		if err := os.WriteFile(file, []byte(newContent), 0644); err != nil {
```
AkihiroSuda, line 474: This is scary too. Needs to be opt-in.

Also, the content of `/etc/subuid` and `/etc/subgid` has to be backed up and has to be restored with `defer()`

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -324,3 +328,160 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+			t.Logf("Failed to write to %s: %v", file, err)
+		}
+	}
+}
+
+func delUser(username string) error {
+	cmd := exec.Command("sudo", "userdel", username)
```
AkihiroSuda, line 481: sudo shouldn't be needed when the test is running as the root

go.mod
```diff
@@ -10,13 +10,13 @@ go 1.22.7
...
-	github.com/compose-spec/compose-go/v2 v2.4.8
-	github.com/containerd/accelerated-container-image v1.3.0
+	github.com/Microsoft/hcsshim v0.13.0-rc.3
+	github.com/compose-spec/compose-go/v2 v2.4.7
+	github.com/containerd/accelerated-container-image v1.2.3
 	github.com/containerd/cgroups/v3 v3.0.5
 	github.com/containerd/console v1.0.4
 	github.com/containerd/containerd/api v1.8.0
-	github.com/containerd/containerd/v2 v2.0.2
+	github.com/containerd/containerd/v2 v2.0.1-0.20250211161307-525332b29211
```
AkihiroSuda, line 19: Would it be possible to cherry-pick the relevant commit to v2.0?

pkg/api/types/container_types.go
```diff
@@ -264,6 +264,9 @@ type ContainerCreateOptions struct {
 
 	// ImagePullOpt specifies image pull options which holds the ImageVerifyOptions for verifying the image.
 	ImagePullOpt ImagePullOptions
+
+	// Userns name for user namespace mapping of container
+	Userns string
```
AkihiroSuda, line 269:
```suggestion
	UserNS string
```

Because `Userns` looks like `Users`

pkg/cmd/container/create.go
```diff
@@ -185,6 +185,27 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
 	opts = append(opts, rootfsOpts...)
 	cOpts = append(cOpts, rootfsCOpts...)
 
+	if options.Userns != "" {
+		if runtime.GOOS != "linux" || rootlessutil.IsRootless() {
+			return nil, generateRemoveStateDirFunc(ctx, id, internalLabels), errors.New("userns is only supported on linux")
```
AkihiroSuda, line 190: The error is confusing when running in rootless mode on linux

cmd/nerdctl/container/container_run.go
```diff
@@ -279,6 +279,7 @@ func setCreateFlags(cmd *cobra.Command) {
 	cmd.Flags().String("ipfs-address", "", "multiaddr of IPFS API (default uses $IPFS_PATH env variable if defined or local directory ~/.ipfs)")
 
 	cmd.Flags().String("isolation", "default", "Specify isolation technology for container. On Linux the only valid value is default. Windows options are host, process and hyperv with process isolation as the default")
+	cmd.Flags().String("userns", "", "Support idmapping of containers. This options is only supported on linux. If `host` is passed, no idmapping is done. if a user name is passed, it does idmapping based on the uidmap and gidmap ranges specified in /etc/subuid and /etc/subgid respectively")
```
AkihiroSuda, line 282: Please add `docs/command-reference.md`.

Also, the command line seems incompatible with Docker?
Docker doesn't accept a username here, and the name is hardcoded to "dockremap".
Maybe we should have its equivalent as "nerdremap"?

AkihiroSuda: Podman accepts `--subuidname string --subgidname string` to specify a custom user name

pkg/api/types/container_types.go
Shubhranshu153: Ack

pkg/cmd/container/create.go
Shubhranshu153: Ack

go.mod
Shubhranshu153: may not be the commit entirely but only the relevant functions as it is client side.

cmd/nerdctl/container/container_run.go
Shubhranshu153: Also, the command line seems incompatible with Docker?
if they add `host` it will behave as docker as we check for that string and create the default snapshot. 

For other names it behaves as docker daemon but at a container level rather than at daemon level. Will you suggest we configure it in nerdctl config instead?

cmd/nerdctl/container/container_create_linux_test.go
Shubhranshu153: ack

cmd/nerdctl/container/container_run.go
AkihiroSuda: > For other names it behaves as docker daemon but at a container level rather than at daemon level. Will you suggest we configure it in nerdctl config instead?

Eventually, the both level should be supported, as in Podman: https://github.com/containers/podman/blob/v5.4.1/docs/source/markdown/options/userns.container.md?plain=1

- `podman run --userns=auto` allocates subuids from the "containers" entry in `/etc/subuid`. 
- When `userns=...` is specified in `containers.conf`, Podman enables UserNS globally, unless `--userns=host` is specified.

nerdctl should probably follow the same convention, but s/containers.conf/nerdctl.toml/

Not all the features need to be implemented at once. Can just begin with the easiest one.

pkg/testutil/testutil.go
```diff
@@ -555,6 +556,7 @@ var (
 func M(m *testing.M) {
 	flag.StringVar(&flagTestTarget, "test.target", Nerdctl, "target to test")
 	flag.BoolVar(&flagTestKillDaemon, "test.allow-kill-daemon", false, "enable tests that kill the daemon")
+	flag.BoolVar(&flagTestUserNS, "test.allow-modify-user", false, "enable tests on userns")
```
AkihiroSuda, line 559:
```suggestion
	flag.BoolVar(&flagTestModifyUsser, "test.allow-modify-user", false, "enable tests that creates/deletes user accounts on the host")
```

pkg/testutil/nerdtest/requirements.go
```diff
@@ -360,3 +363,33 @@ var Private = &test.Requirement{
 		}
 	},
 }
+
+var ContainerdV1 = &test.Requirement{
```
AkihiroSuda, line 367: Better to use the introspection API to check the existence of the relevant plugin (which one?)
https://github.com/containerd/nerdctl/blob/39058fbce68e29feecc9cec992ade74b272817d8/pkg/cmd/container/run_restart.go#L36

go.mod
```diff
@@ -5,13 +5,13 @@ go 1.23.0
...
 	github.com/Microsoft/go-winio v0.6.2
-	github.com/Microsoft/hcsshim v0.12.9
+	github.com/Microsoft/hcsshim v0.13.0-rc.3
 	github.com/compose-spec/compose-go/v2 v2.4.9
 	github.com/containerd/accelerated-container-image v1.3.0
 	github.com/containerd/cgroups/v3 v3.0.5
 	github.com/containerd/console v1.0.4
 	github.com/containerd/containerd/api v1.8.0
-	github.com/containerd/containerd/v2 v2.0.4
+	github.com/containerd/containerd/v2 v2.0.1-0.20250211161307-525332b29211
```
AkihiroSuda, line 14:
```suggestion
	github.com/containerd/containerd/v2 v2.1.0-beta.0
```

cmd/nerdctl/container/container_run.go
AkihiroSuda: ping @Shubhranshu153

Shubhranshu153: havent been focusing on this past few weeks, Will focus on it to completion this week.

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,194 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+			require.Not(nerdtest.Docker)),
+		SubTests: []*test.Case{
+			{
+				Description: "Test container start with valid Userns",
+				NoParallel:  true, // Changes system config so running in non parallel mode
+				Setup: func(data test.Data, helpers test.Helpers) {
+					data.Labels().Set("validUserns", "nerdctltestuser")
+					data.Labels().Set("expectedHostUID", "123456789")
+					if err := appendUsernsConfig(data.Labels().Get("validUserns"), data.Labels().Get("expectedHostUID")); err != nil {
+						t.Fatalf("Failed to append Userns config: %v", err)
```
apostasie, line 348: Maybe
```
err := appendUsernsConfig(data.Labels().Get("validUserns"), data.Labels().Get("expectedHostUID"))
assert.NilError(t, err, "Failed to append Userns config")
```

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,194 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					helpers.Ensure("create", "--tty", "--userns", data.Labels().Get("validUserns"), "--name", data.Identifier(), testutil.NginxAlpineImage)
+					return helpers.Command("start", data.Identifier())
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						ExitCode: 0,
+						Output: func(stdout string, info string, t *testing.T) {
+							actualHostUID, err := getContainerHostUID(helpers, data.Identifier())
+							if err != nil {
```
apostasie, line 364: Ibid (assert.NilErrror instead)

pkg/cmd/container/create_userns_opts_linux_test.go
```diff
@@ -0,0 +1,154 @@
...
+			netOpts: types.NetworkOptions{
+				NetworkSlice: []string{"container:"},
+			},
+			expected:    "",
+			expectError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
apostasie, line 143: Nit: shadowing `t` makes reading a bit more difficult.
Suggesting to change it to `subT` (or whatever you want).
Similarly, maybe `tt` could have a more descriptive name? `testCase`?

apostasie:
@Shubhranshu153 I need to do some reading on userns to get a better grasp first, but short term, happy to help you review testing / overall. Hit me up when ready.

cmd/nerdctl/container/container_create_linux_test.go
Shubhranshu153: i need to read another pass, as i was not working on it i have lost touch on it.

feat-userns.patch
AkihiroSuda, line 1: Seems added by mistake

Shubhranshu153:
@AkihiroSuda 
Do we need to run the test against v1.6.38 and it seems we are not running against 1.7.27 containerd.
Was wondering what we need to run against, should we run it against both these versions or update the 1.6.38 version to 1.7.27?

apostasie:
> @AkihiroSuda Do we need to run the test against v1.6.38 and it seems we are not running against 1.7.27 containerd.

We did remove testing for 1.7 recently to keep the testing matrix in check.
The reasoning is that we test the oldest supported version (1.6) and the latest released version (2.0).

Are you facing specific issues with 1.6 that 1.7 would have addressed?

Shubhranshu153:
@apostasie 
userns may not work with 1.6.38 as this call  client.GetSnapshotterCapabilities(ctx, snapshotterName) doesnt return the remap-label capabilities but i think it should work with containerd 1.7 as we expose the snapshotter capabilities via that function. I dont have any other containerd client calls so i believe rest all will be handled as long as we compile nerdctl with containerd 2.1 library. 
I can add back the condition for containerd version.

pkg/testutil/nerdtest/requirements.go
Shubhranshu153: its not a plugin, the client call we make is to get snapshotter capability, i can modify the requirement to check that and skip rather than doing it like this.

pkg/cmd/container/create_userns_opts_linux_test.go
Shubhranshu153: changed tt to testCase
Not sure the need for t to change, i feel t is more referring to the test struct.

apostasie:
> @apostasie userns may not work with 1.6.38 as this call client.GetSnapshotterCapabilities(ctx, snapshotterName) doesnt return the remap-label capabilities but i think it should work with containerd 1.7 as we expose the snapshotter capabilities via that function. I dont have any other containerd client calls so i believe rest all will be handled as long as we compile nerdctl with containerd 2.1 library. I can add back the condition for containerd version.

I see.
Thanks for the clarification.

My (personal opinion) preference would be to keep things as-is testing-matrix-wise (out of concern for size of testing matrix and correlation with increased propensity to flake out).

Let see what @AkihiroSuda think about this ^?

Quick question: are there scenarios where other versions (or specific configurations) of containerd would not support remap-ids?

If yes, then instead of testing against the version, we could test the feature - eg: add a new `test.Requirement`.

Here is how it is done for cgroups for example:

https://github.com/containerd/nerdctl/blob/main/pkg/testutil/nerdtest/requirements.go#L193-L213

AkihiroSuda:
> My (personal opinion) preference would be to keep things as-is testing-matrix-wise (out of concern for size of testing matrix and correlation with increased propensity to flake out).


SGTM.
In July, we will bump up the minimum containerd version to v1.7: https://github.com/containerd/containerd/blob/main/RELEASES.md#current-state-of-containerd-releases

cmd/nerdctl/container/container_run.go
AkihiroSuda: @Shubhranshu153 Following `podman run --userns=auto` sounds good to you?

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,194 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+	for _, file := range files {
+		fileBak := fmt.Sprintf("%s/%s.bak", tempDir, file)
+		s, err := os.Open(fileBak)
+		if err != nil {
+			t.Logf("failed to open %s, Error: %s", fileBak, err)
+			continue
+		}
+		defer s.Close()
+
+		d, err := os.Open(fmt.Sprintf("/etc/%s", file))
```
AkihiroSuda, line 489: `filepath.Join` should be used for path operations

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,194 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+	output, err = cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("useradd failed: %s, %w", string(output), err)
+	}
+	return nil
+}
+
+func removeUsernsConfig(t *testing.T, userns string) {
+	if err := delUser(userns); err != nil {
+		t.Logf("failed to del user %s, Error: %s", userns, err)
```
AkihiroSuda, line 471: t.Fatalf

Shubhranshu153: removeUsernsConfig is a cleanup function do we want to do a fatal error here and not handle the rest of the cleanup task

Shubhranshu153:
FreeBSD one is probably flaky
Checked userns test were run agianst rootful containerd with snapshotters supporting remap-id
```
PASS cmd/nerdctl/container.TestUsernsMappingCreateCmd/Test_container_start_with_valid_Userns (0.47s)
PASS cmd/nerdctl/container.TestUsernsMappingCreateCmd/Test_container_start_with_invalid_Userns (0.05s)
PASS cmd/nerdctl/container.TestUsernsMappingCreateCmd (0.53s)
```

cmd/nerdctl/container/container_run.go
Shubhranshu153: It does sound good to me in terms of supporting all the things that podman supports but would like to create a separate PR as this one has already been quite long. Can probably get in by 2.1.

apostasie:
> FreeBSD one is probably flaky

Yeah, I think it is broken (again) actually. Something about csv packages conflicting versions.

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,192 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+		defer f.Close()
+
+		if _, err := f.WriteString(entry); err != nil {
+			return fmt.Errorf("failed to write to %s: %w", file, err)
+		}
+	}
+	return nil
+}
+
+func addUser(username string, hostID string) error {
```
apostasie, line 453: If possible, I would prefer we use facilities from the new testing system.

eg: pass along `test.Helpers` to your functions and:

```golang

func addUser(helpers test.Helpers, username, hostID string) error {
  helpers.Custom(blablab)...
}

```

This is a nit - it is ok as-is - but using the test tooling does provide you for free with extended debugging information about any command (pwd, environment, exit code, timeout, signals, stdout, stderr), which makes it easier to reproduce / debug.

hack/test-integration.sh
```diff
@@ -39,17 +39,22 @@ if [ "$#" == 0 ]; then
   exit
 fi
 
+allowUserns=""
```
apostasie, line 42: I don't understand this code.

You are iterating over "$@", then checking if it is equal to `-test.allow-modify-users=true` || `-test.allow-modify-users=true` (duplicated?), then if it is, you add it to the args of the command, which are already including $@? 

Furthermore, you are breaking out of the loop, which will prevent proper processing of `-test.only-flaky` if it appears later?

Maybe I am misreading or misunderstanding something?

pkg/portutil/portutil_test.go
```diff
@@ -30,7 +30,7 @@ import (
 
 func TestTestParseFlagPWithPlatformSpec(t *testing.T) {
 	if runtime.GOOS != "Linux" || rootlessutil.IsRootless() {
-		t.Skip("no non-Linux platform or rootless mode in Linux are not supported yet")
+		t.Skip("UserNS mode innon-Linux platform or rootless in Linux is supported")
```
apostasie, line 33: `UserNS mode in non-Linux`?
`is not supported`?

.github/workflows/test.yml
```diff
@@ -140,7 +140,7 @@ jobs:
           docker run --privileged --rm tonistiigi/binfmt --install linux/arm64
           docker run --privileged --rm tonistiigi/binfmt --install linux/arm/v7
       - name: "Run integration tests"
-        run: docker run -t --rm --privileged test-integration ./hack/test-integration.sh -test.only-flaky=false
+        run: docker run -t --rm --privileged test-integration ./hack/test-integration.sh -test.only-flaky=false -test.allow-modify-users=true
```
apostasie, line 143: I do not quite understand why we would need this at all?
If it works only for rootful, then let's just add a rootful verification to the `Requirement`?

hack/test-integration.sh
apostasie: See other note though. Maybe we just do not need `allow-modify-users` at all.

pkg/portutil/portutil_test.go
Shubhranshu153: user namespace afaik is a linux construct.

cmd/nerdctl/container/container_run.go
AkihiroSuda: No need to support all the things,  but `--userns=<USERNAME>` is a new syntax that is not adopted by either Docker nor Podman, and is hard to adopt here unless there is a strong reason

pkg/portutil/portutil_test.go
apostasie: Sure.
Sorry. What I meant is that you have typos:
-  `UserNS mode innon` -> `UserNS mode in non-Linux`
- `is supported` -> `is not supported`

cmd/nerdctl/container/container_run.go
Shubhranshu153: dockerd --userns-remap="testuser:testuser"
 
For docker it is a docker daemon entry although i have not added all the different option and patterns to support. Should i add userns-remap instead of userns?

hack/test-integration.sh
Shubhranshu153: Replied below for the option
and fixed the logic

hack/test-integration.sh
```diff
@@ -39,17 +39,22 @@ if [ "$#" == 0 ]; then
...
 for arg in "$@"; do
   if [ "$arg" == "-test.only-flaky=true" ] || [ "$arg" == "-test.only-flaky" ]; then
     args+=("--rerun-fails=$retries")
-    break
+    continue
+  fi
+  if [ "$arg" == "-test.allow-modify-users=true" ] || [ "$arg" == "-test.allow-modify-users" ]; then
+    allowUserns="-test.allow-modify-users=true"
```
apostasie, line 49: This still duplicates the argument fed to gotestum, right?

cmd/nerdctl/container/container_run.go
AkihiroSuda: `userns-remap` SGTM.

It should be also available in `/etc/nerdctl/nerdctl.toml`
https://github.com/containerd/nerdctl/blob/main/docs/config.md#properties

Shubhranshu153: sure i can add to the nerdctl.toml

hack/test-integration.sh
Shubhranshu153: ah i see. Thanks it should not be needeed to be added here at all

docs/command-reference.md
```diff
@@ -235,6 +235,7 @@ User flags:
 - :nerd_face: `--umask`: Set the umask inside the container. Defaults to 0022.
   Corresponds to Podman CLI.
 - :whale: `--group-add`: Add additional groups to join
+- :nerd_face: `--userns`: Support idmapping of containers. This options is only supported on rootful linux. If `host` is passed, no idmapping is done. if a user name is passed, it does idmapping based on the uidmap and gidmap ranges specified in /etc/subuid and /etc/subgid respectively.
```
AkihiroSuda, line 238: It should be documented that this does not apply to `nerdctl build` yet

docs/command-reference.md
```diff
@@ -235,6 +235,7 @@ User flags:
 - :nerd_face: `--umask`: Set the umask inside the container. Defaults to 0022.
   Corresponds to Podman CLI.
 - :whale: `--group-add`: Add additional groups to join
+- :nerd_face: `--userns-remap=<username>:<groupname>`: Support idmapping of containers. This options is only supported on rootful linux for container create and run. If `host` is passed, no idmapping is done. if a user name is passed, it does idmapping based on the uidmap and gidmap ranges specified in /etc/subuid and /etc/subgid respectively. (format: <name|uid>[:<group|gid>])
```
AkihiroSuda, line 238: Probably no need to support "host" here (unless Docker supports it)

AkihiroSuda: It should be also clarified that this does not apply to `nerdctl build` yet

docs/command-reference.md
```diff
@@ -422,7 +423,7 @@ IPFS flags:
 
 Unimplemented `docker run` flags:
     `--device-cgroup-rule`, `--disable-content-trust`, `--expose`, `--health-*`, `--isolation`, `--no-healthcheck`,
-    `--link*`, `--publish-all`, `--storage-opt`, `--userns`, `--volume-driver`
+    `--link*`, `--publish-all`, `--storage-opt`, `--volume-driver`
```
AkihiroSuda, line 426: `--userns=host` seems missing in the current revision

docs/command-reference.md
AkihiroSuda: (`--userns=host` still makes a sense, when `userns_remap` is configured in the toml)

docs/command-reference.md
Shubhranshu153: Added it to override --userns-remap

docs/command-reference.md
Shubhranshu153: Added text and a warning in nerdctl build if userns-remap is set during builds

docs/command-reference.md
Shubhranshu153: Addressed this, thanks

Shubhranshu153:
Surprised why the rootfs test passed earlier the logic was definitely wrong, fixed it.

Shubhranshu153:
Rebased to resolve conflict.

@AkihiroSuda @apostasie  Let me know if you need anything further. 
Thanks.

apostasie:
Let me fix tigron CI.

apostasie:
#4173 is fixing CI.

mod/tigron/test/command.go
```diff
@@ -301,6 +301,10 @@ func (gc *GenericCommand) Run(expect *Expected) {
 	}
 }
 
+func (gc *GenericCommand) RunAnyhow() {
```
apostasie, line 304: I would rather keep the numbers of helpers strictly in check, and not add more for simple / rarely used operations...

apostasie (changes requested):

mod/tigron/test/command.go
Shubhranshu153: Do we have a better option in the helpers to do this ? If yes i can move, if no and we want the helpers to handle all invocations i dont see ways around it.

apostasie: I guess just:

```golang
func delUser(username string, helpers test.Helpers) error {
	helpers.Custom("userdel", username).Run(&test.Expected{ExitCode: expect.ExitCodeNoCheck})
	return nil
}
```

cmd/nerdctl/builder/builder_build.go
```diff
@@ -209,6 +211,13 @@ func processBuildCommandFlag(cmd *cobra.Command, args []string) (types.BuilderBu
 		return types.BuilderBuildOptions{}, err
 	}
 
+	usernsRemap, err := cmd.Flags().GetString("userns-remap")
+	if err != nil {
+		return types.BuilderBuildOptions{}, err
+	} else if usernsRemap != "" {
+		log.L.Warn("WARNING! userns remap is not supported with nerdctl build. dropping the config.")
```
AkihiroSuda, line 218: This prints WARN(ING) twice

cmd/nerdctl/container/container_create.go
```diff
@@ -454,6 +454,30 @@ func createOptions(cmd *cobra.Command) (types.ContainerCreateOptions, error) {
...
+	userns, err := cmd.Flags().GetString("userns")
+	if err != nil {
+		return opt, err
+	}
+
+	if userns == "host" {
+		opt.UserNS = ""
+	} else if userns != "" {
+		return opt, fmt.Errorf("invalid user mode")
+	}
```
AkihiroSuda, line 472: nit: `switch` can be used

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Needs rebase

go.mod
```diff
@@ -5,13 +5,13 @@ go 1.23.0
...
 	github.com/Microsoft/go-winio v0.6.2
-	github.com/Microsoft/hcsshim v0.12.9
+	github.com/Microsoft/hcsshim v0.13.0-rc.3
 	github.com/compose-spec/compose-go/v2 v2.6.1
 	github.com/containerd/accelerated-container-image v1.3.0
 	github.com/containerd/cgroups/v3 v3.0.5
 	github.com/containerd/console v1.0.4
 	github.com/containerd/containerd/api v1.8.0
-	github.com/containerd/containerd/v2 v2.0.5
+	github.com/containerd/containerd/v2 v2.1.0-beta.1
```
AkihiroSuda, line 14: https://github.com/containerd/containerd/releases/tag/v2.1.0-rc.0

cmd/nerdctl/builder/builder_build.go
Shubhranshu153: wanted to be visible without disallowing, may be there is some better way to do it dont know

go.mod
Shubhranshu153: the schemaV1 support was dropped i will create a separate PR for this migration so that we review the change more closely. Dont want to touch unrelated code here as it might get overlooked

Shubhranshu153: PR for containerd v2 2.1.0 Migration: https://github.com/containerd/nerdctl/pull/3941

cmd/nerdctl/builder/builder_build.go
apostasie: @Shubhranshu153 I guess @AkihiroSuda is suggesting:
`log.L.Warn("userns remap is not supported with nerdctl build. dropping the config.")`

Is the extra word `WARNING!` in the message itself necessary?

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,197 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+		defer f.Close()
+
+		if _, err := f.WriteString(entry); err != nil {
+			return fmt.Errorf("failed to write to %s: %w", file, err)
+		}
+	}
+	return nil
+}
+
+func addUser(username string, hostID string, helpers test.Helpers) error {
```
apostasie, line 472: I guess we can just remove the return `error`?

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,197 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+		_, err = io.Copy(d, s)
+		if err != nil {
+			t.Logf("failed to restore. Copy %s to %s failed, Error %s", fileBak, file, err)
+			continue
+		}
+
+	}
+}
+
+func delUser(username string, helpers test.Helpers) error {
```
apostasie, line 517: Same as above - no need for the return?

cmd/nerdctl/container/container_create_linux_test.go
```diff
@@ -325,3 +329,197 @@ func TestCreateFromOCIArchive(t *testing.T) {
...
+	helpers.Custom("userdel", username).Run(&test.Expected{ExitCode: expect.ExitCodeNoCheck})
+	return nil
+}
+
+func delGroup(groupname string, helpers test.Helpers) error {
```
apostasie, line 522: Ibid

cmd/nerdctl/container/container_run_user_linux_test.go
```diff
@@ -181,3 +188,236 @@ func TestRunAddGroup_CVE_2023_25173(t *testing.T) {
...
+				Setup: func(data test.Data, helpers test.Helpers) {
+					err := appendUsernsConfig(data.Labels().Get("validUserns"), data.Labels().Get("expectedHostUID"), helpers)
+					assert.NilError(t, err, "Failed to append Userns config")
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					helpers.Anyhow("rm", "-f", data.Identifier())
+					removeUsernsConfig(t, data.Labels().Get("validUserns"), helpers)
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("run", "--tty", "-d", "--userns-remap", data.Labels().Get("validUid"), "--name", data.Identifier(), testutil.NginxAlpineImage)
```
apostasie, line 247: Curious if there was a specific reason to use `NginxAlpineImage` instead of `testutil.CommonImage`?
This is a nit - but generally speaking (unless there is a reason not to), tests are faster + easier to port when they use the smaller base common image.

cmd/nerdctl/container/container_run_user_linux_test.go
```diff
@@ -181,3 +188,236 @@ func TestRunAddGroup_CVE_2023_25173(t *testing.T) {
...
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						ExitCode: 1,
+					}
+				},
+			},
+			{
+				Description: "Test container run with invalid Userns",
+				Setup: func(data test.Data, helpers test.Helpers) {
```
apostasie, line 406: Can just be removed I guess.

cmd/nerdctl/container/container_run_user_linux_test.go
```diff
@@ -181,3 +188,236 @@ func TestRunAddGroup_CVE_2023_25173(t *testing.T) {
...
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					helpers.Anyhow("rm", "-f", data.Identifier())
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("run", "--tty", "-d", "--userns-remap", data.Labels().Get("invalidUserns"), "--name", data.Identifier(), testutil.NginxAlpineImage)
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
```
apostasie, line 414: `Expected: test.Expects(1, nil, nil)` is the more compact notation when you do not have advanced expectations on stdout or the need to manipulate data is your expectations.

hack/test-integration.sh
```diff
@@ -47,7 +47,7 @@ for arg in "$@"; do
 done
 
 if [ "$needsudo" == "true" ] || [ "$needsudo" == "yes" ] || [ "$needsudo" == "1" ]; then
-  gotestsum "${args[@]}" -- -timeout="$timeout" -p 1 -exec sudo -args -test.allow-kill-daemon "$@"
+  gotestsum "${args[@]}" -- -timeout="$timeout" -p 1 -exec sudo -args -test.allow-kill-daemon  "$@"
```
apostasie, line 50: Probably not necessary?

pkg/cmd/container/create_userns_opts_linux_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"gotest.tools/v3/assert"
+
+	"github.com/containerd/nerdctl/v2/pkg/api/types"
+	"github.com/containerd/nerdctl/v2/pkg/imgutil"
+)
+
+// TestCreateSnapshotOpts tests the createSnapshotOpts function.
+func TestCreateSnapshotOpts(t *testing.T) {
+	tests := []struct {
```
apostasie, line 31: + t.Parallel()?

pkg/cmd/container/create_userns_opts_linux_test.go
```diff
@@ -0,0 +1,140 @@
...
+			image:       imgutil.EnsuredImage{},
+			uidMaps:     []specs.LinuxIDMapping{},
+			gidMaps:     []specs.LinuxIDMapping{},
+			expectError: true,
+		},
+	}
+
+	for _, testCase := range tests {
+		t.Run(testCase.name, func(t *testing.T) {
+			_, err := createSnapshotOpts(testCase.id, testCase.image, testCase.uidMaps, testCase.gidMaps)
```
apostasie, line 76: + t.Parallel()?

apostasie:
@Shubhranshu153 left a bunch of notes, mostly on tests - they are mostly nitpicks honestly.

I want to do a more in-depth read of the main code later today, but otherwise, this is great!

As for the failures: freeBSD is acting up again - I guess we can just ignore. The other one is gomodjail - this might have gotten merged too fast - though it is likely just fine to ignore it.

cmd/nerdctl/builder/builder_build.go
Shubhranshu153: well i understood it but it was kind of intentional to do [warn] WARNING: ....
But i can remove it.

cmd/nerdctl/container/container_run_user_linux_test.go
Shubhranshu153: nginx images has a by default entry point that keeps it runnign, i can do with any pause container technically

cmd/nerdctl/container/container_run_user_linux_test.go
Shubhranshu153: We are changing the test framework too often to keep track of. Would like it to have a some fixed cadence.

pkg/testutil/nerdtest/requirements.go
```diff
@@ -376,3 +381,40 @@ var Gomodjail = &test.Requirement{
...
+		return false, "allow modify userns is disabled"
+	},
+}
+
+var RemapIDs = &test.Requirement{
+	Check: func(data test.Data, helpers test.Helpers) (ret bool, mess string) {
+		// Create a cobra command for ProcessRootCmdFlags to get globalOptions
+		ctx := context.Background()
+		snapshotterName := defaults.DefaultSnapshotter
+		namespace := namespaces.Default
```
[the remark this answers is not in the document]
Shubhranshu153: defaultNamespace -> yeah this better.

cmd/nerdctl/container/container_run_user_linux_test.go
apostasie: I see. I understand the convenience, but yeah, it would be a tiny bit better for the future with `CommonImage` + `sleep inf`.
Up to you.

cmd/nerdctl/builder/builder_build.go
Shubhranshu153: Removed it, the [Warn] has colored output so i think should be enough

cmd/nerdctl/container/container_run_user_linux_test.go
apostasie: For this specific aspect, it has been here for a while, but either way, you make a fair point.
There is still a few important changes that need to happen (#3448), but after that, agreed we should freeze and bucket changes.

cmd/nerdctl/container/container_run_user_linux_test.go
Shubhranshu153: fixed this one, test efficiency is important as we are adding so many tests.

pkg/cmd/container/create_userns_opts_linux.go
```diff
@@ -0,0 +1,516 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		container, err := findContainer(ctx, client, containerName)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := validateContainerStatus(ctx, container); err != nil {
```
apostasie, line 120: I do not quite understand this.

`validateContainerStatus` is not giving you any guarantee that the task will still be running by the time you call 
`getUserNamespacePath`, or by the time `userNameSpaceSpecOpts` is going to be used, so, what is the point of even checking that the container is running now?

Also, `getUserNamespacePath` which immediately follows will do another call to `container.Task` - so, two calls instead of one.

cmd/nerdctl/container/container_run_user_linux_test.go
apostasie: Sorry about that :/ I do really appreciate the inconvenience of a moving target...
Hopefully, the heavy lifting is mostly done now...
Pinky swear no more breaking change until the next one! :-D

Thanks for your patience with it @Shubhranshu153 - and glad you see value in it.

pkg/cmd/container/create_userns_opts_linux.go
Shubhranshu153: fair enough, technically it is not helping anything

AkihiroSuda:
Needs rebase

.github/workflows/workflow-test.yml
```diff
@@ -129,7 +129,7 @@ jobs:
           docker run --privileged --rm tonistiigi/binfmt --install linux/arm64
           docker run --privileged --rm tonistiigi/binfmt --install linux/arm/v7
       - name: "Run integration tests"
-        run: docker run -t --rm --privileged test-integration ./hack/test-integration.sh -test.only-flaky=false
+        run: docker run -t --rm --privileged test-integration ./hack/test-integration.sh -test.only-flaky=false -test.allow-modify-users=true
```
apostasie, line 132: @Shubhranshu153 heads-up.

GHA workflows have been shaken-up - hopefully for the better.

Here is where you need to port your new parameter:
https://github.com/containerd/nerdctl/blob/main/.github/workflows/job-test-in-container.yml#L158

Shubhranshu153:
Almalinux test seems to be flaky,

apostasie (approved):

Shubhranshu153:
ok this is repeatable.

Shubhranshu153:
```
		{
			Description: "start",
			NoParallel:  true,
			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
				return helpers.Command("start", data.Labels().Get("cID"))
			},
			Expected: test.Expects(0, nil, nil),
		},
		{
			Description: "logs",
			NoParallel:  true,
			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
				return helpers.Command("logs", data.Labels().Get("cID"))
			},
			Expected: test.Expects(0, nil, expect.Contains("foo")),
		},
	}
```
this test is racy in sense we cant guarantee the echo to complete before checking the logs. Needs to wait or retry

Shubhranshu153:
Not to de-sanitize this PR will create a separate PR to improve the TestCreate, for other bugs i still dont know what is the flakiness.

apostasie:
> Not to de-sanitize this PR will create a separate PR to improve the TestCreate, for other bugs i still dont know what is the flakiness.

Thanks a lot @Shubhranshu153 - the log you just identified is #3717 

Our other main PITA (especially on EL) is #4132

Shubhranshu153:
@apostasie ok got it, i think this is great to have a list to be able to focus on. I will try to pick some of these.
Will request a assignment so that we are not re-doing things.

Shubhranshu153:
Flaky Test PR's:
https://github.com/containerd/nerdctl/pull/4196
https://github.com/containerd/nerdctl/pull/4198
https://github.com/containerd/nerdctl/pull/4199

Shubhranshu153:
Rerunning the test for flaky test.

docs/command-reference.md
```diff
@@ -235,6 +235,9 @@ User flags:
 - :nerd_face: `--umask`: Set the umask inside the container. Defaults to 0022.
   Corresponds to Podman CLI.
 - :whale: `--group-add`: Add additional groups to join
+- :nerd_face: `--userns-remap=<username>:<groupname>`: Support idmapping of containers. This options is only supported on rootful linux for container create and run if a user name and optionally group name is passed, it does idmapping based on the uidmap and gidmap ranges specified in /etc/subuid and /etc/subgid respectively. Note: `--userns-remap` is not supported for building containers. Nerdctl Build doesn't support userns-remap feature. (format: <name|uid>[:<group|gid>])
```
AkihiroSuda, line 238: Isn't this a top-level flag?

Shubhranshu153: you are right after the refactor. let me update it.

Shubhranshu153: Updated.

AkihiroSuda (approved):
Thanks

PR #3913, opened 2025-02-23
Refactor: suggesting we normalize all command variables to cmd

apostasie:
This is a very simple, purely cosmetic PR, harmonizing all command variables to `var cmd` (under the cmd folder).

The proposed cleanup enhances code readability and removes what is currently just noise.

If this is something we think is good, more PRs will follow that focus on the same topic overall: cleanups / reorgs fostering simplicity and code readability.

Let me know your thoughts.

PR #3944, opened 2025-02-27
Cleanup/refactor proposal

apostasie:
This is a suggestion following on #3913

These changes are purely cosmetic, and their purpose is to increase code readability, by removing stuttering information from variables and functions name (following the refactor of cmd this last summer), and overall simplifying names.

This applies solely to `cmd/`

~~For now, these changes are solely on `apparmor`, so that we can talk about them.~~

~~If this is the right thing to do and we agree on the changes, I will apply them to the entirety of `cmd` and submit a complete / working PR.~~

The proposed changes are:
- rename all `newThingPackageCommand` function names to `thingCommand`:
- rename all `packageThingAction` function names to `thingAction`
- rename all `processPackageThingOptions` function names to `thingOptions`
- rename top-level `NewPackageCommand` unique functions to `Command`
- expand `Ls`, `Rm`,  to `List`, `Remove` where applicable

Reasoning is that:
- `Package` always stutter now that they are split under `cmd`
- `new`, or `process` does not add much value if any and is just noisy
- acronyms are bad

Thanks.

Tagging @AkihiroSuda for opinion.

### DIFF
--- a/cmd/nerdctl/apparmor/apparmor_inspect_linux.go
+++ b/cmd/nerdctl/apparmor/apparmor_inspect_linux.go
@@ -26,19 +26,19 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/defaults"
 )
 
-func newApparmorInspectCommand() *cobra.Command {
+func inspectCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:           "inspect",
 		Short:         fmt.Sprintf("Display the default AppArmor profile %q. Other profiles cannot be displayed with this command.", defaults.AppArmorProfileName),
 		Args:          cobra.NoArgs,
-		RunE:          apparmorInspectAction,
+		RunE:          inspectAction,
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
 	return cmd
 }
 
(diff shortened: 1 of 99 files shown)


cmd/nerdctl/apparmor/apparmor_linux_test.go
```diff
@@ -14,7 +14,7 @@
    limitations under the License.
 */
 
-package apparmor
+package apparmor_test
```
apostasie, line 17: This is somewhat orthogonal, but also desirable.
Test packages should not be part of the main package.

apostasie: In a shell the big pro is true blackbox testing (unable to access internal methods of the tested package) - makes tests less brittle / susceptible to internal refactoring, forces developers to actually test and use the public API, etc
And number 2: it is much easier to deal with circular dependencies.

Cons, I would say "true blackbox testing", which means losing access to private methods that may be needed (IMHO usually a sign that the public stuff is lacking).

apostasie: | This should be a separate PR

I agree.

This one here is just for discussion anyhow.

AkihiroSuda (approved):
Thanks

PR #3946, opened 2025-02-27
Fix apparmor host check to include aa-parser

### LINKED ISSUE #3945: Regression (maybe?) in AppArmor detection logic
### Description

[This recent commit](https://github.com/containerd/nerdctl/commit/2ecc39445615cdafd8bb7b3cc5a6cad7e5563644) changed nerdctl's AppArmor detection logic, dropping use of [containerd's implementation](https://github.com/containerd/containerd/blob/79309361be72cb679cb46bcb863a2d9b15dfe707/pkg/apparmor/apparmor_linux.go#L29-L45) and replacing it with a custom one based on [runc's implementation](https://github.com/opencontainers/runc/blob/0d49470392206f40eaab3b2190a57fe7bb3df458/libcontainer/apparmor/apparmor_linux.go#L18-L27), which the containerd implementation itself was based on (per the comment there). The one key difference is that the new check drops the check for the `apparmor_parser` binary, which the containerd implementation had added on top of runc's. 

Where I'm running into this is that in versions of Ubuntu prior to 24.04 (I'm running 22.04 in production), it is [sufficient to disable AppArmor](https://documentation.ubuntu.com/server/how-to/security/apparmor/index.html#disabling-or-re-enabling-apparmor) by uninstalling/disabling the systemd service. So my production machines are not actively running apparmor, and do not have the `apparmor-utils` package (containing the `apparmor_parser` binary) installed. But because we have not _explicitly_ disabled AppArmor in the kernel by adding `apparmor=0` to the kernel args, these flags still make it look like AppArmor is enabled:

```
$ ls -l /sys/kernel/security/apparmor
total 0
drwxr-xr-x 16 root root 0 Jan  1  1970 features
lr--r--r--  1 root root 0 Jan  1  1970 policy -> 'apparmorfs:[12575]'
-r--r--r--  1 root root 0 Jan  1  1970 profiles
-r--r--r--  1 root root 0 Jan  1  1970 revision

$ sudo cat /sys/module/apparmor/parameters/enabled
Y
```

On a machine in this state, using nerdctl (built after this commit) to run a container will fail with this error:

```
$ sudo nerdctl run hello-world
FATA[0000] load apparmor profile /tmp/nerdctl-default818158507: parser error(""): exec: "apparmor_parser": executable file not found in $PATH
```

Luckily this is easy to work around by simply installing the `apparmor-utils` package, and we'll look into setting `apparmor=0` explicitly to be on the safe side. 

But I think this is a regression, and the `apparmor_parser` check should be restored. For more background, see [this containerd commit](https://github.com/containerd/containerd/commit/eba94a15c89994f1be5c3bde6802d960971eb755) where the `apparmor_parser` check was added to _its_ implementation in the first place. 

I did this on a fresh install of Ubuntu 22.04:  
 
1. Confirm that `apparmor_parser` doesn't exist: `stat /usr/sbin/apparmor_parser`
2. Confirm that AppArmor service is not running: `systemctl status apparmor`
3. Using a build of nerdctl from https://github.com/containerd/nerdctl/commit/2ecc39445615cdafd8bb7b3cc5a6cad7e5563644 or newer: `sudo nerdctl run hello-world`

## Expected

```
Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
 2. The Docker daemon pulled the "hello-world" image from the Docker Hub.
    (arm64v8)
 3. The Docker daemon created a new container from that image which runs the
    executable that produces the output you are currently reading.
 4. The Docker daemon streamed that output to the Docker client, which sent it
    to your terminal.

To try something more ambitious, you can run an Ubuntu container with:
 $ docker run -it ubuntu bash

Share images, automate workflows, and more with a free Docker ID:
 https://hub.docker.com/

For more examples and ideas, visit:
 https://docs.docker.com/get-started/
```

## Received

```
FATA[0000] load apparmor profile /tmp/nerdctl-default3590477325: parser error(""): exec: "apparmor_parser": executable file not found in $PATH
```

Version `2.0.3-41-g2ecc3944` is the build I used to reproduce the error as described above, and `2.0.3-40-g8ef14b7a` is the "last good" version (built from the previous commit) that I used in my testing. 

None

### Host information

_No response_

apostasie:
Fix #3945

### DIFF
--- a/pkg/apparmorutil/apparmorutil_linux.go
+++ b/pkg/apparmorutil/apparmorutil_linux.go
@@ -34,13 +34,29 @@ var (
 )
 
 // hostSupports returns true if apparmor is enabled for the host
+// We cannot use containerd implementation because it explicitly prevents it from working inside a container.
 func hostSupports() bool {
 	checkAppArmor.Do(func() {
-		// see https://github.com/opencontainers/runc/blob/0d49470392206f40eaab3b2190a57fe7bb3df458/libcontainer/apparmor/apparmor_linux.go
-		if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil {
-			buf, err := os.ReadFile("/sys/module/apparmor/parameters/enabled")
-			appArmorSupported = err == nil && len(buf) > 1 && buf[0] == 'Y'
+		var pth string
+		if _, err := os.Stat("/sys/kernel/security/apparmor"); err != nil {
+			appArmorSupported = false
+			return
 		}
+		// In some rare circumstances, apparmor may be enabled, but the tooling could be missing
+		// containerd implementation shells out to aa-parser, so, require it here.
+		// See https://github.com/containerd/nerdctl/issues/3945 for details.
+		pth, err := exec.LookPath("apparmor_parser")
+		if err != nil {
+			appArmorSupported = false
+			return
+		}
+		if _, err = os.Stat(pth); err != nil {
+			appArmorSupported = false
+			return
+		}
+		var buf []byte
+		buf, err = os.ReadFile("/sys/module/apparmor/parameters/enabled")
+		appArmorSupported = err == nil && len(buf) == 2 && string(buf) == "Y\n"
 	})
 	return appArmorSupported
 }
@@ -73,7 +89,7 @@ var (
 func CanApplyExistingProfile() bool {
 	paramEnabledOnce.Do(func() {
 		buf, err := os.ReadFile("/sys/module/apparmor/parameters/enabled")
-		paramEnabled = err == nil && len(buf) > 1 && buf[0] == 'Y'
+		paramEnabled = err == nil && len(buf) == 2 && string(buf) == "Y\n"
 	})
 	return paramEnabled
 }


apostasie:
@AkihiroSuda this is minor/corner-case, as outlined by @dancavallaro, but it is still a regression.
Suggesting we have this for 2.0.4.

pkg/apparmorutil/apparmorutil_linux.go
```diff
@@ -34,12 +34,19 @@ var (
...
 func hostSupports() bool {
 	checkAppArmor.Do(func() {
 		// see https://github.com/opencontainers/runc/blob/0d49470392206f40eaab3b2190a57fe7bb3df458/libcontainer/apparmor/apparmor_linux.go
 		if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil {
-			buf, err := os.ReadFile("/sys/module/apparmor/parameters/enabled")
-			appArmorSupported = err == nil && len(buf) > 1 && buf[0] == 'Y'
+			// In some rare circumstances, apparmor may be enabled, but the tooling could be missing
+			// containerd implementation shells out to aa-parser, so, require it here.
+			// See https://github.com/containerd/nerdctl/issues/3945 for details.
+			if _, err = os.Stat("/sbin/apparmor_parser"); err == nil {
```
AkihiroSuda, line 45: `exec.LookPath` should be checked?

pkg/apparmorutil/apparmorutil_linux.go
```diff
@@ -34,12 +34,19 @@ var (
...
+				var buf []byte
+				buf, err = os.ReadFile("/sys/module/apparmor/parameters/enabled")
+				appArmorSupported = err == nil && len(buf) > 1 && buf[0] == 'Y'
```
AkihiroSuda, line 48: This comparison is not robust, as `buf` can be like `[]byte("You cannot read this file)"`, in theory

pkg/apparmorutil/apparmorutil_linux.go
apostasie: I was a bit surprised as well - this is what containerd does (https://github.com/containerd/containerd/blob/79309361be72cb679cb46bcb863a2d9b15dfe707/pkg/apparmor/apparmor_linux.go#L29-L45).

Is there a security concern here?
I am happy to relax the above and allow apparmor_parser from anywhere in the path if we are fine with it, but initially elected to not deviate from what they did. 

Thoughts?

pkg/apparmorutil/apparmorutil_linux.go
apostasie: Agreed.

I'll modify.

apostasie:
@dancavallaro it should be noted that things will still be somewhat "weird", even with the patch.

In your specific circumstances (kernel enabled, no userland tool), `nerdctl info` will still happily report apparmor as a security option, and if the profile was loaded already, you may still run containers with it.

What the patch does is prevent (the problem you reported) _loading_ profile, which is the only place `hostSupports` (or aa-parser) is needed.

I think this is fine (given this is indeed a corner case), but the whole thing is a bit shaky IMHO.

apostasie:
@AkihiroSuda your suggestions integrated with latest push.

I do always have a mild panic attack whenever I fiddle with logic involving `/sys/module`.
Verified working of course locally, but wanted to make sure we are 100% comfortable with the new logic here.

(now going to grab coffee outside to cure my kernel anxiety)

dancavallaro:
@apostasie I totally hear you. I'm already installing `apparmor-utils` as a workaround, and I've verified that explicitly setting `apparmor=0` on the kernel command line resolves the real issue (since finally `/sys/module/apparmor/parameters/enabled` will say `N`). 

Definitely agree that this is all pretty hairy, but thanks for the quick fix!

apostasie:
> Definitely agree that this is all pretty hairy, but thanks for the quick fix!

Well, I am operating under the assumption that @AkihiroSuda will tolerate me here as long as I live by the motto: "you broke it, you better fix it", so, better act fast to sweep my mistakes under the carpet. 🤣🤣🤣

AkihiroSuda (approved):
Thanks

pkg/apparmorutil/apparmorutil_linux.go
```diff
@@ -73,7 +89,7 @@ var (
 func CanApplyExistingProfile() bool {
 	paramEnabledOnce.Do(func() {
 		buf, err := os.ReadFile("/sys/module/apparmor/parameters/enabled")
-		paramEnabled = err == nil && len(buf) > 1 && buf[0] == 'Y'
+		paramEnabled = err == nil && len(buf) == 2 && string(buf) == "Y\n"
```
AkihiroSuda, line 92: nit: `bytes.Equal` could be used

apostasie: Never used it.
It is neat.

Thanks.

Issue #3952, opened 2025-02-28
CI: something wrong going on with pty tests


apostasie:
### Description

Looking at TestIssue3568 on this run:
```
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test run --rm -it --detach-keys=ctrl-a,ctrl-b --name testissue3568-41d4a31b ghcr.io/stargz-containers/alpine:3.13-org
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
    command.go:113: assertion failed: expect.ExitCode is not result.ExitCode: Expected exit code: 0
        
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test container inspect testissue3568-41d4a31b
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   time="2025-02-28T19:00:13Z" level=fatal msg="1 errors:\nno such container: testissue3568-41d4a31b"
```

The command is:
```
		cmd := helpers.Command("run", "--rm", "-it", "--detach-keys=ctrl-a,ctrl-b", "--name", data.Identifier(), testutil.CommonImage)
[...]

PR #3953, opened 2025-02-28
Harden test pty error handling

apostasie:
This is a debugging PR to help figure out and fix issues outlined in #3952

There is likely two distinct issues:
- the first one is likely a race between copying from the master and what the container writes out - it is possibly because of the hack necessary because of icmd use of MultiWriter
- the second one is more mysterious: it seems like in certain circumstances the container exits immediately (?)

I will likely restart the CI many times to try and reproduce the issues.


apostasie:
Ok, it is consistently green.

Good to merge.

AkihiroSuda (approved):

PR #3960, opened 2025-03-03
Fix test TestNoneNetworkHostName

apostasie:
TestNoneNetworkHostName failed on an unrelated build:
https://github.com/containerd/nerdctl/actions/runs/13622696958/job/38074815721#step:8:2329

However, the way the test is currently written prevents useful information from being shown.

This PR rewrite part of it to make it simpler / more sensical and actually show information if it fails.


AkihiroSuda (approved):
Thanks

PR #3965, opened 2025-03-03
CI: fix alma targets

apostasie:
- ~~[ ] ensure alma test are ran against the right version of containerd, ubuntu and rootlesskit inside the container~~ testing matrix clarified separately
- ~~[ ] figure out why ipfs is currently broken~~ PEBKAC
- [x] enable caching to speed-up the test-integration build
- [x] workaround the losetup weirdness with an extra mount for dev (that sounds vaguely familiar? some old centos / docker issue?) - presumably docker is faulty here against that (old) kernel - and there is a timing issue with how /dev is exposed, causing losetup to fail in a flaky way
- [x] split flaky/not flaky like for the others so we speed-up the run a bit
- [x] migrated TestRunDevice
- [x] non-flaky test suite is passing for both rootless and rootful
- [x] flaky test suite is passing for both rootless and rootful
- ~~[ ] TestStats has no reason to run on cgroups v1 (this is why it is hanging) - investigating to figure out why it would run~~ unclear why it would have hung - might be a side-effect of mounting /dev that fixed it

Currently on top of trivial #3970 (< merge first).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -429,6 +429,10 @@ jobs:
       matrix:
         mode: ["rootful", "rootless"]
     runs-on: ubuntu-24.04
+    env:
+      MODE: ${{ matrix.mode }}
+      # FIXME: this is only necessary to access the build cache. To remove with build cleanup.
+      CONTAINERD_VERSION: v2.0.3
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -427,15 +429,19 @@ jobs:
       matrix:
         mode: ["rootful", "rootless"]
     runs-on: ubuntu-24.04
+    env:
+      UBUNTU_VERSION: 20.04
+      CONTAINERD_VERSION: v1.6.36
+      ROOTLESSKIT_VERSION: v1.1.1
```
AkihiroSuda, line 435: Why?

apostasie: That is what we had previously for the Ubuntu 20.04 test target.

Short of specifying these, the container you are going to run on Alma will be 24.04 with containerd v2 + rootlesskit v2.

apostasie: But maybe it is a good time to rethink our test matrix: #3966

apostasie:
Ok, CI is green.

I am not 100% confident that I got all of it, but it is an improvement.

Runs are still quite slow (I guess the cost of the extra vm).

Suggesting we merge this and patch more if/when issues crop up.

AkihiroSuda (approved):
Thanks

PR #3969, opened 2025-03-04
Refactor: clean-up cmd filenames

apostasie:
Following-on previous `cmd` cleanups, this one shorten / rename the files.

Anything that stutters is reduced:
`cmd/nerdctl/category/category_(.+).go` > `cmd/nerdctl/category/$1.go`

Top-level files are renamed to `command.go` and `command_test.go`.


apostasie:
At 58 minutes+ the EL8 tests are presumably stuck.

I believe we can ignore for this specific PR - and see if I can get somewhere fixing it in #3965

apostasie:
> Sorry, I'd prefer the file names to be as unique as possible

This is the opposite of what is currently done in `pkg/cmd`. eg: `pkg/cmd/container/commit.go`. 

Are you saying you would prefer them to be `pkg/cmd/package/package_x.go` then?

AkihiroSuda:
> Are you saying you would prefer them to be pkg/cmd/package/package_x.go then?

That would be preferable when there is another `x.go`

apostasie:
> > Are you saying you would prefer them to be pkg/cmd/package/package_x.go then?
> 
> That would be preferable when there is another `x.go`

I see.

PR #3972, opened 2025-03-04
add uninstall_nerdctl_full Script

xiaomizhg:
Script for uninstalling nerdctl_full  on Linux

### DIFF
--- a/extras/uninstall_nerdctl_full.sh
+++ b/extras/uninstall_nerdctl_full.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+# Check if installation path 
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <installation_path>"
+    exit 1
+fi
+
+# Get the installed path
+INSTALL_PATH="$1"
+
+# the files and directories to be deleted
+BIN_FILES=(
+    "${INSTALL_PATH}/bin/buildctl"
+    "${INSTALL_PATH}/bin/buildg"
+    "${INSTALL_PATH}/bin/buildkit-cni-bridge"
+    "${INSTALL_PATH}/bin/buildkit-cni-firewall"
+    "${INSTALL_PATH}/bin/buildkit-cni-host-local"
+    "${INSTALL_PATH}/bin/buildkit-cni-loopback"
+    "${INSTALL_PATH}/bin/buildkitd"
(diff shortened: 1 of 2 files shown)


extras/uninstall_nerdctl_full.sh
```diff
@@ -0,0 +1,93 @@
...
+# Check if installation path 
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <installation_path>"
+    exit 1
+fi
+
+# Get the installed path
+INSTALL_PATH="$1"
+
+# the files and directories to be deleted
```
AkihiroSuda, line 12: This should read the file names and the digests from `$PREFIX/share/doc/nerdctl-full/SHA256SUMS`.
The digests have to be checked (by default, at least) so as to avoid removing files installed by other tools

extras/uninstall_nerdctl_full.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+# Check if installation path 
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <installation_path>"
+    exit 1
+fi
+
+# Get the installed path
+INSTALL_PATH="$1"
```
AkihiroSuda, line 10: This can be automatically guessed from the path of the script

extras/uninstall_nerdctl_full.sh
AkihiroSuda, line 1: Please squash the commits

extras/uninstall_nerdctl_full.sh
Shubhranshu153: This i think cannot be reliably inferred if both hash match. May be it should only be able remove if the install path is specific to nerdctl.

extras/uninstall_nerdctl_full.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
apostasie, line 1: `#!/usr/bin/env bash`

extras/uninstall_nerdctl_full.sh
apostasie: (bash) portable suggestion:
`INSTALL_PATH="$(cd "$(dirname "${BASH_SOURCE[0]:-$PWD}")" 2>/dev/null 1>&2 && pwd)"`

PR #3973, opened 2025-03-04
CI: implement updated test matrix

apostasie:
Fix #3966

PR #3974, opened 2025-03-04
Remove WORKAROUND_ISSUE_622 

apostasie:
On top of #3973 

We no longer test rootlesskit v1 + slirp4netns (actually, we haven't for a while), so this hack is presumably no longer necessary.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,9 +29,6 @@ jobs:
             containerd: v1.6.36
             arch: amd64
           - runner: ubuntu-24.04
-            containerd: v1.7.25
-            arch: amd64
-          - runner: ubuntu-24.04
             containerd: v2.0.3
             arch: amd64
           - runner: ubuntu-24.04-arm
@@ -106,7 +103,7 @@ jobs:
       matrix:
         include:
           - ubuntu: 22.04
-            containerd: v1.7.25
+            containerd: v1.6.36
             runner: "ubuntu-22.04"
             arch: amd64
           - ubuntu: 24.04
@@ -115,7 +112,7 @@ jobs:
             arch: amd64
           - ubuntu: 24.04
             containerd: v2.0.3
-            runner: ubuntu-24.04-arm
+            runner: "ubuntu-24.04-arm"
             arch: arm64
     env:
       CONTAINERD_VERSION: "${{ matrix.containerd }}"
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
> We no longer test rootlesskit v1 + slirp4netns

We no longer test slirp4netns port forwarder, but we still test the slirp4netns network driver, which is allegedly related to #622

apostasie:
https://github.com/containerd/nerdctl/issues/622#issuecomment-2696061378

Didn't you say it would only happen with rootlesskit v2?

.github/workflows/test.yml
```diff
@@ -219,26 +216,35 @@ jobs:
...
-    runs-on: "ubuntu-${{ matrix.ubuntu }}"
+    runs-on: "${{ matrix.runner }}"
     strategy:
       fail-fast: false
       matrix:
         include:
           - ubuntu: 22.04
-            containerd: v1.7.25
+            containerd: v1.6.36
             rootlesskit: v1.1.1  # Deprecated
```
AkihiroSuda, line 226: We still have RootlessKit v1 here, so the workaround still has to be enabled for this

apostasie: We do have rootlesskit v1, but we do _not_ test slirp with it.

The only target we are testing slirp for is https://github.com/containerd/nerdctl/pull/3974/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R242-R246

AkihiroSuda: There are two kinds of the slirp4netns drivers:
- network driver: outgoing connections
- port driver: incoming connections

The code here still uses slirp4netns as the network driver

apostasie:
project-checks breakage is unrelated (I guess Phil just broke it :D)

.github/workflows/test.yml
apostasie: My bad.
Thanks for clarifying!

apostasie: Kind of curious to see it fail though. Can we leave this PR open long enough for the full CI run?
Then I'll close it.

AkihiroSuda:
> project-checks breakage is unrelated (I guess Phil just broke it :D)

No, I broke it 🙇 
- https://github.com/containerd/ltag/pull/19

Being fixed in:
- https://github.com/containerd/project-checks/pull/18

PR #3977, opened 2025-03-04
Fix broken lint-license make task

apostasie:
I broke it when cleaning-up the Makefile.

Bringing it back so that we can run license check locally.


Makefile
```diff
@@ -168,10 +168,16 @@ lint-mod:
 		&& go mod tidy --diff
 	$(call footer, $@)
 
+# FIXME: go-licenses cannot find LICENSE from root of repo when submodule is imported:
+# https://github.com/google/go-licenses/issues/186
+# This is impacting gotest.tools
```
AkihiroSuda, line 173: Please explain the reason of the ignorance for go-base36 too

apostasie: Sure.
Done.

AkihiroSuda (approved):
Thanks

We will have to dedupe the scripts across the GHA yamls and the Makefile

PR #3983, opened 2025-03-04
CI, build 1: disable buildx output for dependency stage

apostasie:

This is the first PR in a series to reduce build time and increase reliability.

**These PRs are broken down in digestible size from a large PR on my fork to ease review.**

-------

The dependencies stage purpose is solely to save cacheable stages into GHA.

There is no reason to load the image into docker store at this point.

The cost of this (useless) operation is about 40 seconds (out of 1 minute 20 seconds for that stage on a warm cache - or ~6 minutes on a cold cache).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,6 @@ jobs:
         run: |
           docker buildx create --name with-gha --use
           docker buildx build \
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #3984, opened 2025-03-05
CI, build 2: Move GHA to zstd

apostasie:
**For context, see #3940, #3924, and the start of this PR cycle at #3983.**

On top of #3983

--------

This moves caching to zstd.

Given the bottleneck with gha is IO and not compute, this is not likely to have an impact on speed - but it will on cache size.

PR #3979, opened 2025-03-04
Increase wait for ensurecontainerstarted

apostasie:
On EL8, TestStop is supposedly failing to bring up containers in time.

https://github.com/containerd/nerdctl/actions/runs/13648882661/job/38152811978?pr=3976#step:11:2044

This increases the timeout.

PR #3985, opened 2025-03-05
CI, build 3: Stop sharding cache on containerd version

apostasie:

On top of #3984 

-------------

We currently shard our cache per-containerd version.

There is more downsides than upsides in doing this. Specifically:
- it does multiply our cache use by as many versions of containerd we are testing (currently three, soon two)
- because we are more likely to hit GHA limits, this triggers cascading cache evictions, forcing unnecessary full rebuilds and further increasing communication with the cache backend, leading to throttling

~~While _not_ sharding and running in parallel will indeed drop some layers that could be cached until the cache is fully warm, this should be out-weighted by the benefits.~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,8 +47,8 @@ jobs:
         run: |
           docker buildx create --name with-gha --use
           docker buildx build \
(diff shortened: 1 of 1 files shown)


apostasie:
EL8 failure is TestStopWithStopSignal - see #3979

apostasie:
rootless failure is #3513

djdongjin (approved):

PR #3986, opened 2025-03-05
CI, build 4: Cleanup Dockerfile base images

apostasie:

On top of #3985 

--------------

We currently depend on two distincts golang image (bookworm and alpine), and we also download a golang archive from go servers.

Depending at all on docker images is a significant source of issues:
- there is delay between availability of go version and availability of go images on the hub - this has regularly broken canary in the past (currently is broken (*))
- go images on the hub have their own versioning scheme that is different from golang's, requiring us to do some shell monkeying to convert one version to the other
- hitting the Hub often leads to the dreaded 429 (too many requests)
- the hub image adds no visible value on top of the archive that we are downloading anyways...

This PR does cleanup our base images.

It bases every build stage on a single image, a naked bookworm, and uses the golang version of go.

That will reduce our network traffic with Hub, improve cacheability, and make our life less miserable when new versions of go arrive.

Finally, this will drastically reduce network com with go servers, as currently the archive is never cached and always downloaded, for every target we build. There should be a speed boost out of that for the later build stages.

(*) this PR does fix the current canary bustage

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,6 @@ ARG TINI_VERSION=v0.19.0
 ARG BUILDG_VERSION=v0.4.1
 
 # Test deps
-ARG GO_VERSION=1.24
 ARG UBUNTU_VERSION=24.04
 ARG CONTAINERIZED_SYSTEMD_VERSION=v0.1.1
 ARG GOTESTSUM_VERSION=v1.12.0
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -42,36 +42,142 @@ ARG TINI_VERSION=v0.19.0
...
+# - CONTAINERIZED_SYSTEMD_VERSION
+########################################################################################################################
+# tooling-base is the single base image we use for all other tooling image
+FROM --platform=$BUILDPLATFORM debian:$DEBIAN_VERSION AS tooling-base
+SHELL ["/bin/bash", "-o", "errexit", "-o", "errtrace", "-o", "functrace", "-o", "nounset", "-o", "pipefail", "-c"]
+ENV DEBIAN_FRONTEND="noninteractive"
+ENV TERM="xterm"
+ENV LANG="C.UTF-8"
+ENV LC_ALL="C.UTF-8"
+ENV TZ="America/Los_Angeles"
```
AkihiroSuda, line 74: ?

Dockerfile
```diff
@@ -42,36 +42,142 @@ ARG TINI_VERSION=v0.19.0
...
+# - DEBIAN_VERSION (=bookworm) or content of it
+# - XX_VERSION
+# - GO_VERSION
+# - CONTAINERIZED_SYSTEMD_VERSION
+########################################################################################################################
+# tooling-base is the single base image we use for all other tooling image
+FROM --platform=$BUILDPLATFORM debian:$DEBIAN_VERSION AS tooling-base
+SHELL ["/bin/bash", "-o", "errexit", "-o", "errtrace", "-o", "functrace", "-o", "nounset", "-o", "pipefail", "-c"]
+ENV DEBIAN_FRONTEND="noninteractive"
+ENV TERM="xterm"
```
AkihiroSuda, line 71: ?

Dockerfile
```diff
@@ -42,36 +42,142 @@ ARG TINI_VERSION=v0.19.0
...
+ENV LANG="C.UTF-8"
```
AkihiroSuda, line 72: Isn't this the default locale?

Dockerfile
```diff
@@ -270,21 +372,18 @@ VOLUME /var/lib/nerdctl
 ENTRYPOINT ["/docker-entrypoint.sh"]
 CMD ["bash", "--login", "-i"]
 
-# convert GO_VERSION=1.16 to the latest release such as "go1.16.1"
```
AkihiroSuda, line 273: Can we keep this

Dockerfile
```diff
@@ -42,36 +42,142 @@ ARG TINI_VERSION=v0.19.0
...
+      exit 1; \
+    }; \
+    arch=amd64; \
+    filename="$(jq -r 'map(select(.arch=="'"$arch"'"))[0].filename' <(printf "$candidates"))"; \
+    mkdir -p /out/usr/local/linux/"$arch"; \
+    [ "$filename" != "" ] && curl -fsSL --proto '=https' --tlsv1.2 https://go.dev/dl/"$filename" | tar xzC /out/usr/local/linux/"$arch" || {  \
+      echo "Failed retrieving go download for $arch: $GO_VERSION"; \
+      exit 1; \
+    }
+
```
AkihiroSuda, line 135: This complicates the Dockerfile too much

AkihiroSuda (changes requested):
Sorry, I don't get the advantages of this

AkihiroSuda:
> there is delay between availability of go version and availability of go images on the hub

Yes, but usually fixed within a day

AkihiroSuda:
Alternative:
- https://github.com/containerd/nerdctl/pull/3997

apostasie:
> Sorry, I don't get the advantages of this

Commented briefly on #3997 

The point is that our current Dockerfile is retrieving massive amount of data for no good reason, and heavily depends on Hub for stuff that is very easy to do without.

We just disagree on the way to tackle these problems, so, I'll just retract this proposal.

PR #3987, opened 2025-03-05
CI, build 5: Extract cacheable part of runtime image

apostasie:
For context, see #3924, and the start of this PR cycle at #3983.

On top of #3986

-------

Currently, we do not cache at all the runtime image.

So, for every supported target, at every run, we download ubuntu:24.04 from docker hub, then a bunch of packages from ubuntu servers.

This PR addresses that by extracting the cacheable part.

We should see a good speed-up on the later build stage, a significant decrease in our communications with Hub and Ubuntu repos, at the cost of a larger cache.

Note that these benefits will NOT be seen until a later PR, as getting this stage into the `dependencies` pipeline requires more modifications on other stages,

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
         run: |
           docker buildx create --name with-gha --use
           docker buildx build \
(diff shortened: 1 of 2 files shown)


apostasie:
See #3986

PR #3989, opened 2025-03-05
CI: update Canary and EL8 titles

### LINKED ISSUE #3988: CI: mark targets as "read for perspective"

Certain CI targets are more problematic than others:
a. canary:
  -  by its very nature it may fail in ways that are unrelated to the PR
  - there are ongoing issues wrt Docker Hub golang images

b. EL8
* Because of how they are implemented (extra VM), they are very slow. As such, they tend to surface timing issues more easily (this is good, to some extent), that are likely unrelated to PRs being submitted.
* They also tend to hang and timeout (for yet not figured-out reasons).

While these failures should NOT be ignored, as they may point to actual issues, they should also be put in perspective so that casual contributors do not take a fail on these as a "hard-no".

Suggestion is to update the title of these targets on the CI, so that they display:

`[see #3988] EL8/Canary`

That would help not freaking out casual contributors and make it clear these targets are problematic in themselves.

apostasie:
Fix #3988

### DIFF
--- a/.github/workflows/test-canary.yml
+++ b/.github/workflows/test-canary.yml
@@ -1,5 +1,5 @@
 # This pipeline purpose is solely meant to run a subset of our test suites against upcoming or unreleased dependencies versions
-name: canary
+name: "[flaky, see #3988] canary"
 
 on:
   push:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -423,7 +423,7 @@ jobs:
   # Do not upgrade this to EL9 (cgroup v2).
   test-integration-el8:
     timeout-minutes: 60
-    name: "EL8 (cgroup v1)"
+    name: "[flaky, see #3988] EL8 (cgroup v1)"
     strategy:
       fail-fast: false
       matrix:


apostasie:
rootful 2.0.3 failure is #3513

.github/workflows/test.yml
```diff
@@ -423,7 +423,7 @@ jobs:
   # Do not upgrade this to EL9 (cgroup v2).
   test-integration-el8:
     timeout-minutes: 60
-    name: "EL8 (cgroup v1)"
+    name: "[see #3988] EL8 (cgroup v1)"
```
fahedouch, line 426: Sorry but I can't get how `EL8` is configured here ?

apostasie: @fahedouch not sure what you are asking? but yeah, this one is somewhat tricky.
Alma 8 (which is the current EL derivative used here) is started in a Lima VM: https://github.com/containerd/nerdctl/blob/d1f4890fbe957e3a784de58ae041bb1310fdb494/.github/workflows/test.yml#L469-L475

Then we run the normal integration test suite inside an Ubuntu container (but of course with the EL 4.18 kernel).

fahedouch: oh yes! thanks.

fahedouch: So if I understand correctly, the MR is mergeable with the current issues ?

apostasie: Well, the maintainers should make a case by case call.

At least the contributors should not spend too much time trying to figure out what is up with these targets, and yes, IMO the maintainers may make judgment calls that "broken canary because of Hub carfuffle" should not prevent merging.

I do _not_ know what are the current merge settings of the repo and if they require a green run though - up to you folks if you want to adjust that.

.github/workflows/test-canary.yml
```diff
@@ -1,5 +1,5 @@
 # This pipeline purpose is solely meant to run a subset of our test suites against upcoming or unreleased dependencies versions
-name: canary
+name: "[see #3988] canary"
```
AkihiroSuda, line 2: Maybe add a brief description in a few words

```suggestion
name: "[May fail; see #3988] canary"
```

.github/workflows/test.yml
fahedouch: CI failure are note related to your PR, So for me it is mergeable!

fahedouch (approved):
Thanks

PR #3991, opened 2025-03-06
[WIP] CI, build NEXT

apostasie:


AkihiroSuda:
Alternative for the Go version hack:
- https://github.com/containerd/nerdctl/pull/3997

AkihiroSuda:
Also, I thought the direction was to remove the integration test stages from the Dockerfile eventually.
So, the Go version for the integration test can be just specified as "1.24.x" in the GHA yamls.

The Go version for the building the binaries will still have to be specified in the Dockerfile, but probably there is no need to add Go installer to the Makefile.

apostasie:
> Also, I thought the direction was to remove the integration test stages from the Dockerfile eventually. So, the Go version for the integration test can be just specified as "1.24.x" in the GHA yamls.

The runtime test stage, yes - not the build part IMHO.

Retracting (see #3986 )

PR #3994, opened 2025-03-06
add sync syscall before get container diff when commit

### LINKED ISSUE #3791: Commit may fail when container write many small files (`failed to create diff tar stream: failed to copy: archive/tar: write too long`)
### Description

Currently we got an error occasionally in our product environment, when we commit a container to an image.
```
failed to create diff: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount260055902: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount4247854495: failed to write compressed diff: failed to create diff tar stream: failed to copy: archive/tar: write too long: unknown
```
It's more likely to occur when writing a large number of small files in a container. I think this maybe related to buffer io, that the file content is changed during tar write it. There is a very low probability that the problem recurs.

```
func (fw *regFileWriter) Write(b []byte) (n int, err error) {
	overwrite := int64(len(b)) > fw.nb
	if overwrite {
		b = b[:fw.nb]
	}
	if len(b) > 0 {
		n, err = fw.w.Write(b)
		fw.nb -= int64(n)
	}
	switch {
	case err != nil:
		return n, err
	case overwrite:
		return n, ErrWriteTooLong
	default:
		return n, nil
	}
}
```

1. start a container to write many small files
2. nerdctl commit this container
3. repeat it thousands of times.

If this is related to the buffer io, I think maybe the only way to avoid this is to add a sleep after Pause so that kernel can flush all data into disk.

nerdctl version 2.0.1-2-g1f812253

None

### Host information

_No response_

abel-von:
fix #3791

### DIFF
--- a/pkg/imgutil/commit/commit.go
+++ b/pkg/imgutil/commit/commit.go
@@ -173,6 +173,9 @@ func Commit(ctx context.Context, client *containerd.Client, container containerd
 	}
 	defer done(ctx)
 
+	// Sync filesystem to make sure that all the data writes in container could be persisted to disk.
+	Sync()
+
 	diffLayerDesc, diffID, err := createDiff(ctx, id, sn, client.ContentStore(), differ)
 	if err != nil {
 		return emptyDigest, fmt.Errorf("failed to export layer: %w", err)
--- a/pkg/imgutil/commit/commit_other.go
+++ b/pkg/imgutil/commit/commit_other.go
@@ -1,0 +1,23 @@
+//go:build !unix
+
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package commit
+
+func Sync() {
+
+}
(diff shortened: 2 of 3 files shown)


pkg/imgutil/commit/commit.go
```diff
@@ -173,6 +173,8 @@ func Commit(ctx context.Context, client *containerd.Client, container containerd
 	}
 	defer done(ctx)
 
+	Sync()
```
AkihiroSuda, line 176: Please add a comment to explain the reason

pkg/imgutil/commit/commit_linux.go
AkihiroSuda, line 1: Maybe s/linux/unix/ , assuming that the issue may happen on FreeBSD too

pkg/imgutil/commit/commit.go
abel-von: done

pkg/imgutil/commit/commit_linux.go
abel-von: done

pkg/imgutil/commit/commit.go
AkihiroSuda: But why do we this only specifically for `commit`?

abel-von: I think it's because we need to do a new mount and compare the diffs to create a new layer. if data is not synced into disk, it may not be the newest data we read when comparing.

abel-von: we have met the error like below with a very low probability, I think it is because we did't sync before comparing.
```
failed to create diff: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount260055902: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount4247854495: failed to write compressed diff: failed to create diff tar stream: failed to copy: archive/tar: write too long: unknown
```

fahedouch:
Can we guarantee that data is not written to the container layer during the flush? I am wondering if we can blocking writing to the disk during the commit, imo it is the safest way to perform a commit.

PR #2337, opened 2023-06-29
fix: shutdown logger after container process exits

mharwani:
Addresses #2313 

Container task is not deleted after it exits, leaving the runtime shim and logger open throughout the lifecycle of a container. This can be a problem when there is unprocessed `stdout` data after a container task finishes and the logger is blocked forever waiting for a newline or EOF.

This PR fixes the issue by piping container output instead of reading directly from `stdout` and `stderr` file descriptors. Closing the io pipes upon task exit allows the logger to shutdown gracefully and process leftover data without a newline .

PR #3995, opened 2025-03-06
[carry #2337] ensure logger completion after container exit 

fahedouch:
carry #2337

Introduced mechanisms to ensure logger completion after container exit using file locks and added robust handling for log streams.


AkihiroSuda (approved):
Thanks

PR #3997, opened 2025-03-07
Dockerfile: remove "goversion" stage

AkihiroSuda:
No need to fetch `https://golang.org/dl/$(cat /GOVERSION)...` when we can just copy `/usr/local/go` from the `golang:${GO_VERSION}` image.

Alternative to:
- #3986

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -270,10 +270,6 @@ VOLUME /var/lib/nerdctl
 ENTRYPOINT ["/docker-entrypoint.sh"]
 CMD ["bash", "--login", "-i"]
 
-# convert GO_VERSION=1.16 to the latest release such as "go1.16.1"
-FROM golang:${GO_VERSION}-alpine AS goversion
-RUN go env GOVERSION > /GOVERSION
-
 FROM base AS test-integration
 ARG DEBIAN_FRONTEND=noninteractive
 # `expect` package contains `unbuffer(1)`, which is used for emulating TTY for testing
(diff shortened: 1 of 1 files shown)


apostasie:
This is increasing our reliance on Hub content instead of decreasing it, and adding unnecessary intermediary layers and dependent infrastructure with no added value.

AkihiroSuda:
> This is increasing our reliance on Hub content instead of decreasing it, and adding unnecessary intermediary layers and dependent infrastructure with no added value.

This PR just reuses an existing `build-base-debian` stage, so the reliance of Hub isn't changed, and it does not add a new intermediate layer.

djdongjin (approved):
LGTM

apostasie:
> > This is increasing our reliance on Hub content instead of decreasing it, and adding unnecessary intermediary layers and dependent infrastructure with no added value.
> 
> This PR just reuses an existing `build-base-debian` stage, so the reliance of Hub isn't changed, and it does not add a new intermediate layer.

The point is:
- we use _two_ distinct golang images from Hub
- we are now copying go from the debian image as-is to an ubuntu image - while _presumably_ it is not a problem _if all they are doing_ is retrieving it from golang server - and then if it is all they are doing, it then begs the question as to why we should rely on an extra intermediary piece of infrastructure (Hub) to do a curl for us (with availability delays and versioning scheme differences)

I guess we just disagree on this overall, so, this is fine, no point in beating this.

I just wish we would look at all of this critically.

We are currently retrieving about 6G of content from the network on a cold cache (out of which about 1G comes from Hub).
This is a very large amount of data that does not make much sense IMHO - and explains in part why we are getting 429 from Hub so regularly.

Anyhow.

Issue #3895, opened 2025-02-18
`nerdctl run --log-driver=none` should not invoke a `_NERDCTL_INTERNAL_LOGGING` process


AkihiroSuda:
Currently `_NERDCTL_INTERNAL_LOGGING` is always invoked on starting a container, even when `--log-driver` is set to `none`

Shubhranshu153:
```
[shubhum@lima-finch nerdctl]$ sudo nerdctl  --debug run -d --log-driver none alpine  e
cho "hello nerdctl"
DEBU[0000] verifying process skipped                    
DEBU[0000] generated log driver: binary:///usr/local/bin/nerdctl?_NERDCTL_INTERNAL_LOGGING=%2Fvar%2Flib%2Fnerdctl%2F1935db59 
DEBU[0000] detected 127.0.0.53 nameserver, assuming systemd-resolved, so using resolv.conf: /run/systemd/resolve/resolv.conf 
DEBU[0000] remote introspection plugin filters           filters="[type==io.containerd.snapshotter.v1, id==overlayfs]"
in detached mode with loguri
2a32b73ff483232d95ecd03e107fb29bb173a3bba00d4572dc06c91ef69b70a3
[shubhum@lima-finch nerdctl]$ sudo nerdctl logs 2a
FATA[0000] log type `none` was selected, nothing to log 
[shubhum@lima-finch nerdctl]$ 
```
Although invoked it seems it invokes the none_logger which doesnt write to file. 
Do we want not to create the uri itself?

AkihiroSuda:
> Do we want not to create the uri itself?

👍

Shubhranshu153:
i can pick this up. Just want to confirm we donot want to create the URI in case of none and handle it outside the logging drivers. currently it gets routed to none driver.

AkihiroSuda:
> i can pick this up. Just want to confirm we donot want to create the URI in case of none and handle it outside the logging drivers. currently it gets routed to none driver.

Thanks, correct

fahedouch:
fixed by https://github.com/containerd/nerdctl/pull/4002

PR #4002, opened 2025-03-11
fix: none log driver creating log uri

Shubhranshu153:
Fix for Issue: https://github.com/containerd/nerdctl/issues/3895

Description:
The log driver none was creating a logURI. In this case we handle it via setting the log driver to none and escaping starting the logging binary for cases where we have set the driver as none.

Another option was to set logURI to "" in case of none but there are hard coded checks to default the log driver to json in that case avoided refactoring that.


AkihiroSuda (approved):
Thanks

PR #4017, opened 2025-03-18
update slirp4netns (1.3.2), containerd (2.0.4), runc (1.2.6)...; verify the git tags with the commit hashes

AkihiroSuda:
Will release v2.0.4 after merging this


Dockerfile
```diff
@@ -15,11 +15,15 @@
 # -----------------------------------------------------------------------------
 # Usage: `docker run -it --privileged <IMAGE>`. Make sure to add `-t` and `--privileged`.
 
-# TODO: verify commit hash
+# Some deps are compiled from the source releases, some ones are installed from the binary releases.
+#
+# To update a dep, one of the following files have to be updated:
+# - Dockerfile.d/COMMITS.d/<COMPONENT>-<VERSION>:    the hash of the source release
+# - Dockerfile.d/SHA256SUMS.d/<COMPONENT>-<VERSION>: the hash of the binary release
```
fahedouch, line 22: good. Thanks

fahedouch (approved):
Thanks

austinvazquez (approved):

Dockerfile
```diff
@@ -74,42 +79,54 @@ RUN git config --global advice.detachedHead false
...
 FROM build-base-debian AS build-bypass4netns
 ARG BYPASS4NETNS_VERSION
+ARG CANARY
 ARG TARGETARCH
+COPY ./Dockerfile.d/COMMITS.d/ /COMMITS.d
 RUN git clone https://github.com/rootless-containers/bypass4netns.git /go/src/github.com/rootless-containers/bypass4netns
 WORKDIR /go/src/github.com/rootless-containers/bypass4netns
-RUN git checkout ${BYPASS4NETNS_VERSION} && \
+RUN set -eux ; git checkout ${BYPASS4NETNS_VERSION} && \
+  [ "${CANARY}" = "1" -o "$(git rev-parse HEAD)" = "$(cat /COMMITS.d/bypass4netns-${BYPASS4NETNS_VERSION})" ] && \
```
apostasie, line 115: Since we are ignoring commit verification for canary, maybe we could also ignore shasum verification - that would simplify the canary script significantly (eg: not having to retrieve archives and create shasum files on the fly).

Dockerfile
```diff
@@ -74,42 +79,54 @@ RUN git config --global advice.detachedHead false
...
 FROM build-base-debian AS build-runc
 ARG RUNC_VERSION
+ARG CANARY
 ARG TARGETARCH
+COPY ./Dockerfile.d/COMMITS.d/ /COMMITS.d
 RUN git clone https://github.com/opencontainers/runc.git /go/src/github.com/opencontainers/runc
 WORKDIR /go/src/github.com/opencontainers/runc
-RUN git checkout ${RUNC_VERSION} && \
+RUN set -eux ; git checkout ${RUNC_VERSION} && \
+  [ "${CANARY}" = "1" -o "$(git rev-parse HEAD)" = "$(cat /COMMITS.d/runc-${RUNC_VERSION})" ] && \
```
apostasie, line 101: Could we have the commits in ARGs inside the Dockerfile instead?
Maintaining separate independent files seem to increase complexity / maintenance? 

https://github.com/farcloser/lepton/blob/main/Dockerfile#L36

Dockerfile
AkihiroSuda, line 1: Checking if we can have a new release of BuildKit soon as well:
- https://github.com/moby/buildkit/pull/5844#issuecomment-2735296815

Dockerfile
AkihiroSuda: I'll update the PR to use a helper script with a single arg like `RUN git-checkout-tag-with-hash.sh v1.2.3@deadbeef`

AkihiroSuda: done

Dockerfile
AkihiroSuda: The variable is now ignored.
Can be discussed in another issue/PR.

hack/git-checkout-tag-with-hash.sh
```diff
@@ -0,0 +1,46 @@
...
+: "${GIT:=git}"
+TAG="$(echo "$1" | cut -d@ -f1)"
+HASH=""
+case "$1" in
+*@*)
+	HASH="$(echo "$1" | cut -d@ -f2)"
+	;;
+esac
+
+"$GIT" checkout "$TAG"
```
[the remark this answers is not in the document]
apostasie: LGTM - but if you feel like doing it in this PR, suggesting we do `quiet` and shallow clone.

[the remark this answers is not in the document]
AkihiroSuda: Cloning is out of the scope of this script.
Probably dockerfile's ADD git:// instruction should be used for cache optimality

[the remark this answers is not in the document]
apostasie: > Cloning is out of the scope of this script.

To clarify: there should be no need for a separate `checkout` operation at all, which is why I am bringing it up here.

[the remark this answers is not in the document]
fahedouch: > git clone --quiet --depth 1 --branch "$VERSION"

LGTM, the cmd is sandboxed in the ci container, no cache is left and no need to fetch the git entire history of the the tag.

hack/git-checkout-tag-with-hash.sh
```diff
@@ -0,0 +1,46 @@
...
+HEAD="$("$GIT" rev-parse HEAD)"
+if [ -z "$HASH" ]; then
+	echo >&2 "WARNING: ${TAG}: commit hash was not specified (got ${HEAD})"
+else
+	if [ "$HEAD" != "$HASH" ]; then
+		echo >&2 "ERROR: ${TAG}: expected ${HASH}, got ${HEAD}"
+		exit 1
+	fi
+fi
```
fahedouch, line 46: shallow fetch makes the `else` block useless. The following `cond` is always true:
  `"$HEAD" == "$HASH"`

Dockerfile
```diff
@@ -15,42 +15,42 @@
...
-# TODO: verify commit hash
-
 # Basic deps
-ARG CONTAINERD_VERSION=v2.0.3
-ARG RUNC_VERSION=v1.2.5
-ARG CNI_PLUGINS_VERSION=v1.6.2
+# @BINARY: the binary checksums are verified via Dockerfile.d/SHA256SUMS.d/<COMPONENT>-<VERSION>
+ARG CONTAINERD_VERSION=v2.0.4@1a43cb6a1035441f9aca8f5666a9b3ef9e70ab20
+ARG RUNC_VERSION=v1.2.6@e89a29929c775025419ab0d218a43588b4c12b9a
+ARG CNI_PLUGINS_VERSION=v1.6.2@BINARY
```
fahedouch, line 22: a comment to mention that module is `BINARY` is not enough ?

fahedouch:
There are people waiting for the release, let's focus on `v2.0.4 and then address all this refactoring later please

hack/git-checkout-tag-with-hash.sh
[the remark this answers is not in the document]
AkihiroSuda: `ADD git://` is available 
since 2022 (https://github.com/moby/buildkit/pull/2799) and out of experimental since 2023 (https://github.com/moby/buildkit/pull/3979)

https://docs.docker.com/reference/dockerfile/#adding-files-from-a-git-repository

This might be more cache-efficient when running tests with several containerd versions

Dockerfile
AkihiroSuda: What do you mean?

apostasie:
> There are people waiting for the release, let's focus on `v2.0.4` and then address all this refactoring later please

Suggesting we just merge this now.
It does bring in important updates, and is good enough in its current shape.
We can refine the controversial bits later.

hack/git-checkout-tag-with-hash.sh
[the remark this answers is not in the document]
apostasie: Moving that part of the conversation to #4022

hack/git-checkout-tag-with-hash.sh
AkihiroSuda: Doesn't seem so?

```bash
$ git clone --depth 1 --branch v2.0.4 https://github.com/containerd/containerd
$ cd containerd
$ ~/gopath/src/github.com/containerd/nerdctl/hack/git-checkout-tag-with-hash.sh v2.0.4@wrong-hash
HEAD is now at 1a43cb6 Merge commit from fork
ERROR: v2.0.4: expected wrong-hash, got 1a43cb6a1035441f9aca8f5666a9b3ef9e70ab20
```

AkihiroSuda:
Merging and making a release, but we will probably to make another new release soon with a new release of BuildKit

PR #4030, opened 2025-03-20
Dockerfile git clones to be quiet and shallow

### LINKED ISSUE #4022: [CI] Dockerfile enhancements: cloning

Breakout of #4021.
Also see overall Dockerfile analysis for context.

Refresher:
- we clone too much in size (full history), while we do not need it - github clones for a cold-cache run account for about 1G of content
- we clone too often - we should only ever clone any given project _once_ per run

Suggesting that:
1. we git clone --quiet (less noise on the logs)
2. we git clone --depth 1 (reduce network traffic and cache size)
3. we separate clone and build - *one clone* when building multi-arch, and use mounts for build to avoid unnecessary copies (reduce network traffic and cache size)

For reference, one way to do it:
- clone _once_ : https://github.com/farcloser/lepton/blob/main/Dockerfile#L278-L290
- build multi-arch with mount: https://github.com/farcloser/lepton/blob/main/Dockerfile#L534

Note that _if_ we want to keep building using other projects build scripts, we would have to bind mounts RW (as most of these scripts do not allow to build out of tree). I personally believe we should build directly ourselves in most cases instead of relying on Makefiles / scripts (see #4021 for reasons).
If we do not want that, or not feasible in certain cases, we just need a simple lock to prevent cross-platforms builds from conflicting: https://github.com/farcloser/lepton/blob/main/Dockerfile#L753-L755

Let me know if that plan (1, 2, 3) is fine or if you prefer something else.

### Additional context

_No response_

apostasie:
Plan A fixes the first two items in #4022 (see discussion for alternatives and context).

Third item can be folded into a more generic "make the Dockerfile friendlier to multi-arch builds", so, let's consider this here does fix #4022.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,7 +75,7 @@ ADD hack/git-checkout-tag-with-hash.sh /usr/local/bin/
 FROM build-base-debian AS build-containerd
 ARG TARGETARCH
 ARG CONTAINERD_VERSION
-RUN git clone https://github.com/containerd/containerd.git /go/src/github.com/containerd/containerd
+RUN git clone --quiet --depth 1 --branch "${CONTAINERD_VERSION%@*}" https://github.com/containerd/containerd.git /go/src/github.com/containerd/containerd
 WORKDIR /go/src/github.com/containerd/containerd
 RUN git-checkout-tag-with-hash.sh ${CONTAINERD_VERSION} && \
   mkdir -p /out /out/$TARGETARCH && \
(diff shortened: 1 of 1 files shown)


apostasie:
Kube failure is Docker Hub 429.

Rootless failure is #3513

fahedouch (approved):
Thanks

AkihiroSuda (approved):
Thanks

Issue #3996, opened 2025-03-07
[CI]: fix commands being stuck problem


apostasie:
Some commands being tested do deadlock sometimes.
This is particularly true for EL8, for yet undiagnosed reasons.

Because we do not timeout on command execution, the runner will just kill everything, and we do not get any information about which command (nor which test) did actually deadlock, and no logs of course.

We should be able to have some command execution default deadline in place at the test tooling level.
That way, commands being stuck for more than say, 5 minutes, should break and we would report a failed "timed-out test" along with appropriate logs.

@AkihiroSuda can you assign this to me please?

apostasie:
@fahedouch can you slate this for milestone 2.x.x?

Issue #3883, opened 2025-02-14
CI: icmd is not usable to test when binary expect a console


apostasie:
### Description

icmd passes an io.Multiwriter for stdout and stderr of the underlying command.

So (even if stdout is a proper pty), the command will still bail out if it requires a console (because "inappropriate ioctl for device").

That, along with the gymnastic in containerd/console - took me half a day to figure out why the eff I would not get anything on out with attach tests.

I am not sure we should continue using icmd.
The added value of it is rather limited (literally a few dozens lines of code that we are still using) , and I am not particularly interested in going over there to hack on it.

I *will* bypass it so that we can get #3591 in either way, so, might be time to bite the bullet and ditch it.

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
/dibs

I'll get us away from icmd once I am done with the CI-build series.

@AkihiroSuda can you assign?

apostasie:
@fahedouch can you slate this for milestone 2.x.x?

PR #4040, opened 2025-03-27
[CI] new command implementation for tests

apostasie:
This PR is a tigron changeset addressing a few important issues:
- #3883
- #3996
- unlogged issue: dangling processes left behind

In a shell: icmd does not provide value for us. Most of its features we do not use, and the simple wrapping around golang command exec is causing more issues than it solves - icmd does replace the process standard streams with its own buffers, which gets in the way of using ptys, does not provide proper timeout logic for stuck processes, and does not seem to kill process groups properly.

This PR does away with it, and replace it with an internal com.Command that does offer:
- proper timeout
- pty handling
- proper process group termination

... along with absorbing a large part of the logic that was previously baked into test/command.go (env filtering, cloning, command wrapping).

While this should technically be (mostly) transparent for tests, some of the logic changes with timeout and pty / stdin handling required adaptation of some functions signatures.

Specifically:
- `Background(timeout)` is now simply `Background()`, as the new `WithTimeout` helper now applies to any command, and not just the background-ed ones
- `WithPTY` no longer supports the extra "writers" argument. To write to stdin, the same methods can now be used whether or not we are working with a PTY
- the previous `WithStdin` method was ill-named/ill-advised, as it is not setting stdin, but writing to stdin, can be used repeatedly, and lacked flexibility. It has been replaced by Feed(io.Reader) for the simpler cases, and WithFeeder(func()io.Reader) that allows for more complex logic before writing to stdin (like Pty(writers) used to)

Finally, this PR also provides stuff that came up as necessary while working on the new command:
- minimal improvements over debugging information output readability
- (cosmetic) changes to prepare for version 2 of golangci
- the testing of internal Command itself did motivate the addition of leak detection and an internal assert library

While on the surface this PR is minimally impactful, this is actually a profound change, and I assume a few rounds will be needed before taking it out of "draft". Specifically, the current default timeout for commands execution is set at 3 minutes, and we might want to adjust that (eg: shorten it generally, and set it explicitly for tests that require longer)

It is ready for review though.


mod/tigron/internal/com/package_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+package com_test
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
```
AkihiroSuda, line 25: Can we avoid mixing up multiple logging libraries?

If some feature is missing in containerd/log, please open an issue in https://github.com/containerd/log

mod/tigron/internal/highk/fileleak.go
```diff
@@ -0,0 +1,117 @@
...
+
+	_ = cmd.Run()
+
+	_, _ = file.Seek(0, 0)
+
+	return io.ReadAll(file)
+}
+
+// Diff will return a slice of strings showing the diff between two strings.
+func Diff(one, two string) []string {
```
AkihiroSuda, line 64: Can we have a unit test?

mod/tigron/internal/assertive/assertive.go
```diff
@@ -0,0 +1,169 @@
...
+	t.Helper()
+
+	if !errors.Is(err, compErr) {
+		t.Log("expected error to be:", compErr, "- instead it is:", err)
+		failNow(t, msg...)
+	}
+}
+
+// IsEqual immediately fails a test if the two interfaces are not equal.
+func IsEqual(t testingT, actual, expected interface{}, msg ...string) {
```
AkihiroSuda, line 53:
```suggestion
func IsEqual(t testingT, expected, actual interface{}, msg ...string) {
```

To follow https://pkg.go.dev/github.com/stretchr/testify/assert#Equal

mod/tigron/internal/com/package_test.go
apostasie: containerd/log is based on logrus - which is in maintenance, has bad performance compared to more modern loggers and is not resource efficient. IMHO logrus is not a good choice for new projects...

mod/tigron/internal/highk/fileleak.go
apostasie: This is a very rough first draft for the fileleak detector that is solely used internally by tigron for its own testing.
It does its job for that limited purpose, but before I turn it on for others, a lot of work should happen on it (probably rewriting parts of lsof in go), so, most of that code is throwable at this point - tests will come later on with better code.

apostasie:
@AkihiroSuda thanks for the early review on this.

I am breaking it down in smaller chunks to simplify the discussion, so, let see what remains here once we are done with the more trivial stuff in the breakout PRs.

mod/tigron/internal/assertive/assertive.go
apostasie: Conversation moved to #4042.

mod/tigron/internal/highk/fileleak.go
apostasie: Moved to #4044

mod/tigron/internal/com/package_test.go
apostasie: @AkihiroSuda I do appreciate that nerdctl does not want a second logging library though (for the very same reason I do not want logrus in my projects...).

What about I come up with an interface for the logger inside Tigron, and then people use whatever logging system they want?

apostasie: Either way Tigron should not be into logging at all and should not have opinions on this. Most of the framework should and does use `t.Log`. The usefulness of the logger is and should stay limited to advanced debugging of internal packages.

AkihiroSuda: > an interface for logger

👍

mod/tigron/expect/exit.go
```diff
@@ -21,9 +21,12 @@ const (
 	ExitCodeSuccess = 0
 	// ExitCodeGenericFail will verify that the command ran and exited with a non-zero error code.
 	// This does NOT include timeouts, cancellation, or signals.
-	ExitCodeGenericFail = -1
+	ExitCodeGenericFail = -10
```
apostasie, line 24: Note: in some cases, go exec may return -1 exit code, which would be colliding with these - which is why this is getting moved to -10 etc

mod/tigron/internal/com/package_test.go
apostasie: Yanked out the logger, so marking as resolved - but please don't waste time reviewing right now until I hammer the kinks out and we merge all the dependents.

apostasie:
Geeeeeeeez github.

@AkihiroSuda moving to #4056

Sorry about that.

Done:
- yanked out the logger
- CI is (should be) green
- cleaned-up the commits

PR #4041, opened 2025-03-27
[CI] Tigron breakout: golangci v2 and cleanups

apostasie:
This upgrades golangci-lint to v2, cleans-up the Makefile a bit, and a few cosmetic cleanups of the codebase in preparation for more drastic changes.

Issue #3971, opened 2025-03-04
[LEGACY] [CI]: NoneNetworkManager is presumably flaky wrt hostname


apostasie:
### Description

https://github.com/containerd/nerdctl/actions/runs/13644906693/job/38141881547?pr=3965#step:11:452

```
=== RUN   TestNoneNetworkHostName
=== PAUSE TestNoneNetworkHostName
=== CONT  TestNoneNetworkHostName
    container_run_network_linux_test.go:944: 
    container_run_network_linux_test.go:944: ======================== Pre-test cleanup ========================
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test rm -f testnonenetworkhostname-e70856d3
    container_run_network_linux_test.go:944: 
    container_run_network_linux_test.go:944: ======================== Test setup ========================
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test run -d --name testnonenetworkhostname-e70856d3 --network none ghcr.io/stargz-containers/nginx:1.19-alpine-org
    container_run_network_linux_test.go:944: 
    container_run_network_linux_test.go:944: ======================== Test Run ========================
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test exec testnonenetworkhostname-e70856d3 cat /etc/hostname
    command.go:113: assertion failed: 
        --- compare
        +++ stdout
        @@ -1,2 +1 @@
        -0ba62cd1c8fe
         
        : 
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test exec testnonenetworkhostname-e70856d3 cat /etc/hostname
        ExitCode: 0
        Stdout:   
        Stderr:   
        Env:
        HOSTNAME=3a911ce1281e
        container=docker
        HOME=/root
        LANG=C.UTF-8
        INVOCATION_ID=6eadd4f97625[456](https://github.com/containerd/nerdctl/actions/runs/13644906693/job/38141881547?pr=3965#step:11:457)0a6a1ac30bc97d4aa
        TERM=xterm
        SHLVL=3
        CGO_ENABLED=0
        _=/usr/local/bin/gotestsum
        PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
        ***
        DOCKER_CONFIG=/tmp/TestNoneNetworkHostName1438939296/001
[...]

PR #4042, opened 2025-03-27
[CI] Tigron breakout 2: remove gotest.tools assert dependency internally

apostasie:
This is on top of #4041.

This PR removes our dependency on `gotest.tools/v3/assert` and replaces it with a much simpler, no-dependency `internal/assertive`.

People can still use whatever assert library they want while writing their tests (gotest, stretchr/testify, or else) - but there is no reason for Tigron internally to use any.

### DIFF
--- a/mod/tigron/internal/assertive/assertive.go
+++ b/mod/tigron/internal/assertive/assertive.go
@@ -1,0 +1,169 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package assertive
+
+import (
+	"errors"
+	"strings"
+	"time"
+)
+
+type testingT interface {
+	Helper()
+	FailNow()
+	Fail()
+	Log(args ...interface{})
+}
+
+// ErrorIsNil immediately fails a test if err is not nil.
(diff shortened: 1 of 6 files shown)


mod/tigron/internal/assertive/assertive.go
```diff
@@ -0,0 +1,169 @@
...
+	t.Helper()
+
+	if !errors.Is(err, compErr) {
+		t.Log("expected error to be:", compErr, "- instead it is:", err)
+		failNow(t, msg...)
+	}
+}
+
+// IsEqual immediately fails a test if the two interfaces are not equal.
+func IsEqual(t testingT, actual, expected interface{}, msg ...string) {
```
AkihiroSuda, line 53: > ```suggestion
> func IsEqual(t testingT, expected, actual interface{}, msg ...string) {
> ```
> 
> To follow the existing standard:
> - https://pkg.go.dev/github.com/stretchr/testify/assert#Equal
> - https://junit.org/junit4/javadoc/4.8/org/junit/Assert.html
> - etc.
> 

_Originally posted by @AkihiroSuda in https://github.com/containerd/nerdctl/pull/4040#discussion_r2015745287_

apostasie: The problem I have with this is that it is not consistent internally.

testify ErrorIs for example is the other way round (first actual error, then target error), which I find confusing everytime.

Also, gotesttool is `actual`, `expected` as well:
https://github.com/gotestyourself/gotest.tools/blob/main/assert/assert.go#L188

Anyhow, I do not intend on making `assertive` public. People can and should use established assert libraries when they write their tests if they need them. This here is meant to stay internal.

apostasie:
EL failure is #3971

apostasie:
Rebased.

mod/tigron/internal/assertive/assertive.go
apostasie: Going to mark this as resolved.
Suggestion is that _if_ in the future `assertive` moves from internal to public, we can, and should, have a discussion on API (and can have that conversation on parameters order).

Lmk if you still think this is an issue right now and I'll un-resolve.

PR #4048, opened 2025-03-27
[CI] Update golangci to v2

apostasie:
This updates golangci to their newly released v2.

1. convert the configuration file to the new format
2. update the makefile install task (also bump version for other tools we use)
3. remove go-imports-reviser - we can now use `gci` (bundled with golangci) with 99% compatibility (save a handful of changes required in code)
4. remove fix-imports, lint-imports (see ^)

I opted to configure the new .golangci.yml for the minimal set of changes to the source code.
However, that meant disabling a bunch of checks from `staticcheck` and a couple of formatters, that we may want to investigate / enable.

PR #4051, opened 2025-03-28
[CI] golangci v2: review, fix, document, rules

apostasie:
@fahedouch @AkihiroSuda 

This comes on top of #4048 of course.

#4048 changed almost nothing in code.

This here:
- fixes a handful of code issues allowing us to re-enable rules
- re-organize the golangci yml file to document rules and priorities to fix

For the first part that actively change code, each fix is a separate commit, so, it should be easy to just drop one if we do not agree that it was necessary:
- revive redundant-import-alias
- revive struct-tag
- revive time-equal
- revive unexported-naming
- revive if-return
- revive filename-format
- revive import-alias-naming
- static check QF1004
- static check ST1005

For the second part, I did organize rules in groups:
- P0: must fix ASAP
- P1: not critical, but should fix
- P2: nice to have
- Disabled: we don't want them

I also added the number of occurrences for each failing rule, giving the reader a sense of how much of an effort that would be.

This of course is just my (informed) personal opinion on things, so, happy to change rules to go to different categories based on feedback, so, hit me up.

I also re-enabled gocritic (and disabled prealloc), but have not reviewed its ruleset yet.


apostasie:
CI failure is unrelated (#3513 )

AkihiroSuda:
Needs rebase

apostasie:
Rebased.

AkihiroSuda (approved):
Thanks

Issue #4046, opened 2025-03-27
[CI]: buildg times-out


apostasie:
### Description

This was first spotted in https://github.com/containerd/nerdctl/pull/4007#issuecomment-2738540499 and seen again in #4044 

Message is definitely coming from https://github.com/ktock/buildg/blob/main/pkg/buildkit/client.go#L83

It is unclear to me if this is somehow builkit timeouting on GHA cache (but why then?), or something else that would be buildg specific.

Tagging @ktock for insight?

### Describe the results you received and expected

```
=== RUN   TestBuilder/Debug
    builder_builder_test.go:159: 
    builder_builder_test.go:159: ======================== Pre-test cleanup ========================
    builder_builder_test.go:159: 
    builder_builder_test.go:159: ======================== Test setup ========================
    builder_builder_test.go:159: 
    builder_builder_test.go:159: ======================== Test Run ========================
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test builder debug /tmp/TestBuilderDebug385954703/001
    command.go:113: Expected exit code: 0
        
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test builder debug /tmp/TestBuilderDebug385954703/001
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   timed out to access cache storage. other debug session is running?
        
        Env:
        SHELL=/bin/bash
        LOGNAME=rootless
        XDG_SESSION_TYPE=tty
        HOME=/home/rootless
        LANG=C.UTF-8
        SSH_CONNECTION=::1 41924 ::1 22
        XDG_SESSION_CLASS=user
        IPFS_PATH=/home/rootless/.local/share/ipfs
        USER=rootless
        SHLVL=1
        XDG_SESSION_ID=5
        XDG_RUNTIME_DIR=/run/user/1001
        SSH_CLIENT=::1 [419](https://github.com/containerd/nerdctl/actions/runs/14111164974/job/39530341996?pr=4044#step:8:420)24 22
        DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1001/bus
        OLDPWD=/home/rootless
        _=/usr/local/bin/gotestsum
        PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
        ***
[...]

PR #4054, opened 2025-03-28
[fix] Avoid console SetRaw

### LINKED ISSUE #4053: nerdctl reliance on containerd/console SetRaw is resulting in spurious carriage returns (unix.OPOST)
### Description

Because of that, when using ptys, we end-up with double CR.

This is likely because of containerd/console SetRaw method which does enable unix.OPOST.
I do not know why console is doing that.
Crosby added this in https://github.com/containerd/console/commit/362934a0b1f1e8e5beed27401328c4c45b1a1df6

I will send a PR here on nerdctl to bypass calling `console.SetRaw` and instead just use `term.MakeRaw` and see if it causes any unexpected side-effects.

```
sudo unbuffer docker run -it --rm --net=host ghcr.io/stargz-containers/alpine:3.13-org echo this was always working | hexdump
0000000 6874 7369 7720 7361 6120 776c 7961 2073
0000010 6f77 6b72 6e69 0d67 000a
```

vs.

```
unbuffer nerdctl run -it --rm --net=host ghcr.io/stargz-containers/alpine:3.13-org echo this was always working | hexdump
0000000 6874 7369 7720 7361 6120 776c 7961 2073
0000010 6f77 6b72 6e69 0d67 0a0d
000001a
```

nerdctl should not do that.

main

None

### Host information

_No response_

apostasie:
Fix #4053

### DIFF
--- a/cmd/nerdctl/container/container_run.go
+++ b/cmd/nerdctl/container/container_run.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 
 	"github.com/containerd/console"
 	"github.com/containerd/log"
@@ -407,7 +408,7 @@ func runAction(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		defer con.Reset()
-		if err := con.SetRaw(); err != nil {
+		if _, err := term.MakeRaw(int(con.Fd())); err != nil {
 			return err
 		}
 	}
--- a/pkg/cmd/container/attach.go
+++ b/pkg/cmd/container/attach.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 
+	"golang.org/x/term"
+
 	"github.com/containerd/console"
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/containerd/containerd/v2/pkg/cio"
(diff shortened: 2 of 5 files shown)


apostasie:
First failure is unrelated (#4046)

apostasie:
Note: this is blocking #4040.

apostasie:
I would really like to know why does containerd/console force OPOST.

@AkihiroSuda any idea?

apostasie:
Note: that looks like the second strike for containerd/console (the first one, we had to fork out console.GetCurrent because it is panicking instead of returning an error).

Now avoiding their SetRaw method, the only thing that is left in there is 2 calls for termios (tcget / tcset).

I just don't see the value anymore.

Suggesting we remove the dependency entirely.

AkihiroSuda (approved):
Thanks, can we have a test?

apostasie:
> Thanks, can we have a test?

Yes. Let me look into this soon once the dust settles with all the PRs.

apostasie:
Actually, we already have a test - albeit an indirect one.

`TestIssue108` (and probably others)

When using icmd, icmd exec will replace the command Stdout with an io.Multiwriter.
Thus it will not be seen as a pty for the command (albeit _Stdin_ is a pty).
Because of that, the bug is "hidden".

When moving to our new Command implementation, Stdout will be *properly* set to the pty, which will make `TestIssue108` fail (as it will carry the double CR).

PR #4045, opened 2025-03-27
[CI] Tigron breakout 4: adding a few tests 

apostasie:
Small breakout.
Just adding some simple tests for data, config, requirements.

PR #4044, opened 2025-03-27
[CI] Tigron breakout 3: leak detection

apostasie:
On top of the two previous breakouts.

This PR adds `internal/highk` in preparation for the command rewrite.

Currently, the implementation relies on Uber library for go routines, and on parsing the output of lsof for fd.

I have used it intensively while rewriting Command. It has been very useful so far, but it certainly needs more work.

Proposal is to first start using it internally for Tigron own testing (as for the upcoming Command), hammer the kinks and then consider adding it as a feature of the test framework.

For an example on how it can be used at this point: https://github.com/containerd/nerdctl/blob/958c566a121d22391e121faf0e0fda39329188ed/mod/tigron/internal/com/package_test.go#L55-L95

PR #4056, opened 2025-03-29
[CI] new command implementation for tests

### LINKED ISSUE #3883: CI: icmd is not usable to test when binary expect a console

### LINKED ISSUE #3996: [CI]: fix commands being stuck problem

apostasie:
Note this is a large PR, though the bulk of the changes are under a couple of files in `mod/tigron/internal`, and probably half of it is just tests.

This PR is a tigron changeset addressing a few important issues:
- fix #3883
- fix #3996
- unlogged issue: dangling processes left behind

In a shell: icmd does not provide value for us. Most of its features we do not use, and the simple wrapping around golang command exec is causing more issues than it solves - icmd does replace the process standard streams with its own buffers, which gets in the way of using ptys, does not provide proper timeout logic for stuck processes, and does not seem to kill process groups properly.

This PR does away with it, and replace it with an internal com.Command that does offer:
- proper timeout
- pty handling
- proper process group termination

... along with absorbing a large part of the logic that was previously baked into test/command.go (env filtering, cloning, command wrapping).

While this should technically be (mostly) transparent for tests, some of the logic changes with timeout and pty / stdin handling required adaptation of some functions signatures.

Specifically:
- `Background(timeout)` is now simply `Background()`, as the new `WithTimeout` helper now applies to any command, and not just the background-ed ones
- `WithPTY` no longer supports the extra "writers" argument. To write to stdin, the same methods can now be used whether or not we are working with a PTY
- the previous `WithStdin` method was ill-named/ill-advised, as it is not setting stdin, but writing to stdin, can be used repeatedly, and lacked flexibility. It has been replaced by `Feed(io.Reader)` for the simpler cases, and `WithFeeder(func()io.Reader)` that allows for more complex logic before writing to stdin (like `Pty(writers)` used to)

Note that we might want to tweak the current default timeout for commands execution (3 minutes rn).

Blocked by:
- #4041 
- #4042
- #4044 
- #4045 
- #4047 
- #4054
- #4058

Preferable to merge after:
- #4048 
- #4051 
- #4062

TODO:
- [x] green CI
- [x] mute the logging from pipes by default


cmd/nerdctl/container/container_inspect_linux_test.go
```diff
@@ -30,6 +30,7 @@ import (
 
 	"github.com/containerd/nerdctl/mod/tigron/expect"
 	"github.com/containerd/nerdctl/mod/tigron/test"
+
```
apostasie, line 33: Should disappear once we get the golangci v2 PR in.

apostasie:
@AkihiroSuda I will deal with the rebasing once we get all the dependent PR merged-in, but this is ready for review for all intent and purposes.

I am not sure we can break this down further in individual PRs (I did break it in individual commits though, that should help review), but LMK if you want me to yank something specific out in a separate PR.

apostasie:
Reviewers: all commits except one are mundane.

The _one that matters_ is: https://github.com/containerd/nerdctl/pull/4056/commits/2bc62a52058dd5be66321e9fb4d18c0d3c88c974#diff-66fe4e3b3287c94bcd9942c86d92bb7259988cbc2aa6345a47dc2b56d14f411b - which contains the new command implementation.
Every other commit is to adapt tests to that.

apostasie:
I guess we are having a birthday party today - for ghcr.io/stargz-containers/alpine:3.13-org

CreatedAt: "2021-03-31T17:21:23Z"

So, this is no longer "3 years ago" - it is now "4 years ago"...

Need to do something for that test.

Vagrantfile.freebsd
```diff
@@ -59,7 +59,7 @@ Vagrant.configure("2") do |config|
         set -eux -o pipefail
         daemon -o containerd.out containerd
         sleep 3
-        /root/go/bin/nerdctl run --rm --net=none dougrabson/freebsd-minimal:13 echo "Nerdctl is up and running."
+        /root/go/bin/nerdctl run --rm --quiet --net=none dougrabson/freebsd-minimal:13 echo "Nerdctl is up and running."
```
AkihiroSuda, line 62: ?

AkihiroSuda (approved):
Thanks

Issue #3509, opened 2024-10-07
content digest not found: platform-reduced push, follow-up to #3425


apostasie:
### Description

Similarly to #3489, it seems like there may be another case that was not listed in #3425 - or something having side-effects there, when we rely on a temp-reduced-plaform to push.

It is not entirely clear to me what is going on in there - especially what is happening with the whole reduced-platform contorsions (and why we even need this...) - so, I do not have a quick fix that I can send immediately.

While this was somewhat expected... https://github.com/containerd/nerdctl/pull/3435#discussion_r1766200030 - we should have a hard look at this code.

### Steps to reproduce the issue

Hammer the tests.

It took about 20 iterations of
```
go test ./cmd/nerdctl/images -count 1
```

to get there.

Note that this was produced with the rewritten tests in #3492.

### Describe the results you received and expected

```
--- FAIL: TestPush (14.40s)
    image_push_linux_test.go:253: ======================== Pre-test cleanup ========================
    image_push_linux_test.go:253: ======================== Test setup ========================
    image_push_linux_test.go:253: ======================== Test Run ========================
    image_push_linux_test.go:253: ======================== Processing subtests ========================
    --- FAIL: TestPush/with_hosts_dir,_with_login (0.88s)
        image_push_linux_test.go:253: ======================== Pre-test cleanup ========================
        command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test rmi
        image_push_linux_test.go:253: ======================== Test setup ========================
        command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test pull ghcr.io/stargz-containers/alpine:3.13-org
        command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test tag ghcr.io/stargz-containers/alpine:3.13-org 192.168.5.15:5002/testpush-with-hosts-dir-with-login-a6c55caa:3.13-org
        command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test --hosts-dir /tmp/TestPush2005531285/007/certs.d2208102383 login -u admin -p badmin 192.168.5.15:5002
        image_push_linux_test.go:253: ======================== Test Run ========================
        command.go:112: /usr/local/bin/nerdctl --namespace=nerdctl-test push --hosts-dir /tmp/TestPush2005531285/007/certs.d2208102383 192.168.5.15:5002/testpush-with-hosts-dir-with-login-a6c55caa:3.13-org
[...]

PR #4062, opened 2025-03-31
[P0] Fix #3513 (`[MAJOR] content digest not found`)

### LINKED ISSUE #3513: [MAJOR] content digest not found

apostasie:
### Current status

Discussion and patch upstream: https://github.com/containerd/containerd/pull/11628

TL;DR here: this PR here on nerdctl _does seem to fix our issues_ (many runs of the CI with no issue), because it entirely bypasses `converter.Convert` `Delete` logic which _seems to trigger some underlying containerd issue related to lease and deletion_ (<- my opinion, unverified).

My patch upstream does ALSO fix our issue, with a clear reproducer, although it probably just avoids the underlying containerd problem.

-------------------------------------------------------------
### Original post below

It looks like containerd `imagestore.Delete` is asynchronous by default.

If you want a synchronous delete, you have to pass an additional opt.

Of course, this seems to be a problem when you:
- check if the image exist
- delete it (async) if it does
- create a new version of it

This seem to be the case (actually, it is worse) for https://github.com/containerd/containerd/blob/af013606262f9df1de4cc80688725a15a375fb3a/core/images/converter/converter.go#L120-L121
... where convert.Convert just forcefully delete, whether or not the image existed, async...

This of course is also something we are doing internally.

This PR addresses these issues:
- ensure we delete SYNCHRONOUSLY where it matters
- honor the async option when there is one
- wrap / workaround containerd (* maybe) faulty implementation of convert

At this point, this is still a hunch - but for the first time I feel confident that we may be reaching the bottom of it, as this explanation closely matches the soul-destroying symptoms we have witnessed in eg: https://github.com/containerd/nerdctl/issues/3509#issuecomment-2398236766 

and more generally, with anything touching containerd `Convert` (see #3513).

I will close an open this PR many times and confirm that the infamous #3513 is finally gone, but I would like the expert opinion of @lingdie on this - if you have time for a quick review.

Upstream PR: https://github.com/containerd/containerd/pull/11628

Note that this PR here does not try to do more than that - there MAY still be places where we would need to `EnsureContent`.

EDIT: looks like containerd maintainer disagree with this, so, maybe I am misreading this.
Anyhow, I will keep closing / opening and run the CI - we are currently at:
- four runs without failure
- fifth run failed on Github cache
- sixth run failure is #4046

### DIFF
--- a/pkg/cmd/builder/build.go
+++ b/pkg/cmd/builder/build.go
@@ -127,7 +127,7 @@ func Build(ctx context.Context, client *containerd.Client, options types.Builder
 			if _, err := imageService.Create(ctx, image); err != nil {
 				// if already exists; skip.
 				if errors.Is(err, errdefs.ErrAlreadyExists) {
-					if err = imageService.Delete(ctx, targetRef); err != nil {
+					if err = imageService.Delete(ctx, targetRef, images.SynchronousDelete()); err != nil {
 						return err
 					}
 					if _, err = imageService.Create(ctx, image); err != nil {
--- a/pkg/imgutil/converter/convert.go
+++ b/pkg/imgutil/converter/convert.go
@@ -1,0 +1,55 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package converter
+
+import (
+	"context"
+
+	"github.com/containerd/containerd/v2/core/images"
+	"github.com/containerd/containerd/v2/core/images/converter"
+)
+
+// Something seems wrong in converter.Convert.
+// When dstRef != srcRef, convert will first forcefully delete dstRef,
+// *asynchronously*, then create the image.
+// This seems to cause a race conditions, and the deletion may kick in after the creation.
+// This here is to workaround the bug, by manually creating the image first,
+// then converting it in place (which avoid the problematic code-path).
+// See containerd upstream discussion https://github.com/containerd/containerd/pull/11628 and
+// nerdctl issues:
+// https://github.com/containerd/nerdctl/issues/3509#issuecomment-2398236766
+// https://github.com/containerd/nerdctl/issues/3513
+// Note this should be remove if/when containerd merges in a fix.
+
+func Convert(ctx context.Context, client converter.Client, dstRef, srcRef string, opts ...converter.Opt) (*images.Image, error) {
+	imageService := client.ImageService()
+
+	img, err := imageService.Get(ctx, srcRef)
+	if err != nil {
+		return nil, err
+	}
+
+	img.Name = dstRef
+
+	_ = imageService.Delete(ctx, img.Name, images.SynchronousDelete())
+
+	if _, err = imageService.Create(ctx, img); err != nil {
+		return nil, err
+	}
+
+	return converter.Convert(ctx, client, dstRef, dstRef, opts...)
+}
--- a/pkg/cmd/image/convert.go
+++ b/pkg/cmd/image/convert.go
@@ -190,7 +190,7 @@ func Convert(ctx context.Context, client *containerd.Client, srcRawRef, targetRa
 	}
 
 	// converter.Convert() gains the lease by itself
-	newImg, err := converter.Convert(ctx, client, targetRef, srcRef, convertOpts...)
+	newImg, err := converterutil.Convert(ctx, client, targetRef, srcRef, convertOpts...)
 	if err != nil {
 		return err
 	}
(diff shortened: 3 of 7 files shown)


apostasie:
Green once.

pkg/cmd/builder/build.go
```diff
@@ -127,7 +127,7 @@ func Build(ctx context.Context, client *containerd.Client, options types.Builder
 			if _, err := imageService.Create(ctx, image); err != nil {
 				// if already exists; skip.
 				if errors.Is(err, errdefs.ErrAlreadyExists) {
-					if err = imageService.Delete(ctx, targetRef); err != nil {
+					if err = imageService.Delete(ctx, targetRef, images.SynchronousDelete()); err != nil {
```
apostasie, line 130: This _may_ explain some of the "internal: not found" errrors we sometimes see when building.

apostasie:
Green twice. 😓

pkg/imgutil/converter/convert.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/containerd/containerd/v2/core/images"
+	"github.com/containerd/containerd/v2/core/images/converter"
+)
+
+// It looks like containerd converter.Convert is faulty.
+// When dstRef != srcRef, convert will first forcefully delete dstRef,
+// apparently *asynchronously*, then create the image.
+// This is racy, and the deletion may kick in after the creation.
+// This here is to workaround the bug, by manually creating the image first,
```
AkihiroSuda, line 30: Could you add a link to the issue ticket?

apostasie: Done.
Massaged the text a bit too.

pkg/imgutil/converter/convert.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+
+	"github.com/containerd/containerd/v2/core/images"
+	"github.com/containerd/containerd/v2/core/images/converter"
+)
+
+// Something seems wrong in converter.Convert.
+// When dstRef != srcRef, convert will first forcefully delete dstRef,
+// *asynchronously*, then create the image.
+// This seems to cause a race conditions, and the deletion may kick in after the creation.
```
fahedouch, line 29: the lease is not maintained during the creation ? 🤔

pkg/imgutil/converter/convert.go
```diff
@@ -0,0 +1,55 @@
...
+	imageService := client.ImageService()
+
+	img, err := imageService.Get(ctx, srcRef)
+	if err != nil {
+		return nil, err
+	}
+
+	img.Name = dstRef
+
+	_ = imageService.Delete(ctx, img.Name, images.SynchronousDelete())
```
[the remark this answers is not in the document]
apostasie: It does not seem like we do need to do it (note that I am not 100% familiar with it yet).

Indeed, my containerd patch does away with the delete (which is the source of the problem), and instead tries to `update`, then if it fails tries to `create`.

This patch here is opting for the least amount of change, but of course happy to clean-up a bit more if we want.

lmk your preference.

pkg/imgutil/converter/convert.go
apostasie: No clue.

My gut intuition is that this thing faceplants when the old ref is actually exactly the same image as new ref.

Note that patching containerd to also use a **sync** delete does NOT fix the issue. 🤔

Anyhow, for me the bottom-line really is:
- do not use leases + delete - clealy buggy
- do not use async delete - counter-intuitive

pkg/imgutil/converter/convert.go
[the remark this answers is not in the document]
fahedouch: LGTM to unblock the situation. But in the short and medium term, I don't want to move the containerd logic to nerdctl.

fahedouch (approved):
lgtm to fix these problems that have been going on for a long time but the real fix should be done on the containerd side

apostasie:
> lgtm to fix these problems that have been going on for a long time but the real fix should be done on the containerd side

Purely for the `converter.Convert` part, proposal is there: https://github.com/containerd/containerd/pull/11628 if you feel like it is right.

I will _eventually_ get to the true bottom of this pit and figure out if there is indeed a bug / race in the deletion mechanism (which I do believe there is), but I need time and motivation...

apostasie:
Suggesting we merge this soon.

I would like to get as much CI watching time with that patch merged-in before it gets into the general audience with the next patch release.

fahedouch:
> Suggesting we merge this soon.
> 
> I would like to get as much CI watching time with that patch merged-in before it gets into the general audience with the next patch release.

I would have preferred to see @AkihiroSuda 's opinion on this, but he is certainly busy with KubeCon. We will merge this and roll back later if needed.

fahedouch (approved):
I believe that the real fix should be done on the daemon side, but to unblock multiple issues we have on the CLI, let's merge the fix and see what the patch on the containerd side will bring.

AkihiroSuda:
Thanks, LGTM

PR #4063, opened 2025-03-31
fix: Pass ps_args as option to container top

coderbirju:
Currently the arguments for the ps command is parsed out in the [top.go file](https://github.com/containerd/nerdctl/blob/00558aabc88e066658e4dc79cf5278c77833ddda/pkg/cmd/container/top.go#L63)
this PR moves the parsing logic to the parent function and passes the parsed values via a new field in the [ContainerTopOptions](https://github.com/containerd/nerdctl/blob/00558aabc88e066658e4dc79cf5278c77833ddda/pkg/api/types/container_types.go#L342)
This will enable nerdctl lib users to pass ps_args values. 

Tests are already present for these in [cmd/nerdctl/container/container_top_test.go](https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container/container_top_test.go) 
I've just moved the parsing from child function to parent, no change in logical behavior

### DIFF
--- a/cmd/nerdctl/container/container_top.go
+++ b/cmd/nerdctl/container/container_top.go
@@ -19,6 +19,7 @@ package container
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
What was the issue?
How to test this PR?

coderbirju:
Updated the PR description

cmd/nerdctl/container/container_top.go
```diff
@@ -66,9 +67,18 @@ func topAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	defer cancel()
-	return container.Top(ctx, client, args, types.ContainerTopOptions{
+
+	containerID := args[0]
```
fahedouch, line 71: Please check len to avoid out-of-range panic

manugupt1: minNArgs should cover it I think

coderbirju: Yes, cobra.MinimumNArgs(1) should prevent this from becoming out of range

fahedouch (approved):
thanks

coderbirju:
@AkihiroSuda PTAL when you get a chance

coderbirju:
Retrying for flaky tests

PR #4076, opened 2025-04-03
[Tigron]: mocks + debugging output enhancements

apostasie:
<!> Blocked by <!>:
- [x] #4056

Should preferably be merged after:
- [x] #4062 (purely for QOL reasons)

----------------------------------

This is a new series of Tigron PRs (FIXME: currently on top of #4056).

Now that the big change with the new command implementation is in, upcoming work on Tigron will focus on:
- **P0**: internal testing, getting to the point we can absolutely trust Tigron to **never** faceplant
- **P0**: debugging output readability and other developers quality of life improvements
- **P1**: code cleanup and simplification

Ideally, none of these PRs will require ANY change in nerdctl, and are either changes to the way we test Tigron itself, or refactoring that does not change signatures and API.

Efforts are being made to break these PRs down into small, digestible changes to ease and fasten review, which explains the chain of PRs.

LMK at any point if a given PR needs to be further broken down.

----------------------------------

This first PR is focused on making Tigron more testable, starting with `assertive`.

Commit 1 is mundane and just disables golangci linters that we do not want.

Second commit does introduce `internal/mimicry`, a mock mechanism that will allow mocking `*testing.T` and other conendrums. See [commit message for details](https://github.com/containerd/nerdctl/pull/4076/commits/ed47d87ba1f6f6b30401aae1f66d69415c3addb3).

Third commit is work on `internal/assertive` (code cleanup, methods generalization), and a raft of new tests (leveraging `mimicry`). See [commit message for details](https://github.com/containerd/nerdctl/pull/4076/commits/f5cf3692e6bf11249fbf76a1b221aa2c36c00be9)

See following comments for preview of how failures will now be presented.

**Note that more work on debugging output quality will very soon come in an upcoming PR** (this time focused on general output, which is currently crowded, noisy and really hard to read) - I would prefer to keep that separate to keep PRs small, though meanwhile **the display will be a bit loosy goosy** because of that.

### DIFF
--- a/mod/tigron/tig/t.go
+++ b/mod/tigron/tig/t.go
@@ -1,0 +1,40 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package tig
+
+// T is what Tigron needs from a testing implementation (*testing.T obviously satisfies it).
+//
+// Expert note: using the testing.TB interface instead is tempting, but not possible, as the go authors made it
+// impossible to implement (by declaring a private method on it):
+// https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/testing/testing.go;l=913-939
+// Generally speaking, interfaces in go should be defined by the consumer, not the producer.
+// Depending on producers' interfaces make them much harder to change for the producer, harder (or impossible) to mock,
+// and decreases modularity.
+// On the other hand, consumer defined interfaces allows to remove direct dependencies on implementation and encourages
+// depending on abstraction instead, and reduces the interface size of what has to be mocked to just what is actually
+// needed.
+// This is a fundamental difference compared to traditional compiled languages that forces code to declare which
+// interfaces it implements, while go interfaces are more a form of duck-typing.
+// See https://www.airs.com/blog/archives/277 for more.
+type T interface {
+	Helper()
+	FailNow()
+	Fail()
+	Log(args ...any)
+	Name() string
+	TempDir() string
+}
--- a/mod/tigron/.golangci.yml
+++ b/mod/tigron/.golangci.yml
@@ -13,11 +13,23 @@ issues:
 linters:
   default: all
   disable:
-    - cyclop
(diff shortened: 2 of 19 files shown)


apostasie:
Here is how the new assertive debug will be presented.

The first line is enriched with the exact content from the source test file that triggered the failure, as a clickable OCS8 link so that the developer can directly open the offending file.

The following lines present clearly what was expected, what was the actual value, with some high-contrast emojis to facilitate reading.

Screenshot 2025-04-03 at 12 18 27 PM

This is to be compared with the normal "assert.Assert" information:

Screenshot 2025-04-03 at 12 23 08 PM

apostasie:
Here is how the captured call traces will show when using mocks:

In that specific case, the test wants to make sure that the mocked method `FailNow` was NOT called.
Since the method **was** called, the test fails, and shows the call stack from where the offending call was made, along with the arguments that were passed to the function (well, `[]` in that case...) and the call time.

Source links are clickable as well.

Screenshot 2025-04-03 at 12 17 30 PM

mod/tigron/tig/t.go
```diff
@@ -0,0 +1,27 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package tig
+
+// T is what we Tigron needs from a testing implementation (*testing.T obviously satisfies it).
```
AkihiroSuda, line 19: Can we just use https://pkg.go.dev/testing#TB

apostasie: No, we cant. https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/testing/testing.go;l=938

The point they are making here is very apt:
Interfaces should be defined by consumers, not by producers.

mod/tigron/internal/mimicry/mocker.go
AkihiroSuda, line 1: "mimicry" == "mocker"?

`mocker.go` should be named `mimicry.go`, or `mimicry/` should be named `mocker/` ?

The same applies to "mimicry" and "mocker" in `doc.md`.

mod/tigron/tig/t.go
AkihiroSuda: Could you add that reason as a code comment?

apostasie: Will do (also will address your other comment).
I am out most of the week for my bday so I’ll be slower than usual.

apostasie: Thanks Akihiro. :-)

Latest push https://github.com/containerd/nerdctl/compare/c2eaeb7ec5714862bd7f8b7153c514c6f3978fe8..00706ad3f3c7bb7221ce51cf14b0681880591481 does address both of your comments and polishes the documentation a little bit.

AkihiroSuda (approved):
Thanks

PR #4077, opened 2025-04-03
[Tigron]: command env WhiteList support and cosmetic changes

apostasie:
<!> Blocked by <!>:
- [x] #4076

----------------------------------

This is a new series of Tigron PRs.

Now that the big change with the new command implementation is in, upcoming work on Tigron will focus on:
- **P0**: internal testing, getting to the point we can absolutely trust Tigron to **never** faceplant
- **P0**: debugging output readability and other developers quality of life improvements
- **P1**: code cleanup and simplification

Ideally, none of these PRs will require ANY change in nerdctl, and are either changes to the way we test Tigron itself, or refactoring that does not change signatures and API.

Efforts are being made to break these PRs down into small, digestible changes to ease and fasten review, which explains the chain of PRs.

LMK at any point if a given PR needs to be further broken down.

----------------------------------

This second PR is very small:

Commit 1 adds support for environment whitelisting for `internal/com`.
The point here is to make the output on the CI a lot less crowded (since we do output the command environment, there is a lot of irrelevant stuff in there - ANDROID stuff, etc - and blacklisting one by one is a wild goose chase).

Second commit is cosmetic / refining linting.

### DIFF
--- a/mod/tigron/internal/com/command.go
+++ b/mod/tigron/internal/com/command.go
@@ -43,16 +43,14 @@ var (
 	ErrFailedStarting = errors.New("command failed starting")
 	// ErrSignaled is returned by Wait() if a signal was sent to the command while running.
 	ErrSignaled = errors.New("command execution signaled")
-	// ErrExecutionFailed is returned by Wait() when a command executes but returns a non-zero error
-	// code.
+	// ErrExecutionFailed is returned by Wait() when a command executes but returns a non-zero error code.
 	ErrExecutionFailed = errors.New("command returned a non-zero exit code")
 	// ErrFailedSendingSignal may happen if sending a signal to an already terminated process.
 	ErrFailedSendingSignal = errors.New("failed sending signal")
 
 	// ErrExecAlreadyStarted is a system error normally indicating a bogus double call to Run().
 	ErrExecAlreadyStarted = errors.New("command has already been started (double `Run`)")
-	// ErrExecNotStarted is a system error normally indicating that Wait() has been called without
-	// first calling Run().
+	// ErrExecNotStarted is a system error normally indicating that Wait() has been called without first calling Run().
 	ErrExecNotStarted = errors.New("command has not been started (call `Run` first)")
 	// ErrExecAlreadyFinished is a system error indicating a double call to Wait().
 	ErrExecAlreadyFinished = errors.New("command is already finished")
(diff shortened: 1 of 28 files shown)


apostasie:
Rebased.

apostasie:
Rebased. Ready for review.

mod/tigron/internal/com/package_example_test.go
```diff
@@ -35,23 +36,23 @@ func ExampleCommand() {
 
 	err := cmd.Run(context.Background())
 	if err != nil {
-		fmt.Println("Run err:", err)
+		_, _ = fmt.Println("Run err:", err)
```
AkihiroSuda, line 39: The linter looks too strict

mod/tigron/internal/com/command.go
```diff
@@ -399,26 +400,55 @@ func (gc *Command) buildCommand(ctx context.Context) *exec.Cmd {
...
 
-	// Build env
+	// Build env.
 	cmd.Env = []string{}
-	// TODO: replace with regexps? and/or whitelist?
+
+	const (
+		star  = "*"
+		equal = "="
+	)
```
AkihiroSuda, line 408: Can we have unit tests?

apostasie: We have some in the second commit: https://github.com/containerd/nerdctl/pull/4077/files#diff-38ae3cf7577e4445244fa918a0b6a0f1f380f4c1dcad81623eb2145c8febfbddR259 - but yeah, I can add more (especially for star matching).

mod/tigron/internal/com/package_example_test.go
apostasie: In that case we could / should disable that specific rule for the entire file, but otherwise I would prefer to keep enforcing handling error returns.

Let me adjust that ^.

AkihiroSuda: Would it be possible to disable the rule for `*Print*`?

apostasie: Yeah, that would be ideal. Let me look.

apostasie: Fixed. Thanks for the suggestion - just better now.

mod/tigron/internal/com/command.go
apostasie: Added more tests for prefix matching (`FOO*`), a test for precedence of Whitelist over Blacklist, plus some comments.

apostasie:
Thanks for the review @AkihiroSuda. Appreciated.

AkihiroSuda (approved):
Thanks

PR #4078, opened 2025-04-03
feat: add cpu-rt-period and cpu-rt-runtime options in nerdctl create/run

swagatbora90:
Adds `cpu-rt-runtime` and `cpu-rt-period` options with nerdctl run and create.

Tested locally on a system with cgroup v1. 

One issue with the current implementation is that the realtime cgroup settings rely on the parent cgroup. So, for a container to set this value, the parent cgroup `/sys/fs/cgroup/cpu,cpuacct/default/cpu.rt_runtime_us` should also be set. Podman also appears to have the similar issue [containers/podman#12444]. Also, it is not a issue in docker since you can set these values at the daemon level [moby/moby#31411]

### DIFF
--- a/pkg/inspecttypes/dockercompat/dockercompat_test.go
+++ b/pkg/inspecttypes/dockercompat/dockercompat_test.go
@@ -417,3 +417,97 @@ func TestNetworkSettingsFromNative(t *testing.T) {
 		})
 	}
 }
+
+func TestCpuSettingsFromNative(t *testing.T) {
+	// Helper function to create uint64 pointer
+	uint64Ptr := func(i uint64) *uint64 {
+		return &i
+	}
+
+	int64Ptr := func(i int64) *int64 {
+		return &i
+	}
+
+	testcases := []struct {
+		name     string
+		spec     *specs.Spec
+		expected *CPUSettings
+	}{
+		{
+			name:     "Test with empty spec",
+			spec:     &specs.Spec{},
+			expected: &CPUSettings{},
+		},
+		{
+			name: "Full CPU Settings",
+			spec: &specs.Spec{
+				Linux: &specs.Linux{
+					Resources: &specs.LinuxResources{
+						CPU: &specs.LinuxCPU{
+							Cpus:            "0-3",
+							Mems:            "0-1",
+							Shares:          uint64Ptr(1024),
+							Quota:           int64Ptr(100000),
+							Period:          uint64Ptr(100000),
+							RealtimePeriod:  uint64Ptr(1000000),
+							RealtimeRuntime: int64Ptr(950000),
+						},
+					},
+				},
+			},
+			expected: &CPUSettings{
+				CPUSetCpus:         "0-3",
+				CPUSetMems:         "0-1",
+				CPUShares:          1024,
+				CPUQuota:           100000,
+				CPUPeriod:          100000,
+				CPURealtimePeriod:  1000000,
+				CPURealtimeRuntime: 950000,
+			},
+		},
+		{
+			name: "Partial CPU Settings",
+			spec: &specs.Spec{
+				Linux: &specs.Linux{
+					Resources: &specs.LinuxResources{
+						CPU: &specs.LinuxCPU{
+							Cpus:   "0,1",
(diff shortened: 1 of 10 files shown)


pkg/inspecttypes/dockercompat/dockercompat_test.go
```diff
@@ -417,3 +417,104 @@ func TestNetworkSettingsFromNative(t *testing.T) {
...
+		return &i
+	}
+
+	testcases := []struct {
+		name     string
+		spec     *specs.Spec
+		expected *CPUSettings
+	}{
+		{
+			name:     "Test with emoty spec",
```
manugupt1, line 437: nit: empty

pkg/inspecttypes/dockercompat/dockercompat_test.go
```diff
@@ -417,3 +417,104 @@ func TestNetworkSettingsFromNative(t *testing.T) {
...
+			spec:     &specs.Spec{},
+			expected: &CPUSettings{},
+		},
+		{
+			name: "Null Linux Resources",
```
manugupt1, line 442: duplicated test case?

manugupt1:
Please add tests as we have cgroup1 tests.

manugupt1 (changes requested):

pkg/inspecttypes/dockercompat/dockercompat_test.go
swagatbora90: fixed

pkg/inspecttypes/dockercompat/dockercompat_test.go
swagatbora90: Removed

swagatbora90:
> Please add tests as we have cgroup1 tests.

I have added a test for this but as you can see the test only calls `container create` instead of `run`. CPU RT settings are not inherited from the parent cgroup, so it needs to be set explicitly. This also makes it harder to test. 

```
% sysctl kernel.sched_rt_period_us
kernel.sched_rt_period_us = 1000000

% sysctl kernel.sched_rt_runtime_us
kernel.sched_rt_runtime_us = 950000
```

Root cgroup settings
```
 % cat /sys/fs/cgroup/cpu,cpuacct/cpu.rt_runtime_us
950000

% cat /sys/fs/cgroup/cpu,cpuacct/cpu.rt_period_us
1000000
```

Then when I try to run a container, it fails:
```
 % sudo nerdctl --namespace test-cgroup container run --cpu-rt-period 1000000 --cpu-rt-runtime 95000 ubuntu sleep infinity
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: failed to write "95000": write /sys/fs/cgroup/cpu,cpuacct/test-cgroup/2988aa23729a8374c996ec4b1bd8bdd0f7a4a56b40e2f76cfac529a856bcc868/cpu.rt_runtime_us: invalid argument: unknown
```

This is because the `test-cgroup` cgroup does not inherit the realtime settings of the parent and is set as 0 by default,  
```
% cat /sys/fs/cgroup/cpu,cpuacct/test-cgroup/cpu.rt_period_us
1000000

 % cat /sys/fs/cgroup/cpu,cpuacct/test-cgroup/cpu.rt_runtime_us
0
```

If I manually set the rt_runtime_us for `test-cgroup`
```
 % echo 950000 | sudo tee /sys/fs/cgroup/cpu,cpuacct/test-cgroup/cpu.rt_runtime_us
950000
```

I can now run the containers in this cgroup.  
```
% sudo nerdctl --namespace test-cgroup container run --cpu-rt-period 1000000 --cpu-rt-runtime 800000 -d ubuntu sleep infinity
db1eca8f91daceb2bad4d799aa8543feff2a48f0c7bd4fbb1f1d11a1946f5821

```
Note that the sum of `cpu.rt_runtime_us` value of all the child cgroup should be less than or equal to the parent cgroup.

manugupt1 (approved):

AkihiroSuda (approved):
Thanks

Issue #4068, opened 2025-04-01
[LEGACY] [CI]: TestRestartWithSignal does not seem to (always?) work correctly (EL8 only?)


apostasie:
### Description

Seen on #4056 

See attached trace (note that PR #4056 has the new timeout logic).

Signal has been sent and received, but the container has not restarted (and was instead terminated by timeout).

@AkihiroSuda the good news here is that the new timeout logic is working as expected, and this test would have previously hung-up on EL8.

Now we just have to figure out what is happening here.

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

```
=== FAIL: cmd/nerdctl/container TestRestartWithSignal (21.97s)
    container_restart_linux_test.go:163: 
    container_restart_linux_test.go:163: ======================== Pre-test cleanup ========================
    container_restart_linux_test.go:163: 
    container_restart_linux_test.go:163: ======================== Test setup ========================
    container_restart_linux_test.go:163: 
    container_restart_linux_test.go:163: ======================== Test Run ========================
    command.go:435: [2025-04-01T16:46:36Z] [command=/usr/local/bin/nerdctl --namespace=nerdctl-test run --name testrestartwithsignal-27bfb0e4 ghcr.io/stargz-containers/alpine:3.13-org sh -c #!/bin/sh
        	set -eu
        
        	sig_msg () {
        		printf "received\n"
        		[ "false" != true ] || exit 0
        	}
        
        	trap sig_msg 10
        	printf "trap ready\n"
        	while true; do
        		printf "waiting...\n"
        		sleep 0.5
        	done
        ] command cancelled
    command.go:115: expected: 137  - to be equal to: -1
    command.go:115: Expected exit code: 137
        
    command.go:115: 
        =================================
        | Command:	/usr/local/bin/nerdctl run --name testrestartwithsignal-27bfb0e4 ghcr.io/stargz-containers/alpine:3.13-org sh -c #!/bin/sh
        	set -eu
        
        	sig_msg () {
        		printf "received\n"
        		[ "false" != true ] || exit 0
        	}
        
        	trap sig_msg 10
        	printf "trap ready\n"
        	while true; do
        		printf "waiting...\n"
        		sleep 0.5
        	done
        
        | Working Dir:	/tmp/TestRestartWithSignal2256824605/001
        | Timeout:	20s
        =================================
        	SHELL=/bin/bash
        	LOGNAME=rootless
        	XDG_SESSION_TYPE=tty
[...]

PR #4081, opened 2025-04-04
[Tigron]: revised Data interface

apostasie:
Blocked by #4111

From testing experience on nerdctl, many tests do require temporary resources (Dockerfiles, compose.yml, etc), and routinely redo the same thing over and over (create separate temp dir, write file, remove temp dir).
At best this adds a lot of boilerplate / helpers functions - at worst, the resources are not properly cleaned-up, or not well isolated from other tests.

The first commit changes the Data interface to allow temp files manipulation (Load, Save, Path, Dir, Exists), and more clearly separates `labels` and temporary resources.

The changes in tigron are minimal, but since the interface does change, the second commit does modify all current nerdctl tests that are making use of data labels.
Alongside this, some tests and helpers will now leverage the new temp file manipulation, along with some other minor cleanup in helpers and builder tests.

The PR looks big because of that ^. 

Finally note that:
- command cwd and temporary directory will no longer be the same
- temporary dir generation has been altered so that basename is sha-ed after the test name to ensure better unicity (to be usable for compose)

### DIFF
--- a/cmd/nerdctl/builder/builder_build_oci_layout_test.go
+++ b/cmd/nerdctl/builder/builder_build_oci_layout_test.go
@@ -18,13 +18,13 @@ package builder
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 	"testing"
 
 	"gotest.tools/v3/assert"
 
+	"github.com/containerd/nerdctl/mod/tigron/expect"
 	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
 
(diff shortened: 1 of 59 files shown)


cmd/nerdctl/builder/builder_build_oci_layout_test.go
```diff
@@ -44,17 +44,17 @@ func TestBuildContextWithOCILayout(t *testing.T) {
 		),
 		Cleanup: func(data test.Data, helpers test.Helpers) {
 			if nerdtest.IsDocker() {
-				helpers.Anyhow("buildx", "stop", data.Identifier("-container"))
```
apostasie, line 47: Leading dashes are unnecessary. Identifier does separate seed string with dash.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -250,21 +262,15 @@ CMD ["echo", "nerdctl-build-test-stdin"]`, testutil.CommonImage)
...
 
+	dockerfile := fmt.Sprintf(`FROM %s
+CMD ["echo", "nerdctl-build-test-dockerfile"]
+	`, testutil.CommonImage)
+
 	testCase := &test.Case{
 		Require: nerdtest.Build,
-		Cleanup: func(data test.Data, helpers test.Helpers) {
-			helpers.Anyhow("rmi", "-f", data.Identifier())
-		},
```
apostasie, line 257: These (and similar) were doing nothing, as there is no such image.

apostasie:
Failure is #4068

apostasie:
Momentarily reverting to draft.
Need to remove `WithFile` which is problematic and not helpful.
Forgot my security key today, so, cannot push right now.

cmd/nerdctl/network/network_list_linux_test.go
```diff
@@ -31,12 +31,12 @@ func TestNetworkLsFilter(t *testing.T) {
...
-		data.Set("net1", data.Identifier("1"))
-		data.Set("net2", data.Identifier("2"))
-		data.Set("netID1", helpers.Capture("network", "create", "--label="+data.Get("label"), data.Get("net1")))
-		data.Set("netID2", helpers.Capture("network", "create", data.Get("net2")))
+		data.Labels().Set("identifier", data.Identifier())
+		data.Labels().Set("label", "mylabel=label-1")
+		data.Labels().Set("net1", data.Identifier("1"))
+		data.Labels().Set("net2", data.Identifier("2"))
+		data.Labels().Set("netID1", helpers.Capture("network", "create", "--label="+data.Labels().Get("label"), data.Labels().Get("net1")))
+		data.Labels().Set("netID2", helpers.Capture("network", "create", data.Labels().Get("net2")))
```
AkihiroSuda, line 39: nit: `Labels()` could be just called once

AkihiroSuda (approved):
Thanks

cmd/nerdctl/network/network_list_linux_test.go
apostasie: Wondering if the compiler manages to inline that?

Issue #4067, opened 2025-04-01
[CI]: fix TestImageHistory to not hardcode duration


apostasie:
### Description

TestImageHistory currently fails every year, as it hard codes a string mentioning "created X years ago".

This was just temporarily addressed in #4058 but will fail again next year.

### Describe the results you received and expected

This test should be modified to remove the hard coded string so that it does not fail every year...

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
@AkihiroSuda +good-first-issue

biubiubiuboomboomboom:
i'm interested in taking on this issue.

apostasie:
@biubiubiuboomboomboom sure, go for it! Thanks.

PR #4083, opened 2025-04-04
[CI]: fix hardcode in TestImageHistory

biubiubiuboomboomboom:
## Motivation

Fix hardcoded `CreatedSince ` values in TestImageHistory .

## Related Issue 
 #4067 

## Change

- Replaced the fixed string value used in `assert.Equal` for the `CreatedSince` field with the value returned by the  `formatter.TimeSinceInHuman` function in relevant unit test cases.

- Added a unit test case for the `formatter.TimeSinceInHuman` function . 

## Others 

I noticed that other functions in `formatter.go` seem to have no unit tests . There may be a risk when these functions are changed .

### DIFF
--- a/pkg/formatter/formatter_test.go
+++ b/pkg/formatter/formatter_test.go
@@ -1,0 +1,71 @@
+package formatter
+
+import (
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestTimeSinceInHuman(t *testing.T) {
+	now := time.Now()
+
+	tests := []struct {
+		name     string
+		input    time.Time
+		expected string
+	}{
+		{
+			name:     "1 second ago",
+			input:    now.Add(-1 * time.Second),
+			expected: "1 second ago",
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/formatter/formatter_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestTimeSinceInHuman(t *testing.T) {
+	now := time.Now()
```
apostasie, line 27: + `t.Parallel()`

apostasie (approved):

apostasie:
Thanks @biubiubiuboomboomboom 

Two nits about using `t.Parallel` but otherwise looks good.

pkg/formatter/formatter_test.go
biubiubiuboomboomboom: got it , thanks for your help

fahedouch (approved):
Thanks

Issue #2179, opened 2023-04-13
nerdctl run -p <host-port>:<container port>  <image> does not check if user defined host port is in use. 


vsiravar:
### Description

Running containers with `-p/--publish` with user provided host port [does not check if the host port is already in use](https://github.com/containerd/nerdctl/blob/main/pkg/portutil/portutil.go#L99) in both rootful and rootless.  This issue is very similar to https://github.com/containerd/nerdctl/issues/2164. 

### Steps to reproduce the issue

```
# Run 2 containers with -p <hostport>:<containerport> option.   
$ nerdctl run -d --name=my-nginx -p 5000:80 nginx
187eda2196935949563e57cb9227da1c9efafef4fda3aefe54a00b53635fa950

$ nerdctl run -d --name=my-httpd  -p 5000:80 httpd

$ nerdctl ps -a
CONTAINER ID    IMAGE                             COMMAND                   CREATED           STATUS     PORTS                   NAMES
187eda219693    docker.io/library/nginx:latest    "/docker-entrypoint.…"    9 minutes ago     Up         0.0.0.0:5000->80/tcp    my-nginx
27232786d099    docker.io/library/httpd:latest    "httpd-foreground"        10 seconds ago    Up         0.0.0.0:5000->80/tcp    my-httpd2
a22d1643e952    docker.io/library/httpd:latest    "httpd-foreground"        33 seconds ago    Created    0.0.0.0:5000->80/tcp    my-httpd
```
This is very similar to [2164](https://github.com/containerd/nerdctl/issues/2164) except this is for user specified host ports.

Both the containers have host port 5000 and is up but only one of them can be reached from the host. 

### Describe the results you received and expected

Error response 
Similar response as docker. 
`Ports are not available: exposing port TCP 0.0.0.0:5000 -> 0.0.0.0:0: listen tcp 0.0.0.0:5000: bind: address already in use.`

### What version of nerdctl are you using?

main branch
```
$ nerdctl version 
Client:
 Version:	v1.3.0-22-g6da9283.m
 OS/Arch:	linux/amd64
 Git commit:	6da9283cb34d09c4d0ca036e63ec01b60fec5f56.m
 buildctl:
  Version:	v0.11.3
  GitCommit:	4ddee42a32aac4cd33bf9c2be4c87c2ffd34747b

Server:
 containerd:
  Version:	v1.6.19
  GitCommit:	1e1ea6e986c6c86565bc33d52e34b81b3e2bc71f
 runc:
[...]

PR #2190, opened 2023-04-18
fix: Add used port check for publish flag

vsiravar:
Add check for used ports while allocating host port in `-p/--publish` to make it compatible with docker. 

Fixes: https://github.com/containerd/nerdctl/issues/2179

PR #4097, opened 2025-04-09
[Bugfix] Check if port is used in publish flag [duplicate of #2190]

swagatbora90:
This PR continues the work from #2190  (originally by @vsiravar) which checks for used ports while allocating host port in -p/--publish. The original PR is already approved and was just waiting for a rebase form the author. Most of the comments have already been addressed.

#### Addressing port cleanup on stop:
The original PR had some open discussion around handling port cleanups on container stop

> > 
> > 
> > I concur with it, but hold reservations about merging this PR. I believe we should first have nerdctl stop manage the release of ports.
> > The rest LGTM.
> 
> cc [vsiravar](https://github.com/vsiravar), could you please fix the `stop` ( it can be a different PR) so we can merge this accordingly? Thanks

This is now already being addressed with this PR https://github.com/containerd/nerdctl/pull/2839 as network cleanup is part of both container stop and kill 

Verified the fix:
```
 % sudo ./_output/nerdctl run -p 8080:80 --name=test -d nginx
57a24f8da7127aeacee8b888a90e0e4eda866b35513844620de4ad1d28200547

 % sudo ./_output/nerdctl run -p 8080:80  -d nginx
FATA[0000] failed to load networking flags: bind for :8080 failed: port is already allocated 

 % sudo nerdctl stop test
test

% sudo ./_output/nerdctl run -p 8080:80  -d nginx
4ad5ca030355b1eb537c3caa91dfd5a9944dd36e8dea1756a4334a986f96c682
```

Fixes: #2179 

cc @fahedouch @vsiravar

### DIFF
--- a/cmd/nerdctl/container/container_run_network_linux_test.go
+++ b/cmd/nerdctl/container/container_run_network_linux_test.go
@@ -349,6 +349,61 @@ func TestUniqueHostPortAssignement(t *testing.T) {
 	}
 }
 
+func TestHostPortAlreadyInUse(t *testing.T) {
+	testCases := []struct {
+		hostPort      string
+		containerPort string
+	}{
+		{
+			hostPort:      "5000",
+			containerPort: "80/tcp",
+		},
+		{
+			hostPort:      "5000",
+			containerPort: "80/tcp",
+		},
+		{
+			hostPort:      "5000",
+			containerPort: "80/udp",
+		},
+		{
+			hostPort:      "5000",
+			containerPort: "80/sctp",
+		},
+	}
+
+	tID := testutil.Identifier(t)
+
+	for i, tc := range testCases {
+		tc := tc
+		tcName := fmt.Sprintf("%+v", tc)
+		t.Run(tcName, func(t *testing.T) {
+			if strings.Contains(tc.containerPort, "sctp") && rootlessutil.IsRootless() {
+				t.Skip("sctp is not supported in rootless mode")
+			}
+			testContainerName1 := fmt.Sprintf("%s-%d-1", tID, i)
+			testContainerName2 := fmt.Sprintf("%s-%d-2", tID, i)
+			base := testutil.NewBase(t)
+			t.Cleanup(func() {
+				base.Cmd("rm", "-f", testContainerName1, testContainerName2).AssertOK()
+			})
+			pFlag := fmt.Sprintf("%s:%s", tc.hostPort, tc.containerPort)
+			cmd1 := base.Cmd("run", "-d",
+				"--name", testContainerName1, "-p",
+				pFlag,
+				testutil.NginxAlpineImage)
+
+			cmd2 := base.Cmd("run", "-d",
+				"--name", testContainerName2, "-p",
+				pFlag,
+				testutil.NginxAlpineImage)
+
+			cmd1.AssertOK()
+			cmd2.AssertFail()
+		})
+	}
+}
+
 func TestRunPort(t *testing.T) {
 	baseTestRunPort(t, testutil.NginxAlpineImage, testutil.NginxAlpineIndexHTMLSnippet, true)
 }
--- a/pkg/portutil/port_allocate_linux.go
+++ b/pkg/portutil/port_allocate_linux.go
@@ -25,11 +25,14 @@ import (
 
 const (
 	// This port range is compatible with Docker, FYI https://github.com/moby/moby/blob/eb9e42a09ee123af1d95bf7d46dd738258fa2109/libnetwork/portallocator/portallocator_unix.go#L7-L12
-	allocateEnd = 60999
+	allocateEnd = uint64(60999)
+
+	tcpTimeWait  = 6 //TIME_WAIT state is represented by the value 6 in /proc/net/tcp
+	tcpCloseWait = 8 //CLOSE_WAIT state is represented by the value 8 in /proc/net/tcp
 )
 
 var (
-	allocateStart = 49153
+	allocateStart = uint64(49153)
 )
 
 func filter(ss []procnet.NetworkDetail, filterFunc func(detail procnet.NetworkDetail) bool) (ret []procnet.NetworkDetail) {
@@ -42,24 +45,56 @@ func filter(ss []procnet.NetworkDetail, filterFunc func(detail procnet.NetworkDe
 }
 
 func portAllocate(protocol string, ip string, count uint64) (uint64, uint64, error) {
-	netprocData, err := procnet.ReadStatsFileData(protocol)
+	usedPorts, err := getUsedPorts(ip, protocol)
 	if err != nil {
 		return 0, 0, err
 	}
-	netprocItems := procnet.Parse(netprocData)
+
+	start := allocateStart
+	if count > allocateEnd-allocateStart+1 {
+		return 0, 0, fmt.Errorf("can not allocate %d ports", count)
+	}
+	for start < allocateEnd {
+		needReturn := true
+		for i := start; i < start+count; i++ {
+			if _, ok := usedPorts[i]; ok {
+				needReturn = false
+				break
+			}
+		}
+		if needReturn {
+			allocateStart = start + count
+			return start, start + count - 1, nil
+		}
+		start += count
+	}
+	return 0, 0, fmt.Errorf("there is not enough %d free ports", count)
+}
+
+func getUsedPorts(ip string, protocol string) (map[uint64]bool, error) {
+	netprocItems := []procnet.NetworkDetail{}
+
+	if protocol == "tcp" || protocol == "udp" {
+		netprocData, err := procnet.ReadStatsFileData(protocol)
+		if err != nil {
+			return nil, err
+		}
+		netprocItems = append(netprocItems, procnet.Parse(netprocData)...)
+	}
+
 	// In some circumstances, when we bind address like "0.0.0.0:80", we will get the formation of ":::80" in /proc/net/tcp6.
 	// So we need some trick to process this situation.
 	if protocol == "tcp" {
 		tempTCPV6Data, err := procnet.ReadStatsFileData("tcp6")
 		if err != nil {
-			return 0, 0, err
+			return nil, err
 		}
 		netprocItems = append(netprocItems, procnet.Parse(tempTCPV6Data)...)
 	}
 	if protocol == "udp" {
 		tempUDPV6Data, err := procnet.ReadStatsFileData("udp6")
 		if err != nil {
-			return 0, 0, err
+			return nil, err
 		}
 		netprocItems = append(netprocItems, procnet.Parse(tempUDPV6Data)...)
 	}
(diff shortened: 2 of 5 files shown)


cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -350,6 +350,62 @@ func TestUniqueHostPortAssignement(t *testing.T) {
...
+		tc := tc
+		tcName := fmt.Sprintf("%+v", tc)
+		t.Run(tcName, func(t *testing.T) {
+			if strings.Contains(tc.containerPort, "sctp") && rootlessutil.IsRootless() {
+				t.Skip("sctp is not supported in rootless mode")
+			}
+			testContainerName1 := fmt.Sprintf("%s-%d-1", tID, i)
+			testContainerName2 := fmt.Sprintf("%s-%d-2", tID, i)
+			base := testutil.NewBase(t)
+			defer base.Cmd("rm", "-f", testContainerName1, testContainerName2).Run()
```
manugupt1, line 390: nit: should be AssertOk()

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -350,6 +350,62 @@ func TestUniqueHostPortAssignement(t *testing.T) {
...
+			hostPort:      "5000",
+			containerPort: "80/sctp",
+		},
+	}
+
+	tID := testutil.Identifier(t)
+
+	for i, tc := range testCases {
+
+		i := i
```
apostasie, line 380: This is no longer necessary.

cmd/nerdctl/container/container_run_network_linux_test.go
apostasie, line 390: Do not use defer - `t.Cleanup` instead.
Also generally, also call your cleanup before the test so that any leftovers are removed.

swagatbora90:
The current implementation breaks in rootless mode which is resulting in the test failures. I am guessing this is mainly due to how Rootlesskit uses slirp4netns to create a separate net ns and possibly the port allocation information is no longer visible in /proc/net/*.

We can either implement some mechanism to track the rootless state or may be add a generic port store to track all port usage.

fahedouch:
> The current implementation breaks in rootless mode which is resulting in the test failures. I am guessing this is mainly due to how Rootlesskit uses slirp4netns to create a separate net ns and possibly the port allocation information is no longer visible in /proc/net/*.
> 
> We can either implement some mechanism to track the rootless state or may be add a generic port store to track all port usage.

Rootless > 2.0 starts with a detached network namespace by default. You need to `nsenter` into the child network namespace, where the container's network is managed, and the ports should be checked there. You have several ways to nsenter, either by executing a [child process](https://github.com/containerd/nerdctl/blob/main/pkg/cmd/container/create.go#L523) or using the library https://pkg.go.dev/github.com/containernetworking/plugins/pkg/ns#WithNetNSPath (I prefer the containernetworking library)

fahedouch:
ping @swagatbora90

swagatbora90:
nerdctl gomodjail tests are consistently failing with errors such as `could not get iptables version /usr/sbin/iptables: bad file descriptor"`:
```
time=2025-06-02T17:15:46.283Z level=WARN msg=***Blocked*** pid=76099 exe=/home/rootless/.cache/g |
        |         | omodjail/1f381f194c295770/nerdctl syscall=fcntl entry=/go/pkg/mod/github.com/coreos/go-iptables@ |
        |         | v0.8.0/iptables/iptables.go:659:github.com/coreos/go-iptables/iptables.getIptablesVersionString  |
        |         | module=github.com/coreos/go-iptables                                                             |
        |         | time="2025-06-02T17:15:46Z" level=fatal msg="failed to load networking flags: could not get ipta |
        |         | bles version: fork/exec /usr/sbin/iptables: bad file descriptor"
                                 |

```

Seems unrelated to the current changes. Is this a known issue @apostasie @fahedouch ?

fahedouch:
My knowledge of [slirp4netns](https://github.com/rootless-containers/slirp4netns) is very limited, but I think it is the culprit. @AkihiroSuda, can you confirm that it is the one opening the socket on the loopback address `127.0.0.1` and not removing it immediately? It might be an issue to fix in slirp or something need to handled in postStop ocihook (wait for the socket to be removed before removing the container)

AkihiroSuda:
> fork/exec /usr/sbin/iptables: bad file descriptor

This line should have `//gomodjail:unconfined` directive
https://github.com/containerd/nerdctl/blob/b8c4b3d852b8e49b59a3061f72a9064a6f78e3c5/go.mod#L31

AkihiroSuda:
> My knowledge of [slirp4netns](https://github.com/rootless-containers/slirp4netns) is very limited, but I think it is the culprit. @AkihiroSuda, can you confirm that it is the one opening the socket on the loopback address `127.0.0.1` and not removing it immediately? It might be an issue to fix in slirp or something need to handled in postStop ocihook (wait for the socket to be removed before removing the container)

The default port driver is still RootlessKit's `builtin`, not `slirp4netns`

fahedouch:
> The builtin driver is fast, but be aware that the source IP is not propagated and always set to 127.0.0.1.

ok ^^

> handled in postStop ocihook (wait for the socket to be removed before removing the container)

WDYT about this. I think it is the responsibility of the `container`/`network driver` to ensure that the sockets are clean before releasing the container.

swagatbora90:
> > fork/exec /usr/sbin/iptables: bad file descriptor
> 
> This line should have `//gomodjail:unconfined` directive
> 
> https://github.com/containerd/nerdctl/blob/b8c4b3d852b8e49b59a3061f72a9064a6f78e3c5/go.mod#L31

Opened a separate PR #4303 to add the directive

swagatbora90:
> WDYT about this. I think it is the responsibility of the `container`/`network driver` to ensure that the sockets are clean before releasing the container.

I was thinking the same, but when I looked into the current implementation found that nerdctl is using [rootlesskit portmanager](https://github.com/rootless-containers/rootlesskit/blob/master/pkg/port/builtin/parent/tcp/tcp.go#L79) to handle port adds/deletes. The proxy connection creation and deletion is handled in rootlesskit, so nerdctl does not have much control over it.

swagatbora90:
There is still some flakiness in some test. Related  #4143 
```
=== RUN   TestComposeUp
    testing_linux.go:69: projectName="nerdctl-compose-test2030087482"
    testing_linux.go:71: assertion failed: res.ExitCode is not exitCode: 
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test compose -f /tmp/nerdctl-compose-test2030087482/docker-compose.yaml up -d
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   time="2025-06-04T20:14:35Z" level=info msg="Creating network nerdctl-compose-test2030087482_default"
        time="2025-06-04T20:14:35Z" level=info msg="Creating volume nerdctl-compose-test2030087482_wordpress"
        time="2025-06-04T20:14:35Z" level=info msg="Creating volume nerdctl-compose-test2030087482_db"
        time="2025-06-04T20:14:35Z" level=info msg="Ensuring image ghcr.io/stargz-containers/wordpress:5.7-org"
        time="2025-06-04T20:14:35Z" level=info msg="Ensuring image ghcr.io/stargz-containers/mariadb:10.5-org"
        time="2025-06-04T20:14:35Z" level=info msg="Creating container nerdctl-compose-test2030087482-wordpress-1"
        time="2025-06-04T20:14:35Z" level=info msg="Running [/usr/local/bin/nerdctl --namespace=nerdctl-test run --cidfile=/tmp/compose-4290938950/cid -l=com.docker.compose.project=nerdctl-compose-test2030087482 -l=com.docker.compose.service=wordpress -d --name=nerdctl-compose-test2030087482-wordpress-1 --pull=never -e=WORDPRESS_DB_NAME=exampledb -e=WORDPRESS_DB_HOST=db -e=WORDPRESS_DB_USER=exampleuser -e=WORDPRESS_DB_PASSWORD=examplepass --net=nerdctl-compose-test2030087482_default --hostname=wordpress -p=8080:80/tcp --restart=always -v=nerdctl-compose-test2030087482_wordpress:/var/www/html ghcr.io/stargz-containers/wordpress:5.7-org]"
        time="2025-06-04T20:14:35Z" level=warning msg="volume \"nerdctl-compose-test2030087482_wordpress\" already exists and will be returned as-is"
        time="2025-06-04T20:14:35Z" level=fatal msg="OCI runtime start failed: cannot start a container that has stopped: unknown"
        time="2025-06-04T20:14:35Z" level=fatal msg="error while creating container nerdctl-compose-test2030087482-wordpress-1: error while creating container nerdctl-compose-test2030087482-wordpress-1: exit status 1"
```

It appears that this test failed and perhaps there was some issue with cleanup. As a result a lot of other tests that relies on host port 8080 started failing with port already allocated errors. This does mean that with by relying on a single host port in all our test (in this case port 8080), there is a higher chance of unrelated tests failing. We can randomize the port used in the test so that it is not always 8080, but would prefer to keep it in a separate PR

pkg/portutil/portutil.go
```diff
@@ -59,6 +59,7 @@ func ParseFlagP(s string) ([]cni.PortMapping, error) {
 	case 2:
 		proto = strings.ToLower(splitBySlash[1])
 		switch proto {
+		// sctp is not a supported protocol
```
fahedouch: The comment suggests `SCTP` isn't supported, but it is present in the case line `63`

pkg/portutil/port_allocate_linux.go
```diff
@@ -25,11 +25,14 @@ import (
 
 const (
 	// This port range is compatible with Docker, FYI https://github.com/moby/moby/blob/eb9e42a09ee123af1d95bf7d46dd738258fa2109/libnetwork/portallocator/portallocator_unix.go#L7-L12
-	allocateEnd = 60999
+	allocateEnd = uint64(60999)
+
+	tcpTimeWait  = 6 //TIME_WAIT state is represented by the value 6 in /proc/net/tcp
```
fahedouch, line 30: may be add a comment to explain why and point to rootlesskit behavior

pkg/portutil/portutil.go
swagatbora90: removed

pkg/portutil/procnet/procnet.go
```diff
@@ -27,6 +27,7 @@ import (
 type NetworkDetail struct {
 	LocalIP   net.IP
 	LocalPort uint64
+	State     int
 }
 
 func Parse(data []string) (results []NetworkDetail) {
```
AkihiroSuda, line 33: Can we have unit tests?

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -349,6 +349,61 @@ func TestUniqueHostPortAssignement(t *testing.T) {
...
+	}
+
+	tID := testutil.Identifier(t)
+
+	for i, tc := range testCases {
+		tc := tc
+		tcName := fmt.Sprintf("%+v", tc)
+		t.Run(tcName, func(t *testing.T) {
+			if strings.Contains(tc.containerPort, "sctp") && rootlessutil.IsRootless() {
+				t.Skip("sctp is not supported in rootless mode")
```
AkihiroSuda, line 382: Do we already support sctp for rootful?

pkg/portutil/procnet/procnet.go
swagatbora90: I can add some unit test as a follow up.

PR #242, opened 2021-06-05
setup retry mechanism for CI

fahedouch:
several tests often fails in the cgroup2 job because a `timeout`  or `flaky test`. This PR aims to automatically try again these failed jobs

related https://github.com/containerd/nerdctl/issues/218

Signed-off-by: fahed dorgaa <fahed.dorgaa@gmail.com>

PR #4103, opened 2025-04-14
[DEBUGGING]: buildg time-out flaky issue (#4046)

apostasie:
See https://github.com/containerd/nerdctl/issues/4046 Root cause is undiagnosed  / unclear, but it seems like in some circumstances, buildg command is too late to read from stdin.
This workaround introduces an arbitrary delay before writing to stdin, as a temporary solution to reduce test flakyness.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -250,6 +250,9 @@ COPY docs /out/share/doc/nerdctl/docs
 RUN (cd /out && find ! -type d | sort | xargs sha256sum > /tmp/SHA256SUMS ) && \
   mv /tmp/SHA256SUMS /out/share/doc/nerdctl-full/SHA256SUMS && \
   chown -R 0:0 /out
+RUN git clone https://github.com/apostasie/buildg.git /tmp/buildg; cd /tmp/buildg; \
+    go build -o /out/bin/buildg .
+
 
 FROM scratch AS out-full
 COPY --from=build-full /out /
--- a/cmd/nerdctl/builder/builder.go
+++ b/cmd/nerdctl/builder/builder.go
@@ -21,9 +21,12 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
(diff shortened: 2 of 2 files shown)


cmd/nerdctl/builder/builder_builder_test.go
```diff
@@ -83,7 +85,12 @@ CMD ["echo", "nerdctl-builder-debug-test-string"]`, testutil.CommonImage)
 					err := os.WriteFile(filepath.Join(buildCtx, "Dockerfile"), []byte(dockerfile), 0o600)
 					assert.NilError(helpers.T(), err)
 					cmd := helpers.Command("builder", "debug", buildCtx)
-					cmd.Feed(bytes.NewReader([]byte("c\n")))
+					cmd.WithFeeder(func() io.Reader {
+						// FIXME: there may be a race issue where we would hook-in buildg command stdin too late,
+						// so, arbitrary wait here to reduce flakyness until we diagnose the underlying issue properly.
```
AkihiroSuda, line 90: Could you add a link to the issue

apostasie: Of course.

Let me do a few more runs on the CI to confirm this patch is actually working as expected, and I will amend.

apostasie:
Failure likely: #3556

apostasie:
Failure is #4119

apostasie:
Still here :-(.

apostasie:
TODO:
Need to fork buildg and use the forked variant to debug this with better logs.

apostasie:
Failure is #4132

apostasie:
@AkihiroSuda 

Current assessment on this is that buildg timeout is just too short.

Suggesting on upstream to increase it ([#242](https://github.com/ktock/buildg/issues/242)).

I do not see what we can do here in nerdctl, so, I am going to close this for now.

If and when we get an updated buildg, we can re-evaluate.

PR #4105, opened 2025-04-14
[DOC]: mod/tigron/require documentation

apostasie:
Doc for `require`.


apostasie:
Test failure is interesting.
It is not critical, as the Err is still `timeout` as expected, and we normally do not check exit codes in that case, but still indicates windows might behave differently _in certain cases_...

AkihiroSuda (approved):
Thanks

PR #4107, opened 2025-04-15
[FIX]: fix issues in rmi code (including #4104 content digest not found)

### LINKED ISSUE #4104: [CI]: content digest not found follow-up
### Description

Following the patch changing async image delete to sync and avoiding `Convert` problematic codepath, `content digest not found` all but disappeared EXCEPT on `TestRemove/Remove_image_with_running_container`, which does surface it when calling `nerdctl images`.

Intuition is that this might be showcasing the underlying containerd issue with garbage collection / deletion of images with (repeat) identical content, but this is an unproven hunch.

Repeat on the CI.

```
=================================
        | Command:	/usr/local/bin/nerdctl images
        | Working Dir:	/tmp/TestRemoveRemove_image_with_running_container_-_with_-f27[956](https://github.com/containerd/nerdctl/actions/runs/14452145394/job/40527165424?pr=4103#step:7:957)83035/001
        | Timeout:	3m0s
        =================================
        	HOSTNAME=df0f0ac2e041
        	MEMORY_PRESSURE_WRITE=c29tZSAyMDAwMDAgMjAwMDAwMAA=
        	SYSTEMD_EXEC_PID=80
        	container=docker
        	HOME=/root
        	LANG=C.UTF-8
        	MEMORY_PRESSURE_WATCH=/sys/fs/cgroup/system.slice/docker-entrypoint.service/memory.pressure
        	INVOCATION_ID=8ceae0c7056f47ea912930034432fc8e
        	TERM=xterm
        	USER=root
        	SHLVL=3
        	CGO_ENABLED=0
        	_=/usr/local/bin/gotestsum
        	PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
        	***
        	DOCKER_CONFIG=/tmp/TestRemoveRemove_image_with_running_container_-_with_-f2795683035/001
        	NERDCTL_TOML=/tmp/TestRemoveRemove_image_with_running_container_-_with_-f2795683035/001/nerdctl.toml
        =================================
        | Stderr:
        =================================
        time="2025-04-14T17:51:45Z" level=warning msg="content digest sha256:ec14c7992a97fc11425907e908340c6c3d6ff602f5f13d899e6b7027c9b4133a: not found"
        
        =================================
        | Stdout:
        =================================
        
        =================================
        | Exit Code: 0
        | Signaled: <nil>
        | Err: <nil>
        =================================
        	👀 testing:		`REPOSITORY    TAG    IMAGE ID    CREATED    PLATFORM    SIZE    BLOB SIZE
        `
        	❌ does not verify:	~= `<none>`
```

main

None

### Host information

_No response_

apostasie:
There are multiple issues with the current code implementation:
1. it does not honor the `--async` flag at all - and will always remove (running) images `async`, never `sync`
2. since the image is recreated AFTER we call delete (so that we prevent deletion of the content used by running containers), the code is racy, as (presumably) deletion may or may not have already completed, resulting in the dreaded content digest not found
3. we create the updated image with name ":" - this unfortunately means that if there is already another unnamed image, containerd will refuse to create it

I do not know how to address point 3 - which I believe is already an issue currently (tracked in #4109).
Note that this patch will change the behavior wrt to this specific aspect. Currently, we fail to create the new image, but still asked for deletion of the old image, which will break things. With this patch, we fail on creation and do not delete the original.

I would still suggest we merge this here if the reviewers are happy with it.
It should fix #4104.

### DIFF
--- a/pkg/cmd/image/remove.go
+++ b/pkg/cmd/image/remove.go
@@ -79,18 +79,22 @@ func Remove(ctx context.Context, client *containerd.Client, args []string, optio
 
 			if cid, ok := runningImages[found.Image.Name]; ok {
 				if options.Force {
-					// FIXME: this is suspicious, but passing the opt seem to break some tests
-					// if err = is.Delete(ctx, found.Image.Name, delOpts...); err != nil {
-					if err = is.Delete(ctx, found.Image.Name); err != nil {
-						return err
-					}
-					fmt.Fprintf(options.Stdout, "Untagged: %s\n", found.Image.Name)
-					fmt.Fprintf(options.Stdout, "Untagged: %s\n", found.Image.Target.Digest.String())
-
+					// This is a running image, so, we need to keep a ref on it so that containerd does not GC the layers
+					// First create the new image with an empty name
+					originalName := found.Image.Name
 					found.Image.Name = ":"
 					if _, err = is.Create(ctx, found.Image); err != nil {
 						return err
 					}
+
+					// Now, delete the original
+					if err = is.Delete(ctx, originalName, delOpts...); err != nil {
+						return err
+					}
+
+					fmt.Fprintf(options.Stdout, "Untagged: %s\n", originalName)
+					fmt.Fprintf(options.Stdout, "Untagged: %s@%s\n", originalName, found.Image.Target.Digest.String())
+
 					return nil
 				}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #4112, opened 2025-04-16
Bump CNI to 1.3 (and minor tooling versions)

apostasie:
Closes #4091


apostasie:
Note: this involves following cni changes to their configuration format (deprecating non-conflist).
This might have an impact for people who use the old format.

pkg/netutil/netutil.go
```diff
@@ -533,21 +532,9 @@ func cniLoad(fileNames []string) (configList []*NetworkConfig, err error) {
 		}
 
 		var netConfigList *libcni.NetworkConfigList
-		if strings.HasSuffix(fileName, ".conflist") {
```
AkihiroSuda, line 536: Maybe we should print a warning for a non-conflist file?

hack/provisioning/windows/cni.sh
```diff
@@ -83,21 +83,25 @@ subnet="$(calculate_subnet "$GATEWAY" "$PREFIX_LEN")"
 # https://docs.microsoft.com/en-us/windows-server/networking/technologies/hcn/hcn-json-document-schemas
 bash -c 'cat >"'"${CNI_CONFIG_DIR}"'"/0-containerd-nat.conf <<EOF
```
AkihiroSuda, line 84: The suffix should be `*.conflist`?

apostasie: Yeah, that make sense.
Will amend.

pkg/netutil/netutil.go
apostasie: Yeah. I find it all confusing.
Non-list configurations were removed from the 1.0 spec anyhow.

I'll add a log message.

apostasie: Looks like we wrap the error with a link pointing to conflist already `"failed marshalling json out of network configuration file %q: %w\nFor details on the schema, see https://pkg.go.dev/github.com/containernetworking/cni/libcni#NetworkConfigList"`.

Is that enough?

apostasie:
Ohhhh Kubernetes... I want to punch you...

AkihiroSuda (approved):
Thanks

PR #4114, opened 2025-04-16
[CI]: fix internal command flaky test (window)

apostasie:
On windows, commands that time-out occasionally have exit code set to 1 instead of (expected) -1.

It is not clear why - intuition is that golang command cancellation on context timeout might not be fully deterministic.

Also, it does not matter, as on timeout there is no reason to check error code, so, this PR just relaxes the test.


AkihiroSuda (approved):
Thanks

Merging, but this seems to need further investigations

PR #4115, opened 2025-04-17
fix: BUILDKIT_HOST env parsing when doing system prune

Shubhranshu153:
Currently when we set "BUILDKIT_HOST" as env variable, nerdctl doesnt take that into consideration for system prune.

```
[root@lima-finch nerdctl]# export BUILDKIT_HOST=unix:///tmp/buildkit-env.sock 
[root@lima-finch nerdctl]# nerdctl system prune --all
INFO[0000] Using buildkit host from command line flag or environment variable. 
WARN[0000] BuildKit is not running. Build caches will not be pruned.  error="flag accessed but not defined: buildkit-host"
```

With the change:
```
[root@lima-finch nerdctl]# export BUILDKIT_HOST=unix:///tmp/buildkit-env.sock 
[root@lima-finch nerdctl]# nerdctl system prune --all
WARNING! This will remove:
  - all stopped containers
  - all networks not used by at least one container
  - all images without at least one container associated to them
  - all build cache
Are you sure you want to continue? [y/N] y
```

Note: helpers.AddStringFlag(cmd, "buildkit-host", nil, "", "BUILDKIT_HOST", "BuildKit address") this masks the logic for build commands but for prune command it doesn't work correctly

Would like to refactor not have the helper commands to parse it as this only works when we have the specific flag but for buildctl config BUILDKIT_HOST can be also set for commands which may not have the buildkit-host flag specifically: eg: system prune. Of we add other options like du etc for buildctl this will be also relevant

### DIFF
--- a/cmd/nerdctl/builder/builder_build.go
+++ b/cmd/nerdctl/builder/builder_build.go
@@ -44,7 +44,7 @@ If Dockerfile is not present and -f is not specified, it will look for Container
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
-	helpers.AddStringFlag(cmd, "buildkit-host", nil, "", "BUILDKIT_HOST", "BuildKit address")
+	cmd.Flags().String("buildkit-host", "", "BuildKit address")
 	cmd.Flags().StringArray("add-host", nil, "Add a custom host-to-IP mapping (format: \"host:ip\")")
 	cmd.Flags().StringArrayP("tag", "t", nil, "Name and optionally a tag in the 'name:tag' format")
 	cmd.Flags().StringP("file", "f", "", "Name of the Dockerfile")
@@ -242,7 +242,7 @@ func processBuildCommandFlag(cmd *cobra.Command, args []string) (types.BuilderBu
 }
 
 func GetBuildkitHost(cmd *cobra.Command, namespace string) (string, error) {
-	if cmd.Flags().Changed("buildkit-host") || os.Getenv("BUILDKIT_HOST") != "" {
+	if cmd.Flags().Changed("buildkit-host") {
 		// If address is explicitly specified, use it.
 		buildkitHost, err := cmd.Flags().GetString("buildkit-host")
 		if err != nil {
@@ -252,6 +252,14 @@ func GetBuildkitHost(cmd *cobra.Command, namespace string) (string, error) {
 			return "", err
 		}
 		return buildkitHost, nil
+	}
+
+	if buildkitHost := os.Getenv("BUILDKIT_HOST"); buildkitHost != "" {
+		if err := buildkitutil.PingBKDaemon(buildkitHost); err != nil {
+			return "", err
+		}
+		return buildkitHost, nil
+
 	}
 	return buildkitutil.GetBuildkitHost(namespace)
 }
--- a/cmd/nerdctl/builder/builder_build_test.go
+++ b/cmd/nerdctl/builder/builder_build_test.go
@@ -19,6 +19,7 @@ package builder
 import (
 	"errors"
 	"fmt"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"testing"
@@ -29,6 +30,7 @@ import (
 	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
 
+	"github.com/containerd/nerdctl/v2/pkg/buildkitutil"
 	"github.com/containerd/nerdctl/v2/pkg/platformutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
@@ -975,5 +977,89 @@ RUN ping -c 5 beta
 		Expected: test.Expects(expect.ExitCodeSuccess, nil, nil),
 	}
 
+	testCase.Run(t)
+}
+
+func TestBuildWithBuildkitConfig(t *testing.T) {
+	nerdtest.Setup()
+
+	testCase := &test.Case{
+		Require: require.All(
+			nerdtest.Build,
+			require.Not(nerdtest.Docker),
+		),
+		Setup: func(data test.Data, helpers test.Helpers) {
+			dockerfile := fmt.Sprintf(`FROM %s
+CMD ["echo", "nerdctl-build-test-string"]`, testutil.CommonImage)
+			data.Temp().Save(dockerfile, "Dockerfile")
+			data.Labels().Set("buildCtx", data.Temp().Path())
+
+		},
+		SubTests: []*test.Case{
+			{
+				Description: "build with buildkit-host",
+				Setup: func(data test.Data, helpers test.Helpers) {
+					// Get BuildkitAddr
+					buildkitAddr, err := buildkitutil.GetBuildkitHost(testutil.Namespace)
+					assert.NilError(helpers.T(), err)
+					buildkitAddr = strings.TrimPrefix(buildkitAddr, "unix://")
+
+					// Symlink the buildkit Socket for testing
+					symlinkedBuildkitAddr := filepath.Join(data.Temp().Path(), "buildkit.sock")
+
+					// Do a negative test to check the setup
+					helpers.Fail("build", "-t", data.Identifier(), "--buildkit-host", fmt.Sprintf("unix://%s", symlinkedBuildkitAddr), data.Labels().Get("buildCtx"))
+
+					// Test build with the symlinked socket
+					cmd := helpers.Custom("ln", "-s", buildkitAddr, symlinkedBuildkitAddr)
+					cmd.Run(&test.Expected{
+						ExitCode: 0,
+					})
+					helpers.Ensure("build", "-t", data.Identifier(), "--buildkit-host", fmt.Sprintf("unix://%s", symlinkedBuildkitAddr), data.Labels().Get("buildCtx"))
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("run", "--rm", data.Identifier())
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					helpers.Anyhow("rmi", "-f", data.Identifier())
+				},
+				Expected: test.Expects(0, nil, expect.Equals("nerdctl-build-test-string\n")),
+			},
+			{
+				Description: "build with env specified",
+				Setup: func(data test.Data, helpers test.Helpers) {
+					// Get BuildkitAddr
+					buildkitAddr, err := buildkitutil.GetBuildkitHost(testutil.Namespace)
+					assert.NilError(helpers.T(), err)
+					buildkitAddr = strings.TrimPrefix(buildkitAddr, "unix://")
+
+					// Symlink the buildkit Socket for testing
+					symlinkedBuildkitAddr := filepath.Join(data.Temp().Path(), "buildkit-env.sock")
+
+					// Do a negative test to ensure setting up the env variable is effective
+					cmd := helpers.Command("build", "-t", data.Identifier(), data.Labels().Get("buildCtx"))
+					cmd.Setenv("BUILDKIT_HOST", fmt.Sprintf("unix://%s", symlinkedBuildkitAddr))
+					cmd.Run(&test.Expected{ExitCode: expect.ExitCodeGenericFail})
+
+					// Symlink the buildkit socket for testing
+					cmd = helpers.Custom("ln", "-s", buildkitAddr, symlinkedBuildkitAddr)
+					cmd.Run(&test.Expected{
+						ExitCode: 0,
+					})
+
+					cmd = helpers.Command("build", "-t", data.Identifier(), data.Labels().Get("buildCtx"))
+					cmd.Setenv("BUILDKIT_HOST", fmt.Sprintf("unix://%s", symlinkedBuildkitAddr))
+					cmd.Run(&test.Expected{ExitCode: expect.ExitCodeSuccess})
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("run", "--rm", data.Identifier())
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					helpers.Anyhow("rmi", "-f", data.Identifier())
+				},
+				Expected: test.Expects(0, nil, expect.Equals("nerdctl-build-test-string\n")),
+			},
+		},
+	}
 	testCase.Run(t)
 }
--- a/cmd/nerdctl/builder/builder_builder_test.go
+++ b/cmd/nerdctl/builder/builder_builder_test.go
@@ -19,12 +19,17 @@ package builder
 import (
 	"errors"
 	"fmt"
+	"path/filepath"
 	"strings"
 	"testing"
 
+	"gotest.tools/v3/assert"
+
+	"github.com/containerd/nerdctl/mod/tigron/expect"
 	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
 
+	"github.com/containerd/nerdctl/v2/pkg/buildkitutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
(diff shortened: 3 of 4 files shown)


cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,109 @@ RUN ping -c 5 beta
...
+					cmd.Run(&test.Expected{
+						ExitCode: 0,
+					})
+					helpers.Ensure("build", "-t", data.Identifier(), "--buildkit-host", fmt.Sprintf("unix://%s", symlinkedBuildkitAddr), data.Get("buildCtx"))
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("run", "--rm", data.Identifier())
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					os.Remove(data.Get("buildkitAddr"))
```
apostasie, line 1035: `data.TempDir` is automatically removed for you at the end of the test, so, no need to do it manually.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,109 @@ RUN ping -c 5 beta
...
+		Cleanup: func(data test.Data, helpers test.Helpers) {
+			helpers.Anyhow("rmi", "-f", data.Identifier())
+		},
+		SubTests: []*test.Case{
+			{
+				Description: "build with buildkit-host",
+				NoParallel:  true,
+				Setup: func(data test.Data, helpers test.Helpers) {
+					buildkitAddr, err := buildkitutil.GetBuildkitHost(testutil.Namespace)
+					if err != nil {
```
apostasie, line 1016: `assert.NilError(helpers.T(), err)`?

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,109 @@ RUN ping -c 5 beta
...
+					helpers.Anyhow("rmi", "-f", data.Identifier())
+				},
+				Expected: test.Expects(0, nil, expect.Equals("nerdctl-build-test-string\n")),
+			},
+			{
+				Description: "build with env specified",
+				NoParallel:  true,
+				Setup: func(data test.Data, helpers test.Helpers) {
+					buildkitAddr, err := buildkitutil.GetBuildkitHost(testutil.Namespace)
+					if err != nil {
```
apostasie, line 1045: ibid

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,109 @@ RUN ping -c 5 beta
...
+
+					symlinkedBuildkitAddr := filepath.Join(data.TempDir(), "buildkit-env.sock")
+					data.Set("buildkitAddr", symlinkedBuildkitAddr)
+
+					oldbuildkitAddr := os.Getenv("BUILDKIT_HOST")
+					if oldbuildkitAddr != "" {
+						defer os.Setenv("BUILDKIT_HOST", oldbuildkitAddr)
+					}
+					data.Set("oldbuildkitAddr", os.Getenv("BUILDKIT_HOST"))
+					os.Setenv("BUILDKIT_HOST", fmt.Sprintf("unix://%s", symlinkedBuildkitAddr))
```
apostasie, line 1058: Mmmm...
Tigron should allow you to manipulate the env inside setup. I'll patch asap so that you can avoid having to touch the global environment.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,109 @@ RUN ping -c 5 beta
...
+					cmd := helpers.Custom("ln", "-s", buildkitAddr, symlinkedBuildkitAddr)
+					cmd.Run(&test.Expected{
+						ExitCode: 0,
+					})
+
+					cmd = helpers.Custom("ls", "-lh", symlinkedBuildkitAddr)
+					cmd.Run(&test.Expected{
+						ExitCode: 0,
+						Output: func(stdout string, info string, t *testing.T) {
+							fmt.Println("build log", stdout)
```
apostasie, line 1070: Recent updates to the test tooling will always display detailed information about all commands, including stdout, so, that should not be necesarry.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,109 @@ RUN ping -c 5 beta
...
+						},
+					})
+
+					cmd = helpers.Command("--debug", "build", "-t", data.Identifier(), data.Get("buildCtx"))
+					cmd.Run(&test.Expected{
+						ExitCode: 0,
+						Output: func(stdout string, info string, t *testing.T) {
```
apostasie, line 1077: ibid

cmd/nerdctl/builder/builder_build_test.go
apostasie: #4116

As soon as this gets merged, you will be able to call `cmd.Setenv()` on a particular command instead of having to manipulate os.Setenv.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,109 @@ RUN ping -c 5 beta
...
+							fmt.Println("build log", stdout)
+						},
+					})
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("run", "--rm", data.Identifier())
+				},
+				Cleanup: func(data test.Data, helpers test.Helpers) {
+					os.Remove(data.Get("buildkitAddr"))
```
apostasie, line 1086: ibid

apostasie:
@Shubhranshu153 
A few nits on test + implementing something to ease your env manipulation.

Also, do we need to keep calling `helpers.AddStringFlag(` instead of a plain `cmd.Flags().String`?

cmd/nerdctl/builder/builder_build_test.go
Shubhranshu153: ack

cmd/nerdctl/builder/builder_build_test.go
Shubhranshu153: was a debug log would remove it

Shubhranshu153:
As the GetBuildkitHost() is kind of high level function used in 3 different places do you think we should move it to inside buildkitutils?

cmd/nerdctl/builder/builder_build_test.go
Shubhranshu153: awesome would replace it once its merged.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,96 @@ RUN ping -c 5 beta
...
+
+		SubTests: []*test.Case{
+			{
+				Description: "build with buildkit-host",
+				NoParallel:  true,
+				Setup: func(data test.Data, helpers test.Helpers) {
+					// Get BuildkitAddr
+					buildkitAddr, err := buildkitutil.GetBuildkitHost(testutil.Namespace)
+					assert.NilError(helpers.T(), err)
+					buildkitAddr = "/" + strings.Trim(buildkitAddr, "unix:/")
```
apostasie, line 1016: So, if `BUILDKIT_HOST=unix:///foo.socket`, `strings.Trim(buildkitAddr, "unix:/")` will equal `//foo.socket` and `buildkitAddr` will equal `///foo.socket`?

It will still work, but that does not seem right?

Shubhranshu153: func main() {
	buildkitAddr := "unix:///foo.sock"
	str := strings.Trim(buildkitAddr, "unix:/")
	fmt.Println(str)
}

this gives out foo.sock, i was also confused when i was writing the test. i think i understand, so it removes the strings in the cutset from prefix and suffix which might be dangerous as it might remove u, n, i,x 

Although it is highly unlikely as the addr has to end with sock

I will change it as technically its wrong.

Shubhranshu153: Fixed This with TrimPrefix

apostasie:
> As the GetBuildkitHost() is kind of high level function used in 3 different places do you think we should move it to inside buildkitutils?

Personal opinion:
We already have one in buildkitutil.
As for the part that processes flags and environment variable, I would rather keep that (somewhere) under `./cmd`.

cmd/nerdctl/builder/builder_builder_test.go
```diff
@@ -71,6 +73,78 @@ CMD ["echo", "nerdctl-test-builder-prune"]`, testutil.CommonImage)
 				Command:  test.Command("builder", "prune", "--force", "--all"),
 				Expected: test.Expects(0, nil, nil),
 			},
+			{
+				Description: "builder with buildkit-host",
+				NoParallel:  true,
+				Setup: func(data test.Data, helpers test.Helpers) {
+					// Get BuildkitAddr
+					buildkitAddr, err := buildkitutil.GetBuildkitHost(testutil.Namespace)
```
apostasie, line 81: This method depends on the existence of buildctl in the path (because of `pingBKDaemon`).
So, this test needs to be marked `test.Require = test.Not(nerdtest.Docker)`.

cmd/nerdctl/builder/builder_builder_test.go
```diff
@@ -71,6 +73,78 @@ CMD ["echo", "nerdctl-test-builder-prune"]`, testutil.CommonImage)
...
+					return helpers.Command("builder", "prune", "--force", "--buildkit-host", fmt.Sprintf("unix://%s", data.Get("symlinkedBuildkitAddr")))
+				},
+				Expected: test.Expects(0, nil, nil),
+			},
+			{
+				Description: "builder with env",
+				NoParallel:  true,
+				Setup: func(data test.Data, helpers test.Helpers) {
+					// Get BuildkitAddr
+					buildkitAddr, err := buildkitutil.GetBuildkitHost(testutil.Namespace)
```
apostasie, line 109: ibid

apostasie:
Linking latest failures to their tickets:
- rootful failure is #3556 
- docker failure I just commented inline of your PR

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -967,6 +968,7 @@ func TestBuildAddHost(t *testing.T) {
 	testCase := &test.Case{
 		Require: require.All(
 			nerdtest.Build,
+			require.Not(nerdtest.Docker),
```
apostasie, line 971: This test does work on docker, right?

apostasie (approved):
One last comment on a test, but otherwise looks fine.

cc @AkihiroSuda @fahedouch for formal maintainer review.

cmd/nerdctl/builder/builder_build.go
```diff
@@ -253,6 +253,14 @@ func GetBuildkitHost(cmd *cobra.Command, namespace string) (string, error) {
 		}
 		return buildkitHost, nil
 	}
+
+	if buildkitHost := os.Getenv("BUILDKIT_HOST"); buildkitHost != "" {
```
fahedouch, line 257: Wouldn't it be more logical to introduce the `--buildkit-host` flag in the `prune` command? 

Unless I'm mistaken, we wanted to have a single entry point for CLI flag and parameters, which is `cmd`. I think this ensures better management of flags, environment variables, and parameters related to the CLI.

AkihiroSuda:
Needs rebase

cmd/nerdctl/builder/builder_build.go
AkihiroSuda: Probably this should be a global flag as in `--address` (`$CONTAINERD_HOST`)

AkihiroSuda:
> Note: helpers.AddStringFlag(cmd, "buildkit-host", nil, "", "BUILDKIT_HOST", "BuildKit address") this masks the logic for build commands but for prune command it doesn't work correctly

How is `buildctl prune` different from `buildctl build` for handling `BUILDKIT_HOST`? 🤔

apostasie:
> > Note: helpers.AddStringFlag(cmd, "buildkit-host", nil, "", "BUILDKIT_HOST", "BuildKit address") this masks the logic for build commands but for prune command it doesn't work correctly
> 
> How is `buildctl prune` different from `buildctl build` for handling `BUILDKIT_HOST`? 🤔

I think @Shubhranshu153 was talking about `nerdctl system prune` (not `nerdctl build prune`).

apostasie:
> > > Note: helpers.AddStringFlag(cmd, "buildkit-host", nil, "", "BUILDKIT_HOST", "BuildKit address") this masks the logic for build commands but for prune command it doesn't work correctly
> > 
> > 
> > How is `buildctl prune` different from `buildctl build` for handling `BUILDKIT_HOST`? 🤔
> 
> I think @Shubhranshu153 was talking about `nerdctl system prune` (not `nerdctl build prune`).

eg: my understanding is that `buildctl` itself never look into  `BUILDKIT_HOST` because we explicitly pass to it a `--addr` argument.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -989,3 +990,99 @@ RUN ping -c 5 beta
...
+					// TODO: Remove this logic once we have a way to pass in env to the helper function
+					oldbuildkitAddr := os.Getenv("BUILDKIT_HOST")
+					if oldbuildkitAddr != "" {
+						defer os.Setenv("BUILDKIT_HOST", oldbuildkitAddr)
+					}
+					data.Set("oldbuildkitAddr", os.Getenv("BUILDKIT_HOST"))
+					os.Setenv("BUILDKIT_HOST", fmt.Sprintf("unix://%s", symlinkedBuildkitAddr))
+
+					// Do a negative test to ensure setting up the env variable is effective
+					helpers.Fail("build", "-t", data.Identifier(), data.Get("buildCtx"))
```
apostasie, line 1063: @Shubhranshu153 tigron PR with per-command Setenv support got merged.

You can now (for main commands):
```
cmd := helpers.Command("blbalbal")
cmd.Setenv("FOO", "BAR")
```

Instead of `helpers.Fail` use:
```
cmd := helpers.Command("blbalbal")
cmd.Setenv("FOO", "BAR")
cmd.Run(&test.Expected{ExitCode: expect.ExitCodeGenericFail})
```

Instead of `helpers.Ensure`:

```
cmd := helpers.Command("blbalbal")
cmd.Setenv("FOO", "BAR")
cmd.Run(&test.Expected{ExitCode: expect.ExitCodeSuccess})
// Or more simply
cmd.Run(&test.Expected{})
```

cmd/nerdctl/builder/builder_build.go
AkihiroSuda: A global flag can be specified for a subcommand too, so the compatibility is not affected AFAICS

apostasie:
@Shubhranshu153 main is bust right now.

Suggesting you wait for #4128 then rebase.

cmd/nerdctl/builder/builder_build.go
Shubhranshu153: ok i can add it to the global flag and refactor the code. Makes sense to me.
Thanks

apostasie:
Tagging unrelated failure: #4068

cmd/nerdctl/builder/builder_build.go
Shubhranshu153: i started doing this change but i feel this requires a larger refactor and some testing, will want to put it in a separate PR.

cmd/nerdctl/builder/builder_build_test.go
```diff
@@ -977,3 +979,90 @@ RUN ping -c 5 beta
...
+			require.Not(nerdtest.Docker),
+		),
+		Setup: func(data test.Data, helpers test.Helpers) {
+			dockerfile := fmt.Sprintf(`FROM %s
+CMD ["echo", "nerdctl-build-test-string"]`, testutil.CommonImage)
+			data.Temp().Save(dockerfile, "Dockerfile")
+			data.Labels().Set("buildCtx", data.Temp().Path())
+
+		},
+		NoParallel: true,
```
apostasie, line 998: Do we still need `NoParallel` on this one (and its subtests)?

apostasie (approved):
One minor question about NoParallel on the first test, but otherwise LGTM.
Thanks @Shubhranshu153 !

Shubhranshu153:
Can we merge this? Thanks

AkihiroSuda (approved):
Thanks

PR #4116, opened 2025-04-17
[Tigron]: allow per-command environment override

apostasie:
While `test.Case` allows declaration of custom environment variables, they must be set during the test declaration, apply to all commands, and cannot be altered during lifecycle operations (eg: Setup). This PR provides the ability to set variables for specific, individual commands.

Use-case seen in #4115


AkihiroSuda (approved):
Thanks

PR #4111, opened 2025-04-16
[CI]: quality of life: minor debugging output fixes

apostasie:
- commit 1 has a few assorted minor fixes and changes (see commit message for details)
- commit 2 does make use of recently introduced env Whitelisting to tame Environ display

Issue #4118, opened 2025-04-17
[CI]: ipfs /root/.ipfs/api not available


apostasie:
### Description

Seen on EL8 - which is notoriously slower than other environment.

Looks like ipfs could be very slow to start - or timeouting somehow.

`/root/.ipfs/api` is not there when we need it.

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

```
    issues_linux_test.go:156: 
        
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl image push ipfs://ghcr.io/stargz-containers/alpine:3.13-org            |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🚫 command returned a non-zero exit code                                                         |
        +------------------------------------------------------------------------------------------------------------+
        |         | ⚠️ 1                                                                                             |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🟠 time="2025-04-17T07:15:41Z" level=info msg="pushing image \"ghcr.io/stargz-containers/alpine: |
        |         | 3.13-org\" to IPFS"                                                                              |
        |         | time="2025-04-17T07:15:41Z" level=info msg="ensuring image contents"                             |
        |         | time="2025-04-17T07:15:41Z" level=warning msg="failed to ensure the existence of image \"ghcr.io |
        |         | /stargz-containers/alpine:3.13-org\"" error="open /root/.ipfs/api: no such file or directory"    |
        |         | time="2025-04-17T07:15:41Z" level=warning msg="ipfs push failed" error="open /root/.ipfs/api: no |
        |         |  such file or directory"                                                                         |
        |         | time="2025-04-17T07:15:41Z" level=fatal msg="open /root/.ipfs/api: no such file or directory"    |
        +------------------------------------------------------------------------------------------------------------+
        | 🌱      | HOME=/root                                                                                       |
        |         | PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin |
        |         | :/bin                                                                                            |
        |         | NERDCTL_TOML=/tmp/TestIssue3425with_ipfs840600782/001/nerdctl.toml                               |
        |         | DOCKER_CONFIG=/tmp/TestIssue3425with_ipfs840600782/001                                           |
        |         | IPFS_PATH=/root/.ipfs                                                                            |
[...]

PR #4122, opened 2025-04-17
[CI]: rewrite some container and compose tests

apostasie:
I have been delaying engaging with compose and container tests so far because of the daunting amount of work...

But at this point, they are the number one contributor to overall slowness (and most of remaining flakyness).
Moving these will cut through the total runtime and make a dent in the bucket of "tests with retries".

~~Note that we should merge first, before this:~~
~~- #4111~~
~~- #4081~~
~~- #4116~~

~~and preferably #4114.~~

~~... as this PR does leverage some of the recent changes in Tigron (specifically temp files manipulation).~~

~~I will obviously take care of rebasing once they are in, though the last commit of this PR could be reviewed already at this point.~~

Note that this PR also adopt / cleans-up / remove deprecated helpers (some of them wrongly located under `./cmd`, some under `./pkg/testutil`).

I know it is painful to review large PRs like that, but I believe there is really no other way here - and at the end of the day, this is only tests - the CI being green is proof in the pudding.

### DIFF
--- a/pkg/testutil/nerdtest/utilities.go
+++ b/pkg/testutil/nerdtest/utilities.go
@@ -18,12 +18,15 @@ package nerdtest
 
 import (
 	"encoding/json"
+	"net"
+	"strings"
 	"testing"
 	"time"
 
 	"gotest.tools/v3/assert"
 
 	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
 	"github.com/containerd/nerdctl/mod/tigron/tig"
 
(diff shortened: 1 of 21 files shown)


apostasie:
Failures are:
- #4068 
- #4118 
- windows not happy with TestLogsOutStreamsSeparated anymore

AkihiroSuda:
Needs rebase

apostasie:
Failure is #4118

apostasie:
Rebasing for good measure.

apostasie:
Failure is again buildg timeout.

I do not have a good explanation for why we are seeing this so much now.
Intuition would suggest:
- recent update in buildkitd that trips this (also buildg is linking a quite old buildkit fork, so, compat may be an issue)
- recent changes in test.Command which may slightly alter behavior when the streams are attached to buildg (seems unlikely, but then?)
- changes in tests (or otherwise github runner sizing change) that would increase the pressure on the github instance, making for a slower response, that would (now) trip over the timeout (rather short, at 3 secs)

#4103 should get to the root cause eventually, but it is an excruciating process to nail flaky behavior in third party binaries...

pkg/testutil/nerdtest/utilities.go
```diff
@@ -132,3 +135,59 @@ func EnsureContainerStarted(helpers test.Helpers, con string) {
...
+
+	cmd := helpers.Custom("cosign", "generate-key-pair")
+	cmd.WithCwd(data.Temp().Path(path))
+	cmd.Setenv("COSIGN_PASSWORD", password)
+	cmd.Run(&test.Expected{})
+
+	return data.Temp().Path(path, "cosign.key"), data.Temp().Path(path, "cosign.pub")
+}
+
+func FindIPv6(output string) net.IP {
```
djdongjin, line 180: what's the `output`? Can we use a more descriptive name or add comments?

djdongjin (approved):
I hope we can try to reduce PR sizes in future PRs.

> I know it is painful to review large PRs like that, but I believe there is really no other way here - and at the end of the day, this is only tests - the CI being green is proof in the pudding.

I would suggest at least split the PR by areas, e.g., one for `compose`, one for `image`, one for `container`.

pkg/testutil/nerdtest/utilities.go
apostasie: This function has been moved as-is.

https://github.com/containerd/nerdctl/pull/4122/files/ff8b599ae6bf66eb5f38a256f256da18fdb3caec#diff-5403777981474698937be66c4c7636b4d2a131fe7ef93afebb500f50e9bf8c25L36-L49

Eventually, I will do a pass on all these helpers and document / enhance param names when done rewriting tests.

If that is fine with you, I would rather do that later rather than as part of this here.
Lmk your thoughts.

apostasie:
> I hope we can try to reduce PR sizes in future PRs.
> 
> > I know it is painful to review large PRs like that, but I believe there is really no other way here - and at the end of the day, this is only tests - the CI being green is proof in the pudding.
> 
> I would suggest at least split the PR by areas, e.g., one for `compose`, one for `image`, one for `container`.

That is fair - will do.

Thanks @djdongjin

pkg/testutil/nerdtest/utilities.go
djdongjin: If it's existing code then it's okay to keep as is (or improve in follow up)

AkihiroSuda (approved):
Thanks

PR #4127, opened 2025-04-18
[PRIORITY] [CI]: Fix EL8 tests execution

apostasie:
Pretty sure I wrote that code... :/

If rootless fails, it will ALSO run rootful.
PR addresses this.

Kids: say no to bash.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -489,20 +489,20 @@ jobs:
         # This might be due to the old kernel shipped with Alma (4.18), or something else between centos/docker.
         run: |
           set -eux
-          [ "$MODE" = "rootless" ] && {
+          if [ "$MODE" = "rootless" ]; then
             echo "rootless"
             docker run -t -v /dev:/dev --rm --privileged test-integration /test-integration-rootless.sh ./hack/test-integration.sh -test.only-flaky=false
-          } || {
+          else
             echo "rootful"
             docker run -t -v /dev:/dev --rm --privileged test-integration ./hack/test-integration.sh -test.only-flaky=false
-          }
+          fi
       - name: "Run integration tests (flaky)"
         run: |
           set -eux
-          [ "$MODE" = "rootless" ] && {
+          if [ "$MODE" = "rootless" ]; then
             echo "rootless"
             docker run -t -v /dev:/dev --rm --privileged test-integration /test-integration-rootless.sh ./hack/test-integration.sh -test.only-flaky=true
-          } || {
+          else
             echo "rootful"
             docker run -t -v /dev:/dev --rm --privileged test-integration ./hack/test-integration.sh -test.only-flaky=true
-          }
+          fi


apostasie:
Failure is #4046

.github/workflows/test.yml
```diff
@@ -501,7 +501,7 @@ jobs:
           set -eux
           [ "$MODE" = "rootless" ] && {
             echo "rootless"
-            docker run -t -v /dev:/dev --rm --privileged test-integration /test-integration-rootless.sh ./hack/test-integration.sh -test.only-flaky=true
+            docker run -t -v /dev:/dev --rm --privileged test-integration /test-integration-rootless.sh ./hack/test-integration.sh -test.only-flaky=true || exit
```
AkihiroSuda, line 504: This basically disables `set -e` and makes the test always green? 🤔

.github/workflows/test.yml
```diff
@@ -501,7 +501,7 @@ jobs:
           set -eux
           [ "$MODE" = "rootless" ] && {
```
AkihiroSuda, line 502: Why don't we use `if EXPR then` here ?

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/4127/files#r2050754558

.github/workflows/test.yml
apostasie: No.
`exit` will exit wit the exit code of the previous command.
eg: what it will really do is:
`blablabla || exit $?`

Nevertheless, I see this is confusing to everybody (and I got tripped as well, clearly, so...) - will change to a plain old `if else` to make this more readable.

AkihiroSuda: 👀 

Thanks, but `if else` seems more readable

apostasie: Yeah, agreed (and done).

apostasie: @AkihiroSuda do you remember https://en.wikipedia.org/wiki/Underhanded_C_Contest ?

Every time I write bash, I feel like it was designed to do this sort of evil things 😂.

AkihiroSuda (approved):
Thanks

apostasie:
Failure is #4046

.github/workflows/test.yml
AkihiroSuda: Wasn't aware of that.
Similar to IOCCC but the codes are ostensibly-readable?

apostasie: Yeah, pretty similar, but indeed made to look readable / honest mistake.

I believe this is the kind of attacks that are the most dangerous, because it is so easy to overlook.
(eg: https://www.infoq.com/news/2013/10/Linux-Backdoor/ )

apostasie:
And rootful failure is a compose IPFS inability to start the container (we have seen that before - just not sure it is a separate bug or a variant of one of the already logged-one - I will focus on it when done with the other flakies)

apostasie:
Rootless failure is again #4046. That's the last straw. I am going after this one now.

PR #4134, opened 2025-04-19
Mark kube target as flaky

apostasie:
In light of the amount of failures we are getting, which are presumably a problem with async delete in CRI, and the fact we cannot easily swap out a different containerd version, suggesting we mark this target as flaky and just ignore it.


AkihiroSuda (approved):
Thanks

PR #4129, opened 2025-04-18
[CI]: Evergreen / Everred

apostasie:
This PR is a persistent playground for future improvements to the CI, build, Dockerfile, test tooling (Tigron), or tests themselves.

It is likely messy, is never meant to be merged, but does showcase future direction and sandboxes improvements to be evaluated.

Actual mergeable PRs will be broken out of it.

Issue #4132, opened 2025-04-19
[CI]: events filter sometimes does not work


apostasie:
### Description

Here is an example from the CI where event filter does not produce on stdout the expected output.
(ignore the timeout aspect, which is by design)

It is unclear if this is a test issue, or a code issue related to events.

I have seen this before, ~~but it is awfully rare~~ it happens regularly on EL, so, this is likely going to be ~~very hard~~ a bit of a trek to debug.

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

```
test target: "nerdctl"
PASS cmd/nerdctl/system.TestSystemPrune/buildkit (2.04s)
PASS cmd/nerdctl/system.TestSystemPrune/volume_prune_all_success (2.33s)
PASS cmd/nerdctl/system.TestSystemPrune (2.04s)
PASS cmd/nerdctl/system.TestInfo/info_convenience_form (0.02s)
PASS cmd/nerdctl/system.TestInfo/info (0.02s)
PASS cmd/nerdctl/system.TestEventFilters/CapitalizedFilter (10.01s)
PASS cmd/nerdctl/system.TestEventFilters/UnsupportedEventFilter (10.01s)
PASS cmd/nerdctl/system.TestInfo/info_with_namespace (0.01s)
PASS cmd/nerdctl/system.TestInfo/info_with_namespace_env_var (0.02s)
PASS cmd/nerdctl/system.TestInfo (0.00s)
PASS cmd/nerdctl/system.TestEventFilters/UnsupportedStatusFilter (10.01s)
PASS cmd/nerdctl/system.TestEventFilters/StatusFilter (10.00s)
=== RUN   TestEventFilters/StartEventFilter
=== PAUSE TestEventFilters/StartEventFilter
=== CONT  TestEventFilters/StartEventFilter
    command.go:461: [2025-04-19T01:00:54Z] [command=/usr/local/bin/nerdctl --namespace=nerdctl-test events --filter event=start --format json] command cancelled
    system_events_linux_test.go:36: 
        
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl run --rm ghcr.io/stargz-containers/alpine:3.13-org                     |
        +------------------------------------------------------------------------------------------------------------+
        | 🌱      | HOME=/root                                                                                       |
[...]

PR #4137, opened 2025-04-19
[Tigron]: data SaveToWriter method

apostasie:
On top of #4122.
Broken out of #4129.

This is a small addition to the DataTemp interface, allowing tests author to get a Writer to a managed temporary file.

This is of particular interest for tests that need to create temporary resources with systems expecting an io.Writer. For example: `pem.Encode`.

Next PR will re-jiggle x509 manipulation and leverage this.


apostasie:
Failure is #4132

AkihiroSuda (approved):
Thanks

PR #4140, opened 2025-04-20
[DEBUGGING]: evaluate what happens if we disable gotestsum retries

apostasie:
A lot of work has improved stability since I last tried this.
Though we still have some flaky tests in the "non-retry" bucket, we need more info at this point on the state of the "retry" bucket that is likely hiding more issues.

This is a debugging PR at this point, not meant to be merged.

### DIFF
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -41,7 +41,7 @@ fi
 
 for arg in "$@"; do
   if [ "$arg" == "-test.only-flaky=true" ] || [ "$arg" == "-test.only-flaky" ]; then
-    args+=("--rerun-fails=$retries")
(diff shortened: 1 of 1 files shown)


apostasie:
Ok... this is not a pleasing sight...

apostasie:
Second run brings:
- one count of #4143 
- one count of #4146 
- several counts of #4142 
- one count of #4147
- several counts of compose ipfs failure to communicate with database
- multiple counts of TestIPFSCompNoBuild content digest failure

apostasie:
Third run brings:
- one count of #4143 on TestComposeUpWithBypass4netns
- four counts of #4142 on the same run <- we have a serious problem here

PR #4154, opened 2025-04-23
[WIP] [CI] cleanup series end-goal

apostasie:
NOTE: this is (mostly) the desired end-result here. It is not meant to be merged as-is, but to beget smaller PR and be broken down.

### Motivation

Github workflows file are hard to maintain, read and understand at this point, because of how they grew organically over time:
1. certain files are overcrowded
2. there is a lot of duplication / variants / repetition
3. variables / versions are peppered all over the place

Suggestion is that we leverage GitHub reusable workflows to encapsulate independent tasks, with proper `inputs`, and to reduce the number of entrypoint workflow files in a more logical way.

### About the remainder here

Highlights:
- integration testing jobs have been broken out into "how" they are being run (on host, in container, in lima, in vagrant)
- streamlined `in-container` testing now has a single job able to accommodate all variants including `ipv6`, `canary`, `gomodjail`, instead of having separate jobs
- `in-lima` is now expanded to allow selecting the lima template, so that we can _also_ test EL9 (at least temporarily), see rationale inside the file for why
-  there is now three main nerdctl top-level workflows: `lint`, `test`, and `flaky`

-------------------------------
   - `lint` now look like this:
 
Screenshot 2025-05-02 at 1 08 30 PM

-------------------------------
   - `test` contains unit-tests, integration tests in-container and on-host

Screenshot 2025-05-02 at 1 09 32 PM

-------------------------------
   - `flaky` contains ad-hoc stuff: in-vagrant testing, kube, in-lima testing

Screenshot 2025-05-02 at 2 08 21 PM

-------------------------------

Efforts have been made to wrangle the "titles/description" to be as concise as possible while still providing relevant context info.
Pretty sure this is not perfect, so, **_let me know if any title needs adjusting_**.

I am not particularly attached to how exactly we organize the top-level workflows and what they contain (what I care about is that jobs are cleanly expressed, separated, and can have their matrix easily modified or expanded, providing an easy to look at high level view), so, **_if you have opinion on better distribution accross the top-level workflows, go for it_**.

Finally note that:
- on-host test for nerdctl-linux (instead of in-container) is structurally there, but not functional yet - more work has to be done for provisioning of dependencies, and this requires thinking+consolidation with Dockerfile so that we do not duplicate work and have one way of retrieving dependencies and verifying their shas
- **windows base containerd version has been bumped from 1.7 to 2.0** (discussed in prior testing matrix ticket)
- I have not touched at all `release` and `ghcr-image-build` - there is little upside in changing them, and I do not like messing-up with release critical things for frivolous reasons - we can still rename the workflow files if we want - just let me  know what would be the right names


.github/workflows/environment.yml
AkihiroSuda, line 1: Why do we need this?

.github/workflows/workflow-lint.yml
```diff
@@ -0,0 +1,42 @@
...
+    uses: ./.github/workflows/job-lint-go.yml
+    with:
+      timeout: 5
+      go-version: 1.24
+      runner-for-linux: ubuntu-24.04
+      runner-for-freebsd: ubuntu-24.04
+      runner-for-macos: macos-15
+      # FIXME: this is currently failing in a nonsensical way, so, running on linux instead...
+      # runner-for-windows: windows-2022
+      runner-for-windows: ubuntu-24.04
```
AkihiroSuda, line 26: Why do we need these `with` arguments?
Shouldn't we just use a `matrix` for `runs-on`?

.github/workflows/environment.yml
apostasie: Still WIP sorry - should have marked draft.

.github/workflows/workflow-lint.yml
apostasie: Let me finish this.
Marked it draft now.
Sorry.

apostasie:
This is done - everything has been broken out.

Thanks a lot @AkihiroSuda as usual for all the reviews and feedback.

Closing.

PR #4155, opened 2025-04-23
[Tigron]: shorten max lines in logs to 50 and add message

apostasie:
@AkihiroSuda latest run on your gomodjail PR I cannot even get the logs on the GitHub UI.

It is possible that the UI is just choking because of the amount of information we are printing there.

This PR reduces the "maxlength" of stdout/stderr being displayed to 50 lines instead of 100, and also adds information for the user about the total number of lines, that should help us figure out when we have abnormally large quantities and maybe help poor GitHub UI?

### DIFF
--- a/mod/tigron/internal/formatter/formatter.go
+++ b/mod/tigron/internal/formatter/formatter.go
@@ -25,7 +25,7 @@ import (
 
 const (
 	maxLineLength = 110
-	maxLines      = 100
+	maxLines      = 50
 	kMaxLength    = 7
 	spacer        = " "
 )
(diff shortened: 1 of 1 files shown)


apostasie:
Alternatively, we could have a `godzillaThreshold` const (maybe `500` lines) that would make the logger not display anything at all, but instead print a warning.

That way "normal" logs could still be kept at 100 lines max (so we do not loose too much information), while preventing crazy stuff from happening.

lmk if you think this would be better than the current proposal.

AkihiroSuda (approved):
Thanks, let's try to see if this works

PR #4156, opened 2025-04-23
[CI]: remove arm/v7 from testing

apostasie:
The corresponding tests serve little to no-purpose (eg: as for adding value specifically for armv7) and are being replaced by equivalent tests against `arm64`, `linux/arm64`, `linux/arm64/v8`.

We do not test any other (s390x, etc) while we could.

Since setting up emulation for arm/v7 is not entirely trivial (because of how tonis image works), suggesting we just get rid of this:
- remove the binfmt contorsions and likelyhood of tests failing locally
- remove the need for an alpine arm v7 availability (important for upcoming work on better controlling test images)
- arm/v7 is going the way of dinosaurs anyhow (even embedded these days gets v8) and our current testing is not doing much anyhow

### DIFF
--- a/.github/workflows/test-canary.yml
+++ b/.github/workflows/test-canary.yml
@@ -34,13 +34,8 @@ jobs:
           sudo losetup -lv
       - name: "Register QEMU (tonistiigi/binfmt)"
         run: |
-          # `--install all` will only install emulation for architectures that cannot be natively executed
-          # Since some arm64 platforms do provide native fallback execution for 32 bits,
-          # armv7 emulation may or may not be installed, causing variance in the result of `uname -m`.
-          # To avoid that, we explicitly list the architectures we do want emulation for.
           docker run --privileged --rm tonistiigi/binfmt --install linux/amd64
           docker run --privileged --rm tonistiigi/binfmt --install linux/arm64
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
> arm/v7 is going the way of dinosaurs anyhow 

Not sure about this.
The golang and the OCI ecosystem still supports arm/v7.

apostasie:
> > arm/v7 is going the way of dinosaurs anyhow
> 
> Not sure about this. The golang and the OCI ecosystem still supports arm/v7.

Sure.
They also support arm/v6 (I think?) - along with s390x, ppc, etc - and we are/should too (by the way, we might want to verify that we can at least build nerdctl for these targets).

I guess the point I would like to outline here is that the tests I am changing here are doing:
- validate that we can run multiple platforms when we have emulation (and we can test that with just arm64, we do not need arm/v7 specifically - we just need a non-native, emulated platform)
- validate that platform name aliasing works (we can do that with linux/arm64/v8 as well) (and also that is probably better suited for unit testing)


I am now wondering if I did not suggest this change earlier actually...
If I did, apologies for that.

Lmk your thoughts (no strong opinion here - just preparing for some CI cleanup)

apostasie:
Closing.

PR #4160, opened 2025-04-25
[CI]: cleanup, breakout 2: small cleanups and fixes

apostasie:
Commit 1: go-setup cleanups
- remove useless .x suffixes
- consistently enforce check latest
- consistent step title

Commit 2:
- remove freebsd broken var

Commit 3:
- fix EL8 cache path

On top of #4159 - broken out of #4154

PR #4161, opened 2025-04-25
[CI]: cleanup, breakout 3: separate jobs

apostasie:
On top of #4160 - breakout of #4154.

TL;DR: workflows have grown organically to the point of being hard to read and reason about.
This PR shows a sample of what is proposed in #4154 so that we can discuss here on a reasonably-sized PR.

The perceived problems are:
- workflow files are overcrowded
- there is a lot of duplication
- variables and versions are all over the place

The proposal is to have:
- top level `workflow-X` files that only contain calls to jobs and are responsible for passing along versions, meant to outline the overall structure of the CI
- individual `jobs-Y` files for the jobs, in their own "logical" file, where X describe clearly what the _function_ of the job is

This specific PR is limited to solely `unit testing` and provides a single `job` that will now take care of unit-testing, including for canary instead of having it duplicated in two places.

------------

There has been concerns over #4154 about the use of `inputs` for jobs.

I do personally find it better (which is why I am proposing it :-) ):
- explicit versions will be in one single place, instead of being duplicated accross
- I find it easier to read / more logical to manipulate `product-version-old` and  `product-version-stable`, as I feel it conveys the _intent_, rather than squinting on the difference between `ubuntu-24.04` and `ubuntu-22.04`

Evidently I appreciate this may boil down to a matter of personal taste and that we may not agree.

If there is indeed sentiment against using variables and a preference to just embed versions, fine with me - we can still have the rest (separation and cleanup of jobs and workflows).

LMK your thoughts.

### DIFF
--- a/.github/workflows/workflow-test.yml
+++ b/.github/workflows/workflow-test.yml
@@ -52,40 +52,31 @@ jobs:
             --target build-dependencies --build-arg CONTAINERD_VERSION=${CONTAINERD_VERSION} .
 
   test-unit:
-    # FIXME:
-    # Supposed to work: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-returning-a-json-data-type
-    # Apparently does not
-    # timeout-minutes: ${{ fromJSON(env.SHORT_TIMEOUT) }}
-    timeout-minutes: 10
-    name: unit | ${{ matrix.goos }}
-    runs-on: "${{ matrix.os }}"
-    defaults:
-      run:
-        shell: bash
+    name: "unit${{ inputs.hack }}"
+    uses: ./.github/workflows/job-test-unit.yml
     strategy:
       fail-fast: false
       matrix:
+        # Run on all supported platforms but freebsd
+        # Additionally run on canary for linux
         include:
-          - os: windows-2022
-            goos: windows
-          - os: ubuntu-24.04
-            goos: linux
-          - os: macos-15
-            goos: darwin
-    steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
-        with:
-          fetch-depth: 1
-      - name: "Install go"
-        uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b  # v5.4.0
-        with:
-          go-version: ${{ env.GO_VERSION }}
-          check-latest: true
-      - if: ${{ matrix.goos=='windows' }}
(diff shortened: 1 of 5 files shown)


apostasie:
Rootful failure is Hub 429

apostasie:
Second commit is to fix #4167

.github/workflows/workflow-test.yml
```diff
@@ -52,40 +52,19 @@ jobs:
...
+    name: "unit"
+    uses: ./.github/workflows/job-test-unit.yml
+    with:
+      # Windows routinely go over 5 minutes
+      timeout: 10
+      go-version: 1.24
+      runner-for-linux: ubuntu-24.04
+      runner-for-macos: macos-15
+      # 2022 is incredibly slow to retrieve go mod cache, so, for unit testing go with 2025
+      runner-for-windows: windows-2025
```
AkihiroSuda, line 64: runner-for-OS variables are not needed if we define a matrix here

.github/workflows/workflow-test.yml
```diff
@@ -52,40 +52,19 @@ jobs:
...
+      windows-cni-version: v0.3.1
+      linux-cni-version: v1.7.1
+      linux-cni-sha: 1a28a0506bfe5bcdc981caf1a49eeab7e72da8321f1119b7be85f22621013098
```
AkihiroSuda, line 67: Eventually this should be consolidated to SHA256SUMS.d

pkg/netutil/netutil_test.go
```diff
@@ -34,7 +34,7 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 )
 
-const testBridgeIP = "10.1.100.1/24" // nolint:unused
+const testBridgeIP = "10.42.100.1/24" // nolint:unused
```
AkihiroSuda, line 37: ?

.github/workflows/workflow-test.yml
apostasie: Agreed.

pkg/netutil/netutil_test.go
apostasie: Conflicts with the runner address space.

.github/workflows/workflow-test.yml
apostasie: Ok.
`runner-for-X` variables are now replaced by `runner` and `canary` input variables.

apostasie:
God this is awful. Editing github yaml is exactly like punched cards programming. Whoever came-up with this needs to be sent back to the 60s.

AkihiroSuda (approved):
Thanks

PR #4165, opened 2025-04-25
[FIX]: tentatively fix concurrent access to locked CNI directory (#3783)

### LINKED ISSUE #3783: [Cilium] Executing nerdctl run in k8 environment is stuck
### Description

Executing nerdctl run in the k8 environment is stuck, but k8s can create pods normally

1.[root@m1 ~]# nerdctl ps
CONTAINER ID    IMAGE                                                                                                       COMMAND                   CREATED           STATUS    PORTS    NAMES
f0571a9094ce    quay.io/cilium/hubble-ui-backend@sha256:0e0eed917653441fded4e7cdb096b7be6a3bddded5a2dd10812a27b1fc6ed95b    "/usr/bin/backend"        6 minutes ago     Up                 k8s://kube-cilium/hubble-ui-77555d5dcf-pj77v/backend
046ba04231f7    docker.io/wangyanglinux/myapp:v1                                                                            "nginx -g daemon off;"    6 minutes ago     Up                 k8s://default/test-z2gms/test
5c6c52541c37    docker.io/wzxmtlw/metrics-server:v0.6.3                                                                     "/metrics-server --c…"    6 minutes ago     Up                 k8s://kube-system/metrics-server-5c7b6df7d8-md58r/metrics-server
fcb24a33d77a    quay.io/cilium/hubble-relay@sha256:d352d3860707e8d734a0b185ff69e30b3ffd630a7ec06ba6a4402bed64b4456c         "hubble-relay serve"      7 minutes ago     Up                 k8s://kube-cilium/hubble-relay-7bc7544857-95dqm/hubble-relay
....

2.[root@m1 ~]# nerdctl run --name test --rm -it busybox:1.28 /bin/sh        
Executing the above command gets stuck

3.Can nerdctl run be executed outside the k8s environment

null

[root@m1 ~]# nerdctl version
Client:
 Version:	v2.0.2
 OS/Arch:	linux/amd64
 Git commit:	1220ce7ec2701d485a9b1beeea63dae3da134fb5
 buildctl:
  Version:	v0.17.1
  GitCommit:	8b1b83ef4947c03062cdcdb40c69989d8fe3fd04

Server:
 containerd:
  Version:	v2.0.1
  GitCommit:	88aa2f531d6c2922003cc7929e51daf1c14caa0a
 runc:
  Version:	1.2.2
  GitCommit:	v1.2.2-0-g7cb36325

None

### Host information

[root@m1 ~]# nerdctl info
Client:
 Namespace:	k8s.io
 Debug Mode:	false

Server:
 Server Version: v2.0.1
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Storage: native overlayfs
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version:   5.14.0-427.13.1.el9_4.x86_64
 Operating System: Rocky Linux 9.4 (Blue Onyx)
 OSType:           linux
 Architecture:     x86_64
 CPUs:             4
 Total Memory:     3.793GiB
 Name:             m1
 ID:               b26f2865-ca8a-49fa-a3a2-ec66adae9813

apostasie:
See #3783 for discussion.

TL;DR: based on investigation above, the hypothesis is that other systems may add locks over the CNI directory.
Since we do as well, this seem to deadlock Cilium.

This patch is moving our lock for CNIEnv to a sub-directory (which appears to be the least disruptive option - see unhappy comment below :-) ), which should allow Cilium (and others) to go through.

HOWEVER:
- I believe this is a shortcoming of CNI plugins in the first place (ie: that they are enumerating the filesystem without locking), and we should not have to do that ourselves
- it is unclear to me if that will really fix the issue with Cilium, or if it is going to create **more** (as in: since we no longer prevent concurrent access, what will happen if both Cilium and nerdctl touch the same resources, ignoring each other lock?)

Nevertheless, it feels like the investigation in #3783 is conclusive enough, and this should _only_ enhance the situation at worse (even if it does not fix everything).

Suggesting we consider for inclusion in 2.0.5.

### DIFF
--- a/pkg/lockutil/lockutil_unix.go
+++ b/pkg/lockutil/lockutil_unix.go
@@ -28,6 +28,7 @@ import (
 )
 
 func WithDirLock(dir string, fn func() error) error {
+	_ = os.MkdirAll(dir, 0700)
 	dirFile, err := os.Open(dir)
 	if err != nil {
 		return err
@@ -55,6 +56,7 @@ func flock(f *os.File, flags int) error {
 }
 
 func Lock(dir string) (*os.File, error) {
+	_ = os.MkdirAll(dir, 0700)
 	dirFile, err := os.Open(dir)
 	if err != nil {
 		return nil, err
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -56,7 +56,8 @@ func (e *CNIEnv) ListNetworksMatch(reqs []string, allowPseudoNetwork bool) (list
 	var err error
 
 	var networkConfigs []*NetworkConfig
-	err = lockutil.WithDirLock(e.NetconfPath, func() error {
+	// NOTE: we cannot lock NetconfPath directly, as Cilium (maybe others) are also locking it.
+	err = lockutil.WithDirLock(filepath.Join(e.NetconfPath, ".nerdctl.lock"), func() error {
 		networkConfigs, err = e.networkConfigList()
 		return err
 	})
@@ -220,7 +221,7 @@ func (e *CNIEnv) NetworkList() ([]*NetworkConfig, error) {
 		netConfigList, err = e.networkConfigList()
 		return err
 	}
-	err = lockutil.WithDirLock(e.NetconfPath, fn)
+	err = lockutil.WithDirLock(filepath.Join(e.NetconfPath, ".nerdctl.lock"), fn)
 
 	return netConfigList, err
 }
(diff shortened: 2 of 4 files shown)


apostasie:
Geez...
I hate this already.

apostasie:
- putting lock files inside the directories is going to fuckup store implementations that will now see extra stuff in their directory
- adding sidecar lock files alongside the directory will leak files as we cannot safely remove them without overly complicating all this, which is currently breaking tests (and well, is bad because creating leftover files plenty)
- locking out of tree will have the same problem

(╯°□°)╯︵ ┻━┻

AkihiroSuda (approved):
Thanks

Issue #4018, opened 2025-03-18
nerdctl restart Fata[0000] 1 errors: failed to start binary process with cmdArgs []: exec: no command


AshinWu:
Environment: Debian11
Kernel version: 5.10.0-26-amd64
k8s version: 1.32.2
nerdctl version: [v2.0.3](https://github.com/containerd/nerdctl/tree/v2.0.3)
cni：cilium1.7 and set kube-proxy-replacement: true

Installed wiit `nerdctl-full`:
> Update containerd (2.0.2), runc (1.2.4), CNI plugins (1.6.2), RootlessKit (2.3.2), bypass4netns (0.4.2), containerd-fuse-overlayfs (2.1.1), BuildKit (0.19.0)

Configuration：

```
# etc/nerdctl/nerdctl.toml
address        = "unix:///run/containerd/containerd.sock"
namespace      = "k8s.io"
cgroup_manager = "systemd"
```

Reproduction conditions:

When I execute `nerdctl ps`, I can query the containers on the k8s node normally. 
However, **when I execute `nerdctl restart ${containerid}`, there is always an error**：

```
FATA[0000] 1 errors:
failed to start binary process with cmdArgs []: exec: no command
```

Similarly, `nerdctl start` and `nerdctl stop` also report the same errors.

Is this a bug or is there a configuration fix? Thx

apostasie:
@AkihiroSuda you have a repro?

apostasie:
@AshinWu

Would you be able to provide the output of the following?

- `nerdctl restart --debug-full ${containerid}`
- `nerdctl inspect ${containerid}`

Thanks.

AshinWu:
> [@AshinWu](https://github.com/AshinWu)
> 
> Would you be able to provide the output of the following?
> 
> * `nerdctl restart --debug-full ${containerid}`
> * `nerdctl inspect ${containerid}`
> 
> Thanks.

`nerdctl restart --debug-full ${containerid}` also outputs the same error: 

```
FATA[0000] 1 errors:
failed to start binary process with cmdArgs []: exec: no command
```

 `nerdctl inspect ${containerid}` can output normally ：

```
[
  {
    "Id": "xxx",
    "Created": "2025-xx",
    "Path": "xxx",
    "Args": [
      "xxx"
    ],
    "State": {
      "Status": "",
      "Running": false,
      "Paused": false,
      "Restarting": false,
      "Pid": 0,
      "ExitCode": 0,
[...]

PR #4166, opened 2025-04-25
[QOL]: enhance error information for failing logURI

apostasie:
We likely have incomplete logic in how we handle borked loguri labels (see: #4018 for eg).

The current erroring is not helpful and we can't really derive anything out of it.

This PR suggests we add the actual logURI to help.


AkihiroSuda (approved):
Thanks

Issue #1415, opened 2022-10-05
-a for stats command is not working when there is an exited container


azhouwd:
### Description

When using `stats` command with -a, no output is generated and the terminal stuck there forever.

```
~/Desktop ✓ ➭ ./_output/lima/bin/limactl shell default
[user@lima-default default]$
```

```
[user@lima-default default]$ nerdctl ps -a
CONTAINER ID    IMAGE                              COMMAND                CREATED         STATUS                     PORTS    NAMES
371ecb033c10    docker.io/library/alpine:latest    "sh -c nc -l -p 80"    34 hours ago    Up                                  alpine-371ec
a091b2be8eb0    docker.io/library/alpine:latest    "/bin/sh"              34 hours ago    Exited (0) 34 hours ago             alpine-a091b
```

`[user@lima-default default]$ nerdctl stats -a`

### Steps to reproduce the issue

1. create and shell into lima instance (Fedora).
2. run nerdctl stats -a.
3. The terminal stuck there and doesn't generate any output.

### Describe the results you received and expected

Received: no output is generated and terminal stuck forever.

Expected: Able to see the usage statistics for all the containers.

### What version of nerdctl are you using?

```
[user@lima-default default]$ nerdctl -v
nerdctl version 0.22.0
```

### Host information

```
~/Desktop ✓ ➭ uname -a
Darwin 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64
```

Sprlia:
v1.4.0  is still have this problem

apostasie:
Confirmed still a problem with nerdctl v2.

fahedouch:
do you face the same issue in main branch

from my side every thing is ok:

````
faheddorgaa@lima-default:/Users/faheddorgaa/go/src/github.com/nerdctl$ _output/nerdctl ps -a
CONTAINER ID    IMAGE                              COMMAND             CREATED          STATUS    PORTS    NAMES
1abd6ea9629e    docker.io/library/alpine:latest    "sleep infinity"    3 minutes ago    Up                 alpine-1abd6

CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT     MEM %     NET I/O         BLOCK I/O     PIDS
[...]

Issue #4043, opened 2025-03-27
Stats panic (`send on closed channel`)


apostasie:
### Description

Seen on CI. Stats panic sending to a closed channel.

### Describe the results you received and expected

```
=== RUN   TestStats/container_stats
=== PAUSE TestStats/container_stats
=== CONT  TestStats/container_stats
    container_stats_test.go:110: 
    container_stats_test.go:110: ======================== Pre-test cleanup ========================
    container_stats_test.go:110: 
    container_stats_test.go:110: ======================== Test setup ========================
    container_stats_test.go:110: 
    container_stats_test.go:110: ======================== Test Run ========================
    command.go:113: /usr/local/bin/nerdctl --namespace=nerdctl-test container stats --no-stream --no-trunc
    command.go:113: assertion failed: expect.ExitCode is not result.ExitCode: Expected exit code: 0
        
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test container stats --no-stream --no-trunc
        ExitCode: 2
        Error:    exit status 2
        Stdout:   CONTAINER ID                                                       NAME                                 CPU %     MEM USAGE / LIMIT     MEM %     NET I/O         BLOCK I/O     PIDS
        56c2e9ebefbb3836c2bf0576fc2b7dbdd35d886e04d2cf799673d29936d767ef   testexectty-3d80652a                 0.00%     352KiB / 15.62GiB     0.00%     7.12kB / 612B   0B / 0B       1
        ec83950c27fe969597be9c97f9f8c8bd71383de4474139adff9317b0f6e7b37d   teststats-memlimited-2f829aaf        0.00%     352KiB / 1GiB         0.03%     9.52kB / 612B   0B / 0B       1
        ed8d898251f830ac5aeb36e4807f97d84792d422c97b43cd6d69034083dec2f4   teststats-container-8514[604](https://github.com/containerd/nerdctl/actions/runs/14100070551/job/39494696009?pr=4041#step:7:605)0         0.00%     344KiB / 15.62GiB     0.00%     10kB / 612B     0B / 0B       1
        f9b366abcd25db264bceabf7e9949117f8a0d1379ba871f9c645d27ffe6a6bee   testsharednetworkwithnone-7c41c3a8   0.00%     3.621MiB / 15.62GiB   0.02%     0B / 0B         0B / 12.3kB   5
        
        Stderr:   panic: send on closed channel
        
        goroutine 27 [running]:
[...]

PR #4172, opened 2025-04-28
fix stats for exited container and introduce some refacto

fahedouch:
* Updated `statsutil.NewStats` to take both `containerID` and `containerName` as parameters, ensuring the container name is initialized in the stats entry before data collection. This simplifies the collection process by removing the need to fetch labels in each iteration.
=> Removed redundant code in the collect function, such as fetching container labels inside a loop, to make the logic more efficient.

* (fix https://github.com/containerd/nerdctl/issues/1415) Enhanced error handling in the `collect` function by adding a specific check for "no such file or directory" errors to avoid unnecessary error propagation (fix ticket) . The errors is not wrapped in the original [err](https://github.com/containernetworking/plugins/blob/v1.7.1/pkg/ns/ns_linux.go#L129).

### DIFF
--- a/pkg/cmd/container/stats.go
+++ b/pkg/cmd/container/stats.go
@@ -93,7 +93,7 @@ func Stats(ctx context.Context, client *containerd.Client, containerIDs []string
 	closeChan := make(chan error)
 
 	var err error
-	var w = options.Stdout
+	w := options.Stdout
 	var tmpl *template.Template
 	switch options.Format {
 	case "", "table":
@@ -128,7 +128,6 @@ func Stats(ctx context.Context, client *containerd.Client, containerIDs []string
 				return
 			}
 		}
-
 	}
 
 	// getContainerList get all existing containers (only used when calling `nerdctl stats` without arguments).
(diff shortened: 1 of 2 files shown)


pkg/cmd/container/stats.go
```diff
@@ -176,7 +177,11 @@ func Stats(ctx context.Context, client *containerd.Client, containerIDs []string
 					return
 				}
 			}
-			s := statsutil.NewStats(datacc.ID)
+			// Load the container to get its ID for retrieving the container name from Labels.
+			// if an error occurs, the ID alone is sufficient for the stats screen.
+			container, _ := client.LoadContainer(ctx, datacc.ID)
+			clabels, _ := container.Labels(ctx)
```
apostasie, line 183: So, we could have `clables == nil` and `containerutil.GetContainerName(clabels) == ""`  at this point.
Is this ok?

apostasie:
@fahedouch while we are in here, do you have a top of the head for #4043?

pkg/cmd/container/stats.go
fahedouch: > So, we could have clabels == nil and containerutil.GetContainerName(clabels) == "" at this point.
Is this ok?

yes it is ok

AkihiroSuda (approved):
Thanks

PR #2108, opened 2023-03-16
feat: add 'nerdctl container attach'

davidhsingyuchen:
## Summary

PR is unblocked by #2128. After this PR, for a container started in attached mode, a user can detach from it (enabled by #2128) and then re-attach to it using `nerdctl attach` (enabled by this PR).

However, since ["dual logging"](https://github.com/containerd/nerdctl/issues/1946#issuecomment-1443649307) is not implemented yet, containers spun up in detached mode, namely via `nerdctl run -d` or `nerdctl start` (without `-a`), cannot be `nerdctl attach`ed, which is mentioned in `nerdctl attach --help` and `command-reference.md`. To be specific, for a container started in detached mode, `cio.LogURI` instead of `cio.NewCreator` is used, which means the FIFOs that store the container's stdin/stdout/stderr and will be used by `cio.NewAttach` won't be there.

## Notes

- `ContainerAttachOptions.GOptions` is not used by `container.Attach` at the moment, but I still added it for consistency across commands.
- To keep this PR as simple as possible, only the bare minimum CLI parameters are supported in it. In other words, out of [all the parameters supported by `docker attach`](https://docs.docker.com/engine/reference/commandline/attach/#options), only `--detach-keys` is supported, and both `--no-stdin` and `sig-proxy` are not supported yet. We can add support for them in follow-up PRs.
- `cio.Terminal` is not really used by `cio.NewAttach` (i.e., whether the container is with a terminal is already decided when the container is created/started), so I omit the logic to decide if it should be passed to `cio.NewAttach`.
- Currently only one attach session is supported because [only one reader can be attached to a task](https://github.com/containerd/containerd/blob/e208c24256a0435e4a22c64f31dda19ef14724a5/container.go#L67-L68) under `containerd`'s current design. Supporting multiple attach sessions at nerdctl side implies quite a bit complexity, so maybe we can leave it to another PR, or even consider doing it at containerd side.

PR #4176, opened 2025-04-30
feat: add --no-stdin flag to container attach

subashkotha:
### Summary

This PR adds support for a new flag in the nerdctl container attach command:

```
--no-stdin: disables attaching stdin to the container
```
### Notes

We are not adding `--sig-proxy` in this PR due to a limitation in nerdctl: attach only works for containers started with `-it` ref - https://github.com/containerd/nerdctl/pull/2108
When `-it` is used, signals like Ctrl+C (SIGINT) are sent directly to the container process by the TTY, bypassing nerdctl.
As a result, `--sig-proxy=false` has no effect in such cases and cannot be used/tested reliably under current constraints.

### Testing
* Added an integration test to validate that input is not sent to the container when --no-stdin is used.
* Verified behavior across re-attached containers.

### DIFF
--- a/cmd/nerdctl/container/container_attach_linux_test.go
+++ b/cmd/nerdctl/container/container_attach_linux_test.go
@@ -211,3 +211,44 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
 
 	testCase.Run(t)
 }
+
+func TestAttachNoStdin(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "-it", "--detach-keys=ctrl-p,ctrl-q", "--name", data.Identifier(),
+			testutil.CommonImage, "sleep", "5")
+		cmd.WithPseudoTTY()
+		cmd.Feed(bytes.NewReader([]byte{16, 17})) // Ctrl-p, Ctrl-q to detach (https://en.wikipedia.org/wiki/C0_and_C1_control_codes)
+		cmd.Run(&test.Expected{
+			ExitCode: 0,
+			Output: func(stdout string, info string, t *testing.T) {
+				assert.Assert(t, strings.Contains(helpers.Capture("inspect", "--format", "{{.State.Running}}", data.Identifier()), "true"))
+			},
+		})
+	}
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		cmd := helpers.Command("attach", "--no-stdin", data.Identifier())
+		cmd.WithPseudoTTY()
+		cmd.Feed(strings.NewReader("should-not-appear\n"))
+		cmd.Feed(bytes.NewReader([]byte{16, 17}))
+		return cmd
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: 0, // Since it's a normal exit and not detach.
+			Output: func(stdout string, info string, t *testing.T) {
+				logs := helpers.Capture("logs", data.Identifier())
+				assert.Assert(t, !strings.Contains(logs, "should-not-appear"))
+			},
+		}
+	}
+
+	testCase.Run(t)
+}
--- a/cmd/nerdctl/container/container_attach.go
+++ b/cmd/nerdctl/container/container_attach.go
@@ -17,6 +17,8 @@
 package container
 
 import (
+	"io"
+
 	"github.com/spf13/cobra"
 
 	containerd "github.com/containerd/containerd/v2/client"
@@ -56,6 +58,7 @@ Caveats:
 		SilenceErrors:     true,
 	}
 	cmd.Flags().String("detach-keys", consoleutil.DefaultDetachKeys, "Override the default detach keys")
+	cmd.Flags().Bool("no-stdin", false, "Do not attach STDIN")
 	return cmd
 }
 
@@ -68,9 +71,18 @@ func attachOptions(cmd *cobra.Command) (types.ContainerAttachOptions, error) {
 	if err != nil {
 		return types.ContainerAttachOptions{}, err
 	}
+	noStdin, err := cmd.Flags().GetBool("no-stdin")
+	if err != nil {
+		return types.ContainerAttachOptions{}, err
+	}
+
+	var stdin io.Reader
+	if !noStdin {
+		stdin = cmd.InOrStdin()
+	}
 	return types.ContainerAttachOptions{
 		GOptions:   globalOptions,
-		Stdin:      cmd.InOrStdin(),
+		Stdin:      stdin,
 		Stdout:     cmd.OutOrStdout(),
 		Stderr:     cmd.ErrOrStderr(),
 		DetachKeys: detachKeys,
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -686,8 +686,9 @@ Usage: `nerdctl attach CONTAINER`
 Flags:
 
 - :whale: `--detach-keys`: Override the default detach keys
+- :whale: `--no-stdin`: Do not attach STDIN
 
-Unimplemented `docker attach` flags: `--no-stdin`, `--sig-proxy`
+Unimplemented `docker attach` flags: `--sig-proxy`
 
 ### :whale: nerdctl container prune
 
--- a/pkg/cmd/container/attach.go
+++ b/pkg/cmd/container/attach.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 
 	"golang.org/x/term"
 
@@ -114,9 +115,12 @@ func Attach(ctx context.Context, client *containerd.Client, req string, options 
 			}
 			io.Cancel()
 		}
-		in, err := consoleutil.NewDetachableStdin(con, options.DetachKeys, closer)
-		if err != nil {
-			return err
+		var in io.Reader
+		if options.Stdin != nil {
+			in, err = consoleutil.NewDetachableStdin(con, options.DetachKeys, closer)
+			if err != nil {
+				return err
+			}
 		}
 		opt = cio.WithStreams(in, con, nil)
 	} else {


cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,89 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "-it", "--detach-keys=ctrl-p,ctrl-q", "--name", data.Identifier(),
+			testutil.CommonImage, "sh", "-c", "sleep 1; echo 'after cat'; sleep 1")
+		cmd.WithPseudoTTY()
+		cmd.Feed(bytes.NewReader([]byte{16, 17}))
```
AkihiroSuda, line 226: Could you add a comment to explain these magic numbers

AkihiroSuda:
Please squash the commits

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,89 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+func TestAttachNoStdin(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "-it", "--detach-keys=ctrl-p,ctrl-q", "--name", data.Identifier(),
+			testutil.CommonImage, "sh", "-c", "sleep 1; echo 'after cat'; sleep 1")
```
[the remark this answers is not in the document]
subashkotha: We're using detach-keys here specifically to immediately detachment after attach. hence there is no need for passing in env variables.

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,89 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+			Output: func(stdout string, info string, t *testing.T) {
+				assert.Assert(t, strings.Contains(helpers.Capture("inspect", "--format", "json", data.Identifier()), "\"Running\":true"))
+			},
+		})
+	}
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		cmd := helpers.Command("attach", "--sig-proxy=false", data.Identifier())
+		cmd.WithPseudoTTY()
+		var stdin bytes.Buffer
```
apostasie, line 281: Maybe the more compact:
```
cmd.Feed(bytes.NewReader([]byte{3, 16, 17}))
```
?

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,89 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "-it", "--detach-keys=ctrl-p,ctrl-q", "--name", data.Identifier(),
+			testutil.CommonImage, "sh", "-c", `trap "exit 42" INT; while true; do sleep 1; done`)
+		cmd.WithPseudoTTY()
+		cmd.Feed(bytes.NewReader([]byte{16, 17}))
+		cmd.Run(&test.Expected{
+			ExitCode: 0,
+			Output: func(stdout string, info string, t *testing.T) {
+				assert.Assert(t, strings.Contains(helpers.Capture("inspect", "--format", "json", data.Identifier()), "\"Running\":true"))
```
apostasie, line 273: Your other way to check this `helpers.Capture("inspect", "--format", "{{.State.Running}}", data.Identifier())` seems more elegant?
Any reason to not use it here as well?

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,89 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+
+func TestAttachNoSigProxy(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "-it", "--detach-keys=ctrl-p,ctrl-q", "--name", data.Identifier(),
```
apostasie, line 266: Unfortunately, this is quite tricky to get right.

By the time you start writing to stdin, you have no guarantee that your bash trap is set and ready.

eg: this in its current state will be racy.

We do have a helper to deal with the same issue (but limited to sending signals: `RunSigProxyContainer`), that shows what needs to be done to ensure non-racy behavior.

So, in your case here, instead of writing directly, pass along a feeder method that will wait before writing - that should do the trick. Something in the line of:

```golang

cmd.WithFeeder(func() io.Reader {
    // Ensure container is started before moving forward
    EnsureContainerStarted(helpers, data.Identifier())

    // Test logs until we see a message from the bash script that the trap is set (NOTE: modify your trap code so that you output something when ready)
	for {
		out := helpers.Capture("logs", data.Identifier())
		if strings.Contains(out, "trap said it was ready") {
			break
		}
		time.Sleep(100 * time.Millisecond)
	}

// Now you can write:
  return bytes.NewReader([]byte{.....})
})

```

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,89 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+		stdin.Write([]byte{16, 17})
+		cmd.Feed(&stdin)
+		return cmd
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: 0,
+			Output: expect.All(
+				expect.Contains("after cat"),
```
apostasie, line 249: Confused. How is this guaranteed to have happened when we are here?
Isn't it possible that we reach here before the first sleep 1 gets completed?

cmd/nerdctl/container/container_attach_linux_test.go
Shubhranshu153: Seems this can get stuck in the for loop. Probably needs a try number or we just sleep for some time and then check

Shubhranshu153: Can we not use a refactor version of: RunSigProxyContainer
Seems like it will be more robust.

apostasie: Adding arbitrary limits or arbitrary sleeps is biting us hard on slower env (eg: el8) and we should really avoid them.

if this get stuck, the command will timeout and the test will fail.

we can still have a upper limit on the number of iterations though, but it should be big

apostasie: > Can we not use a refactor version of: RunSigProxyContainer Seems like it will be more robust.

Sure, if there are more than one case for this that should definitely be preferable

cmd/nerdctl/container/container_attach_linux_test.go
subashkotha: Ack, will fix it.

cmd/nerdctl/container/container_attach_linux_test.go
subashkotha: We can do the same here, but in other tests I see we are asserting in a single line 

```
assert.Assert(t, strings.Contains(helpers.Capture("inspect", "--format", "json", data.Identifier()), "\"Running\":true"), info)
```

To keep it consistent across tests, I'll instead modify line 293-294

apostasie: Oh sure. I was just referring to the `"--format", "{{.State.Running}}"` part that matches to true rather that matching the full json to `"\"Running\":true`.

Anyhow, nit.

cmd/nerdctl/container/container_attach_linux_test.go
subashkotha: Yep the initial sleep added a race. Will removed it and assert the echo output via logs to ensure it actually happened.

cmd/nerdctl/container/container_attach_linux_test.go
subashkotha: My bad, will add a comment :)

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,54 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+func TestAttachNoStdin(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "-it", "--detach-keys=ctrl-p,ctrl-q", "--name", data.Identifier(),
+			testutil.CommonImage, "sh", "-c", `echo ready; sleep 1`)
```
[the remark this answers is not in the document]
subashkotha: The test hangs with `sleep inf` because we're using `--no-stdin`, so there's no way to detach via input (Ctrl+P Ctrl+Q) after attach without using a timeout. Since the container won't exit on its own, the test never completes. I'll use `sleep 5` instead to avoid racy conditions from a shorter sleep.

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,54 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+			Output: func(stdout string, info string, t *testing.T) {
+				assert.Assert(t, strings.Contains(helpers.Capture("inspect", "--format", "{{.State.Running}}", data.Identifier()), "true"))
+			},
+		})
+	}
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		cmd := helpers.Command("attach", "--no-stdin", data.Identifier())
+		cmd.WithPseudoTTY()
+		cmd.WithTimeout(5 * time.Second)
```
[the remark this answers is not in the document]
subashkotha: Makes sense, I'll remove the timeout.

cmd/nerdctl/container/container_attach_linux_test.go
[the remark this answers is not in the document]
apostasie: I am not on a laptop right now so cannot really focus on this but an alternative would be to background the command and Signal() it maybe?
Anyhow this is fine, lets move this forward.

Shubhranshu153:
Flaky Test PR's:
https://github.com/containerd/nerdctl/pull/4196
https://github.com/containerd/nerdctl/pull/4198
https://github.com/containerd/nerdctl/pull/4199

subashkotha:
Hi @apostasie , @AkihiroSuda 

I have updated the PR addressing comments, can you please take a look.

apostasie:
Thanks a lot @subashkotha

Issues on the CI are AFAIK:
- rekor down
- gomodjail (tentative fix in #4198) 
- windows slow (tentative fix in #4199 )

I am bit concerned about rootless old getting to timeout at 45 mins+ (clearly getting stuck in container tests).

Suggestion would be that we first merge a raft of fixes for the CI (pending), and then you could rebase on top of main to get a cleaner run.

Does that work for you?

subashkotha:
Sure, @apostasie 
I'll rebase after fixes for CI are merged. Can you approve the PR if there are no concerns, thanks!

apostasie (approved):

apostasie:
@subashkotha thanks.
Code looks good to me, but I would like to see a clean CI run with no timeout before we merge.
Also note that I am not a maintainer: approval from me is just advisory, and you will need a maintainer approval for merge.

AkihiroSuda:
Could you try rebasing?

apostasie:
This does not feel right. EL8 does not get past the new test.

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,53 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+				assert.Assert(t, strings.Contains(helpers.Capture("inspect", "--format", "{{.State.Running}}", data.Identifier()), "true"))
+			},
+		})
+	}
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		cmd := helpers.Command("attach", "--no-stdin", data.Identifier())
+		cmd.WithPseudoTTY()
+		cmd.WithFeeder(func() io.Reader {
+			for {
```
apostasie, line 239: Does the loop serve a purpose here?

If I am reading this correctly, we attach to the container, and we will not return until the sleep 5 has lapsed.
Once it has elapsed, we are guaranteed to have "ready" in the log, right?

apostasie:
Thanks for keeping up with this @subashkotha 
Appreciated.

Docker Hub is now flipping a 429 on us (<- this is rate limiting from Docker Hub) - unfortunately there is nothing to do about that but wait for it to chill (@AkihiroSuda so tired of this - need to act)

Let's give it a minute then we will re-run.

cmd/nerdctl/container/container_attach_linux_test.go
subashkotha: Yes, I tried removing the loop, but the test failed because it couldn't reliably assert "ready" in the logs ([example failure](https://github.com/containerd/nerdctl/actions/runs/14890135501/job/41819915827?pr=4176)). The loop ensures we wait until the container is actually ready before feeding input — without it, the test becomes racy.

AkihiroSuda:
Could you rebase to see if passes the CI?

apostasie:
> Seems hanging up on almalinux-8 (cgroup v1)

Reading tea-leaves: we get stuck in the `Feed` loop waiting for the logs to show `ready` (which never happens for whatever reason <- this is the culprit).

While tigron does drop the main command after 3 minutes, we might still be waiting (too politely) on the ioGroup (which is dead looping):
https://github.com/containerd/nerdctl/blob/main/mod/tigron/internal/com/command.go#L327

Maybe I could add a reasonable timeout of the ioGroup Wait to deal with these conditions.

@subashkotha so, beside the problem in tigron ^ that prevents this from timeouting (I will deal with that), could you add a counter in your loop waiting for `ready` so that we stop iterating after - let's say - 500 iterations (or 50 iterations if you increase the sleep to 1 second, which I would recommend to reduce pressure on the runtime).

apostasie:
@fahedouch 
Here is the scenario:
```
nerdctl run -ti --name foo image "sh" "-c" `echo ready; sleep 5`
```

Then detach (ctrl+p ctrl+q).

Then we `nerdctl logs foo`, but in some cases we never see `ready` in the logs after an extended period of time.

Do you have an intuition on this? Could it be a bug in logging with detach?

cmd/nerdctl/container/container_attach_linux_test.go
```diff
@@ -211,3 +211,61 @@ func TestAttachForAutoRemovedContainer(t *testing.T) {
...
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", data.Identifier())
+	}
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		cmd := helpers.Command("run", "-it", "--detach-keys=ctrl-p,ctrl-q", "--name", data.Identifier(),
+			testutil.CommonImage, "sh", "-c", `echo ready; sleep 5`)
+		cmd.WithPseudoTTY()
+		time.Sleep(1 * time.Second)
+		cmd.Feed(bytes.NewReader([]byte{16, 17})) // Ctrl-p, Ctrl-q to detach (https://en.wikipedia.org/wiki/C0_and_C1_control_codes)
```
apostasie, line 227: @subashkotha let's test my hypothesis that there is a bug in logging.

So, instead of sending `cmd.Feed(bytes.NewReader([]byte{16, 17}))` after one second, instead do your `logs` / wait on `ready` BEFORE sending the escape sequence.

eg:
```
cmd.Feeder(func()io.Reader{

for {
				out := helpers.Capture("logs", data.Identifier())
				if strings.Contains(out, "ready") {
					break
				}
}

return bytes.NewReader([]byte{16, 17})

}
)
```

apostasie: Pseudo-code ^ just add back the sleep and make this proper - thanks a lot! :-)

apostasie:
Looks indeed like a log bug: #4233

@subashkotha suggested change above should allow you to work-around it.

cmd/nerdctl/container/container_attach_linux_test.go
subashkotha: Thanks, let me try this out.

Shubhranshu153:
so i did some digging i think overall we are approaching the stdin nil case wrongly.
no-stdin has the work of not attaching the stdin of the container, but once its done we terminate the stdin.
The target is to still keep the stdin. this will let you enter ctrl c to end the attach process. but also we need to be careful not to forward the proxy signal

AkihiroSuda (approved):
Thanks

PR #4177, opened 2025-04-30
[CI]: cleanup, breakout 4: separate jobs

apostasie:
On top of #4161, integrating feedback received over there.

This PR focuses on the previous workflows `build`, `project checks` and all linting tasks.

Besides cleanup / pure refactoring, this PR offers a few additional changes:
- regroup all jobs mentioned above under `workflow-lint` - instead of `build` under `test` and `project-checks` in its own workflow
- `build` is expanded
  - add canary
  - build on other platforms and architectures instead of just linux current arch


apostasie:
@AkihiroSuda rebasing and proof-reading.
Should be good to review in a minute.

.github/workflows/workflow-test.yml
```diff
@@ -61,16 +61,13 @@ jobs:
...
           - runner: "macos-15"
-            canary: false
           - runner: "windows-2025"
-            canary: false
           - runner: "ubuntu-24.04"
             canary: true
     with:
       runner: ${{ matrix.runner }}
-      canary: ${{ matrix.canary }}
+      canary: ${{ matrix.canary && true || false }}
```
apostasie, line 70: Aligning with the rest to make the matrix more readable by omitting "default" values.
Note that in github yaml world, passing `matrix.canary` is not the same thing as passing `matrix.canary && true || false`.

Makefile
```diff
@@ -102,7 +103,7 @@ help:
 binaries: $(CURDIR)/_output/$(BINARY)$(BIN_EXT)
 
 $(CURDIR)/_output/$(BINARY)$(BIN_EXT):
-	$(call title, $@)
+	$(call title, $@: $(GOOS)/$(GOARCH))
```
apostasie, line 106: Purely to enhance logging now that we build multiple architectures during testing.

.github/workflows/job-build.yml
```diff
@@ -0,0 +1,69 @@
+# This job just builds nerdctl for the golang versions we support (as a smoke test)
+name: build-for-go
```
AkihiroSuda, line 2: The name and the yaml file name should be synced

.github/workflows/job-build.yml
```diff
@@ -0,0 +1,69 @@
...
+      - if: ${{ env.GO_VERSION != '' }}
+        name: "Run: make binaries"
+        run: |
+          # We officially support these
+          GOOS=linux make binaries
+          GOOS=windows make binaries
+          GOOS=freebsd make binaries
+          GOOS=darwin make binaries
+          GOARCH=arm GOARM=6 make binaries
+          # These architectures are not released, but we still verify that we can at least compile
```
apostasie, line 66: To be discussed.
While we do not ship binaries for these architectures, nor officially support them, I think there is no downside in checking that we can at least build for them and ensure we know if we break them.

.github/workflows/workflow-test.yml
AkihiroSuda: > passing matrix.canary is not the same thing as passing matrix.canary && true || false.

`canary: false` is a string and a negative boolean simultaneously, and yet it is different from a `false` boolean literal ? 🤔

AkihiroSuda: Also please add an explanation in a code comment

.github/workflows/job-build.yml
apostasie: Ok.
Will do a full pass for that aspect then.

.github/workflows/job-build.yml
AkihiroSuda: SGTM

.github/workflows/job-build.yml
apostasie: Done.
I enforced it for job-X files, but not for workflow-Y files (in the case of workflows, the name is displayed).

.github/workflows/workflow-test.yml
apostasie: > > passing matrix.canary is not the same thing as passing matrix.canary && true || false.
> 
> `canary: false` is a string and a negative boolean simultaneously, and yet it is different from a `false` boolean literal ? 🤔

🤷‍♂️

I think if `canary` is not defined, it does coerce to `false` when tested, but not when passed as an boolean input?

apostasie: > Also please add an explanation in a code comment

Done.

.github/workflows/workflow-lint.yml
```diff
@@ -0,0 +1,80 @@
...
+          # Additionally lint for canary
+          - runner: ubuntu-24.04
+            goos: linux
+            canary: true
+    with:
+      timeout: 5
+      go-version: "1.24"
+      runner: ubuntu-24.04
+      # Note: in GitHub yaml world, if `matrix.canary` is undefined, and is passed to `inputs.canary`, the job
+      # will not run. However, if you test it, it will coerce to `false`, hence:
```
AkihiroSuda, line 41: 😄 
Maybe this should be reported as a bug? (in https://github.com/actions/toolkit/issues ?)

AkihiroSuda (approved):
Thanks

apostasie:
@AkihiroSuda hold on a minute - do not merge yet.

apostasie:
> @AkihiroSuda hold on a minute - do not merge yet.

False alarm, we good.

PR #4181, opened 2025-05-02
[Regression] [Tigron]: fix prepend args

apostasie:
Presumably the regression was introduced when I got rid of icmd.

This went unnoticed so far because we never make use of this in a way that would trigger a problem (eg: multiple successive calls to `PrependArgs`).
It would happen if we were using some of the advanced helpers (eg: config `HostsDir`, which we do not - yet).

I just noticed it while looking for something else and re-reading that part.

Clearly, cleanup of `test.Command` (post icmd) is overdue, but at least this patch should be merged now.

### DIFF
--- a/mod/tigron/test/command.go
+++ b/mod/tigron/test/command.go
@@ -160,7 +160,7 @@ func (gc *GenericCommand) WithTimeout(timeout time.Duration) {
 }
 
 func (gc *GenericCommand) PrependArgs(args ...string) {
-	gc.cmd.PrependArgs = args
+	gc.cmd.PrependArgs = append(gc.cmd.PrependArgs, args...)
 }
 
 func (gc *GenericCommand) Background() {


mod/tigron/test/command.go
```diff
@@ -160,7 +160,7 @@ func (gc *GenericCommand) WithTimeout(timeout time.Duration) {
 }
 
 func (gc *GenericCommand) PrependArgs(args ...string) {
-	gc.cmd.PrependArgs = args
+	gc.cmd.PrependArgs = append(gc.cmd.PrependArgs, args...)
```
AkihiroSuda, line 163: Shouldn't this be prepended, not appended?

apostasie: The list `gc.cmd.PrependArgs` will get added first, before `gc.cmd.Args` when we will build the final arguments list before starting the command:
https://github.com/containerd/nerdctl/blob/main/mod/tigron/internal/com/command.go#L385

This method here just adds more stuff to the internal list of `PrependArgs`.

`test/command` was designed when we were depending on icmd - since the rewrite (in `internal/command`) does provide some of these features by default now, `test/command` is now a confusing - almost empty - shell... need a full cleanup as mentioned, but I do not want to break the API too often, so, I will likely batch that for later.

Anyhow, your question clearly demonstrates cleanup is needed :(.

AkihiroSuda (approved):
Thanks

PR #4180, opened 2025-05-01
[FIX]: #4012 follow-ups (fixes and target test binaries)

apostasie:
This addresses a few issues from #4012 as a follow-up.

Besides a couple of mundane fixes, the bulk of it focuses on the ability to test binaries that are not named `docker` nor `nerdctl`.

This feature is especially interesting:
- for people building forks, or custom cli on top of nerdctl as a library, that may use a different binary name
- for gomodjail variants of course
- for busy contributors or bisectors who want to be able to quickly test a bunch of differently named nerdctl binaries

Here is a list of things I wish I could test:
- `~/docker.xyz`
- `nerdctl.gomodjail`
- `lepton`
- `/somerandompath/nerdctl.xyz`

The proposal here is to expand on the existing `--test.target` flag, that currently only allows `docker` and `nerdctl`.

The upside is that it strictly minimizes the amount of change required by the rest of the test tooling, is strictly backward compatible, and does not introduce new knobs.

Of course we still need a way to differentiate between docker, and "nerdishctl" binaries (cli that are expected to behave like nerdctl), as we do have specialized test behavior.

We do that here by checking `strings.HasPrefix(filepath.Base(binary), "docker")`.

So:

`go test ./cmd/nerdctl/... -test.target=~/docker.xyz` is treated as docker, while any other variant above is treated as a "nerdishctl".

NOTE that:
- there were a bunch of hardcoded assumptions in the tests assuming that the binary was named `nerdctl`, so, they got modified
- the `Target` type and a couple of (now useless) consts have been removed

There is still some level of duplication between testutil / testutil/nerdtest - this is not new, and is transitory while we move to the new testing tooling - can be cleaned-up later on to keep this small.

PR #4183, opened 2025-05-02
[GOMODJAIL]: suggesting to replace indirect module syndtr/gocapability

apostasie:
See README for detailed rationale.

TL;DR:
This indirect dependency has an unavoidable `init()` routine that will systematically `openat` and `fcntl` (reading `/proc/sys/kernel/cap_last_cap`), even / while it is not necessary.
`gomodjail` will (righfully) block this, resulting in a series of loud warnings - or we would have to `unconfine` this dependency.
Precisely because this project has been abandoned for multiple years and the maintainer seem to have vanished, marking it unconfined feels exactly like the kind of things we should _not_ do - as a likely candidate for a supply chain compromise.

This PR suggests we replace it.
The only part we do need is mundane (the mere list of linux capabilities), which is what we implement here from https://github.com/torvalds/linux/blob/master/include/uapi/linux/capability.h

Note that the state of this dependency has already raised concerns enough that most recent contributors have closed their PRs over there, and at least RH Kolyshkin did fork it: https://github.com/kolyshkin/capability

An alternative future / better solution would be for opencontainers/runtime-tools to address this on their side, since the dependency is getting pulled by them:

graph-main

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -152,3 +152,5 @@ require (
 )
 
 replace github.com/containerd/nerdctl/mod/tigron v0.0.0 => ./mod/tigron
+
+replace github.com/syndtr/gocapability => ./internal/replace/gocapability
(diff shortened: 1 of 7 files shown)


apostasie:
This was uncovered while working on #4180

apostasie:
Failures are docker hub 429, gomodjail concurrency issue, and one of the EL8 flakes.

AkihiroSuda:
> An alternative future / better solution would be for opencontainers/runtime-tools to address this on their side, since the dependency is getting pulled by them:

👍 

Also CDI shouldn't depend on runtime-tools as runtime-tools is also relatively inactive

apostasie:
@AkihiroSuda you did mean to close this, right? (just checking if you are saying "this problem **must** go upstream", or if you closed accidentally - either way is fine with me).

AkihiroSuda:
Let me close this. I don't think we should fork the repo just for eliminating the single `//gomodjail:unconfined` line.

Thanks anyway for heading this up.

apostasie:
Sure. No problem.

PR #4184, opened 2025-05-02
fix: enhance event monitoring by adding close event channel

fahedouch:
fix https://github.com/containerd/nerdctl/issues/4043


apostasie (approved):

AkihiroSuda (approved):
Thanks,  can we have a test?

apostasie:
@fahedouch related: #4132
This one has been plaguing us for some time.
It looks like sometimes we miss events.

Would be nice to have it looked into at some point... :)

fahedouch:
> Thanks, can we have a test?

The impacted func is a nested function, so it's complicated to test it unitarily unless I refactor the code and transform the inner function into a standalone function. Here, I don't see the point of doing that.

Issue #4016, opened 2025-03-18
rootless | runc failing to start container with `--sysctl net.core.somaxconn`


apostasie:
### Description

In rootless, runc fails to start a container that uses `--sysctl net.core.somaxconn`

This is possibly an effect of https://github.com/torvalds/linux/commit/464dc801c76aa0db88e16e8f5f47c6879858b9b2

This might be a runc issue - depending on how you look at this. If it is, there seem to be some related discussion here: https://github.com/opencontainers/runc/issues/3770

Maybe we should not hard error on this?

Not entirely sure why we do not catch that on the CI, as we do have TestContainerInspectHostConfig that is tripping this bug on my local machine.

### Steps to reproduce the issue

1. `nerdctl run -ti --sysctl net.core.somaxconn=1024 debian bash`

### Describe the results you received and expected

`FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: open /proc/sys/net/core/somaxconn: no such file or directory`

Expected: maybe complain about that, but do not hard fail?

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

Linux lima-on-debian 6.1.0-32-cloud-arm64 #1 SMP Debian 6.1.129-1 (2025-03-06) aarch64 GNU/Linux

AkihiroSuda:
Expected behavior.
Can be discussed in https://github.com/opencontainers/runtime-spec/issues/1270 to add the `allowSysctlErrors` (bool? []string?) flag

apostasie:
Or should we filter in nerdctl?

AkihiroSuda:
Sorry, intended to link https://github.com/opencontainers/runtime-spec/issues

Issue #4182, opened 2025-05-02
[CI]: gomodjail catch-all issues


apostasie:
### Description

The newly introduced gomodjail target likely has a few rough edges that we need to blunt.

This here is to discuss them in one place without overloading other PRs.

### Describe the results you received and expected

A first one seen on #4180 

```
    main_test.go:80: 
        
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl.gomodjail non-existent-command                                         |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🚫 command returned a non-zero exit code                                                         |
        +------------------------------------------------------------------------------------------------------------+
        |         | ⚠️ 1                                                                                             |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🟠 time=2025-05-02T01:11:18.448Z level=ERROR msg="exiting with an error" error="open /home/rootl |
        |         | ess/.cache/gomodjail/1f381f194c295770/go.mod: no such file or directory" exitCode=1              |
        +------------------------------------------------------------------------------------------------------------+
        | 🌱      | XDG_SESSION_TYPE=tty                                                                             |
        |         | HOME=/home/rootless                                                                              |
        |         | XDG_SESSION_CLASS=user                                                                           |
        |         | XDG_SESSION_ID=4                                                                                 |
        |         | XDG_RUNTIME_DIR=/run/user/1001                                                                   |
        |         | PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin |
        |         | :/bin:/usr/games:/usr/local/games:/snap/bin                                                      |
        |         | DOCKER_CONFIG=/tmp/TestUnknownCommandnon-existent-command3111744727/003                          |
        |         | NERDCTL_TOML=/tmp/TestUnknownCommandnon-existent-command3111744727/003/nerdctl.toml              |
        +------------------------------------------------------------------------------------------------------------+
        | ⏰      | <1s (limit: 3m0s)                                                                                |
        +------------------------------------------------------------------------------------------------------------+
        | 📁      | /tmp/TestUnknownCommandnon-existent-command3111744727/002         
```

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
[...]

PR #4186, opened 2025-05-02
[CI]: removing net.core.somaxconn from test

apostasie:
See https://github.com/containerd/nerdctl/issues/4016

Currently, the test is flagged as running for rootless+cgroupv2, but it does fail on:
- debian on lima
- EL9 on lima

In that context, I am not sure what we are testing with it here.
Suggesting we remove that line - but of course open to any better alternative here?

### DIFF
--- a/cmd/nerdctl/container/container_inspect_linux_test.go
+++ b/cmd/nerdctl/container/container_inspect_linux_test.go
@@ -263,7 +263,6 @@ func TestContainerInspectHostConfig(t *testing.T) {
 		"--read-only",
 		"--shm-size", "256m",
 		"--uts", "host",
-		"--sysctl", "net.core.somaxconn=1024",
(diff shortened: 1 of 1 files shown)


apostasie:
Random kube flake - #3556 - and #4182

cmd/nerdctl/container/container_inspect_linux_test.go
```diff
@@ -263,7 +263,6 @@ func TestContainerInspectHostConfig(t *testing.T) {
 		"--read-only",
 		"--shm-size", "256m",
 		"--uts", "host",
-		"--sysctl", "net.core.somaxconn=1024",
```
AkihiroSuda, line 266: Ideally this should be another sysctl and the inspect result should be validated

AkihiroSuda (approved):
Thanks

Issue #4185, opened 2025-05-02
[CI]: TestRunBlkioSettingCgroupV2 appears non functioning on EL9


apostasie:
### Description

PR #4154 tentatively introduces additional testing on EL9 (as a way to bisect failures and figure out which are EL8-specific or in-lima-specific).

EL9 is cgroupv2, however it does not look it is quite working.

It could be the testing environment, or we have a bug.

Is there a cgroup jedi-master around who could advise on the following error?

### Steps to reproduce the issue

TestRunBlkioSettingCgroupV2

### Describe the results you received and expected

```

        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl run -d --name testrunblkiosettingcgroupv2-blkio-weight-device-888e71b0 |
        |         |  --blkio-weight-device /dev/dummy-zero:100 ghcr.io/stargz-containers/alpine:3.13-org sleep infin |
        |         | ity                                                                                              |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🚫 command returned a non-zero exit code                                                         |
        +------------------------------------------------------------------------------------------------------------+
        |         | ⚠️ 1                                                                                             |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🟠 time="2025-05-02T20:17:50Z" level=fatal msg="failed to create shim task: OCI runtime create f |
        |         | ailed: runc create failed: unable to start container process: error during container init: error |
        |         |  setting cgroup config for procHooks process: setting device weight \"1:5 100\": write /sys/fs/c |
        |         | group/system.slice/nerdctl-84d7508c134fd3489e413241ac4c09bb9110d7f5273177152ae916e48cf77049.scop |
        |         | e/io.bfq.weight: operation not supported"                                                        |
        +------------------------------------------------------------------------------------------------------------+
        | 🌱      | HOME=/root                                                                                       |
        |         | PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin |
        |         | :/bin                                                                                            |
        |         | DOCKER_CONFIG=/tmp/TestRunBlkioSettingCgroupV2blkio-weight-device1124629430/003                  |
        |         | NERDCTL_TOML=/tmp/TestRunBlkioSettingCgroupV2blkio-weight-device1124629430/003/nerdctl.toml      |
[...]

PR #4187, opened 2025-05-02
[CI]: disable blkio test for older kernels

apostasie:
See #4185 for what is known.

Also note this PR improves a bit `RequireKernelVersion` which is not working currently for EL kernels.


AkihiroSuda (approved):
Thanks

Issue #4143, opened 2025-04-20
[CI]: OCI runtime start failed: cannot start a container that has stopped: unknown


apostasie:
### Description

This is seen in the context of compose (again on #3303 which is meant to surface our hidden issues).
Specifically in the case of the attached trace, for `TestIPFSCompNoBuild`.

The message is coming from containerd runc shim, probably from runc itself.
It is unclear to me if that would be a runc/containerd bug, or if nerdctl should do something to catch the underlying problem earlier.

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

```
=== Failed
=== FAIL: cmd/nerdctl/ipfs TestIPFSCompNoBuild/with-stargz-no-byAddr (100.26s)
    requirements.go:238: 
        
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl info --format {{ json . }}                                             |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🟢 {"ID":"424471f2-16a8-4116-b393-0d1844a51eac","Driver":"overlayfs","Plugins":{"Log":["fluentd" |
        |         | ,"journald","json-file","none","syslog"],"Storage":["native","overlayfs","stargz","soci"]},"Memo |
        |         | ryLimit":true,"SwapLimit":true,"CpuCfsPeriod":true,"CpuCfsQuota":true,"CPUShares":true,"CPUSet": |
        |         | true,"CPURealtime":false,"PidsLimit":true,"IPv4Forwarding":true,"BridgeNfIptables":true,"BridgeN |
        |         | fIp6tables":true,"OomKillDisable":false,"SystemTime":"2025-04-20T18:52:35.058013148Z","LoggingDr |
        |         | iver":"json-file","CgroupDriver":"systemd","CgroupVersion":"2","KernelVersion":"6.8.0-1026-azure |
        |         | ","OperatingSystem":"Ubuntu 22.04.5 LTS","OSType":"linux","Architecture":"x86_64","NCPU":4,"MemT |
        |         | otal":16766754816,"Name":"25f5fde766b2","ServerVersion":"v1.6.38","SecurityOptions":["name=appar |
        |         | mor","name=seccomp,profile=builtin","name=cgroupns"],"Warnings":null}                            |
        +------------------------------------------------------------------------------------------------------------+
        | 🌱      | HOME=/root                                                                                       |
[...]

PR #4189, opened 2025-05-03
[CI]: cleanup, breakout 5: separate jobs

apostasie:
This continues the work started on cleaning workflows.

The desired end-result is still #4154 (head there for full explanation and additional details.) - which while currently green (save gomodjail) is also still big, so, this here is another breakout to reduce the size per-PR to ease review.

Note that introduction of EL9 is temporary, though we can decide to keep it in the long run if we want (see context in file).


.github/workflows/job-test-in-host.yml
```diff
@@ -0,0 +1,179 @@
...
+      ${{ contains(inputs.runner, '22.04') && '(old ubuntu)' || '' }}
+      ${{ inputs.canary && ' (canary)' || '' }}
+    timeout-minutes: ${{ inputs.timeout }}
+    runs-on: "${{ inputs.runner }}"
+    defaults:
+      run:
+        shell: bash
+
+    env:
+      SHOULD_RUN: "yes"
```
apostasie, line 62: I wish there was a way to interrupt a job programmatically.
This variable is used to decide if further steps should run or not.

apostasie:
Rebased and proof-read.

apostasie:
Failures are #3556 and #4143

AkihiroSuda (approved):
Thanks

PR #4194, opened 2025-05-05
[EXPERIMENTAL] proofing

apostasie:
Ignore for now.


pkg/testutil/images.env
```diff
@@ -0,0 +1,28 @@
...
+KUBO=ghcr.io/stargz-containers/ipfs/kubo:v0.16.0-org
+MARIADB=ghcr.io/stargz-containers/mariadb:10.5-org
+NANOSERVER=mcr.microsoft.com/windows/nanoserver:ltsc2022
+NGINX=ghcr.io/stargz-containers/nginx:1.19-alpine-org
+# FIXME: update to v3 soon. Right now, this is breaking tests.
+# REGISTRY=ghcr.io/distribution/distribution:3.0.0@sha256:4ba3adf47f5c866e9a29288c758c5328ef03396cb8f5f6454463655fa8bc83e2
+REGISTRY=ghcr.io/stargz-containers/registry:2-org
+STARGZ=ghcr.io/containerd/stargz-snapshotter:0.15.1-kind
+WORDPRESS=ghcr.io/stargz-containers/wordpress:5.7-org
+FEDORAESGZ=ghcr.io/stargz-containers/fedora:30-esgz
```
AkihiroSuda: Maybe
```suggestion
FEDORA_ESGZ=ghcr.io/stargz-containers/fedora:30-esgz
```

apostasie: Sure. Let me fix it.

pkg/testutil/images.env
AkihiroSuda, line 1: Can we use YAML/TOML so that we can add more metadata (OS, arch, etc.) in future?

apostasie: Yeah, that makes sense.

I will need a minute to get this done so feel free to either move the milestone for this - or merge this as-is and I'll update in a follow-up PR.

Whatever is preferable.

apostasie:
YAML variant incoming.
Reverting to draft until the CI is happy.

pkg/testutil/images.env
apostasie: YAML variant done.

pkg/testutil/images_linux.go
```diff
@@ -0,0 +1,54 @@
...
+var testImages map[string]TestImage
+
+func getImage(key string) string {
+	var im TestImage
+	var ok bool
+
+	err := yaml.Unmarshal([]byte(rawImagesList), &testImages)
+	if err != nil {
+		fmt.Printf("Error unmarshaling test images YAML file: %v\n", err)
+		panic("testing is broken")
```
fahedouch: Using `panic` is not a good practice for production code and should be replaced with proper error handling.

pkg/testutil/images_linux.go
```diff
@@ -0,0 +1,54 @@
...
+	Variants []string `yaml:"variants,omitempty"`
+}
+
+var testImages map[string]TestImage
+
+func getImage(key string) string {
+	var im TestImage
+	var ok bool
+
+	err := yaml.Unmarshal([]byte(rawImagesList), &testImages)
```
fahedouch: Cache parsed YAML data to avoid repeated parsing:

```suggestion
    var testImagesOnce sync.Once
    testImagesOnce.Do(func() {
        testImages = make(map[string]TestImage)
        testImagesErr = yaml.Unmarshal([]byte(rawImagesList), &testImages)
    })
```

pkg/testutil/images_linux.go
```diff
@@ -0,0 +1,54 @@
...
+type TestImage struct {
+	Ref      string   `yaml:"ref"`
+	Tag      string   `yaml:"tag,omitempty"`
+	Digest   string   `yaml:"digest,omitempty"`
+	Variants []string `yaml:"variants,omitempty"`
+}
+
+var testImages map[string]TestImage
+
+func getImage(key string) string {
```
fahedouch, line 41: can we have a test for this func `getImage` :) !

pkg/testutil/images.yaml
```diff
@@ -0,0 +1,77 @@
+# Current schema (defined in images_linux.go) allows for ref, tag, (index) digest and platform variants.
+# Right now, digest and variants are not used for anything, but they should / could be in the future.
+# Also note that changing the schema should be easy and straight-forward for now, so,
+# this might evolve in the near future.
```
fahedouch, line 4: may be document how to add new images or update existing ones.

pkg/testutil/images_linux.go
apostasie: This is not production code - this is test tooling

pkg/testutil/images_linux.go
apostasie: I think testing test tools is a bit much :) (well, core testing like tigron sure, it should be tested - but this here is just a yaml file getting parsed).
Also this is going to evolve soon with upcoming PRs so let see when this is getting finalized?

pkg/testutil/images.yaml
apostasie: Will do in upcoming PRs once we stabilize the schema?

pkg/testutil/images_linux.go
apostasie: Yep, will do.

apostasie: Done. Thanks!

pkg/testutil/images_linux.go
fahedouch: I mean unit test but ok not mandatory  regarding the impact

pkg/testutil/images_linux.go
fahedouch: thanks

fahedouch (approved):
Thanks

AkihiroSuda (approved):
Thanks

PR #4195, opened 2025-05-05
[GOMODJAIL]: unconfine fsnotify and cdi

apostasie:
Looks like we might need these unconfined?

For reference, failures for these are documented in:
- https://github.com/containerd/nerdctl/issues/4182#issuecomment-2848008980 for fsnotify
- https://github.com/containerd/nerdctl/issues/4182#issuecomment-2848727417 for CDI


apostasie:
Note: I am not sure why these failures are intermittent - aka why are these code paths being hit only occasionally

AkihiroSuda (approved):
Thanks

W.r.t. the flakiness gomodjail may have a bug related to goroutines?

PR #4198, opened 2025-05-06
[CI]: workaround gomodjail concurrency issue

apostasie:
This adds a preflight call to nerdctl.gomodjail that will ensure extraction is done before we test, (hopefully) avoiding concurrency issues.

This is a workaround for https://github.com/AkihiroSuda/gomodjail/issues/50 meant to reduce the disruptive effect on nerdctl CI, and must be removed once fixed upstream.

### DIFF
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -26,6 +26,13 @@ if [[ "$(id -u)" = "0" ]]; then
   fi
 fi
 
+# This is a temporary workaround for https://github.com/AkihiroSuda/gomodjail/issues/50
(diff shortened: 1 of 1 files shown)


apostasie:
Failures are:
- #3556 
- windows is (now) getting over 3 seconds for rm tests (in `TestRunRmTime`) - this is likely a consequence of upgrading from containerd 1.7 to 2.0 - we probably just want to push the timeout from 3 seconds to 10 seconds (maybe only for windows)

AkihiroSuda (approved):
Thanks

PR #4199, opened 2025-05-06
[CI]: small fixes

apostasie:
This adds a few minor tweaks to the CI:
- commit 1 adds darwin as a target for some makefile tasks
- commit 2 makes the Makefile more resilient in case `git` is missing
- commit 3 is a minor simplification of the `in-container` CI job
- commit 4 increases the timeout for Windows on TestRunRmTime

Though unrelated to one another, these are small and minor.
Still LMK if you prefer this PR to be broken down in individual chunks.


Dockerfile.d/test-integration-rootless.sh
```diff
@@ -26,7 +26,7 @@ if [[ "$(id -u)" = "0" ]]; then
 	fi
 
 	: "${WORKAROUND_ISSUE_622:=}"
-	if [[ "$WORKAROUND_ISSUE_622" = "1" ]]; then
+	if [[ "$WORKAROUND_ISSUE_622" != "" ]]; then
```
AkihiroSuda, line 29: This breaks `WORKAROUND_ISSUE_622=0` but I guess we do not need to keep compatibility for this workaround

Makefile
```diff
@@ -39,9 +39,9 @@ DOCDIR  ?= $(DATADIR)/doc
 
 BINARY ?= "nerdctl"
 MAKEFILE_DIR := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
-VERSION ?= $(shell git -C $(MAKEFILE_DIR) describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+VERSION ?= $(shell git -C $(MAKEFILE_DIR) describe --match 'v[0-9]*' --dirty='.m' --always --tags 2>/dev/null || echo no_git_information)
```
AkihiroSuda, line 42: Bikeshedding: an empty string should suffice for no_git_information

AkihiroSuda (approved):
Thanks

Issue #4147, opened 2025-04-20
[CI]: TestLogsWithoutNewlineOrEOF flaky


apostasie:
### Description

This is yet another log related test that fails randomly.

### Describe the results you received and expected

```
=== RUN   TestLogsWithoutNewlineOrEOF
=== PAUSE TestLogsWithoutNewlineOrEOF
=== CONT  TestLogsWithoutNewlineOrEOF
    container_logs_test.go:194: assertion failed: error is not nil: expected stdout to contain "Hello World!"
--- FAIL: TestLogsWithoutNewlineOrEOF (3.38s)
FAIL cmd/nerdctl/container.TestLogsWithoutNewli
```

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
@AkihiroSuda can you reopen this? (cannot do it myself apparently).

This just got spotted in https://github.com/containerd/nerdctl/actions/runs/15710306096/job/44266315074?pr=4290#step:10:2451

PR #4200, opened 2025-05-06
[CI]: rewrite TestLogsWithoutNewlineOrEOF

apostasie:
This rewrites TestLogsWithoutNewlineOrEOF - ideally fixing #4147

Note that this still relies on an arbitrary timeout, which (although increased) is by its very nature likely to flake.

UPDATE: test is now disabled for nerdctl until we fix nerdctl (see #4201 )


AkihiroSuda (approved):
Thanks

PR #4215, opened 2025-05-08
testutil: bump up GolangImage to 1.23

AkihiroSuda:
Fix #4214

PR #4217, opened 2025-05-08
update containerd version to 2.1.0

Shubhranshu153:
Upgrading containerd version to 2.1.0


pkg/testutil/testutil_darwin.go
```diff
@@ -31,5 +31,5 @@ var (
 	BusyboxImage     = "ghcr.io/containerd/busybox:1.36"
 	AlpineImage      = mirrorOf("alpine:3.13")
 	NginxAlpineImage = mirrorOf("nginx:1.19-alpine")
-	GolangImage      = mirrorOf("golang:1.18")
+	GolangImage      = mirrorOf("golang:1.23")
```
apostasie, line 34: Heads-up @Shubhranshu153 .
See https://github.com/containerd/nerdctl/blob/main/docs/multi-platform.md#macos--lima 
Tread carefully with this.
Unfortunately, the mirrored images we are using are not multiplatform.
Since this specifically is for Darwin (which is not using it, actually), we are fine, but the same change in _linux would break hell loose on arm.

Shubhranshu153: thanks. But in that case the test would fail using that image right or do we expect some unrecoverable issue?

apostasie: Any test involving building with that image would hard fail.
@AkihiroSuda go bitten by this: #4215

(though again that would be if you change it for linux as well, which you are not doing here)

apostasie:
Linking failures:
- gomodjail failure is the obnoxious #3556 
- build fail is ubuntu servers being inconsistent (no ticket, transient)
- docker failure is (recently showing up a lot more) TestRunSigProxy failure (no ticket yet)

apostasie:
Docker failure is #4219

pkg/testutil/testutil_darwin.go
fahedouch: this  is already set for all platform here https://github.com/containerd/nerdctl/pull/4215

AkihiroSuda: Could you rebase and revert this line, then LGTM

Shubhranshu153: reverted this change and rebased

AkihiroSuda (approved):
Thanks

PR #4222, opened 2025-05-08
[CI BUST FIX]: fix docker target (disable docker testing for SigProxyFalse)

### LINKED ISSUE #4219: [CI]: docker now fails (a lot) TestLogsWithStartContainer
### Description

```
docker rm -f foo; docker run -ti --sig-proxy=false --name foo ghcr.io/stargz-containers/alpine:3.13-org sh -c -- "while true; do echo lol; sleep 1; done"
```

```
kill -s INT DOCKER_PID
```

Will work fine in docker 26.

No longer works in Docker 28.

Maybe this is related to [#5302 ](https://github.com/docker/cli/pull/5302)

apostasie:
See #4219 for details.

Note that it is not 100% clear to me if docker fixed a legitimate issue and our test is wrong, or if this is unintended on their part.

One way or the other, disabling the test for now for docker so that we uncork the CI.

(fix #4219 )

### DIFF
--- a/cmd/nerdctl/container/container_run_linux_test.go
+++ b/cmd/nerdctl/container/container_run_linux_test.go
@@ -423,6 +423,10 @@ func TestRunSigProxy(t *testing.T) {
 		{
 			Description: "SigProxyFalse",
 
+			// Docker behavior changed sometimes with Docker 27
+			// See https://github.com/containerd/nerdctl/issues/4219 for details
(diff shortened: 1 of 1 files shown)


apostasie:
Rootful failure is #3556

apostasie:
Tagging @Shubhranshu153 so that we do not duplicate effort on this.

AkihiroSuda (approved):
Thanks

Issue #4201, opened 2025-05-06
[LOGS]: nerdctl logs -f adds extraneous line feed


apostasie:
### Description

`logs -f` adds a trailing 0a compared with docker.

### Steps to reproduce the issue

```
docker rm -f foo >/dev/null
docker run --name foo debian printf 'bar\nfoo' >/dev/null
docker logs -f foo | hexdump
```

vs.
```
nerdctl rm -f foo >/dev/null
nerdctl run --name foo debian printf 'bar\nfoo' >/dev/null
nerdctl logs -f foo | hexdump
```

### Describe the results you received and expected

```
0000000 6162 0a72 6f66 006f
```

vs

```
0000000 6162 0a72 6f66 0a6f
```

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Uncovered while looking into #4200

PR #4226, opened 2025-05-09
fix: remove trailing newline from log entries on last decode iteration

fahedouch:
fix https://github.com/containerd/nerdctl/issues/4201


AkihiroSuda (approved):
Thanks

PR #4229, opened 2025-05-09
[CI]: pin docker to a specific version

### LINKED ISSUE #4223: [CI]: should we pin docker on the CI to a specific version?

As our testing becomes sharper, accounting for more complicated / corner cases, it seems like we are more likely to hit issues with updated docker changing behavior.

Case in point: #4219 

While this is somewhat nice if our testing does also catch issues for Docker that we can forward to them, this is still a problem for us as it is causing head scratching instability (as we are currently testing a moving target).

Suggesting we pin docker installation on the CI to a specific version.

We can (should) still update it regularly, and when we do we can focus on what has changed and possibly adapt nerdctl or our testing consciously, instead of being confronted with sudden issues seemingly out of nowhere...

Thoughts?

### Additional context

_No response_

apostasie:
Fix #4223

TL;DR we should not be testing a moving target.
This forces downgrade of docker to a specific version.


.github/workflows/job-test-in-host.yml
```diff
@@ -105,7 +108,18 @@ jobs:
             echo "::group:: configure cdi for docker"
             sudo mkdir -p /etc/docker
             sudo jq '.features.cdi = true' /etc/docker/daemon.json | sudo tee /etc/docker/daemon.json.tmp && sudo mv /etc/docker/daemon.json.tmp /etc/docker/daemon.json
-            sudo systemctl restart docker
+            echo "::endgroup::"
+            echo "::group:: downgrade docker to the specific version we want to test (${{ inputs.docker-version }})"
+            sudo apt-get update -qq
+            sudo apt-get install -qq ca-certificates curl
+            sudo install -m 0755 -d /etc/apt/keyrings
+            sudo cp ./hack/provisioning/docker/gpg /etc/apt/keyrings/docker.asc
```
AkihiroSuda, line 116:
```suggestion
            sudo cp ./hack/provisioning/gpg/docker /etc/apt/keyrings/docker.asc
```

So as to collect all the gpg keys in the same directory

AkihiroSuda (approved):
Thanks

Issue #4163, opened 2025-04-25
[CI]: enforce hash verification for all retrieved artifacts.


apostasie:
### What is the problem you're trying to solve

All network retrieved resources should be verified against a sha (or otherwise pinned to a digest).

### Describe the solution you'd like

This applies to:
- hack/provisioning helpers, which currently retrieves a variety of binaries
- vagrant provisioning in .github workflow, which does not verify the key it is installing
- all test images, that are not not using digest pinning

PR #4230, opened 2025-05-09
[CI]: carry a copy of vagrant gpg key

apostasie:
Part of #4163


.github/workflows/job-test-in-vagrant.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           # from https://github.com/containerd/containerd/blob/v2.0.2/.github/workflows/ci.yml#L583-L596
           # which is based on https://github.com/opencontainers/runc/blob/v1.1.8/.cirrus.yml#L41-L49
           # FIXME: https://github.com/containerd/nerdctl/issues/4163
-          curl -fsSL --proto '=https' --tlsv1.2 https://apt.releases.hashicorp.com/gpg | sudo gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
+          cat ./hack/provisioning/vagrant/gpg | sudo gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
```
AkihiroSuda, line 38:
```suggestion
          cat ./hack/provisioning/gpg/hashicorp | sudo gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
```

Because we may add more `gpg/*`, and the hashicorp key is not specific to vagrant

apostasie: Done.

AkihiroSuda (approved):
Thanks

PR #4232, opened 2025-05-09
[CI]: fix flaky diff test

apostasie:
On slow environments (EL8), starting with -d will be racy as there is no guarantee the script has run by the time we test.

Seen in #4176


AkihiroSuda (approved):
Thanks

Issue #4231, opened 2025-05-09
[Tigron]: looping inside pipes reading or writing will deadlock


apostasie:
### Description

On a command timeout, we still `Wait` for the ioGroup.
This works fine in normal circumstances, but if a feeder to stdin does loop forever, we just wait there.

Possible solutions:
- separate stdin / feeders in their own group and force drop it on `wrap`
- add a timeout to the ioGroup wait

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

Timeout properly, even with a faulty feeder to would get stuck.

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
The ioGroup uses a context with timeout though. That should not happen...

apostasie:
Meh.

https://github.com/golang/go/issues/26099

So, errgroup with a timeout-ing context does not handle it on its own and will just wait (possibly forever)... You are supposed to terminate the group yourself when the timeout runs out.

@AkihiroSuda did you know that (I did not)? ^

Are there places in our codebase where we pass a context with timeout to an errgroup assuming the errgroup will self terminate once the timeout runs-out?

apostasie:
Closing this as wontfix.

Unless I am mistaken, there is nothing to be done if a func in an errgroup gets in a deadloop.
LMK if there is a solution to forcefully close a go routine.

PR #4234, opened 2025-05-09
[DOC]: clarify corner case on Tigron WithFeeder

apostasie:
See #4231 for context.

I did erroneously believed that `errgroup` would cancel on their own when using a context WithTimeout (and then some misunderstanding on my part of go routines).

The bottom-line is: if a feeder function is passed that does continuously loop, we will deadlock and it does not seem there is anything to do.

This PR just clarifies that fact.

Maybe in the future tigron could offer a `WaitFor` parameter to feeder functions, that would automatically run in a loop, properly accounting for context timeout.


AkihiroSuda (approved):
Thanks

PR #4238, opened 2025-05-10
Add loong64 as target for job-build + completion

apostasie:
This is a small follow-up to the recent merge of loong64, actually building the arch on the CI as part of job-build and adding the arch for completion.

### DIFF
--- a/.github/workflows/job-build.yml
+++ b/.github/workflows/job-build.yml
@@ -88,9 +88,10 @@ jobs:
           build linux arm64
           build windows
           build freebsd
+          # These architectures are not released, but we still verify that we can at least compile
           build darwin
           build linux arm 6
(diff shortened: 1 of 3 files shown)


.github/workflows/job-build.yml
```diff
@@ -91,6 +91,7 @@ jobs:
           build darwin
           build linux arm 6
           # These architectures are not released, but we still verify that we can at least compile
```
AkihiroSuda, line 93: This comment should be moved to darwin and arm v6

AkihiroSuda (approved):
Thanks

.github/workflows/job-build.yml
apostasie: Done.

AkihiroSuda (approved):
Thanks

PR #4245, opened 2025-05-14
[REFACTOR]: consolidate filesystem ops into internal pkg

apostasie:
This suggests that we consolidate low-level filesystem related operations into one package (currently pepperred over under `lockutil` and `store`).

This includes path validation, atomic writes, locks.

Also suggesting that said package should be made internal. This is _not_ nerdctl API, but merely low-level primitives, and generally speaking the nerdctl API is just too big, exposing too much irrelevant stuff.

This refactor is in preparation for future PRs that will focus on:
- enhancing locks to be more flexible and allow locking of files AND dirs, and also allow for RO locks
- enhancements to atomic writefile to support in-place writing (current rename approach does change inode), with a rollback / disaster recovery mechanism
- consistently use atomic writes to ensure better resilience and data consistency
(they will be broken out of #4240 as well in small chuncks)

This PR here sticks with purely refactoring.

### DIFF
--- a/cmd/nerdctl/main.go
+++ b/cmd/nerdctl/main.go
@@ -248,12 +248,12 @@ Config file ($NERDCTL_TOML): %s
 		}
 
 		// Since we store containers' stateful information on the filesystem per namespace, we need namespaces to be
-		// valid, safe path segments. This is enforced by store.ValidatePathComponent.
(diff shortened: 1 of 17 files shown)


AkihiroSuda (approved):
Thanks

Issue #4253, opened 2025-05-15
[Network]: Failed container creation will f-up iptables rules


apostasie:
### Description

Start a container:
```
dmp@lima-main-2025-05-15:/Users/dmp/Projects/go/nerd/nerdctl$ nerdctl run -d --name nerdctl-diagnostics2 -p 9044:80 nginx
d1364e3fb6f5bbc1de316656c9d7df483959e0c3e7202008f212ebe6c5bcd8d2
```

Check iptables:
```
dmp@lima-main-2025-05-15:/Users/dmp/Projects/go/nerd/nerdctl$ while read -r line; do sudo nsenter --net=$line iptables-save; done < <(lsns -n -u -t net -o PATH) | grep " 9044 "
-A CNI-DN-4095fe68ef51b84d4d4bd -s 10.4.0.0/24 -p tcp -m tcp --dport 9044 -j CNI-HOSTPORT-SETMARK
-A CNI-DN-4095fe68ef51b84d4d4bd -s 127.0.0.1/32 -p tcp -m tcp --dport 9044 -j CNI-HOSTPORT-SETMARK
-A CNI-DN-4095fe68ef51b84d4d4bd -p tcp -m tcp --dport 9044 -j DNAT --to-destination 10.4.0.83:80
-A CNI-HOSTPORT-DNAT -p tcp -m comment --comment "dnat name: \"bridge\" id: \"default-d1364e3fb6f5bbc1de316656c9d7df483959e0c3e7202008f212ebe6c5bcd8d2\"" -m multiport --dports 9044 -j CNI-DN-4095fe68ef51b84d4d4bd
```

So far, all fine.

Start a new container, different name, same port binding:

```
dmp@lima-main-2025-05-15:/Users/dmp/Projects/go/nerd/nerdctl$ nerdctl run -d --name nerdctl-diagnostics3 -p 9044:80 nginx
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running createRuntime hook #0: exit status 1, stdout: , stderr: time="2025-05-15T15:11:33-07:00" level=fatal msg="failed to expose ports in rootless mode: conflict with ID 16"
```

Not happy about port conflict - expected.

Now check iptables again:

```
[...]

Issue #4246, opened 2025-05-15
Unable to access port in rootful mode.


Jezza:
### Description

I'm unable to access any containers.
A simple example would be nginx.
Bind it to 8080, and curl is unable to reach it.

![Image](https://github.com/user-attachments/assets/667768bb-7b94-4857-b12c-d27e4d280c0d)

### Steps to reproduce the issue

1. Run `sudo nerdctl run -d -p 8080:80 --name=nginx --restart=always nginx` 
2. Run `curl localhost:8080`

### Describe the results you received and expected

I would expect the host to be able to reach the container.
Instead, it seems like it times out.

### What version of nerdctl are you using?

nerdctl version 2.0.3

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v2.0.2
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Log:     fluentd journald json-file none syslog
  Storage: btrfs native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version:   6.5.0-0.deb12.1-amd64
 Operating System: Debian GNU/Linux 12 (bookworm)
 OSType:           linux
 Architecture:     x86_64
 CPUs:             20
 Total Memory:     30.87GiB
 Name:             peeriot-00X
 ID:               0f398317-967c-4237-9898-5be7c24e7429

WARNING: bridge-nf-call-iptables is disabled
WARNING: bridge-nf-call-ip6tables is disabled

fahedouch:
while waiting for the fix on the CNI side to be merged, I     am working on PR that cleanup container iptable rule using post ici Hook.

apostasie:
@fahedouch check #4254

Jezza:
Sorry for the delay.
Lemme post the outputs, and then catch up with the messages.

```
> cat /etc/resolv.conf

# Generated by NetworkManager
search fritz.box
nameserver 127.0.0.1
options edns0 trust-ad
```

```
> cat /etc/hosts

# Standard host addresses
127.0.0.1  localhost
::1        localhost ip6-localhost ip6-loopback
ff02::1    ip6-allnodes
ff02::2    ip6-allrouters
# This host address
127.0.1.1  peeriot-00X
```

```
[...]

Issue #3488, opened 2024-10-02
[rootless] bridge network degrading ending in failure to expose port


apostasie:
### Description

After some time testing:

```
nerdctl run -d --name bar -p 5003:80 nginx
6c1fa26eba42ff417d193acd3b7097ba33a3764a8a52efad49c4a1a99c8c4435

curl localhost:5003
curl: (56) Recv failure: Connection reset by peer
```

This is specific to that port - that has been used (past tense) by other containers (not specific to `5003` - just specific to whichever port has been used heavily).

It looks like after these containers got destroyed, something in cni (?) does not completely release the port (maybe in iptables?).

This is tricky to reproduce - I usually trigger this with repeatedly running test suite.

- cni 1.5.1
- containerd v2rc5
- nerdctl main

apostasie:
Restarting containerd does fix the issue - until repeat usage will trigger it again.

@AkihiroSuda is containerd maintaining the list of mapped ports??
Any pointer for me on this issue?

apostasie:
```
sudo nsenter --net=/proc/410080/ns/net iptables-save

-A CNI-DN-49eb952835f33818786f4 -p tcp -m tcp --dport 5000 -j DNAT --to-destination 10.4.0.146:80
-A CNI-DN-7574ca47ce906c42f658c -s 10.4.0.0/24 -p tcp -m tcp --dport 5000 -j CNI-HOSTPORT-SETMARK
-A CNI-DN-7574ca47ce906c42f658c -s 127.0.0.1/32 -p tcp -m tcp --dport 5000 -j CNI-HOSTPORT-SETMARK
-A CNI-DN-7574ca47ce906c42f658c -p tcp -m tcp --dport 5000 -j DNAT --to-destination 10.4.0.124:5000
-A CNI-DN-d7f4e2470d8fbcbaa6504 -s 10.4.2.0/24 -p tcp -m tcp --dport 5004 -j CNI-HOSTPORT-SETMARK
-A CNI-DN-d7f4e2470d8fbcbaa6504 -s 127.0.0.1/32 -p tcp -m tcp --dport 5004 -j CNI-HOSTPORT-SETMARK
-A CNI-DN-d7f4e2470d8fbcbaa6504 -p tcp -m tcp --dport 5004 -j DNAT --to-destination 10.4.2.9:80
[...]

Issue #3487, opened 2024-10-02
Network degrading over time?


apostasie:
### Description

After heavy, prolonged usage and testing of nerdclt, network inside lima seems to be degrading, with a very large proportion of all requests ending with i/o timeout.

This is affecting the entire VM networking, not just nerdctl. Rebooting the VM does not help.

The same requests ran from the host (or from another VM) are just fine.

It is unclear to me if this would be a lima issue, a cni issue, or a nerdctl issue?

Something as simple as
`curl https://ghcr.io/v2/stargz-containers/registry/manifests/2-org`

Will intermittently (~70% of the time) fail with:

```
curl: (28) Failed to connect to ghcr.io port 443 after 132561 ms: Couldn't connect to server
```

tcpdump:
```
11:56:15.724007 IP 192.168.5.15.42684 > 140.82.116.33.443: Flags [S], seq 847856034, win 64240, options [mss 1460,sackOK,TS val 2029123835 ecr 0,nop,wscale 7], length 0
11:56:16.769207 IP 192.168.5.15.42684 > 140.82.116.33.443: Flags [S], seq 847856034, win 64240, options [mss 1460,sackOK,TS val 2029124880 ecr 0,nop,wscale 7], length 0
11:56:17.794140 IP 192.168.5.15.42684 > 140.82.116.33.443: Flags [S], seq 847856034, win 64240, options [mss 1460,sackOK,TS val 2029125905 ecr 0,nop,wscale 7], length 0
11:56:18.812797 IP 192.168.5.15.42684 > 140.82.116.33.443: Flags [S], seq 847856034, win 64240, options [mss 1460,sackOK,TS val 2029126924 ecr 0,nop,wscale 7], length 0
11:56:19.842554 IP 192.168.5.15.42684 > 140.82.116.33.443: Flags [S], seq 847856034, win 64240, options [mss 1460,sackOK,TS val 2029127953 ecr 0,nop,wscale 7], length 0
11:56:20.860340 IP 192.168.5.15.42684 > 140.82.116.33.443: Flags [S], seq 847856034, win 64240, options [mss 1460,sackOK,TS val 2029128971 ecr 0,nop,wscale 7], length 0
11:56:22.906894 IP 192.168.5.15.42684 > 140.82.116.33.443: Flags [S], seq 847856034, win 64240, options [mss 1460,sackOK,TS val 2029131018 ecr 0,nop,wscale 7], length 0
[...]

PR #4254, opened 2025-05-15
[FIX]: ensure cleanup of cni config in case of failure in createRuntime

apostasie:
This is a fix for #4253, possibly for #4246 - likely for others as well:
- #3488

Maybe indirectly responsible for #3487 and possibly some of #3253 

Possibly as well:
- #2947

### DIFF
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -418,7 +418,7 @@ func getIP6AddressOpts(opts *handlerOpts) ([]cni.NamespaceOpts, error) {
 	return nil, nil
 }
 
-func applyNetworkSettings(opts *handlerOpts) error {
+func applyNetworkSettings(opts *handlerOpts) (err error) {
 	portMapOpts, err := getPortMapOpts(opts)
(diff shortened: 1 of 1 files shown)


apostasie:
We need to review all of this lifecycle - and I would put that in the same bucket as "rewrite CNIenv".

What should happen if I restart such a container after deleting the one that was port conflicting?
Right now I am getting `mkdir /run/containerd/io.containerd.runtime.v2.task/default/ebc4ab0968ba6d8ae8455bae8b11298b50573a2213d099d233d84b6a07bfe31c: file exists`

We are fine with the "happy path", but as soon as things go south, we just break the system with no recovery hope but trash it all.

apostasie:
> Hi @apostasie, I think it is more relevant to clean the netns on `task.Kill()`, there is no point to keep iptables rules after container kill

Would that cover the scenario described here though? (eg: failing during onCreateRuntime with a `run -d`)

apostasie:
Closing after discussion with @fahedouch who will incorporate commit into #4255.

Issue #1872, opened 2023-01-18
Passing an empty Netns causes leak iptable rules


qkboy:
### Description

Run or start a container will leak three iptables rules when using cni bridge plugin and set `"ipMasq": true` .

### Steps to reproduce the issue

1.nerdctl run container
2.nerdctl stop container
3.nerdctl start container
This action will leak six iptables rules in the host.

### Describe the results you received and expected

It'll create three iptable rules when run task.Start. But it didn't remove when run task.Delete.

The reason is that cni.Remove() just pass an empty Netns path to the bridge plugin. 
```
if err := opts.cni.Remove(ctx, opts.fullID, "", namespaceOpts...); err != nil {
	logrus.WithError(err).Errorf("failed to call cni.Remove")
	return err
}
```
Bridge plugin just reclaim ipam pool and skip teardown ip masq when Netns is empty.
```
if args.Netns == "" {
	return ipamDel()
}
```
Why not pass the correct sandbox path into cni.Remove() ?

### What version of nerdctl are you using?

latest

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

```
Client:
 Namespace:	default
 Debug Mode:	false

Server:
 Server Version: v1.6.6
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Log: fluentd journald json-file
  Storage: aufs native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.13.0-40-generic
 Operating System: Ubuntu 20.04.1 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.59GiB
 Name: env2-node02
 ID: eb7dbbb8-64a4-40c7-9a64-07e0b1de9d04
```

qkboy:
In fact, nerdctl only execute cni.Remove() when run rm/start container instead of stopping container.

At that time, it has run task.Kill, the container progress has exited. So bridge plugin can't obtain the ip address from Netns even we pass the correct sandbox path to it. It'll fail to get ipns then skip to teardown ip masq.

I think this problem may not be easy to solve in nerdctl. However it can be handled in the bridge plugin.

refer: https://github.com/containernetworking/plugins/issues/810

fahedouch:
fixe by https://github.com/containerd/nerdctl/pull/4255

PR #4255, opened 2025-05-16
fix: clean up iptables rules for stopped containers

fahedouch:
fixing this issue https://github.com/containerd/nerdctl/issues/1872 or at least a part of the issue

Based on the[ runtime-spec](https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle) documentation, the post-hook is executed after the container task is deleted. This ensures that resources are properly cleaned up when the container is removed, allowing sufficient iptables space for the next container to be created.

I think this should replace https://github.com/containerd/nerdctl/pull/4254 ( cc @apostasie)

### DIFF
--- a/cmd/nerdctl/container/container_remove_linux_test.go
+++ b/cmd/nerdctl/container/container_remove_linux_test.go
@@ -1,0 +1,121 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package container
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/require"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+
+	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/portlock"
+)
+
+// iptablesCheckCommand is the shell command to check iptables rules
+const iptablesCheckCommand = "iptables -t nat -S && iptables -t filter -S && iptables -t mangle -S"
+
+// testContainerRmIptablesExecutor is a common executor function for testing iptables rules cleanup
+func testContainerRmIptablesExecutor(data test.Data, helpers test.Helpers) test.TestableCommand {
+	t := helpers.T()
+
+	// Get the container ID from the label
+	containerID := data.Labels().Get("containerID")
+
+	// Remove the container
+	helpers.Ensure("rm", "-f", containerID)
+
+	time.Sleep(1 * time.Second)
+
+	// Create a TestableCommand using helpers.Custom
+	if rootlessutil.IsRootless() {
+		// In rootless mode, we need to enter the rootlesskit network namespace
+		if netns, err := rootlessutil.DetachedNetNS(); err != nil {
+			t.Fatalf("Failed to get detached network namespace: %v", err)
+		} else {
+			if netns != "" {
+				// Use containerd-rootless-setuptool.sh to enter the RootlessKit namespace
+				return helpers.Custom("containerd-rootless-setuptool.sh", "nsenter", "--", "nsenter", "--net="+netns, "sh", "-ec", iptablesCheckCommand)
+			}
+			// Enter into :RootlessKit namespace using containerd-rootless-setuptool.sh
+			return helpers.Custom("containerd-rootless-setuptool.sh", "nsenter", "--", "sh", "-ec", iptablesCheckCommand)
+		}
+	}
+
+	// In non-rootless mode, check iptables rules directly on the host
+	return helpers.Custom("sh", "-ec", iptablesCheckCommand)
+}
+
+// TestContainerRmIptables tests that iptables rules are cleared after container deletion
+func TestContainerRmIptables(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	// Require iptables and containerd-rootless-setuptool.sh commands to be available
+	testCase.Require = require.All(
+		require.Binary("iptables"),
+		require.Binary("containerd-rootless-setuptool.sh"),
+		require.Not(require.Windows),
+		require.Not(nerdtest.Docker),
+	)
+
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "Test iptables rules are cleared after container deletion",
+			Setup: func(data test.Data, helpers test.Helpers) {
+				// Get a free port using portlock
+				port, err := portlock.Acquire(0)
+				if err != nil {
+					helpers.T().Fatalf("Failed to acquire port: %v", err)
+				}
+				data.Labels().Set("port", strconv.Itoa(port))
+
+				// Create a container with port mapping to ensure iptables rules are created
+				containerID := helpers.Capture("run", "-d", "--name", data.Identifier(), "-p", fmt.Sprintf("%d:80", port), testutil.NginxAlpineImage)
+				data.Labels().Set("containerID", containerID)
+				nerdtest.EnsureContainerStarted(helpers, data.Identifier())
+			},
+			Cleanup: func(data test.Data, helpers test.Helpers) {
+				// Make sure container is removed even if test fails
+				helpers.Anyhow("rm", "-f", data.Identifier())
+
+				// Release the acquired port
+				if portStr := data.Labels().Get("port"); portStr != "" {
+					port, _ := strconv.Atoi(portStr)
+					_ = portlock.Release(port)
+				}
+			},
+			Command: testContainerRmIptablesExecutor,
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				// Get the container ID from the label
+				containerID := data.Labels().Get("containerID")
+				return &test.Expected{
+					ExitCode: expect.ExitCodeSuccess,
+					// Verify that the iptables output does not contain the container ID
+					Output: expect.DoesNotContain(containerID),
+				}
+			},
+		},
+	}
+
+	testCase.Run(t)
+}
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -418,7 +419,7 @@ func getIP6AddressOpts(opts *handlerOpts) ([]cni.NamespaceOpts, error) {
 	return nil, nil
 }
 
-func applyNetworkSettings(opts *handlerOpts) error {
+func applyNetworkSettings(opts *handlerOpts) (err error) {
 	portMapOpts, err := getPortMapOpts(opts)
 	if err != nil {
 		return err
@@ -475,10 +476,19 @@ func applyNetworkSettings(opts *handlerOpts) error {
 	// See https://github.com/containerd/nerdctl/issues/3355
 	_ = opts.cni.Remove(ctx, opts.fullID, "", namespaceOpts...)
 
+	// Defer CNI configuration removal to ensure idempotency of oci-hook.
+	defer func() {
+		if err != nil {
+			log.L.Warn("Container failed starting. Removing allocated network configuration.")
+			_ = opts.cni.Remove(ctx, opts.fullID, nsPath, namespaceOpts...)
+		}
+	}()
+
 	cniRes, err := opts.cni.Setup(ctx, opts.fullID, nsPath, namespaceOpts...)
 	if err != nil {
 		return fmt.Errorf("failed to call cni.Setup: %w", err)
 	}
+
 	cniResRaw := cniRes.Raw()
 	for i, cniName := range opts.cniNames {
 		hsMeta.Networks[cniName] = cniResRaw[i]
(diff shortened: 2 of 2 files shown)


apostasie:
It does not feel like this would fix #4253 or any of the same ilk, or will it?

apostasie:
Also be sure to check out: https://github.com/containerd/nerdctl/issues/3357

In case of failure during onCreateRuntime, onPostStop is either wrongly dispatched while the container will not be deleted, or the container should have been deleted but is not.
While this might be a bug in containerd, or otherwise some unpolished corner of the oci spec, the end-result is the same: we need to prevent onPostStop from proceeding further as the container is not destroyed (this is why we have `shouldExit`).

Your post-cleanup here (while useful for "normal" exits) will not work in case of failures onCreateRuntime because of ^

apostasie:
Suggestion is:
1. isolate cni in its own file, and at least provide a `Remove` method that incorporates your post-cleanup here
2. incorporate #4254 into this patch here, using the updated Remove method

Then we might want to have a look overall at what we do in oci hooks.
There are definitely multiple issues (maybe some outside of nerdctl) - the whole thing is untested, and the code does not look too good...

fahedouch:
> Suggestion is:
> 
> 1. isolate cni in its own file, and at least provide a `Remove` method that incorporates your post-cleanup here
> 2. incorporate [[FIX]: ensure cleanup of cni config in case of failure in createRuntime #4254](https://github.com/containerd/nerdctl/pull/4254) into this patch here, using the updated Remove method
> 
> Then we might want to have a look overall at what we do in oci hooks. There are definitely multiple issues (maybe some outside of nerdctl) - the whole thing is untested, and the code does not look too good...

SGTM, 
Indeed, OCI operations, especially creation, should be `idempotent`.
Regarding your first point, I agree. However, I'd like to wait for the CNI fix related to purging iptable rules with empty netNs path before refactoring this part, as my function will be affected in that case.
For your second point, I agree and will cherry-pick your commit. There's no need for a separate PR for this.

apostasie:
> * Regarding your first point, I agree. However, I'd like to wait for the CNI fix related to purging iptable rules with empty netNs path before refactoring this part, as my function will be affected in that case.

The original PR has been opened over two years ago.
I don't mean to be negative nor judgmental - what I am saying is that we might prefer to act now, assuming it is not going to be merged anytime soon, and revert later when/if it happens.

pkg/ocihook/ocihook.go
```diff
@@ -640,6 +650,55 @@ func onPostStop(opts *handlerOpts) error {
...
+				if len(parts) < 2 {
+					continue
+				}
+
+				// Create delete command
+				deleteArgs := []string{"-t", table, "-D"}
+				deleteArgs = append(deleteArgs, parts[1:]...)
+
+				// Execute delete command
+				deleteCmd := exec.Command("iptables", deleteArgs...)
```
fahedouch, line 689: @apostasie 

I'm struggling to manage the escape characters introduced by Go and the shell.
When I print the command, it looks like this:

```
msg="cmd: /usr/sbin/iptables -t nat -D POSTROUTING -s 10.4.0.230/32 -m comment --comment \"name: \\\"bridge\\\" id: \\\"default-a090a1eca8409688f58138782af91891743d9d7a42e41a15c31734d92584b799\\\"\" -j CNI-97c2473ecc8978c6084c824e" error="exit status 1"
```

Apparently, there are extra escape characters.

Do you have an idea how to fix this?

apostasie: Top of the head, _you are not going to have fun_ trying to parse the output of `iptables -t X -S`.

What I would suggest is to not even try:

```golang

exec.Command("sh", "-c", "--", fmt.Sprintf(`iptables -t %s -D %s`, table, rule[3:]))
```

(and drop the strings.Fields)

fahedouch: thanks @apostasie

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// In non-rootless mode, check iptables rules directly on the host
+	return helpers.Custom("sh", "-c", "iptables -t nat -S && iptables -t filter -S && iptables -t mangle -S")
+}
+
+// TestContainerRmIptables tests that iptables rules are cleared after container deletion
+func TestContainerRmIptables(t *testing.T) {
+	if _, err := exec.LookPath("iptables"); err != nil {
+		t.Skip("iptables command not found, skipping test")
```
AkihiroSuda, line 91: `testCase.Require` should be used to require the command

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	testCase := nerdtest.Setup()
+
+	// This test requires Linux
+	testCase.Require = require.Not(require.Windows)
```
AkihiroSuda, line 97: The test is meaningless on Docker?

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,125 @@
...
+					"--", "sh", "-c", fmt.Sprintf("nsenter --net=%s sh -c 'iptables -t nat -S && iptables -t filter -S && iptables -t mangle -S'", netnsPath))
+			}
+			// Enter the user and mount namespace
+			return helpers.Custom("nsenter", "-t", strconv.Itoa(childPid), "-U", "-m", "-n", "--preserve-credentials",
+				"--", "sh", "-c", "iptables -t nat -S && iptables -t filter -S && iptables -t mangle -S")
+		}
+	}
+
+	// In non-rootless mode, check iptables rules directly on the host
+	return helpers.Custom("sh", "-c", "iptables -t nat -S && iptables -t filter -S && iptables -t mangle -S")
```
AkihiroSuda, line 85: The shell script should be defined as a constant to avoid the code clone across rootless and rootful

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,128 @@
...
+		if netns, err := rootlessutil.DetachedNetNS(); err != nil {
+			t.Fatalf("Failed to get detached network namespace: %v", err)
+		} else {
+			if netns != "" {
+				// First enter the user and mount namespace, then the network namespace
+				return helpers.Custom("nsenter", "-t", strconv.Itoa(childPid), "-U", "-m", "--preserve-credentials",
+					"--", "sh", "-c", fmt.Sprintf("nsenter --net=%s sh -c '%s'", netnsPath, iptablesCheckCommand))
+			}
+			// Enter the user and mount namespace using containerd-rootless-setuptool.sh
+			return helpers.Custom("containerd-rootless-setuptool.sh", "nsenter", "--", "sh", "-c", iptablesCheckCommand)
```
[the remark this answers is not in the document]
fahedouch: nvm

pkg/ocihook/ocihook.go
```diff
@@ -477,6 +478,15 @@ func applyNetworkSettings(opts *handlerOpts) error {
 	if err != nil {
 		return fmt.Errorf("failed to call cni.Setup: %w", err)
 	}
+
+	// Defer CNI configuration removal to ensure idempotency of oci-hook.
+	defer func() {
+		if err != nil {
+			log.L.Warn("Container failed starting. Removing allocated network configuration.")
+			_ = opts.cni.Remove(ctx, opts.fullID, nsPath, namespaceOpts...)
```
apostasie, line 486: Do we want to _also_ `cleanupIptablesRules` here?

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,127 @@
...
+		require.Not(nerdtest.Docker),
+	)
+
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "Test iptables rules are cleared after container deletion",
+			Setup: func(data test.Data, helpers test.Helpers) {
+				// Create a container with port mapping to ensure iptables rules are created
+				helpers.Ensure("run", "-d", "--name", data.Identifier(), "-p", "8080:80", testutil.NginxAlpineImage)
+				time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
fahedouch: fixed

fahedouch:
@AkihiroSuda  @apostasie  PTAL

pkg/ocihook/ocihook.go
```diff
@@ -479,6 +480,15 @@ func applyNetworkSettings(opts *handlerOpts) error {
 	if err != nil {
 		return fmt.Errorf("failed to call cni.Setup: %w", err)
 	}
+
+	// Defer CNI configuration removal to ensure idempotency of oci-hook.
+	defer func() {
```
apostasie, line 485: I am now wondering if this should be moved above `opts.cni.Setup`? (eg: would there be conditions where cni setup would fail and leave partially configured network above?)

apostasie (approved):

apostasie:
@fahedouch couple of non critical nits:
- we can ensure the port is free to avoid future collisions instead of using 8080
- minor comments on testing

Anyhow, LGTM - let's have this in, it is a big deal.

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+		// Construct the path to the network namespace
+		uid := os.Getuid()
+		netnsPath := fmt.Sprintf("/run/user/%d/containerd-rootless/netns", uid)
+
+		if netns, err := rootlessutil.DetachedNetNS(); err != nil {
+			t.Fatalf("Failed to get detached network namespace: %v", err)
+		} else {
+			if netns != "" {
+				// First enter the user and mount namespace, then the network namespace
```
AkihiroSuda, line 76: Probably you can use `containerd-rootless-setuptool.sh nsenter -- nsenter --net=...` to simplify the code

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		childPid, err := rootlessutil.RootlessKitChildPid(stateDir)
+		if err != nil {
+			t.Fatalf("Failed to get rootlesskit child pid: %v", err)
+		}
+
+		// Construct the path to the network namespace
+		uid := os.Getuid()
+		netnsPath := fmt.Sprintf("/run/user/%d/containerd-rootless/netns", uid)
```
AkihiroSuda, line 70: This variable is redundant. The path is provided in the `netns` string below

fahedouch: fixed

cmd/nerdctl/container/container_remove_linux_test.go
fahedouch: fixed

pkg/ocihook/ocihook.go
fahedouch: fixed

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,127 @@
...
+func testContainerRmIptablesExecutor(data test.Data, helpers test.Helpers) test.TestableCommand {
+	t := helpers.T()
+
+	// Get the actual container ID from the container name and store it in a label
+	containerID := data.Labels().Get("containerID")
+	if containerID == "" {
+		container := nerdtest.InspectContainer(helpers, data.Identifier())
+		containerID = container.ID
+		data.Labels().Set("containerID", containerID)
+	}
```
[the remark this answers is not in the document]
fahedouch: fixed

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,127 @@
...
+		require.Not(require.Windows),
+		require.Not(nerdtest.Docker),
+	)
+
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "Test iptables rules are cleared after container deletion",
+			Setup: func(data test.Data, helpers test.Helpers) {
+				// Create a container with port mapping to ensure iptables rules are created
+				helpers.Ensure("run", "-d", "--name", data.Identifier(), "-p", "8080:80", testutil.NginxAlpineImage)
```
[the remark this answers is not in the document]
fahedouch: fixed

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,122 @@
...
+	// Create a TestableCommand using helpers.Custom
+	if rootlessutil.IsRootless() {
+		// In rootless mode, we need to enter the rootlesskit network namespace
+		if netns, err := rootlessutil.DetachedNetNS(); err != nil {
+			t.Fatalf("Failed to get detached network namespace: %v", err)
+		} else {
+			if netns != "" {
+				// Use containerd-rootless-setuptool.sh to enter the RootlessKit namespace
+				return helpers.Custom("containerd-rootless-setuptool.sh", "nsenter", "--", "sh", "-c",
+					fmt.Sprintf("nsenter --net=%s sh -c '%s'", netns, iptablesCheckCommand))
```
AkihiroSuda: No need to use `sh -c` twice.
The first one can be replaced with `"c-r-s.sh", "nsenter", "--", "nsenter", "--net="+netns, ...`

AkihiroSuda: Also probably `sh -c` should be `sh -ec`

pkg/ocihook/ocihook.go
```diff
@@ -642,6 +661,43 @@ func onPostStop(opts *handlerOpts) error {
...
+			log.L.WithError(err).Warnf("failed to list iptables rules for table %s", table)
+			continue
+		}
+
+		// Find and delete rules related to the container
+		rules := strings.Split(string(output), "\n")
+		for _, rule := range rules {
+			if strings.Contains(rule, containerID) {
+				// Execute delete command
+				deleteCmd := exec.Command("sh", "-c", "--", fmt.Sprintf(`iptables -t %s -D %s`, table, rule[3:]))
```
AkihiroSuda, line 688: No need to use `sh -c`

fahedouch: I need to pass string cmd with `sh -c` to avoid escaping issues. see https://github.com/containerd/nerdctl/pull/4255#discussion_r2096626718

cmd/nerdctl/container/container_remove_linux_test.go
```diff
@@ -0,0 +1,121 @@
...
+func testContainerRmIptablesExecutor(data test.Data, helpers test.Helpers) test.TestableCommand {
+	t := helpers.T()
+
+	// Get the container ID from the label
+	containerID := data.Labels().Get("containerID")
+
+	// Remove the container
+	helpers.Ensure("rm", "-f", containerID)
+
+	time.Sleep(1 * time.Second)
```
apostasie, line 48: Curious if this is necessary. Is cni teardown going to be asynchronous?

apostasie (approved):

PR #4256, opened 2025-05-16
[REFACTOR] Filesystem, part 2: revamp locks

apostasie:
This is the next PR in filesystem refactor (broken out of #4240).

This one is focused strictly on locks.

1. add tests
2. lock is now meant for either dir or file - if the lock does not exist, by default a file is created 
3. rename WithDirLock to WithLock
4. introduce RO locks - this will be good for performance
5. streamline platform specific logic


pkg/internal/filesystem/lock.go
```diff
@@ -0,0 +1,124 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+// Portions from internal go
+//
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
AkihiroSuda, line 21: Probably this should be a separate package.
Also needs permalinks to "internal go" and the LICENSE file

pkg/internal/filesystem/lock.go
```diff
@@ -0,0 +1,124 @@
...
+
+package filesystem
+
+// Package filelock provides a platform-independent API for advisory file
```
AkihiroSuda, line 25: Doesn't match the package name

pkg/internal/filesystem/lock.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func commonlock(path string, mode lockType) (file *os.File, err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Join(ErrLockFail, err, file.Close())
+		}
+	}()
+
+	if runtime.GOOS == "windows" {
```
AkihiroSuda, line 57: This should be in `_windows.go`?

pkg/internal/filesystem/lock_windows.go
```diff
@@ -0,0 +1,56 @@
...
+   limitations under the License.
+*/
+
+// Portions from internal go
+//
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/lockedfile/internal/filelock/filelock_windows.go
```
AkihiroSuda, line 23: s/master/permanent link/

pkg/internal/filesystem/lock_windows.go
```diff
@@ -0,0 +1,56 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+// Portions from internal go
+//
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
AkihiroSuda, line 21: Same as my comments above

pkg/internal/filesystem/lock.go
apostasie: Thanks.

pkg/internal/filesystem/lock.go
apostasie: Will do.
Honestly, this is me being a bit overzealous on this - mainly acknowledging that I did _read_ that package but yeah, let's be thorough.
I ll add these ^.

pkg/internal/filesystem/lock.go
apostasie: The note below elaborates a bit on why it is not.

If this is really a problem, we could introduce an additional platform specific method (something like `linkPath()`,  just to do the path transform?

LMK if you would prefer that ^ or if we can live with the runtime check.

pkg/internal/filesystem/lock_windows.go
apostasie: Done.

pkg/internal/filesystem/lock.go
apostasie: Done.

apostasie: > Probably this should be a separate package. Also needs permalinks to "internal go" and the LICENSE file

Let's continue the work on filesystem primitives and see if we are overcrowded in the end and need to split out in subpackages?

apostasie:
Comments addressed.

Second push does rebase.

apostasie:
Failure is Hub 429.

AkihiroSuda (approved):
Thanks

PR #4259, opened 2025-05-17
Use only one build base in Dockerfile

### LINKED ISSUE #4037: Dockerfile: consistently use Alpine or Debian for building the binaries
> IIRC static linking didn't work well with glibc, at least in the past.
Basically we should rather try to use Alpine (musl) for building the binaries

_Originally posted by @AkihiroSuda in https://github.com/containerd/nerdctl/pull/4012#discussion_r2014426549_

apostasie:
Fix #4037


AkihiroSuda (approved):
Thanks

Issue #4027, opened 2025-03-20
nerdctl compose file map multiple ports


chawson:
### What is the problem you're trying to solve
when compose.yml file include follows code will get error:
#### compose.yaml: service part
```yaml
mozi-zlm:
    image: nginx:1.27.4
    ports:
      - '32000-32200:32000-32200'
      - '32000-32200:32000-32200/udp'
```
#### nerdctl compose up -d received
![image](https://private-user-images.githubusercontent.com/29857648/423841318-269963c7-6f3f-4cad-846b-1a15fbb1b1a0.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDI0NTkwMzcsIm5iZiI6MTc0MjQ1ODczNywicGF0aCI6Ii8yOTg1NzY0OC80MjM4NDEzMTgtMjY5OTYzYzctNmYzZi00Y2FkLTg0NmItMWExNWZiYjFiMWEwLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTAzMjAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwMzIwVDA4MTg1N1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWNlMGRlMWQxZDFkNGM5MzljMjI2ZjMyMWFkM2M2N2Q5M2IxODU2NDQ5MTExNmYwNDNlMDA0N2M0MGFiZmRhZGQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.NPl0xkDHFVKiC8kpFMgyhSJPPp0kFvGnaYfm-Yw5qJ0)
[...]

PR #4260, opened 2025-05-17
test: update test logic in TestParsePortsLabel

haytok:
The current implementation does not compare `[]cni.PortMapping{}` obtained from `labelMap` and `ParsePortsLabel()` and want.
Therefore, this commit will update so that the evaluation is performed.


haytok:
I am currently working on #4027 and created this PR because I need to modify `ParsePortsLabel()` in the future and want to test it properly.

AkihiroSuda (approved):
Thanks, nice catch

PR #4261, opened 2025-05-17
test: update test logic in TestTestParseFlagPWithPlatformSpec

haytok:
This test does not run in the rootfull Linux environment. 
Also, there are some cases where the test does not run. 
Therefore, this commit modifies all tests to perform the evaluation.

Note that when the host ports are not specified in the -p option of `nerdctl run`, ports are randomly assigned, but it is not known what port number will be assigned in each test.
Therefore, this commit modifies not to compare host ports.

### DIFF
--- a/pkg/portutil/portutil_test.go
+++ b/pkg/portutil/portutil_test.go
@@ -29,7 +29,7 @@ import (
 )
 
 func TestTestParseFlagPWithPlatformSpec(t *testing.T) {
-	if runtime.GOOS != "Linux" || rootlessutil.IsRootless() {
+	if runtime.GOOS != "linux" || rootlessutil.IsRootless() {
 		t.Skip("no non-Linux platform or rootless mode in Linux are not supported yet")
 	}
 	type args struct {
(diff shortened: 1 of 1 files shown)


haytok:
Similar Case : https://github.com/containerd/nerdctl/pull/4260

haytok:
The test I modified will not run in `rootless`. So the test failure is not related to this PR fix.

> EL / almalinux-8 rootless (pull_request)

So, could maintainers retry the CI ?

AkihiroSuda (approved):
Thanks

pkg/portutil/portutil_test.go
AkihiroSuda, line 1: The tests should use gotest.tools/v3/assert

haytok: Thanks for comments! 
Ok, I will try to rewrite it in another PR.

AkihiroSuda: Thanks 🙏

PR #4265, opened 2025-05-18
[CI]: move away from raw github domain to API

### LINKED ISSUE #4264: [CI]: github 429 break the build
### Description

In recent days, Github started answering 429 (too many requests) while building the Dockerfile.

It seems that this only applies to raw.githubusercontent and releases downloads - not git clone.

This is worrisome, as the inability to build would prevent any test to run on the CI, and as this is new - possibly indicative that GH is tightening the belt.

Solutions:
1. stop using binary releases and build everything from source (this is desirable in itself)
2. issue a ro github token and set it up on the CI so that we are authenticated and (possibly) have higher limits

1. will require significant work and take some time

@AkihiroSuda any chance you could setup 2. as a stop-gap measure?

_No response_

na

main

None

### Host information

_No response_

apostasie:
Fix (part of?) #4264
See ^ for context.

TL;DR: recent changes in GitHub rate limits (presumably https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/) now break the build for us, apparently for requests to raw.githubusercontent.

This PR does a couple of things:
- ensure we pass along the GITHUB_TOKEN into the Dockerfile so that we can authenticate GitHub queries
- provide a helper that can retrieve (authenticated) raw content from GitHub API instead of using the raw domain

Note that it is unclear if we are going to get throttled on other GitHub requests as well - but at least we now have the infrastructure in place to perform authenticated requests, so, we can change what needs be later on piece-meal.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,6 +61,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -qq && apt-get install -qq --no-install-recommends \
   make \
   git \
+  jq \
   curl \
   dpkg-dev
 ARG TARGETARCH
@@ -75,6 +76,7 @@ RUN xx-apt-get update -qq && xx-apt-get install -qq --no-install-recommends \
   pkg-config
 RUN git config --global advice.detachedHead false
 ADD hack/git-checkout-tag-with-hash.sh /usr/local/bin/
+ADD hack/scripts/lib.sh /usr/local/bin/http::helper
 
 FROM build-base AS build-containerd
 ARG TARGETARCH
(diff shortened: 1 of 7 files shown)


apostasie:
Failure is not random logs failure.

Dockerfile
```diff
@@ -52,6 +52,8 @@ ARG NYDUS_VERSION=v2.3.1
 ARG SOCI_SNAPSHOTTER_VERSION=0.9.0
 ARG KUBO_VERSION=v0.34.1
 
+ARG GITHUB_TOKEN=""
```
AkihiroSuda, line 55: 🚨 This leaks the token via  `docker history`

AkihiroSuda: `RUN --secret` is the right option

AkihiroSuda: Relevant proposal
- https://github.com/moby/buildkit/issues/5978

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/4265#discussion_r2095091105

Dockerfile
apostasie: Sure...
This token can be accessed and manipulated by anyone who sends a pull request in the first place.
This is not a secret at all.

Am I missing something about how this works?

apostasie: I can do it on principles - though, it does not seem like this would protect us from anything or not leak  something that is not already public.

AkihiroSuda: The token has permissions here:

https://github.com/containerd/nerdctl/blob/3c8411b33994dbc7f161d34d92c60b98a9f261a3/.github/workflows/ghcr-image-build-and-publish.yml#L28-L30

https://github.com/containerd/nerdctl/blob/3c8411b33994dbc7f161d34d92c60b98a9f261a3/.github/workflows/release.yml#L21-L24


These yamls weren't touched in this PR, but basically it should be always assumed that the token is privileged and unsafe to publish

apostasie: Agreed on principle.

apostasie: (resolving as latest push from earlier today had the changes)

AkihiroSuda: Do you want to update `ghcr-*.yaml` and `release.yaml` too?

apostasie: Given the recent snafoo with 2.1.0, suggesting we run on a test tag before next release.

PR #4209, opened 2025-05-07
[CI]: rewrite logs tests

apostasie:
Commit 1 relaxes RunSigProxy refresh frequency.
Suspicion is that currently this is putting extra pressure on the runtime - especially for slow environments like EL.

Commit 2 rewrites all remaining logs tests with the new framework.

IMPORTANT NOTES:
- fix #4147
- some tests conditions have been made more stringent, since #4201 got fixed
- tests are disabled for Windows - it seems like logging for Windows only works intermittently
- this rewrite does away with `-d`, as this is a significant source of flakyness - if this is not acceptable and we do believe that all these tests **must** be run with `-d`, let me know and I will come up with something else

Note that while torturing tests, the following issues showed-up:
- #4243
- #4237
- and to a lesser extent: #4233

Because of these, logs tests may/will fail randomly.

PR #4203, opened 2025-05-06
[CI]: rewrite some compose tests

apostasie:
This PR is rewriting some compose tests (as many as I can stomach in one sitting) to the new test framework.

Obviously I am mostly interested in:
- making them running parallel to reduce total CI runtime
- moving them out of the "with-retry" bucket

A few notes:
- removing `version: XYZ` in most cases from compose yml, as this is deprecated and is just printing additional noise
- removing ports being exposed when not necessary for the test (as this was hardcoded to 8080, it would prevent tests from running in parallel)
- no longer manually poke image name and instead use referenceutils to parse them properly when needed (in preparation for upgrading test images to digest-pinning)
- some tests have been consolidated into subtests (cut the additional cost on tear-up / tear-down when not needed)

PR #4266, opened 2025-05-18
[CI] more tests cleanup

apostasie:
~~On top of #4203 and #4209~~

In a shell:
- remove compose yaml `version` field (deprecated, only adds noise to the logs)
- remove useless port bindings preventing parallelization
- move some tests from `Alpine` to `Common`
- use data.Temp() for assets to simplify temp files handling


AkihiroSuda (approved):
Thanks

PR #4278, opened 2025-05-21
[cleanup]: golang linter QF1012 to enforce fmt.Fprintf usage

yankay:
Enable staticcheck linter QF1012 to detect and replace `x.Write(fmt.Sprintf(...))` patterns
And updated non-compliant code.


AkihiroSuda (approved):
Thanks

PR #4284, opened 2025-05-23
<fix> stats: CPU perusage use the wrong systemUsage

zzzzzzzzzy9:
<fix> stats: CPU perusage use the wrong systemUsage

### DIFF
--- a/pkg/statsutil/stats_linux.go
+++ b/pkg/statsutil/stats_linux.go
@@ -38,8 +38,8 @@ func calculateMemPercent(limit float64, usedNo float64) float64 {
 	return 0
 }
 
-func SetCgroupStatsFields(previousStats *ContainerStats, data *v1.Metrics, links []netlink.Link) (StatsEntry, error) {
-	cpuPercent := calculateCgroupCPUPercent(previousStats, data)
+func SetCgroupStatsFields(previousStats *ContainerStats, data *v1.Metrics, links []netlink.Link, systemInfo SystemInfo) (StatsEntry, error) {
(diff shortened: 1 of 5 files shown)


pkg/statsutil/stats_linux.go
```diff
@@ -114,18 +119,21 @@ func getHostMemLimit() float64 {
...
+		systemDelta = float64(systemInfo.SystemUsage) - float64(previousStats.CgroupSystem)
+		onlineCPUs  = systemInfo.OnlineCPUs
 	)
 
+	if onlineCPUs == 0 {
+		onlineCPUs = uint32(len(metrics.CPU.Usage.PerCPU))
+	}
 	if systemDelta > 0.0 && cpuDelta > 0.0 {
-		cpuPercent = (cpuDelta / systemDelta) * onlineCPUs * 100.0
+		cpuPercent = (cpuDelta / systemDelta) * float64(systemInfo.OnlineCPUs) * 100.0
```
fahedouch, line 136: There's a potential inconsistency in the `calculateCgroupCPUPercent` function where `onlineCPUs` is used in two places:

````
if onlineCPUs == 0 {
    onlineCPUs = uint32(len(metrics.CPU.Usage.PerCPU))
}
// ...
cpuPercent = (cpuDelta / systemDelta) * float64(systemInfo.OnlineCPUs) * 100.0
`````
The second usage directly accesses `systemInfo.OnlineCPUs` instead of using the local onlineCPUs variable that might have been updated. This could lead to inconsistent behavior if the fallback was triggered.

PR #4289, opened 2025-05-24
Prevent --name="" for container

### LINKED ISSUE #4288: Explicit, empty string container name should be replaced by generated name
### Description

`nerdctl run -d --name="" debian` vs `docker run -d --name="" debian`

Docker will generate a name - nerdctl will create a container with name "".

Note that this also likely effup the namestore and probably has side-effects.

_No response_

na

main

None

### Host information

_No response_

apostasie:
Fix #4288


AkihiroSuda (approved):
Thanks

PR #4290, opened 2025-05-25
fix: allow containers to start using a large numbers of ports

haytok:
Suppose we have a compose.yaml that allocates a large numbers of ports as follows.

```
> cat compose.yaml
services:
  svc0:
    image: alpine
    command: "sleep infinity"
    ports:
      - '32000-32060:32000-32060'
```

When we run `nerdctl compose up -d` using this compose.yaml, we will get the following error.

```
FATA[0000] create container failed validation: containers.Labels: label key and value length (4711 bytes) greater than maximum size (4096 bytes), key: nerdctl/ports: invalid argument
FATA[0000] error while creating container haytok-svc0-1: error while creating container haytok-svc0-1: exit status 1
```

This issue is reported in the following issue.

- https://github.com/containerd/nerdctl/issues/4027

This issue is considered to be the same as the one with errors when trying to perform many port mappings, such as `nerdctl run -p 80:80 -p 81:81 ~ -p 1000:1000 ...`

The current implementation is processing to create a container with the information specified in `-p` to the label.
And as can be seen from the error message, as the number of ports to be port mapped increases, the creation of the container fails because it violates the limit of the maximum number of bytes on the containerd side that can be allocated for a label.

Therefore, this PR modifies the container creation process so that containers can be launched without having to assign the information specified in the `-p` option to the labels.

Specifically, port mapping information is stored in the following path, and when port mapping information is required, it is retrieved from this file.

```
<DATAROOT>/<ADDRHASH>/containers/<NAMESPACE>/<CID>/port-mappings.json
```

### DIFF
--- a/cmd/nerdctl/container/container_run_network_linux_test.go
+++ b/cmd/nerdctl/container/container_run_network_linux_test.go
@@ -36,7 +36,6 @@ import (
 
 	"github.com/containerd/containerd/v2/defaults"
 	"github.com/containerd/containerd/v2/pkg/netns"
-	"github.com/containerd/errdefs"
 	"github.com/containerd/nerdctl/mod/tigron/expect"
 	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
@@ -409,21 +408,21 @@ func TestRunPort(t *testing.T) {
 	baseTestRunPort(t, testutil.NginxAlpineImage, testutil.NginxAlpineIndexHTMLSnippet, true)
 }
 
-func TestRunWithInvalidPortThenCleanUp(t *testing.T) {
+func TestRunWithManyPortsThenCleanUp(t *testing.T) {
 	testCase := nerdtest.Setup()
 	// docker does not set label restriction to 4096 bytes
 	testCase.Require = require.Not(nerdtest.Docker)
 
 	testCase.SubTests = []*test.Case{
 		{
-			Description: "Run a container with invalid ports, and then clean up.",
+			Description: "Run a container with many ports, and then clean up.",
 			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
 				return helpers.Command("run", "--data-root", data.Temp().Path(), "--rm", "-p", "22200-22299:22200-22299", testutil.CommonImage)
 			},
 			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
 				return &test.Expected{
-					ExitCode: 1,
-					Errors:   []error{errdefs.ErrInvalidArgument},
+					ExitCode: 0,
+					Errors:   []error{},
 					Output: func(stdout string, t tig.T) {
 						getAddrHash := func(addr string) string {
 							const addrHashLen = 8
--- a/pkg/containerutil/container_network_manager.go
+++ b/pkg/containerutil/container_network_manager.go
@@ -893,12 +893,6 @@ func NetworkOptionsFromSpec(spec *specs.Spec) (types.NetworkOptions, error) {
 	}
 	opts.NetworkSlice = networks
 
-	if portsJSON := spec.Annotations[labels.Ports]; portsJSON != "" {
-		if err := json.Unmarshal([]byte(portsJSON), &opts.PortMappings); err != nil {
-			return opts, err
-		}
-	}
-
 	return opts, nil
 }
 
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -57,6 +57,7 @@ const (
 	// Currently, the length of the slice must be 1.
 	Networks = Prefix + "networks"
 
+	// DEPRECATED : https://github.com/containerd/nerdctl/pull/4290
 	// Ports is a JSON-marshalled string of []cni.PortMapping .
 	Ports = Prefix + "ports"
 
--- a/pkg/netutil/networkstore/networkstore.go
+++ b/pkg/netutil/networkstore/networkstore.go
@@ -1,0 +1,110 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package networkstore
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"path/filepath"
+
+	"github.com/containerd/go-cni"
+
+	"github.com/containerd/nerdctl/v2/pkg/store"
+)
+
+const (
+	containersDirBaseName = "containers"
+	networkConfigName     = "network-config.json"
+)
+
+var ErrNetworkStore = errors.New("network-store error")
+
+func New(dataStore, namespace, containerID string) (ns *NetworkStore, err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Join(ErrNetworkStore, err)
+		}
+	}()
+
+	if dataStore == "" || namespace == "" || containerID == "" {
+		return nil, fmt.Errorf("either dataStore or namespace or containerID is empty")
+	}
+
+	st, err := store.New(filepath.Join(dataStore, containersDirBaseName, namespace, containerID), 0, 0o600)
+	if err != nil {
+		return nil, err
+	}
+
+	return &NetworkStore{
+		safeStore: st,
+	}, nil
+}
+
+type NetworkStore struct {
+	safeStore store.Store
+
+	PortMappings []cni.PortMapping
+}
+
+func (ns *NetworkStore) Acquire(portMappings []cni.PortMapping) (err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Join(ErrNetworkStore, err)
+		}
+	}()
+
+	portsJSON, err := json.Marshal(portMappings)
+	if err != nil {
+		return fmt.Errorf("failed to marshal port mappings to JSON: %w", err)
+	}
+
+	return ns.safeStore.WithLock(func() error {
+		return ns.safeStore.Set(portsJSON, networkConfigName)
+	})
+}
+
+func (ns *NetworkStore) Load() (err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Join(ErrNetworkStore, err)
+		}
+	}()
+
+	return ns.safeStore.WithLock(func() error {
+		doesExist, err := ns.safeStore.Exists(networkConfigName)
+		if err != nil || !doesExist {
+			return err
+		}
+
+		data, err := ns.safeStore.Get(networkConfigName)
+		if err != nil {
+			if errors.Is(err, store.ErrNotFound) {
+				err = nil
+			}
+			return err
+		}
+
+		var ports []cni.PortMapping
+		if err := json.Unmarshal(data, &ports); err != nil {
+			return fmt.Errorf("failed to parse port mappings %v: %w", ports, err)
+		}
+		ns.PortMappings = ports
+
+		return err
+	})
+}
--- a/pkg/portutil/portutil.go
+++ b/pkg/portutil/portutil.go
@@ -28,6 +28,7 @@ import (
 	"github.com/containerd/log"
 
 	"github.com/containerd/nerdctl/v2/pkg/labels"
+	"github.com/containerd/nerdctl/v2/pkg/netutil/networkstore"
 	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
 )
 
@@ -139,16 +140,35 @@ func ParseFlagP(s string) ([]cni.PortMapping, error) {
 	return mr, nil
 }
 
-// ParsePortsLabel parses JSON-marshalled string from label map
-// (under `labels.Ports` key) and returns []cni.PortMapping.
-func ParsePortsLabel(labelMap map[string]string) ([]cni.PortMapping, error) {
-	portsJSON := labelMap[labels.Ports]
-	if portsJSON == "" {
-		return []cni.PortMapping{}, nil
+func GeneratePortMappingsConfig(dataStore, namespace, id string, portMappings []cni.PortMapping) error {
+	ns, err := networkstore.New(dataStore, namespace, id)
+	if err != nil {
+		return err
 	}
+	return ns.Acquire(portMappings)
+}
+
+func LoadPortMappings(dataStore, namespace, id string, containerLabels map[string]string) ([]cni.PortMapping, error) {
 	var ports []cni.PortMapping
-	if err := json.Unmarshal([]byte(portsJSON), &ports); err != nil {
-		return nil, fmt.Errorf("failed to parse label %q=%q: %s", labels.Ports, portsJSON, err.Error())
+
+	ns, err := networkstore.New(dataStore, namespace, id)
+	if err != nil {
+		return ports, err
 	}
+	if err = ns.Load(); err != nil {
+		return ports, err
+	}
+	if len(ns.PortMappings) != 0 {
+		return ns.PortMappings, nil
+	}
+
+	portsJSON := containerLabels[labels.Ports]
+	if portsJSON == "" {
+		return ports, nil
+	}
+	if err := json.Unmarshal([]byte(portsJSON), &ports); err != nil {
+		return ports, fmt.Errorf("failed to parse label %q=%q: %s", labels.Ports, portsJSON, err.Error())
+	}
+	log.L.Warnf("container %s (%s) is using legacy port mapping configuration. To ensure compatibility with the new port mapping logic, please recreate this container. For more details, see: https://github.com/containerd/nerdctl/pull/4290", containerLabels[labels.Name], id[:12])
 	return ports, nil
 }
--- a/cmd/nerdctl/compose/compose_port.go
+++ b/cmd/nerdctl/compose/compose_port.go
@@ -88,11 +88,18 @@ func portAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	dataStore, err := clientutil.DataStore(globalOptions.DataRoot, globalOptions.Address)
+	if err != nil {
+		return err
+	}
+
 	po := composer.PortOptions{
 		ServiceName: args[0],
 		Index:       index,
 		Port:        port,
 		Protocol:    protocol,
+		DataStore:   dataStore,
+		Namespace:   globalOptions.Namespace,
 	}
 
 	return c.Port(ctx, cmd.OutOrStdout(), po)
--- a/cmd/nerdctl/compose/compose_port_linux_test.go
+++ b/cmd/nerdctl/compose/compose_port_linux_test.go
@@ -20,7 +20,11 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 )
 
 func TestComposePort(t *testing.T) {
@@ -74,4 +78,43 @@ services:
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "port", "svc0", "9999").AssertFail()
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "port", "--protocol", "udp", "svc0", "10000").AssertFail()
 	base.ComposeCmd("-f", comp.YAMLFullPath(), "port", "--protocol", "tcp", "svc0", "10001").AssertFail()
+}
+
+// TestComposeMultiplePorts tests whether it is possible to allocate a large
+// number of ports. (https://github.com/containerd/nerdctl/issues/4027)
+func TestComposeMultiplePorts(t *testing.T) {
+	var dockerComposeYAML = fmt.Sprintf(`
+services:
+  svc0:
+    image: %s
+    command: "sleep infinity"
+    ports:
+    - '32000-32060:32000-32060'
+`, testutil.AlpineImage)
+
+	testCase := nerdtest.Setup()
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		compYamlPath := data.Temp().Save(dockerComposeYAML, "compose.yaml")
+		data.Labels().Set("composeYaml", compYamlPath)
+
+		helpers.Ensure("compose", "-f", compYamlPath, "up", "-d")
+	}
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("compose", "-f", data.Temp().Path("compose.yaml"), "down", "-v")
+	}
+
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "Issue #4027 - Allocate a large number of ports.",
+			NoParallel:  true,
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("compose", "-f", data.Labels().Get("composeYaml"), "port", "svc0", "32000")
+			},
+			Expected: test.Expects(expect.ExitCodeSuccess, nil, expect.Contains("0.0.0.0:32000")),
+		},
+	}
+
+	testCase.Run(t)
 }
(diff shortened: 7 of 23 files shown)


haytok:
I'll check the root cause of CI failurers ...

haytok:
Hi, @apostasie Thanks for commnets!

> Why are we not storing the range itself in the label?



I also think of a way to keep the range of ports in the label, but the following edge case would violate the upper limit of the number of labels.
Specifically, when host, port and protocol are random, it is difficult to combine ranges.

```bash
nerdctl run -p 127.0.0.1:32000:32000 -p 127.0.0.2:32001:32001 ~ -p 127.0.0.254:32253:32253 nginx
```



Considering these cases, I think that when the port mapping information is given to the Label, it would violate the upper limit, so I implemented the way of this PR so that the port mapping information is not retained in the Label.

---

When nerdctl compose up is executed, the ports specified in compose.yaml are expanded and executed by the following function: 

- https://github.com/compose-spec/compose-go/blob/main/cli/options.go#L448-L449

The ports are expanded and executed as `nerdctl run -p 32000:32000 ~ -p 32060:32060 ...`.

pkg/portutil/portutil.go
```diff
@@ -129,16 +130,56 @@ func ParseFlagP(s string) ([]cni.PortMapping, error) {
...
+	return filepath.Join(dataStore, "containers", ns, id, "port-mappings.json")
+}
+
+func GeneratePortMappingsConfig(dataStore, ns, id string, portMappings []cni.PortMapping) error {
+	portsJSON, err := json.Marshal(portMappings)
+	if err != nil {
+		return fmt.Errorf("failed to marshal port mappings to JSON: %w", err)
+	}
+	portMappingsPath := portMappingsPath(dataStore, ns, id)
+	if err := os.WriteFile(portMappingsPath, portsJSON, 0644); err != nil {
```
apostasie, line 143: This (and others) direct access to the filesystem are problematic. These writes are not atomic (eg: they could leave a broken JSON in place in some cases), and they are not safe to use concurrently.

Direct access to the filesystem should generally be avoided.

There is an abstraction: `store.Store` that does provide primitives to read and write with locking, and writes are atomic.

The `state.Store` uses it (and others, like the namestore, volumestore, etc).

You could either modify the `state.Store` to also store port information, or if necessary, create your own store abstraction.

Also, `containerutil.ContainerStateDirPath` should be used to get the root.

AkihiroSuda:
> That overall does beg the question: should we stop using labels/annotations to store network configuration altogether, and just use a filesystem store for that instead?

👍 

We have to keep existing containers functional, though, if we are going to merge this in v2.1.x.

pkg/portutil/portutil.go
haytok: Thanks for your advices!!!

I'll modify the code to use `store.Store` so that file read/write operations can be performed atomically!

https://github.com/containerd/nerdctl/blob/b8c4b3d852b8e49b59a3061f72a9064a6f78e3c5/pkg/store/store.go#L62

haytok:
Thanks for comments !!!

I understand that it is acceptable to implement a policy of storing port mapping information in a local file.
Therefore, I'll proceed with the implementation in this PR and make corrections so that file writing can be performed atomically.

haytok:
Hi @AkihiroSuda @apostasie 

Could you review when you have time ? 🙏

pkg/labels/labels.go
```diff
@@ -57,9 +57,6 @@ const (
 	// Currently, the length of the slice must be 1.
 	Networks = Prefix + "networks"
 
-	// Ports is a JSON-marshalled string of []cni.PortMapping .
-	Ports = Prefix + "ports"
```
AkihiroSuda, line 61: This change seems to break ports of containers created with an existing version of nerdctl ?

Then probably we should postpone this to v2.2.0

haytok: Thanks for comments!!!

This change is **not backward compatible**, so these backward incompatible changes should be postponed until the update from `2.1` to `2.2`. 
If so, is it okay to leave this PR open for now until the release of `2.2` approaches? 
I'll rebase as necessary and ask you to review this PR when the release of `2.2` approaches.

AkihiroSuda: Yes, please.

There should be also a human-friendly warning message when an incompatible container exists.

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -366,8 +365,8 @@ func TestRunWithInvalidPortThenCleanUp(t *testing.T) {
 			},
 			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
 				return &test.Expected{
-					ExitCode: 1,
-					Errors:   []error{errdefs.ErrInvalidArgument},
+					ExitCode: 0,
```
apostasie, line 424: So, is this test now working for both nerdctl and docker?
Should we lift the requirement?

pkg/containerutil/container_network_manager.go
```diff
@@ -892,12 +892,6 @@ func NetworkOptionsFromSpec(spec *specs.Spec) (types.NetworkOptions, error) {
 	}
 	opts.NetworkSlice = networks
 
-	if portsJSON := spec.Annotations[labels.Ports]; portsJSON != "" {
```
apostasie, line 896: PortMappings no longer set here?

pkg/netutil/networkstore/networkstore.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			err = errors.Join(ErrNetworkStore, err)
+		}
+	}()
+
+	loc, err := ns.safeStore.Location(networkConfigName)
+	if err != nil {
+		return err
+	}
+	if _, err := os.Stat(loc); err != nil {
```
apostasie: `ns.SafeStore.Exists()` instead of using os.Stat.
Also, do not ignore errors.

apostasie: And move `ns.SafeStore.Exists()` inside the locked section.

apostasie:
@haytok 

What if we make this change backward compatible?

eg: 
```go
func LoadPortMappings(dataStore, namespace, id string) ([]cni.PortMapping, error) {
      // if there is no port information from NetworkStore, look into the labels
```

apostasie:
@haytok a few comments above - also I think we should make this backward compatible (at least for reading the ports from the labels if they are there).

pkg/labels/labels.go
haytok: I'll add handling for cases where there are containers that are not backward compatible (e.g., when containers created using the -p option are already exist).

haytok:
Hi @apostasie 

Thanks for commens!!!

I hadn't considered backward compatibility at all, so I'll update based on your advices to make backward compatible.

pkg/netutil/networkstore/networkstore.go
haytok: Thanks for your advices!!! fixed.

cmd/nerdctl/container/container_run_network_linux_test.go
haytok: > So, is this test now working for both nerdctl and docker?
> Should we lift the requirement?

No, the current implementation of this test does not work with `docker`.

To begin with, it is possible to launch a container with the `-p 22200-22299:22200-22299` option in docker.

<p>

```bash
> docker --version
Docker version 25.0.8, build 0bab007

> docker run -d -p 22200-22299:22200-22299 --name alpine alpine sleep infinity
6ecd77073469ed362773eacb54a0af9e474ff2f0b58ada9553a668d98ee3ee51

> docker ps
CONTAINER ID   IMAGE     COMMAND            CREATED          STATUS          PORTS                                                                   NAMES
6ecd77073469   alpine    "sleep infinity"   14 seconds ago   Up 11 seconds   0.0.0.0:22200-22299->22200-22299/tcp, :::22200-22299->22200-22299/tcp   alpine
```

</p>
</details> 

So the scenario in this test will also work in docker.

However, in the current implementation of the test, the test relies on **nerdctl-specific** processing, as it takes into account the use of `the --data-root` option and the existence of the `etchosts` that are evaluated in the Output below, so the `testCase.Require = require. Not(nerdtest.Docker)` cannot be commented out.

<p>

```golang
func TestRunWithInvalidPortThenCleanUp(t *testing.T) {
...
			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
				return helpers.Command("run", "--data-root", data.Temp().Path(), "--rm", "-p", "22200-22299:22200-22299", testutil.CommonImage)
			},
			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
...
					Output: func(stdout string, info string, t *testing.T) {
...

						dataRoot := data.Temp().Path()
						h := getAddrHash(defaults.DefaultAddress)
						dataStore := filepath.Join(dataRoot, h)
						namespace := string(helpers.Read(nerdtest.Namespace))
						etchostsPath := filepath.Join(dataStore, "etchosts", namespace)

						etchostsDirs, err := os.ReadDir(etchostsPath)

						assert.NilError(t, err)
						assert.Equal(t, len(etchostsDirs), 0)
...
```

</p>
</details>

haytok:
I've added the logic to maintain backward compatibility.

However, maybe all failures are flaky ...

`canary` / `go canary` failures are related to #4318.

- `[flaky, see #3988] / EL / almalinux-9 rootful (pull_request) Failing after 16m`

<p>

```bash
2025-06-17T14:53:34.7593504Z     container_logs_test.go:313: 
2025-06-17T14:53:34.7593560Z         
2025-06-17T14:53:34.7593715Z         +------------------------------------------------------------------------------------------------------------+
2025-06-17T14:53:34.7594063Z         | ➡️      | ⚙️ /usr/local/bin/nerdctl logs -f testlogswithoutnewlineoreof-f625832d                           |
2025-06-17T14:53:34.7594222Z         +------------------------------------------------------------------------------------------------------------+
2025-06-17T14:53:34.7594417Z         | 🌱      | HOME=/root                                                                                       |
2025-06-17T14:53:34.7594667Z         |         | PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin |
2025-06-17T14:53:34.7594770Z         |         | :/bin                                                                                            |
2025-06-17T14:53:34.7595027Z         |         | DOCKER_CONFIG=/tmp/TestLogsWithoutNewlineOrEOF3385429920/003                                     |
2025-06-17T14:53:34.7595290Z         |         | NERDCTL_TOML=/tmp/TestLogsWithoutNewlineOrEOF3385429920/003/nerdctl.toml                         |
2025-06-17T14:53:34.7595526Z         +------------------------------------------------------------------------------------------------------------+
2025-06-17T14:53:34.7595727Z         | ⏰      | <1s (limit: 3m0s)                                                                                |
2025-06-17T14:53:34.7595888Z         +------------------------------------------------------------------------------------------------------------+
2025-06-17T14:53:34.7596214Z         | 📁      | /tmp/TestLogsWithoutNewlineOrEOF3385429920/002                                                   |
2025-06-17T14:53:34.7596372Z         +------------------------------------------------------------------------------------------------------------+
2025-06-17T14:53:34.7596434Z         
2025-06-17T14:53:34.7596977Z     container_logs_test.go:313: 	🔗 ]8;;file:///go/src/github.com/containerd/nerdctl/cmd/nerdctl/container/container_logs_test.go#313:1testCase.Run(t)]8;;[0m
2025-06-17T14:53:34.7597042Z         
2025-06-17T14:53:34.7597110Z         <<<<<<<<<<<<<<<<<<<<
2025-06-17T14:53:34.7597242Z         	🖊️ Inspecting output (equals)
2025-06-17T14:53:34.7597339Z         	👀 testing:		``
2025-06-17T14:53:34.7597450Z         	❌ FAILED!		= `'Hello World!
2025-06-17T14:53:34.7597532Z         There is no newline'`
2025-06-17T14:53:34.7597595Z         >>>>>>>>>>>>>>>>>>>>
```

</p>
</details> 

- `lint / go / linux (go canary) (pull_request) Failing after 1m`

<p>

```
Error: ../../../go/pkg/mod/golang.org/x/tools@v0.12.0/internal/tokeninternal/tokeninternal.go:78:9: invalid array length -delta * delta (constant -256 of type int64)
make: *** [Makefile:221: install-dev-tools] Error 1
Error: Process completed with exit code 2.
```

</p>
</details> 

- `test / in-container / rootful (canary) (pull_request) Failing after 5m`

<p>

```
Dockerfile:321
--------------------
 319 |     ENV PATH=/usr/local/go/bin:$PATH
 320 |     ARG GOTESTSUM_VERSION
 321 | >>> RUN GOBIN=/usr/local/bin go install gotest.tools/gotestsum@${GOTESTSUM_VERSION}
 322 |     COPY . /go/src/github.com/containerd/nerdctl
 323 |     WORKDIR /go/src/github.com/containerd/nerdctl
--------------------
ERROR: failed to solve: process "/bin/sh -c GOBIN=/usr/local/bin go install gotest.tools/gotestsum@${GOTESTSUM_VERSION}" did not complete successfully: exit code: 1
Error: Process completed with exit code 1.
```

</p>
</details> 

- `test / in-host / windows (canary) (pull_request) Failing after 3m`

<p>

```
Error: C:\Users\runneradmin\go\pkg\mod\golang.org\x\tools@v0.12.0\internal\tokeninternal\tokeninternal.go:78:9: invalid array length -delta * delta (constant -256 of type int64)
make: *** [Makefile:221: install-dev-tools] Error 1
Error: Process completed with exit code 2.
```

</p>
</details>

apostasie:
Yes, all `invalid array length -delta * delta (constant -256 of type int64)` you can ignore.

`TestLogsWithoutNewlineOrEOF` is likely #4147

pkg/portutil/portutil.go
```diff
@@ -129,16 +130,34 @@ func ParseFlagP(s string) ([]cni.PortMapping, error) {
...
+	}
+
+	if portsJSON == "" {
+		return ports, nil
+	}
 	if err := json.Unmarshal([]byte(portsJSON), &ports); err != nil {
-		return nil, fmt.Errorf("failed to parse label %q=%q: %s", labels.Ports, portsJSON, err.Error())
+		return ports, fmt.Errorf("failed to parse label %q=%q: %s", labels.Ports, portsJSON, err.Error())
 	}
+	log.L.Warnf("container %s is using legacy port mapping configuration. To ensure compatibility with the new port mapping logic, please recreate this container. For more details, see: https://github.com/containerd/nerdctl/pull/4290", id[:12])
```
apostasie: Maybe we actually do not need this warning?
We **are** compatible with legacy port mapping, so, it is transparent for users - eg: there is no real reason to urge them to recreate containers until we break compatibility / remove the compat layer in 2.X.

Let see what others think about this though ^

pkg/netutil/networkstore/networkstore.go
```diff
@@ -0,0 +1,110 @@
...
+func (ns *NetworkStore) Load() (err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Join(ErrNetworkStore, err)
+		}
+	}()
+
+	return ns.safeStore.WithLock(func() error {
+		doesExist, err := ns.safeStore.Exists(networkConfigName)
+		if err != nil {
```
apostasie: Nit:
```golang
if err != nil || !doesExist {
  return err
}
```

pkg/netutil/networkstore/networkstore.go
```diff
@@ -0,0 +1,110 @@
...
+			return err
+		}
+		if !doesExist {
+			return nil
+		}
+
+		data, err := ns.safeStore.Get(networkConfigName)
+		if err == nil {
```
apostasie: Nit:

Personally, I find it more readable to first deal with error conditions (and return).
Then the rest does not have to be inside an if/then test.

Eg:

```golang

if err != nil {
  if errors.Is(err, store.ErrNotFound) {
    err = nil
  }
  return err
}

var ports []cni.PortMapping
if err := json.Unmarshal(data, &ports); err != nil {
  return fmt.Errorf("failed to parse port mappings %v: %w", ports, err)
}
ns.PortMappings = ports

return nil

```

This is really personal preference, so, if you don't like it, fine :)

apostasie:
Thanks @haytok - this is good progress on solving this.

1. I have a couple of (minor) nits, but otherwise looks good
2. I am still annoyed at the fact we are unable to store internally port ranges
eg:

```
nerdctl ps
CONTAINER ID    IMAGE                              COMMAND             CREATED          STATUS    PORTS                                                                                                                                                                                                                                                                                           NAMES
93a2736dd462    docker.io/library/alpine:latest    "sleep infinity"    3 seconds ago    Up        0.0.0.0:22200->22200/tcp, 0.0.0.0:22201->22201/tcp, 0.0.0.0:22202->22202/tcp, 0.0.0.0:22203->22203/tcp, 0.0.0.0:22204->22204/tcp, 0.0.0.0:22205->22205/tcp, 0.0.0.0:22206->22206/tcp, 0.0.0.0:22207->22207/tcp, 0.0.0.0:22208->22208/tcp, 0.0.0.0:22209->22209/tcp, 0.0.0.0:22210->22210/tcp    alpine
```

This is really ugly ^ - docker does a much better job.

I appreciate this aspect would be more complicated and possibly involve additional changes in compose, but yeah, I would much prefer that:
`nerddctl -p 22200-22210:22200-22210` to be _stored_ as ranges and not expanded.

apostasie:
Note, I am not suggesting that we try to reconstitute ranges from discrete ports specs, but that we do store the ranges when ranges are provided

pkg/netutil/networkstore/networkstore.go
haytok: Thanks !!! fixed.

pkg/netutil/networkstore/networkstore.go
haytok: The code you suggested is easier to read, so I've fixed based on your advice! Thank you!

pkg/portutil/portutil.go
haytok: Based on the advice I received below, this warning message is displayed.

- https://github.com/containerd/nerdctl/pull/4290#discussion_r2132633766

If at some stage (release) the logic for obtaining port mapping information from the label is to be deprecated, I think this warning message should be displayed.

However, I would like to hear the opinions of other maintainers as well.

haytok:
Thanks for comments!!! @apostasie 

I also think the `PORTS` display in `nerdctl ps` and `nerdctl compose ps` results is hard to read... :(

First, I believe this is caused by the current implementation where the logic for displaying `PORTS` cannot aggregate ranges.

https://github.com/containerd/nerdctl/blob/b60fbeaad1fb4c43ab29ea4751d78d6a4a735686/pkg/formatter/formatter.go#L115-L128

I think this behavior occurs even in the **latest** version, but to fundamentally fix this readability issue, rather than saving ranges when port ranges are specified in `-p`, I think rewriting `FormatPorts()` would lead to a more fundamental solution to the problem.

For example, when we specify port ranges in `compose.yaml` as shown below, each port mapping gets expanded to `-p` and `nerdctl run` is executed. 
Therefore, whether ranges are specified in `-p` or multiple `-p` options are specified, it would be better if `PORTS` could display them as ranges.

<p>

```bash
> cat compose.yaml
services:
  nginx:
    image: nginx:latest
    container_name: nginx
    ports:
      - '80-81:80-81'
      - '80-81:80-81/udp'

> sudo nerdctl compose up
...
INFO[0000] Running [/usr/local/bin/nerdctl run --cidfile=/tmp/compose-2285964703/cid -l=com.docker.compose.project=ec2-user -l=com.docker.compose.service=nginx -d --name=nginx --pull=never --net=ec2-user_default --hostname=nginx -p=80:80/tcp -p=81:81/tcp -p=80:80/udp -p=81:81/udp --restart=no nginx:latest]
...
```

</p>
</details> 

However, I think fixing this `PORTS` display is beyond the scope of this PR, so how about creating a separate issue and addressing it there? 
I will continue to work on it. 
If it would be better to add the logic for merging `PORTS` display to this PR's commit, I can handle it here!

pkg/portutil/portutil.go
apostasie: I think the advice was because at that time your PR was _not_ backward compatible.
Since you did make it backward compatible now, it is less useful.

Anyhow, I don't care, this is a nit, so, fine with it :-)

apostasie:
> However, I think fixing this `PORTS` display is beyond the scope of this PR, so how about creating a separate issue and addressing it there? I will continue to work on it. If it would be better to add the logic for merging `PORTS` display to this PR's commit, I can handle it here!

Yeah, that works too.
Also, _if_ we decide to store ranges in certain cases in the future, it should be backward compatible, so, we can look into this later.

Thanks @haytok

AkihiroSuda:
Needs rebase

haytok:
Rebased. Thanks!

AkihiroSuda:
Sorry, needs rebase again

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -421,8 +420,8 @@ func TestRunWithInvalidPortThenCleanUp(t *testing.T) {
...
```
AkihiroSuda, line 424: "Run a container with invalid ports, and then clean up." is no longer true?
Same applies to the test name too.

pkg/portutil/portutil.go
```diff
@@ -139,16 +140,34 @@ func ParseFlagP(s string) ([]cni.PortMapping, error) {
...
```
AkihiroSuda: The name should be printed in addition to the hex ID?

cmd/nerdctl/container/container_run_network_linux_test.go
haytok: Thanks, fixed!

pkg/portutil/portutil.go
haytok: Thanks, fixed!

haytok:
I've resolved conflict, but some tests have failed ...
I'll check whether similar failures are or not.

- `test / in-container / rootful (arm) (pull_request)[test / in-container / rootful (arm) (pull_request) Failing after 45m`

<p>

```
2025-06-25T10:03:00.8563767Z === RUN   TestComposeUpWithExternalNetwork
2025-06-25T10:03:00.8564720Z     compose_up_linux_test.go:425: assertion failed: 
2025-06-25T10:03:00.8565831Z         Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test exec nerdctl-testcomposeupwithexternalnetwork-1 wget -qO- http://nerdctl-testcomposeupwithexternalnetwork-2
2025-06-25T10:03:00.8566488Z         ExitCode: 1
2025-06-25T10:03:00.8566688Z         Error:    exit status 1
2025-06-25T10:03:00.8566874Z         Stdout:   
2025-06-25T10:03:00.8567116Z         Stderr:   wget: can't connect to remote host (10.4.1.4): Connection refused
2025-06-25T10:03:00.8567504Z         time="2025-06-25T10:02:59Z" level=fatal msg="exec failed with exit code 1"
2025-06-25T10:03:00.8568061Z         
2025-06-25T10:03:00.8568192Z         
2025-06-25T10:03:00.8568329Z         Failures:
2025-06-25T10:03:00.8568489Z         ExitCode was 1 expected 0
2025-06-25T10:03:00.8568757Z         Expected stdout to contain "<title>Welcome to nginx!</title>"
2025-06-25T10:03:00.8569050Z         Expected no error
2025-06-25T10:03:00.8569270Z --- FAIL: TestComposeUpWithExternalNetwork (2.91s)

...

2025-06-25T10:17:40.2864006Z     ipfs_compose_linux_test.go:287: 	🔗 ]8;;file:///go/src/github.com/containerd/nerdctl/cmd/nerdctl/ipfs/ipfs_compose_linux_test.go#146:1composeUP(data, helpers, fmt.Sprintf(`]8;;[0m
2025-06-25T10:17:40.2864620Z         
2025-06-25T10:17:40.2864768Z         <<<<<<<<<<<<<<<<<<<<
2025-06-25T10:17:40.2864996Z         	🖊️ Command should succeed
2025-06-25T10:17:40.2865532Z         	👀 testing:		`command returned a non-zero exit code`
2025-06-25T10:17:40.2865908Z         	❌ FAILED!		is `<nil>`
2025-06-25T10:17:40.2866093Z         >>>>>>>>>>>>>>>>>>>>
```

</p>
</details> 

- `test / in-container / rootless-port-slirp4netns (pull_request) Failing after 13m`

<p>

```bash
2025-06-25T09:51:34.1214885Z     image_push_linux_test.go:282: 	
2025-06-25T09:51:34.1215101Z         
2025-06-25T09:51:34.1215368Z         <<<<<<<<<<<<<<<<<<<<
2025-06-25T09:51:34.1215618Z         	🖊️ Command should succeed
2025-06-25T09:51:34.1216106Z         	👀 testing:		`command returned a non-zero exit code`
2025-06-25T09:51:34.1216414Z         	❌ FAILED!		is `<nil>`
2025-06-25T09:51:34.1216606Z         >>>>>>>>>>>>>>>>>>>>
```

</p>
</details>

AkihiroSuda (approved):
Thanks

apostasie:
🎉

haytok:
Thanks for your reviews and CI retries!!!

PR #4293, opened 2025-05-26
fix: improve nsenter command handling for rootless containers

fahedouch:
The `containerd-rootless-setuptool.sh nsenter` command does not allow direct access to the network namespace. You must first access the user namespace and then the network namespace. Let me fix this beforehand.


AkihiroSuda (changes requested):
The default behavior should not be changed

fahedouch:
> The default behavior should not be changed

the default behavior is not changing, it is

> The default behavior should not be changed

Sorry, I misread the comments. Let me close this

Issue #4157, opened 2025-04-24
Add Health Check Support in nerdctl


subashkotha:
### What is the problem you're trying to solve

Docker provides built-in support for container health checks through its daemon, allowing users to monitor container health and take automated actions like restarts. nerdctl currently lacks native health check support, which limits its ability to support similar workflows—especially when running containers defined via Dockerfiles or Compose files that include health check instructions. This feature aims to close that gap by enabling health checks in a way that aligns with nerdctl’s daemonless design.

### Describe the solution you'd like

We’d like to propose adding health check support in nerdctl, similar to what Docker offers, but designed to work in nerdctl’s daemonless architecture. This feature will enable users to define and run application-level health checks using:
1. Docker-compatible CLI flags (e.g., --health-cmd, --health-interval, etc.)
2. Health check configuration defined in Dockerfiles (HEALTHCHECK instruction)
3. Health checks defined in Compose files

As part of this support, we will introduce a new subcommand:
```
nerdctl healthcheck run <container>
```
This command acts as the entry point for performing the health check—it runs the configured health check command inside the container and determines the result.

Since nerdctl does not have a persistent daemon like Docker, we use systemd to invoke this command at the configured intervals, effectively handling the scheduling aspect externally.

Related discussion - https://github.com/containerd/nerdctl/discussions/4032

swagatbora90:
Related #2386, https://github.com/runfinch/finch/issues/494

apostasie:
Is this complete with #4302?

swagatbora90:
[#4302] only adds support for healthcheck option in nerdctl run, nerdctl create, and introduces a new nerdctl container healthcheck command. However, periodic execution of health checks against running containers is still WIP. 

This issue tracks the overall healthcheck support in nerdctl, so we should keep it open.

swagatbora90:
This issue can be closed now since both the healthcheck command implementation and orchestration PRs are merged now.

tom-rm-meyer-ISST:
> This issue can be closed now since both the healthcheck command implementation and orchestration PRs are merged now.

Cool, if I see things correctly, this is part of version [2.1.3](https://github.com/containerd/nerdctl/releases/tag/v2.1.3), isn't it?

swagatbora90:
Healthcheck command was added in 2.1.3. Orchestration PR was recently merged https://github.com/containerd/nerdctl/pull/4427 and will be available in the next 2.1.5 release.

PR #4302, opened 2025-05-29
feat: Add healthcheck command in nerdctl

subashkotha:
This PR adds support for Docker-compatible health checks in `nerdctl run`, `nerdctl create`, and introduces a new `nerdctl container healthcheck` command to manually trigger health checks.

Key features:

1. Health check configuration via CLI flags (--health-cmd, --health-interval, --health-timeout, etc.)
2. Support for health checks defined in Docker images (HEALTHCHECK in the Dockerfile)
3. Merging logic for health checks:
  * CLI flags override image-defined health checks
  * Fallback to image-defined health checks if CLI flags not set
  *  No health check configured by default
4. `nerdctl container healthcheck` command to manually run health checks
5. Health check configuration is stored as an internal label `nerdctl/healthcheck`
6. Health state (status and failing streak) is stored in internal label `nerdctl/healthstate` for quick access 
7. Health check results stored in a `health.json` file in the container’s runtime state directory
8. Tests added for CLI flag parsing, merging behavior, execution of health checks, and health state inspection in nerdctl inspect

Future work (WIP):
1. Automate periodic health checks using systemd.
2. Generate systemd timer and service units based on container lifecycle events.
3. Improve logic to store/fetch health check results. 

Related issue - https://github.com/containerd/nerdctl/issues/4157

cc: @Shubhranshu153 @AkihiroSuda


pkg/healthcheck/log.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	data, err := json.MarshalIndent(result, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal health log: %w", err)
+	}
+
+	path := filepath.Join(stateDir, HealthLogFilename)
+
+	return os.WriteFile(path, data, 0o600)
```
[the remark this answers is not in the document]
apostasie: There is a `state.Store` already that could be used / expanded to manage that log file as well.

cmd/nerdctl/container/container_health_check_test.go
```diff
@@ -0,0 +1,375 @@
...
+	"github.com/containerd/nerdctl/mod/tigron/test"
+
+	"github.com/containerd/nerdctl/v2/pkg/healthcheck"
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
+)
+
+func TestContainerHealthCheckBasic(t *testing.T) {
+	testCase := nerdtest.Setup()
+	testutil.DockerIncompatible(t)
```
AkihiroSuda, line 37:
```suggestion
        // (Reason goes here)
	testCase.Require = require.Not(nerdtest.Docker)
```

pkg/cmd/container/create.go
```diff
@@ -828,9 +840,108 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
...
+			options.HealthStartInterval != 0
+
+	if options.NoHealthcheck {
+		if options.HealthCmd != "" || healthFlagsSet {
+			return fmt.Errorf("--no-healthcheck conflicts with --health-* options")
+		}
+	}
+
+	// Note: HealthCmd can be empty with other healthcheck flags set cause healthCmd could be coming from image.
+	if options.HealthInterval < 0 {
```
Shubhranshu153, line 926: these can be checked when the params are passed in cli?

pkg/cmd/container/health_check.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	if status.Status != containerd.Running {
+		return nil, fmt.Errorf("container is not running (status: %s)", status.Status)
+	}
+
+	return task, nil
+}
+
+// If configuredValue is zero, use defaultValue instead.
+func timeoutWithDefault(configuredValue time.Duration, defaultValue time.Duration) time.Duration {
```
Shubhranshu153, line 91: feel the defaults should be part of container config so when you inspect you get the values for it. 
This seems too down the stream.

cmd/nerdctl/container/container_health_check.go
```diff
@@ -0,0 +1,85 @@
...
+	n, err := walker.Walk(ctx, containerID)
+	if err != nil {
+		return err
+	} else if n == 0 {
+		return fmt.Errorf("no such container %s", containerID)
+	}
+	return nil
+}
+
+func healthCheckShellComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
Shubhranshu153, line 81: can we not have this shell completion, i feel this command is not really going to be used in a shell rather part of systemd command.

pkg/cmd/container/create.go
```diff
@@ -828,9 +840,108 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
...
+			}
+			hc = parsed
+		}
+	}
+
+	// Apply CLI overrides
+	if options.HealthCmd != "" {
+		hc.Test = []string{"CMD-SHELL", options.HealthCmd}
+	}
+	if options.HealthInterval != 0 {
```
Shubhranshu153, line 884: if == 0 can it not be overridden?

pkg/healthcheck/health.go
```diff
@@ -0,0 +1,95 @@
...
+type HealthcheckResult struct {
+	Start    time.Time // Start is the time this check started
+	End      time.Time // End is the time this check ended
+	ExitCode int       // ExitCode meanings: 0=healthy, 1=unhealthy, 2=reserved (considered unhealthy), else=error running probe
+	Output   string    // Output from last check
+}
+
+// Healthcheck represents the health check configuration
+type Healthcheck struct {
+	Test          []string      `json:"Test,omitempty"`          // Test is the test to perform to check that the container is healthy
```
Shubhranshu153, line 59: nit: Test -> Check

pkg/healthcheck/executor.go
```diff
@@ -0,0 +1,200 @@
...
+			Output:   outputBuf.String(),
+			End:      time.Now(),
+		}, nil
+	}
+}
+
+// updateHealthStatus updates the health status based on the health check result
+func updateHealthStatus(ctx context.Context, container containerd.Container, hcConfig *Healthcheck, hcResult *HealthcheckResult) error {
+	// Get current health status from health log
+	currentHealth, err := readHealthLog(ctx, container)
```
Shubhranshu153, line 120: reading logs to get current health seems not so robust. i think we should atleast check the possibility of having a db. 
@AkihiroSuda thoughts? or having a db for it is over engineering the problem?

AkihiroSuda: How could the DB improve the robustness?

pkg/imgutil/imgutil.go
```diff
@@ -464,3 +473,28 @@ func GetDanglingImages(ctx context.Context, client *containerd.Client, filters .
 
 	return ApplyFilters(allImages, filters...)
 }
+
+// addHealthCheckToImageConfig extracts health check information from the image content store and adds it to the labels
+func addHealthCheckToImageConfig(rawConfigContent []byte, config *ocispec.ImageConfig) error {
```
Shubhranshu153, line 478: Do we need to do this one in the sense its already part of rawConfigContent so when we require it we can have a function that extracts it.

cmd/nerdctl/container/container_health_check_test.go
subashkotha: Ack, will do.

cmd/nerdctl/container/container_health_check.go
subashkotha: You're right, it's not critical for this command. But having shell completion helps a bit for testing, so if it's okay can we keep it for now. We can remove it later if it feels unnecessary.

pkg/cmd/container/create.go
```diff
@@ -828,9 +840,108 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
...
+	}
+	if options.HealthStartPeriod != 0 {
+		hc.StartPeriod = options.HealthStartPeriod
+	}
+	if options.HealthStartInterval != 0 {
+		hc.StartInterval = options.HealthStartInterval
+	}
+
+	// If no effective health check is configured, return nil
+	if hc.Test == nil && hc.Interval == 0 && hc.Timeout == 0 && hc.Retries == 0 && hc.StartPeriod == 0 && hc.StartInterval == 0 {
```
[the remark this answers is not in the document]
subashkotha: Default values will be populated at runtime when performing the actual healthcheck

pkg/cmd/container/create.go
```diff
@@ -828,9 +840,108 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
...
+
+func validateHealthcheckFlags(options types.ContainerCreateOptions) error {
+	healthFlagsSet :=
+		options.HealthInterval != 0 ||
+			options.HealthTimeout != 0 ||
+			options.HealthRetries != 0 ||
+			options.HealthStartPeriod != 0 ||
+			options.HealthStartInterval != 0
+
+	if options.NoHealthcheck {
```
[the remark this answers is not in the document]
subashkotha: We still need to persist `NONE` as part of the health check config, so that when the user inspects, they see `NONE` in the health check configuration (similar to Docker). This way, the user can confirm that they explicitly disabled health checks using CLI flags, even if the underlying image has health checks configured.

pkg/cmd/container/create.go
subashkotha: Thats right, will move the validation logic to create

pkg/cmd/container/health_check.go
subashkotha: No, we intentionally don't store defaults in the container config. Storing defaults when they haven't been explicitly set can lead to confusion during inspect, users might think those values were explicitly configured. This approach also aligns with Docker's behavior, where defaults are applied at runtime, not persisted in the container config.

pkg/imgutil/imgutil.go
subashkotha: I think it makes it cleaner to add health check info to the image config during ensure image. Since the health check field is not present in the OCI image spec, we store it in labels. This also ensures consistency and ease of access for health config during container create/run for merging health info and when inspecting images.

docs/healthchecks.md
```diff
@@ -0,0 +1,49 @@
...
+configured health check inside the container and reports the result. It serves as the entry point for executing
+health checks, especially in scenarios where external scheduling is used.
+
+Example:
+
+```
+nerdctl container healthcheck <container-id>
+```
+
+### Future Work (WIP)
```
swagatbora90, line 45: Since the documentation states that `nerdctl` supports Docker-compatible health checks for containers, the orchestraction of health checks should be included as part of this change and not as Future work. Otherwise we should update the documentation to say that this PR only adds support for manual healthchecks on container.

docs/command-reference.md
```diff
@@ -321,6 +321,16 @@ Metadata flags:
 - :whale: :blue_square: `--cidfile`: Write the container ID to the file
 - :nerd_face: `--pidfile`: file path to write the task's pid. The CLI syntax conforms to Podman convention.
 
+Health check flags:
+
+- :whale: :blue_square: `--health-cmd`: Command to run to check container health
+- :whale: :blue_square: `--health-interval`: Time between running the check (e.g., 30s, 1m)
+- :whale: :blue_square: `--health-timeout`: Time to wait before considering the check failed (e.g., 5s)
+- :whale: :blue_square: `--health-retries`: Number of failures before container is considered unhealthy
+- :whale: :blue_square: `--health-start-period`: Time to wait after container starts before running the first check (e.g., 10s)
```
swagatbora90: In docker and podman, start-period does not actually delay health checks, instead it delays reporting failures on health checks and starting the retry count. If we are not running health checks during start period, I don't see the value of having a `--health-start-interval`

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -629,6 +648,15 @@ func ImageFromNative(nativeImage *native.Image) (*Image, error) {
 		ExposedPorts: portSet,
 	}
 
+	// Add health check if present in labels
+	if healthStr, ok := imgOCI.Config.Labels[labels.HealthCheck]; ok && healthStr != "" {
```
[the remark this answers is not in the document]
subashkotha: OCI image spec doesn’t include a healthcheck field, that’s why we add the healthcheck info to labels when ensuring the image and during docker-compat inspect we add it to image config.

Not sure if I got your question right

cmd/nerdctl/container/container_health_check.go
Shubhranshu153: my concern is with adding overhead, may be its not much of a concern if the frequency is low but if the health check frequency is high it might be as it is something that can run continually.

pkg/cmd/container/health_check.go
Shubhranshu153: does docker inspect if health check is configured returns 0 for mandatory params?

pkg/imgutil/imgutil.go
Shubhranshu153: my concern is its mutating the image config when its not really required to do it. 
Lets say somebody uses the same image labels then we might override them, unlike container labels these are not marked as internal labels.

pkg/cmd/container/create.go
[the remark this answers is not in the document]
Shubhranshu153: i dont understand this part what its checking in the sense why everything needs to be 0 for this to be correct?

pkg/cmd/container/health_check.go
subashkotha: Docker inspect will not contain hc params that are not configured. 

Example - 
```
{
  "Test": [
    "CMD-SHELL",
    "curl -f http://localhost || exit 1"
  ]
}
```

cmd/nerdctl/container/container_health_check.go
subashkotha: Makes sense, I'll remove it before merging the PR.

docs/command-reference.md
subashkotha: We do run health-checks during start period, but as you mentioned we just ignore the failures.

`--health-start-interval` will be used to define the time between running the check during the start period [docker docs](https://docs.docker.com/reference/cli/docker/container/run/), [Podman docs](https://docs.podman.io/en/stable/markdown/podman-create.1.html#health-startup-interval-interval)

pkg/cmd/container/create.go
subashkotha: Docker doesn't allow overriding healthcheck related fields with 0 values, in-case user passes in 0 values, docker defaults to using those values from image if present. Ensuring the same behavior in nerdctl as well.

pkg/cmd/container/create.go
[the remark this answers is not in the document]
subashkotha: If no healthcheck related fields are present in the image or the CLI flags during create/run, we return empty string here so we don't add healthcheck to container config.

docs/command-reference.md
swagatbora90: Okay, in that case this is incorrect `:whale: :blue_square: `--health-start-period`: Time to wait after container starts before running the first check (e.g., 10s)`

pkg/healthcheck/executor.go
apostasie: > db would support atomicity of the writes.

There is an effort to consolidate atomic, concurrency-safe filesystem operations in `internal/filesystem`. If needed, these can be used instead of filesystem access in almost all cases.

Generally IMHO we should avoid using the filesystem as an API and instead rely on a higher-level abstraction (eg: `store.Store` for eg). Leaking implementation details into the consumer (filesystem or db) is bad API, and instead having a storage API will allow swapping out the fs implementation to something else, if need-be.

docs/command-reference.md
subashkotha: Ack, will fix the comment.

pkg/healthcheck/executor.go
subashkotha: Thanks for the feedback. I'll update the PR to store current health state and failing streak in the container labels, so we no longer need to read the log file during each health check to get current state. 

Health check results are still written to a log file, and during inspect we fetch the last five entries to provide recent history. For `health.json` operations, we’ll use internal/filesystem along with proper locking as recommended to ensure concurrency safety.

pkg/healthcheck/log.go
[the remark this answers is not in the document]
subashkotha: Ack, I'll switch to using internal/filesystem with locking to ensure atomic and concurrency-safe read/write operations.

docs/healthchecks.md
subashkotha: Makes sense, will update the doc accordingly.

Shubhranshu153:
LGTM. Lets fix the windows test.

pkg/cmd/container/create.go
```diff
@@ -828,9 +840,108 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
...
+		if err != nil {
+			return "", fmt.Errorf("failed to serialize disabled healthcheck config: %w", err)
+		}
+		return hcJSON, nil
+	}
+
+	// Start with health checks in image if present
+	hc := &healthcheck.Healthcheck{}
+	if ensuredImage != nil && ensuredImage.ImageConfig.Labels != nil {
+		if label := ensuredImage.ImageConfig.Labels[labels.HealthCheck]; label != "" {
```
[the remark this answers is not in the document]
swagatbora90: Looks like the ImageConfig in the ensuredImage is read from [getImageConfig](https://github.com/containerd/nerdctl/blob/221e3e009174bb1ec19a7a9e2867eab705a3787d/pkg/imgutil/imgutil.go#L258) which unmarshals the config blob into ocispec.ImageConfig type. In order to retrieve the Healthcheck config, can we not use[DockerOCIImageConfig](https://github.com/moby/docker-image-spec/blob/main/specs-go/v1/image.go#L20C6-L20C26) instead?

pkg/cmd/container/health_check.go
```diff
@@ -0,0 +1,96 @@
...
+	containerd "github.com/containerd/containerd/v2/client"
+
+	"github.com/containerd/nerdctl/v2/pkg/healthcheck"
+	"github.com/containerd/nerdctl/v2/pkg/labels"
+)
+
+// HealthCheck executes the health check command for a container
+func HealthCheck(ctx context.Context, client *containerd.Client, container containerd.Container) error {
+	// verify container status and get task
+	task, err := verifyContainerStatus(ctx, container)
```
swagatbora90: nit: perhaps a better name will be`isContainerRunning()` since it specifically checks for the Running state.

pkg/cmd/container/create.go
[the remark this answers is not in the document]
subashkotha: The ImageConfig in the ensured image is of type `ocispec.ImageConfig`, which doesn’t include healthcheck field. While we could unmarshal the raw config using `DockerOCIImageConfig` to extract healthcheck data, that structure is not used to construct the ensured image. Instead, we extract the healthcheck config separately using `*healthcheck.Healthcheck` and embed it in the image labels. So even though ImageConfig itself doesn’t carry the healthcheck field, the relevant info is still preserved and accessible through labels.

pkg/healthcheck/log.go
```diff
@@ -0,0 +1,261 @@
...
+	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/log"
+
+	"github.com/containerd/nerdctl/v2/pkg/internal/filesystem"
+	"github.com/containerd/nerdctl/v2/pkg/labels"
+)
+
+// writeHealthLog writes the latest health check result to the log file, appending it to existing logs.
+func writeHealthLog(ctx context.Context, container containerd.Container, result *HealthcheckResult) error {
+	stateDir, err := getContainerStateDir(ctx, container)
```
swagatbora90, line 40: Instead of the writing the log file to the container state directory, should we create a health specific store? That way we avoid unnecessary locking the state dir, preventing other threads to write to it.

pkg/healthcheck/log.go
```diff
@@ -0,0 +1,261 @@
+/*
```
swagatbora90, line 1: Nit: calling it log.go is not very accurate. May be we can split the log specific methods from the containerd metadata update methods into their own files.

pkg/cmd/container/create.go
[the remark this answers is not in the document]
swagatbora90: Is there a reason to preserve the healthcheck in the image labels. From what I understand we only need to check the healthcheck info once during container create/run, can we not do it by just implementing  and calling a new `getImageConfigWithHealthCheck`. Seems simpler this way.

pkg/cmd/container/health_check.go
subashkotha: Ack, will rename.

pkg/cmd/container/create.go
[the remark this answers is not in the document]
subashkotha: Refactoring to introduce `getImageConfigWithHealthCheck()` is turning out to be a bit tricky. The main issue is that EnsuredImage doesn't retain enough context specifically, it doesn't store `ImageConfigDesc`, which is needed to locate and unmarshal the raw image config blob.

Since EnsureImage is the only thing available during both create/run and inspect (thats where we need healthcheck config), we'd need to thread additional data through or change its structure, which adds complexity.

Shubhranshu153:
For the canary errors: compiler issue which the canary is running with go1.25rc. More details available in link
https://github.com/google/go-licenses/issues/312

Shubhranshu153:
LGTM. Once the fixes are merged we can rebase and do a rebase and run the tests before merge.
Open PR for fixing the failing tests: https://github.com/containerd/nerdctl/pull/4318

@AkihiroSuda can we get a +1 for this. Thank You

swagatbora90 (approved):

pkg/healthcheck/executor.go
```diff
@@ -0,0 +1,200 @@
...
+
+	exitStatusC, err := process.Wait(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to wait for health check: %w", err)
+	}
+
+	select {
+	case <-time.After(hc.Timeout):
+		_ = process.Kill(ctx, syscall.SIGKILL)
+		go func() { <-exitStatusC }()
```
AkihiroSuda, line 92: This doesn't do anything after waiting for the channel?

pkg/inspecttypes/dockercompat/dockercompat_test.go
```diff
@@ -41,6 +47,16 @@ func TestContainerFromNative(t *testing.T) {
 	os.WriteFile(filepath.Join(tempStateDir, "resolv.conf"), []byte(""), 0644)
 	defer os.RemoveAll(tempStateDir)
 
+	hc := &healthcheck.Healthcheck{
+		Test:        []string{"CMD-SHELL", "curl -f http://localhost || exit 1"},
+		Interval:    30000000000,
```
AkihiroSuda: Can't we use `time.Second` here?

pkg/healthcheck/log.go
```diff
@@ -0,0 +1,255 @@
...
+	healthLogPath := filepath.Join(stateDir, HealthLogFilename)
+
+	// Ensure container state directory exists
+	if _, err := os.Stat(stateDir); os.IsNotExist(err) {
+		return fmt.Errorf("container state directory does not exist: %s", stateDir)
+	}
+
+	// Create health.json if it doesn't exist
+	if _, err := os.Stat(healthLogPath); os.IsNotExist(err) {
+		return os.WriteFile(healthLogPath, []byte{}, 0600)
```
apostasie: @subashkotha can you rebase and replace `os.WriteFile` with `filesystem.WriteFile`?
Thanks!

pkg/inspecttypes/dockercompat/dockercompat_test.go
subashkotha: Sure, these durations are stored as nanoseconds in container labels internally, but I'll express the flag parameters in seconds for improved readability.

pkg/healthcheck/executor.go
subashkotha: Correct, it doesn't do anything with the value it receives from the channel, it simply drains the channel to prevent a goroutine leak after the process is killed, allowing the goroutine that sends to the channel to complete.

pkg/healthcheck/log.go
subashkotha: Ack, rebased with suggested changes.

subashkotha:
Hi @AkihiroSuda @Shubhranshu153
Can I get approval on this PR if there are no comments, thanks!

AkihiroSuda (approved):
Thanks

PR #4305, opened 2025-06-04
feat: add support for DockerfileInline

### LINKED ISSUE #4108: `nerdctl compose` does not support `dockerfile_inline`

The `compose` specification contains a [`dockerfile_inline`](https://github.com/compose-spec/compose-spec/blob/main/build.md#dockerfile_inline) feature that is used instead of an external `dockerfile`. Currently this is not supported in `nerdctl`.

Added support for the feature.

### Additional context

Given the `compose.yaml`:
```yaml
services:
  foo:
    build:
      dockerfile_inline: |
        FROM scratch
```
`nerdctl compose build` says:
```
WARN[0000] Ignoring: build: [DockerfileInline]          
INFO[0000] Building image a-foo                         
FATA[0000] lstat /a/Containerfile: no such file or directory 
FATA[0000] error while building image a-foo: exit status 1 
Error: exit status 1
```

nerdctl version 2.0.4

tushar5526:
Fixes #4108

As nerdctl currently uses "nerdctl build" to build the service images, write the inline file to a temporary file and use "-f" to specify the temporary dockerfile.

### DIFF
--- a/pkg/composer/serviceparser/build.go
+++ b/pkg/composer/serviceparser/build.go
@@ -19,6 +19,7 @@ package serviceparser
 import (
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -34,7 +35,7 @@ import (
 
 func parseBuildConfig(c *types.BuildConfig, project *types.Project, imageName string) (*Build, error) {
 	if unknown := reflectutil.UnknownNonEmptyFields(c,
-		"Context", "Dockerfile", "Args", "CacheFrom", "Target", "Labels", "Secrets", "AdditionalContexts",
+		"Context", "Dockerfile", "Args", "CacheFrom", "Target", "Labels", "Secrets", "DockerfileInline", "AdditionalContexts",
 	); len(unknown) > 0 {
 		log.L.Warnf("Ignoring: build: %+v", unknown)
 	}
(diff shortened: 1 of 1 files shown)


tushar5526:
I was a bit confused about where to place this logic: writing the inline file to a temporary path and then modifying the build command to build using that file. 

It seems like `Build` object on `Service` is something that is meant to be reusable, so I placed the logic of temporary file generation at the time of parsing the service and generating the `Build` config.

tushar5526:
@apostasie, can I get a cursory review of the approach?

apostasie:
Hey @tushar5526 
Looks fine to me.

You do need to make sure that dockerfile and dockerfileinline cannot be specified together though:
https://docs.docker.com/reference/compose-file/build/

tushar5526:
It looks like `composecli` package "github.com/compose-spec/compose-go/v2/cli" is already doing that job for me.

AkihiroSuda (changes requested):
Please squash and sign off the commits.
(make sure that the `Signed-off-by: NAME <EMAIL>` line with your real name is included in the commit message)

tushar5526:
Done, thanks!

AkihiroSuda:
CI is failing
https://github.com/containerd/nerdctl/actions/runs/15929816575/job/44936014200?pr=4305

> FAIL - has whitespace errors. See `git show --check 5651e76d85e12092b65f3e35b8994d2244e35698`.

tushar5526:
I think its good to go now.

pkg/composer/serviceparser/build.go
```diff
@@ -86,6 +87,20 @@ func parseBuildConfig(c *types.BuildConfig, project *types.Project, imageName st
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create temp file for DockerfileInline: %w", err)
+		}
+		defer tmpFile.Close()
+		if _, err := tmpFile.Write([]byte(c.DockerfileInline)); err != nil {
+			return nil, fmt.Errorf("failed to write DockerfileInline: %w", err)
+		}
+		b.BuildArgs = append(b.BuildArgs, "-f="+tmpFile.Name())
+	}
+
```
AkihiroSuda: The tmp file has to be removed on completion of the build?

tushar5526: Hmm, at the moment, it does not clear it up. We are creating this file at the "service parser" phase, the actual build command just gets back the modified `BuildArgs`. I will take a look and see if I can clear up the `tmp` file.

tushar5526: It seems like I can store the `dockerfile_inline` value in the `Build` struct, and then during the build phase, I can create the temporary file, use it for the build, and then clear it up. 

```go
type Build struct {
	Force     bool     // force build even if already present
	BuildArgs []string // {"-t", "example.com/foo", "--target", "foo", "/path/to/ctx"}
	// something like follows
	DockerfileInline string
       // TODO: call BuildKit API directly without executing `nerdctl build`
}
```

Do you have any objections to putting it there?

AkihiroSuda: No objection

tushar5526: Done.

AkihiroSuda (approved):
Thanks

PR #4308, opened 2025-06-06
internal/filesystem consolidation, part 2 

apostasie:
This is the second part of the effort to consolidate filesystem operations into `internal/filesystem`.

Problem refresher: `os.FileWrite` is not atomic, nor durable.
This has led in the past to a variety of issues (files with broken content after interruption, missing data, etc).
Though we did fix these issues, contributors are reintroducing (and will reintroduce) `os.FileWrite` in new code, reintroducing these issues unless done carefully.

This PR does offer a compatible drop-in replacement for FileWrite (`filesystem.FileWrite`) that guarantees atomicity and durability and configures the linter to prevent uses of `os.FileWrite`

Note that `filesystem.FileWrite` uses file rename to guarantee atomicity, which does change the inode.
This is fine in most cases, except one (when we want to modify in-place a containerd-mounted file).

About changes in this PR:
- first commit is a trivial (unrelated) typo
- second commit does strengthen `FileWriteWithRename` to be fully compatible with os.WriteFile (specifically on permissions - which is why the new `umask` helpers are added) and to `sync`
- third commit does use forbidigo to blacklist `os.FileWrite`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,6 +31,7 @@ linters:
     - revive
     # Gocritic
     - gocritic
+    - forbidigo
 
     # 3. We used to use these, but have now removed them
 
@@ -41,6 +42,12 @@ linters:
     # - nakedret
 
   settings:
+    forbidigo:
+      forbid:
+        # FIXME: there are still calls to os.WriteFile in tests under `cmd`
+        - pattern: ^os\.WriteFile.*$
+          pkg: github.com/containerd/nerdctl/v2/pkg
+          msg: os.WriteFile is neither atomic nor durable - use nerdctl filesystem.WriteFile instead
     staticcheck:
       checks:
         # Below is the default set
--- a/docs/dev/auditing_dockerfile.md
+++ b/docs/dev/auditing_dockerfile.md
@@ -255,7 +255,7 @@ On a warm cache, it is still over 150MB and 30+ seconds.
 In and of itself, this is hard to reduce, as we need these...
 
 Actions:
-- [ ] we could cache the module download location to reduce round-trips on modules that are shared accross
+- [ ] we could cache the module download location to reduce round-trips on modules that are shared across
 different projects
 - [ ] we are likely installing nerdctl modules six times - (once per architecture during the build phase, then once per
 ubuntu version and architecture during the tests runs (this is not even accounted for in the audit above)) - it should
--- a/docs/dev/store.md
+++ b/docs/dev/store.md
@@ -23,7 +23,7 @@ containers can be named the same), etc.
 However, storing data on the filesystem in a reliable way comes with challenges:
 - incomplete writes may happen (because of a system restart, or an application crash), leaving important structured files
 in a broken state
-- concurrent writes, or reading while writing would obviously be a problem as well, be it accross goroutines, or between
+- concurrent writes, or reading while writing would obviously be a problem as well, be it across goroutines, or between
 concurrent executions of the nerdctl binary, or embedded in a third-party application that does concurrently access resources
 
 The `pkg/store` package does provide a "storage" abstraction that takes care of these issues, generally providing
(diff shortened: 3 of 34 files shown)


apostasie:
@Shubhranshu153 @AkihiroSuda et al - this is ready for review.
Sorry it took a while to get this into shape - been busy.

Shubhranshu153:
This is critical change for data reliability, few questions. ( still reviewing)

1. Will this be a breaking change? If we install it in system already having nerdctl metadata will it get impacted?
2. Overall it seems to me we are having to re-implement features of database it to avoid using a database which offers the functionality of atomicity and reliability of data in situation of concurrent access.  
Are we so strongly against using a database for storing the metadata needed for nerdctl?

apostasie:
> This is critical change for data reliability, few questions. ( still reviewing)
> 
> 1. Will this be a breaking change? If we install it in system already having nerdctl metadata will it get impacted?

No.
This PR purely replaces `os.FileWrite` with `filesystem.FileWrite`, which is functionally identical (save for the inodes question).

> 2. Overall it seems to me we are having to re-implement features of database it to avoid using a database which offers the functionality of atomicity and reliability of data in situation of concurrent access.
>    Are we so strongly against using a database for storing the metadata needed for nerdctl?

Ultimately, that's a decision for maintainers.

Now:
1. there are certain things that will forever have to stay as files (anything mounted inside a container - anything shared with cni, or other systems - logs - etc). We should still make sure that anything dealing with these is ACID, regardless.
2. For the pieces that _could_ be stored inside a database instead, my 2 cents opinion is:
  a. "hard no" on having to run a daemon, so, that would limit us to daemonless systems (presumably sqlite then?) - **_and_** this system must work on all supported platforms (darwin, freebsd, windows, linux).
  b. I don't quite care what we end-up choosing in the end ^ - what I care about is that we expose an API that is abstracting said system so that it could easily be hot swapped without changing any code besides the concrete implementations (eg: instead of letting casual contributors hang themselves with it) - this is what `store.Store` is meant to provide: a storage abstraction that is safe to use and hides away how it is implemented to the developer

Shubhranshu153:
nerdctl is not truly daemonless as both buildkit and containerd are daemon.
Afaik containerd already uses db so not being able to use db on the systems may not be the case. 

I do agree certain things will remain as file and we need to be following ACID for it. Just the fact that we are writing a file system library to solve this makes me think it should be its own package if it doesnt exist as this is technically a generic problem.
For example: https://github.com/natefinch/atomic (i dont know how good or bad it is)

apostasie:
> nerdctl is not truly daemonless as both buildkit and containerd are daemon. Afaik containerd already uses db so not being able to use db on the systems may not be the case.

They do use bolt AFAIK (note: it is daemonless, which is good).
A few things:
- I am not sure freebsd is supported, or that Darwin works well?
- while not "huge", bolt is still ~5k lines of code - the tradeoff of taking on a (somewhat large) dependency should be evaluated against the value we really derive out of it - if this is just about have "safe" filesystem primitives, this might not be worth it
- several aspects should be evaluated precisely: for example does it support multiple rw locking, or a single write lock (<- my guess)? <- this type of things will have an impact on the performance profile / might be a deal killer - we _do need_ long-running locks, for eg.

> I do agree certain things will remain as file and we need to be following ACID for it. Just the fact that we are writing a file system library to solve this makes me think it should be its own package if it doesnt exist as this is technically a generic problem. For example: https://github.com/natefinch/atomic (i dont know how good or bad it is)

It is definitely a problem IMHO - though most people do brush it away, do not have to deal with concurrency, and/or are fine ignoring the occasional failed writes.

_Maybe_ one day we split it out - but for now, I think keeping it internal is better for the project while we figure out what we truly need in there and while we iterate on it? (disaster recovery and rollback are likely next)

Anyhow, interesting conversation!
Thanks @Shubhranshu153 !

apostasie:
Reading bolt: you apparently can have only one rw lock at a time.
This might not work well for us.

PR #4309, opened 2025-06-07
fix: display containers belonging to multiple networks in nerdctl net…

haytok:
…work inspect

When a container belongs to multiple networks, running the nerdctl network inspect command on the network to which the container belongs does not display the container in the current implementation.

Specifically, it is displayed as follows.

```
$ sudo nerdctl run -d --name net --net=foo --net=bar nginx
d88e878f0c60823bd0c361bad250f27b19ad117fb3336fcf18fa26ab1910c367

$ sudo nerdctl network inspect foo | jq .[0].Containers
{}

$ sudo nerdctl network inspect bar | jq .[0].Containers
{}
```

Ideally, running the nerdctl network inspect command on the networks to which the contaienr belongs should display the container name as follows.

```
$ sudo nerdctl network inspect foo | jq .[0].Containers
{
  "d88e878f0c60823bd0c361bad250f27b19ad117fb3336fcf18fa26ab1910c367": {
    "Name": "net"
  }
}

$ sudo nerdctl network inspect bar | jq .[0].Containers
{
  "d88e878f0c60823bd0c361bad250f27b19ad117fb3336fcf18fa26ab1910c367": {
    "Name": "net"
  }
}
```

Therefore, this behaviour is fixed in this PR.


cmd/nerdctl/network/network_inspect_test.go
```diff
@@ -320,6 +320,40 @@ func TestNetworkInspect(t *testing.T) {
...
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					Output: func(stdout string, info string, t *testing.T) {
+						var dc []dockercompat.Network
+						err := json.Unmarshal([]byte(stdout), &dc)
+
+						assert.NilError(t, err, "Unable to unmarshal output\n"+info)
+						assert.Equal(t, 1, len(dc), "Unexpectedly got multiple results\n"+info)
+						assert.Equal(t, dc[0].Name, data.Identifier("nginx-network-1"))
+						// Assert only the "running" containers on the same network are returned.
```
tushar5526: We can delete this comment.

pkg/cmd/network/inspect.go
```diff
@@ -58,25 +58,39 @@ func Inspect(ctx context.Context, client *containerd.Client, options types.Netwo
 		}
 
 		network := netList[0]
-		var filters = []string{fmt.Sprintf("labels.%q==%q", labels.Networks, []string{network.Name})}
```
tushar5526, line 61: nit: You can also use `regex` to create filters. `labels.%q~=%q`.

tushar5526: The current approach also looks acceptable :)

fahedouch: The nested loops and conditions make the code slightly more complex than before. I prefer fixing the regex  to avoid performance issue  with large numbers of containers

pkg/cmd/network/inspect.go
```diff
@@ -58,25 +58,39 @@ func Inspect(ctx context.Context, client *containerd.Client, options types.Netwo
...
-			containers = append(containers, nativeContainer)
+			info, err := container.Info(ctx)
+			if err != nil {
+				return err
+			}
+			networkLabels, ok := info.Labels[labels.Networks]
+			if !ok {
+				continue
+			}
+			var networks []string
```
fahedouch: please Consider extracting the container network membership check into a helper function for better readability

pkg/cmd/network/inspect.go
```diff
@@ -58,25 +58,39 @@ func Inspect(ctx context.Context, client *containerd.Client, options types.Netwo
...
+			if err := json.Unmarshal([]byte(networkLabels), &networks); err != nil {
+				return err
+			}
+			for i := range len(networks) {
```
fahedouch:
```suggestion
			for _, network := range networks
```
for better readability ;)

cmd/nerdctl/network/network_inspect_test.go
haytok: Thanks !!! fixed.

pkg/cmd/network/inspect.go
haytok: Thanks for comments!!! I've updated based on your advices.

pkg/cmd/network/inspect.go
haytok: OK, I've created another helper functon and call it.

pkg/cmd/network/inspect.go
haytok: Thanks!!!

pkg/cmd/network/inspect.go
```diff
@@ -58,16 +59,14 @@ func Inspect(ctx context.Context, client *containerd.Client, options types.Netwo
 		}
 
 		network := netList[0]
-		var filters = []string{fmt.Sprintf("labels.%q==%q", labels.Networks, []string{network.Name})}
 
+		var filters = []string{fmt.Sprintf("labels.%q~=%q", labels.Networks, network.Name)}
```
tushar5526, line 63: Can't the regex be stricter? This will also match substrings of a word instead of the complete word :thinking: ?

pkg/cmd/network/inspect.go
```diff
@@ -76,7 +75,14 @@ func Inspect(ctx context.Context, client *containerd.Client, options types.Netwo
 			if nativeContainer.Process == nil || nativeContainer.Process.Status.Status != containerd.Running {
 				continue
 			}
-			containers = append(containers, nativeContainer)
+
+			isNetworkMember, err := isContainerInNetwork(ctx, container, network.Name)
```
tushar5526, line 79: If the regex filter is stricter, I don't think we need to double-check whether the container is in the network, as we are already filtering it using the regex filter.

pkg/cmd/network/inspect.go
tushar5526: I think we can use `\b` to specify [word boundary](https://www.regular-expressions.info/wordboundaries.html) in the regex.

cmd/nerdctl/network/network_inspect_test.go
```diff
@@ -320,6 +320,39 @@ func TestNetworkInspect(t *testing.T) {
...
 		},
+		{
+			Description: "Display containers belonging to multiple networks in the output of nerdctl network inspect",
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("network", "create", data.Identifier("nginx-network-1"))
+				helpers.Ensure("network", "create", data.Identifier("nginx-network-2"))
+
+				helpers.Ensure("run", "-d", "--name", data.Identifier(), "--network", data.Identifier("nginx-network-1"), "--network", data.Identifier("nginx-network-2"), testutil.NginxAlpineImage)
+
+				data.Labels().Set("containerID", strings.Trim(helpers.Capture("inspect", data.Identifier(), "--format", "{{.Id}}"), "\n"))
```
apostasie, line 331: `id = helpers.Capture("run", "-d", "--name", data.Identifier(), "--network", ...)` from above would give you the id directly without having to do another inspect.

cmd/nerdctl/network/network_inspect_test.go
```diff
@@ -320,6 +320,39 @@ func TestNetworkInspect(t *testing.T) {
 				}
 			},
 		},
+		{
+			Description: "Display containers belonging to multiple networks in the output of nerdctl network inspect",
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("network", "create", data.Identifier("nginx-network-1"))
+				helpers.Ensure("network", "create", data.Identifier("nginx-network-2"))
+
+				helpers.Ensure("run", "-d", "--name", data.Identifier(), "--network", data.Identifier("nginx-network-1"), "--network", data.Identifier("nginx-network-2"), testutil.NginxAlpineImage)
```
apostasie, line 329: I am a bit late on this, sorry.
Maybe we can have a follow-on PR though?

testutil.NginxAlpineImage -> use testutil.CommonImage instead

cmd/nerdctl/network/network_inspect_test.go
```diff
@@ -320,6 +320,39 @@ func TestNetworkInspect(t *testing.T) {
...
+				helpers.Anyhow("rm", "-f", data.Identifier())
+				helpers.Anyhow("network", "remove", data.Identifier("nginx-network-1"))
+				helpers.Anyhow("network", "remove", data.Identifier("nginx-network-2"))
+			},
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("network", "inspect", data.Identifier("nginx-network-1"))
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					Output: func(stdout string, info string, t *testing.T) {
```
apostasie, line 343: You can use `Output: expect.JSON([] dockercompat.Network{}, func(dc []dockercompat.Network, info string, t tig.T) {
    // ...
})`

pkg/cmd/network/inspect.go
apostasie: Agreed. What happens if network1="my-network" and network2="my-network-as-well"?

cmd/nerdctl/network/network_inspect_test.go
haytok: Thanks for commnts !!! I'll create the PR to enhance tests asap.

pkg/cmd/network/inspect.go
haytok: > Can't the regex be stricter? This will also match substrings of a word instead of the complete word 🤔 ?
> I think we can use \b to specify word boundary in the regex.

Thanks for comments !!!

Referring only to regular expressions, strictly speaking, I think it would be good to define a filter like the one below.(ref : [The Go Playground](https://go.dev/play/p/J0O8SBi-48Z))

```
regexp.Compile(fmt.Sprintf("(\\[|,)\\s*\"%s\"\\s*(\\]|,)", regexp.QuoteMeta(networkName)))
```

Therefore, let us assume that the following filters are defined.

```golang
var filters = []string{fmt.Sprintf("labels.%q~=(\\[|,)\\s*\"%s\"\\s*(\\]|,)", labels.Networks, regexp.QuoteMeta(network.Name))}
```

However, such a filter is limited by the containerd side.

```bash
> sudo nerdctl network inspect my-network
FATA[0000] filters: parse error: [labels."nerdctl/networks"~=(\[|, >|)|< \s*"my-network"\s*(\]|,)]: expected field or quoted: invalid argument: invalid argument
```

So, I can't think of a good regular expression at the moment, but if you have any other better regular expressions, I would appreciate if you could give me some advices 🙏

pkg/cmd/network/inspect.go
haytok: > Agreed. What happens if network1="my-network" and network2="my-network-as-well"?
> If the regex filter is stricter, I don't think we need to double-check whether the container is in the network, as we are already filtering it using the regex filter.

As you know, when `container A` belongs to the network `my-network-as-well` and we perform `nerdctl network insect my-network`, the retrieved containers (`filteredContainers`) includes `container A`.

```golang
var filters = []string{fmt.Sprintf("labels.%q~=%q", labels.Networks, network.Name)}
filteredContainers, err := client.Containers(ctx, filters...)
```

Since the defined filter only cannot exclude `container A`, we use `isContainerInNetwork()` as the logic to exclude this unnecessary container.

pkg/cmd/network/inspect.go
tushar5526: As the network names are effectively an array which are stored as a string. I think this should work 

```
`"%s"` 
```
I tested in your example and it seems to be working fine: https://go.dev/play/p/bliOU3VRWC9

haytok: Sorry, I've missed your comments ...
And, thanks for your PR related to this comment 🙏

tushar5526: Ah, don’t worry. It was a bit trickier than I expected to get the regex right. Thanks

Issue #117, opened 2021-03-15
Flaky test: TestLogsWithFailingContainer


AkihiroSuda:
```
=== RUN   TestLogsWithFailingContainer
    logs_test.go:46: assertion failed: 
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test logs -f nerdctl-test-logs
        ExitCode: 0
        Stdout:   
        Stderr:   
        
        Failures:
        Expected stdout to contain "bar"
--- FAIL: TestLogsWithFailingContainer (0.50s)
```

PR #4311, opened 2025-06-10
feat: add support for com.docker.network.bridge.enable_icc network option

swagatbora90:
This PR adds support for docker compatible `com.docker.network.bridge.enable_icc` or `--icc` network create option. This option is used to enable/disable intercontainer connectivity. 

By default, any bridge network allows connectivity between containers attached to the same network, while containers in different networks are isolated. The enable_icc feature can be further used to enable/disable intra bridge container connectivity.

Requires firewall plugin >= **v1.7.1**

### Testing

1. Create a network with `com.docker.network.bridge.enable_icc` set to false, say `test-net`
2. Run a container and attach it to `test-net`
3. Run a second container and attach it to `test-net`
4. Ping from container 1 to container 2 should not succeed. 

```
sudo nerdctl network create  -o "com.docker.network.bridge.enable_icc=false" test-net          
254ddbe0ec1e5f3ed4b492876e2b6b9a051436b44c40d00ebd5e72bf7aa88435

sudo nerdctl run -d  --network test-net --name C1 ubuntu  sleep infinity  
6f22e651b9a4f3988e1ebce35f13806f65d0be942bfe9a2325c668a53cafcc9b

 sudo nerdctl run --network test-net alpine ping -c 1 -W 1 C1
PING C1 (10.4.4.4): 56 data bytes

--- C1 ping statistics ---
1 packets transmitted, 0 packets received, 100% packet loss
```

With icc=true, it will fallback to default behavior
```
sudo nerdctl network create  -o "com.docker.network.bridge.enable_icc=true" test-net1   
cf021853e2fd31f5ef29264e5fd834d45bf1ea364c3fef8ff60edec326899b77
 
sudo nerdctl run -d  --network test-net1 --name C2 ubuntu  sleep infinity  
2c95ae6ad9da8a07de97c609d9e65acdc95f3b343ea6e4c46ccd4c9b3df79b90

sudo nerdctl run --network test-net1 alpine ping -c 1 -W 1 C2
PING C2 (10.4.5.2): 56 data bytes
64 bytes from 10.4.5.2: seq=0 ttl=127 time=0.115 ms

--- C2 ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 0.115/0.115/0.115 ms
```

### DIFF
--- a/cmd/nerdctl/network/network_create_linux_test.go
+++ b/cmd/nerdctl/network/network_create_linux_test.go
@@ -26,6 +26,7 @@ import (
 	"gotest.tools/v3/assert"
 
 	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
 	"github.com/containerd/nerdctl/mod/tigron/tig"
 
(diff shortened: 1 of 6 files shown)


cmd/nerdctl/network/network_create_linux_test.go
```diff
@@ -106,6 +106,52 @@ func TestNetworkCreate(t *testing.T) {
...
+				helpers.Anyhow("container", "rm", "-f", data.Identifier("c1"))
+				helpers.Anyhow("network", "rm", data.Identifier())
+			},
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				// Try to ping the other container in the same network
+				// This should fail when ICC is disabled
+				return helpers.Command("run", "--rm", "--net", data.Identifier(),
+					testutil.CommonImage, "ping", "-c", "1", "-W", "1", data.Identifier("c1"))
+			},
+			Expected: test.Expects(1, nil, nil), // Expect ping to fail with exit code 1
```
apostasie: Wondering if `ping` portably returns that same exit code.
Maybe use `expect.ExitCodeGenericFail` to check for failure without being specific on exit code?

pkg/netutil/netutil_unix.go
```diff
@@ -310,8 +332,13 @@ func firewallPluginGEQ110(firewallPath string) (bool, error) {
...
 		return guessed, fmt.Errorf("failed to guess the version of %q: %w", firewallPath, err)
 	}
-	ver110 := semver.MustParse("v1.1.0")
-	return ver.GreaterThan(ver110) || ver.Equal(ver110), nil
+	targetVer := semver.MustParse(versionStr)
+	return ver.GreaterThan(targetVer) || ver.Equal(targetVer), nil
+}
+
+// firewallPluginGEQ110 checks if the firewall plugin is greater than or equal to v1.1.0
+func firewallPluginGEQ110(firewallPath string) (bool, error) {
```
apostasie: Maybe we can just remove this helper and call directly `firewallPluginGEQVersion(firewallPath, "v1.1.0")` where needs be?

cmd/nerdctl/network/network_create_linux_test.go
apostasie: eg: on macOS, it seems to exit 2

cmd/nerdctl/network/network_create_linux_test.go
```diff
@@ -106,6 +106,52 @@ func TestNetworkCreate(t *testing.T) {
...
+				helpers.Anyhow("network", "rm", data.Identifier())
+			},
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				// Try to ping the other container in the same network
+				// This should succeed when ICC is enabled
+				return helpers.Command("run", "--rm", "--net", data.Identifier(),
+					testutil.CommonImage, "ping", "-c", "1", "-W", "1", data.Identifier("c1"))
+			},
+			Expected: test.Expects(0, nil, nil), // Expect ping to succeed with exit code 0
+		},
```
apostasie, line 230: Maybe add a third case where enable_icc is not specified, to verify the default behavior?

pkg/netutil/netutil_unix.go
```diff
@@ -129,9 +135,24 @@ func (e *CNIEnv) generateCNIPlugins(driver string, name string, ipam map[string]
 		if ipv6 {
 			bridge.Capabilities["ips"] = true
 		}
-		plugins = []CNIPlugin{bridge, newPortMapPlugin(), newFirewallPlugin(), newTuningPlugin()}
+
+		// Determine the appropriate firewall ingress policy based on icc setting
```
apostasie, line 143: This only matter for non-default network, right?

Maybe the entire section should be moved inside the `if name != DefaultNetworkName {` section?

pkg/netutil/netutil_unix.go
```diff
@@ -111,6 +112,11 @@ func (e *CNIEnv) generateCNIPlugins(driver string, name string, ipam map[string]
 				if err != nil {
 					return nil, err
 				}
+			case "icc", "com.docker.network.bridge.enable_icc":
```
AkihiroSuda, line 115: Needs documentation

pkg/netutil/netutil_unix.go
swagatbora90: Yes, currently network options are only applied to non default network, which makes sense since the default network is created by nerdctl. That being said, there isn't a hard rule that these network options can never be applied to the default network. Docker actually allows you to configure the default network with these options on daemon startup. I think we can keep this as is, and let the callers decide how they want to invoke this.

cmd/nerdctl/network/network_create_linux_test.go
swagatbora90: Sure, will add

cmd/nerdctl/network/network_create_linux_test.go
swagatbora90: I will change this to just check for any non-zero exit code

pkg/testutil/nerdtest/requirements.go
```diff
@@ -439,3 +442,23 @@ func ContainerdVersion(v string) *test.Requirement {
 		},
 	}
 }
+
+// CNIFirewallVersionGE checks if the CNI firewall plugin version is greater than or equal to the specified version
+func CNIFirewallVersionGE(requiredVersion string) *test.Requirement {
```
AkihiroSuda: nit: "GE" suffix isn't consistent with `ContainerdVersion()`

cmd/nerdctl/network/network_create_linux_test.go
```diff
@@ -110,3 +111,97 @@ func TestNetworkCreate(t *testing.T) {
 
 	testCase.Run(t)
 }
+
+func TestNetworkCreateICC(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Require = require.All(
+		require.Linux,
+		nerdtest.Rootful,
```
AkihiroSuda: Why?

swagatbora90:
Looks like the test disabling ICC is failing on a few platform. Taking a look.

AkihiroSuda:
Moving to the v2.1.4 milestone

AkihiroSuda:
Needs rebase, in addition to fixing the CI

swagatbora90:
> Needs rebase, in addition to fixing the CI

@AkihiroSuda I had a hard time debugging the CI failures since this was working on my local test host. Finally, figured out that the the test are failing because the ubuntu runners do not have `br-netfilter` module loaded, which is required for this feature to work. We need the bridged traffic to go through netfilter. 

We will need to load the module in our ubuntu runners first for this test to work. Given we have other features that also rely on cni firewall rules, I am thinking of enabling it across all the runners. let me know what you think?

AkihiroSuda:
> We will need to load the module in our ubuntu runners first for this test to work. Given we have other features that also rely on cni firewall rules, I am thinking of enabling it across all the runners. let me know what you think?

👍, but maybe you do not need to modprobe it explicitly if you mount `-v /lib/modules:/lib/modules:ro` here:

https://github.com/containerd/nerdctl/blob/832c4556e0b82789f687b70d1e394b892e035722/.github/workflows/job-test-in-container.yml#L166

(and in the similar places in other `test-in-*.yaml`)

swagatbora90:
> but maybe you do not need to modprobe it explicitly if you mount -v /lib/modules:/lib/modules:ro here

I tried this option, but ran into the same errors https://github.com/swagatbora90/nerdctl/pull/6/commits/06d49d011aa236c271864ea810446a21e4dc4eeb. Looks like I need to load the br_netfilter module explicitly as they are not loaded by default.

swagatbora90:
Opened [https://github.com/containerd/nerdctl/pull/4481] to load the br_netfilter module inn the runners.

swagatbora90:
Test failure related to one of the flaky test https://github.com/containerd/nerdctl/issues/117

swagatbora90:
Test failures are not related and appear to be one of the flaky ones.

swagatbora90:
@AkihiroSuda @apostasie  Tests are passing now after latest CI update to load the br_netfilter module. I think this PR is good to be merged now. Still seeing some random flaky test failures though, but appears to be unrelated.

cmd/nerdctl/network/network_create_linux_test.go
swagatbora90: I have removed the require for 'rootful'. Test should pass for both rootful and rootless.

AkihiroSuda (approved):
Thanks

PR #4324, opened 2025-06-18
internal/filesystem consolidation, part 3 (no rename WriteFile, disaster recovery and rollbacks)

apostasie:
This is part 3 of internal/FS series.

<!> unlike the previous PRs that were essentially consolidation and enforcing atomic writes, this here is more ambitious, and adds net new features (eg: rollbacks and disaster recovery) - as such, it _may_ be more controversial, so, open to discussion here, as usual <!>

This expands on the previous goal (Atomicity and Durability of `WriteFile`) by providing a new version of `WriteFile` that no longer relies on file renames - hence no longer changing inode - making it now 100% compatible with `os.WriteFile` and usable everywhere (including when files are mounted).

Furthermore, it also focuses on enabling _Consistency_, by introducing:
- disaster recovery:
  - if `WriteFile` fails, the file is immediately reverted to its previous version (or removed if it did not exist prior)
  - if `WriteFile` is hard interrupted for any reason, the _next_ `ReadFile` or `WriteFile` call will first revert it to its original content (or remove it if it did not exist)
- rollback: on a successful write, `WriteFileWithRollback` will return a rollback function that can revert changes

`rollback` should be used in the future to ensure higher-level operations _consistency_ and to facilitate cleanup in case of failures. This should make higher-level operations failures able to revert state / cleanup leftover easier.

I do appreciate this is additional complexity, and some of the code is not trivial.
But then:
- the complexity is already here in things like `container.Create` and other places that are manually rollbacking in a baroque / ad-hoc form
- IMHO this is what it takes to implement higher-level consistent data operations

Finally note that this changeset is purely internal to the `filesystem` package and does not change anything in the rest of nerdctl codebase.

Let me know your thoughts overall, and if this is desirable or not.

### DIFF
--- a/pkg/fs.go
+++ b/pkg/fs.go
@@ -1,0 +1,27 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package pkg
+
+import "github.com/containerd/nerdctl/v2/pkg/internal/filesystem"
+
+// InitFS will set the root location to store `internal/filesystem` ops files.
+// These files are used to allow `WriteFile` to backup and rollback content.
+// While they are transient in nature, they should still persist OS crashes / reboots, so, preferably under something
+// like XDGData, rather than tmp.
+func InitFS(path string) error {
+	return filesystem.SetFilesystemOpsDirectory(path)
+}
--- a/pkg/internal/filesystem/writefile_rollback_test.go
+++ b/pkg/internal/filesystem/writefile_rollback_test.go
@@ -1,0 +1,206 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+//nolint:forbidigo
+package filesystem
+
+import (
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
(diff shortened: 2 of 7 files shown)


AkihiroSuda:
No merge conflict, but could you rebase just in case

pkg/fs.go
```diff
@@ -0,0 +1,23 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package pkg
+
+import "github.com/containerd/nerdctl/v2/pkg/internal/filesystem"
+
+func InitFS(path string) error {
```
AkihiroSuda, line 25: Needs a explanation comment

apostasie: Sure.

apostasie:
Rebased and comment added.

AkihiroSuda:
Comment seems still missing

apostasie:
> Comment seems still missing

My bad. Added it to the call: https://github.com/containerd/nerdctl/pull/4324/files#diff-202c9b24bcda33888c16a71750bd47809ffd474e126bfd2be11857033c0272f0R149

Will add more to the func declaration then.

apostasie:
CI failures:
- EL8 rootful: `TestSharedNetworkSetup` flaky 
- EL8 rootless - newly introduced `TestContainerHealthCheckAdvance` seems flaky
- rootful canary timeouting - looks like the runs are taking longer lately, so, maybe we need to adjust timeouts
- windows failure I have seen on another PR - something fishy going on with windows networking

apostasie:
> Comment seems still missing

Added expanded comment to function declaration (along with comment on function call).

pkg/internal/filesystem/writefile_rollback_test.go
```diff
@@ -0,0 +1,206 @@
...
+	fp := filepath.Join(dir, "pre-existing-file")
+	_ = os.WriteFile(fp, []byte("original content"), 0o600)
+
+	// Artificially create leftover marker
+	_ = markerCreate(fp, "")
+
+	err := ensureRecovery(fp)
+	assert.NilError(t, err)
+
+	cn, _ := os.ReadFile(fp)
```
AkihiroSuda, line 204: Any reason not to check err?

pkg/internal/filesystem/writefile_rollback_test.go
```diff
@@ -0,0 +1,206 @@
...
+	assert.Assert(t, os.IsNotExist(err))
+}
+
+func TestDisasterRecoveryNoBackup2(t *testing.T) {
+	// Test that a file with a more recent marker leftover and no backup will be left untouched.
+
+	// Create file
+	dir := t.TempDir()
+	fp := filepath.Join(dir, "pre-existing-file")
+	_ = os.WriteFile(fp, []byte("original content"), 0o600)
```
AkihiroSuda, line 196: Any reason not to check err?

apostasie: It is indeed a bit lazy - though, given the (test) codepath, if _this_ is failing, presumably everything is failing...
I see you merged.
If this is important, I can update in a follow on.

PR #4334, opened 2025-06-20
[FIX]: revert atomic write for logs

apostasie:
It _seems_ like #4308 _maybe_ is wrecking logs.

While #4324 would fix it (by introducing a better version of WriteFile that does not rely on Rename), we should still fix this now for the current state of affairs.

Looking at CI now to confirm.


apostasie:
Closing, as #4324 got merged.

Issue #4339, opened 2025-06-21
Nerdctl as a library (documentation and API work)


apostasie:
### What is the problem you're trying to solve

See https://github.com/containerd/nerdctl/discussions/4337 for context.

Building a simple cli that just runs a container is not trivial for newcomers, and certainly requires significant familiarity with nerdctl internals to get it to work.

Furthermore, the "API" is largely undocumented, and there are a number of extra things to implement on-top of using `pkg` (hooking of logging and oci-hooks).

### Describe the solution you'd like

Suggesting:
1. documentation
2. express a better / simpler API hiding away some of the complexity

PR #4340, opened 2025-06-21
Add example of using nerdctl as a library

apostasie:
See #4339 and linked discussion for context.

This is a first step.
Actual documentation and making the API better is still needed.

### DIFF
--- a/examples/nerdctl-as-a-library/README.md
+++ b/examples/nerdctl-as-a-library/README.md
@@ -1,0 +1,3 @@
+# Using nerdctl as a library
+
+This directory contains examples showing how to implement a cli communicating with containerd, using nerdctl as a library.
--- a/examples/nerdctl-as-a-library/run-container/main.go
+++ b/examples/nerdctl-as-a-library/run-container/main.go
@@ -1,0 +1,109 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


cmd/examples/run-container/main.go
AkihiroSuda, line 1: Can we place this as /examples/nerdctl-as-a-library/main.go

apostasie: Sure.

cmd/examples/run-container/main.go
```diff
@@ -0,0 +1,109 @@
...
+		LogDriver:   "json-file",
+		StopSignal:  "SIGTERM",
+		Restart:     "unless-stopped",
+		Interactive: true,
+	}
+
+	// Create client
+	client, ctx, cancel, err := clientutil.NewClient(context.Background(), globalOpt.Namespace, globalOpt.Address)
+	if err != nil {
+		fmt.Println(err)
```
AkihiroSuda: os.Exit(1) on error

cmd/examples/run-container/main.go
apostasie: Now ` examples/nerdctl-as-a-library/run-container/main.go`.
Keeping the `run-container` segment, as additional, other examples should be added performing other operations?

cmd/examples/run-container/main.go
apostasie: Because of defer, this will make the linter complain (in other places where we return).

Agreed this overall could be made somewhat better - maybe future work on this can come up with better code and examples? (feeling lazy Sunday :-))

AkihiroSuda (approved):
Thanks

apostasie:
Failures are:
- logs
- some compose up issue
- hitting total duration timeout

Issue #4343, opened 2025-06-22
[CI]: TestContainerHealthCheckAdvance is flaky


apostasie:
### Description

https://github.com/containerd/nerdctl/actions/runs/15811078048/job/44562430914?pr=4327

```
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ C:\Users\runneradmin\go\bin\nerdctl.exe container inspect testcontainerhealthcheckadvance-hea |
        |         | lthcheck-using-cmd-shell-format-b70c8418                                                         |
        +------------------------------------------------------------------------------------------------------------+
        |         | Actual content is 147 lines long and has been abbreviated to 50
         |
        |         |                                                                                                  |
        |         | 🟢 [                                                                                             |
        |         |     {                                                                                            |
        |         |         "Id": "2cb068792e63e09b0ca9b6e19af60a5518853d56498a0187bd02384b407df8e8",                |
        |         |         "Created": "2025-06-22T21:51:07.1924498Z",                                               |
        |         |         "Path": "cmd.exe",                                                                       |
        |         |         "Args": [                                                                                |
        |         |             "/s",                                                                                |
        |         |             "/c",                                                                                |
        |         |             "sleep",                                                                             |
        |         |             "3600"                                                                               |
        |         |         ],                                                                                       |
        |         |         "State": {                                                                               |
        |         | ... 97 lines are being ignored...                                                                |
        |         |                 "image_version": "1.36.1-1",                                                     |
        |         |                 "io.containerd.image.config.stop-signal": "SIGTERM",                             |
        |         |                 "nerdctl/auto-remove": "false",                                                  |
        |         |                 "nerdctl/dns": "{\"DNSServers\":null,\"DNSResolvConfOptions\":null,\"DNSSearchDo |
        |         | mains\":null}",                                                                                  |
        |         |                 "nerdctl/extraHosts": "[]",                                                      |
        |         |                 "nerdctl/healthcheck": "{\"Test\":[\"CMD-SHELL\",\"echo shell-format\"],\"Interv |
        |         | al\":1000000000}",                                                                               |
        |         |                 "nerdctl/healthstate": "{\"Status\":\"healthy\",\"FailingStreak\":0}",           |
        |         |                 "nerdctl/host-config": "{\"BlkioWeight\":0,\"CidFile\":\"\",\"Devices\":null}",  |
[...]

PR #4346, opened 2025-06-23
Bump CI timeout

apostasie:
Canary builds have been getting over 45 minutes lately.

Unclear to me why (and we may want to look into that later on), but meanwhile bumping the timeout to 60 mins to at least figure out if it is stuck or simply getting over.

### DIFF
--- a/.github/workflows/workflow-test.yml
+++ b/.github/workflows/workflow-test.yml
@@ -102,7 +102,7 @@ jobs:
             canary: true
 
     with:
-      timeout: 45
+      timeout: 60
       runner: ${{ matrix.runner }}
       target: ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


apostasie:
Failures:
- #4343 
- TestContainerRmIptables somehow tries to bind on a used port

AkihiroSuda (approved):
Thanks

.github/workflows/workflow-test.yml
```diff
@@ -102,7 +102,7 @@ jobs:
             canary: true
 
     with:
-      timeout: 45
+      timeout: 60
```
fahedouch, line 105: it's getting crazy LOL

PR #4350, opened 2025-06-23
Pass along GITHUB_TOKEN to canary

apostasie:
We have been getting lots of 403 from Github lately.

This PR does pass along the GITHUB_TOKEN where it matters so we do authenticated requests.

(scripts are already wired to use the token if provided)


AkihiroSuda (approved):
Thanks

Issue #4347, opened 2025-06-23
[CI]: TestContainerRmIptables fails binding on port


apostasie:
### Description

Happens regularly on the CI.

The intriguing part is that the port is obtained through `portlock`, so...

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

Not flake.

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

Issue #4146, opened 2025-04-20
[CI]: TestComposeUpWithExternalNetwork network failure


apostasie:
### Description

Test output is minimal.
It is unclear if the test failed to bring up the stack, or if there actually is a networking issue (note: this is rootful).

Test must be rewritten for better debugging information.

### Describe the results you received and expected

```
=== RUN   TestComposeUpWithExternalNetwork
    compose_up_linux_test.go:447: assertion failed: 
        Command:  /usr/local/bin/nerdctl --namespace=nerdctl-test exec nerdctl-testcomposeupwithexternalnetwork-1 wget -qO- http://nerdctl-testcomposeupwithexternalnetwork-2
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   wget: can't connect to remote host (10.4.1.4): Connection refused
        time="2025-04-20T20:00:27Z" level=fatal msg="exec failed with exit code 1"
        
        
        Failures:
        ExitCode was 1 expected 0
        Expected stdout to contain "<title>Welcome to nginx!</title>"
        Expected no error
--- FAIL: TestComposeUpWithExternalNetwork (2.99s)
FAIL cmd/nerdctl/compose.TestComposeUpWithExternalNetwork (2.99s)
```

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #4362, opened 2025-06-24
Simplify healthcheck file handling + fix locking

apostasie:
Reading hc code to figure out why the tests are failing:
- removing unnecessary extension to append-filelock (note: generally a bad idea to apply a lock on the same file you want to manipulate)
- simplify code and remove useless helper
- fix logic fault that is probably leading to TOCTOU issue resetting the log file, as the lock is not applied there


apostasie:
Unrelated failures:
- #4146
- #4347

apostasie:
Failure is:
- #4146

Issue #4322, opened 2025-06-17
[CI]: TestNetworkInspect is flaky in windows


Shubhranshu153:
### Description

Some of the subtest are flaky in network inspect test in windows.
Need to root cause.

### Steps to reproduce the issue
Running the CI pipelines will occasionally reproduce this error.

### Describe the results you received and expected

Expected test to pass.

### What version of nerdctl are you using?

latest.

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
@Shubhranshu153 can you add the log output? + the link to the build that failed?

apostasie:
eg:
- linking the PR(s) that failed with this helps evaluating how prevalent this is
- concrete log output avoids linking unrelated issues that may affect the same test
:)

apostasie:
```
          +------------------------------------------------------------------------------------------------------------+
          | ➡️      | ⚙️ C:\Users\runneradmin\go\bin\nerdctl.exe create --name testnetworkinspect-verify-that-only-act |
          |         | ive-containers-appear-in-the-0a54a379 --network testnetworkinspect-verify-that-only-active-conta |
          |         | iners-appear-in-the-844e8a21 registry.k8s.io/e2e-test-images/nginx:1.14-2                        |
          +------------------------------------------------------------------------------------------------------------+
          |         | 🚫 command returned a non-zero exit code                                                         |
          +------------------------------------------------------------------------------------------------------------+
          |         | ⚠️ 1                                                                                             |
          +------------------------------------------------------------------------------------------------------------+
          |         | 🟠 time="2025-06-22T21:52:24Z" level=warning msg="default network named \"nat\" does not have an |
          |         |  internal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl"  |
          |         | time="2025-06-22T21:52:24Z" level=warning msg="default network named \"nat\" does not have an in |
          |         | ternal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl"     |
          |         | time="2025-06-22T21:52:25Z" level=warning msg="networking setup error has occurred" error="plugi |
          |         | n type=\"nat\" failed (add): hcnCreateNetwork failed in Win32: The object already exists. (0x139 |
          |         | 2) {\"Success\":false,\"Error\":\"The object already exists. \",\"ErrorCode\":2147947410}"       |
[...]

PR #4368, opened 2025-06-24
refactor: move BUILDKIT_HOST to buildkitutil

pendo324:
Previously, only the build command would attempt to use the `BUILDKIT_HOST` environment variable, where as other usage of buildkitutil's GetBuildkitHost function would not

This split behavior can be confusing to users, so I think we should centralize the logic


apostasie:
Unrelated failures:
- #4347 
- #4322

AkihiroSuda (approved):
Thanks

PR #4376, opened 2025-06-26
rename `network-config.json` to `ports.json` to match reality

AkihiroSuda:
Also rename `networkstore` to `portstore` 

Follow-up to PR #4290 
cc @haytok

### DIFF
--- a/docs/dir.md
+++ b/docs/dir.md
@@ -35,7 +35,7 @@ Files:
 - `<CID>-json.log`: used by `nerdctl logs`
 - `oci-hook.*.log`: logs of the OCI hook
 - `lifecycle.json`: used to store stateful information about the container that can only be retrieved through OCI hooks
-- `network-config.json`: used to store port mapping information for containers run with the `-p` option.
+- `ports.json`: used to store port mapping information for containers run with the `-p` option.
 
 ### `<DATAROOT>/<ADDRHASH>/names/<NAMESPACE>`
 e.g. `/var/lib/nerdctl/1935db59/names/default`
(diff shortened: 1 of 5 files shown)


haytok:
Thanks!!! LGTM

apostasie:
Hey @AkihiroSuda 
I am wondering about this.
We are going to need to put more stuff in there soon, as we will want to move more config away from labels to here.
Particularly thinking about multiple IPs for different networks, which will require significant changes.

apostasie:
Point being: I think we might want to leave it named `network-config.json`.

haytok:
The reason why I named `network-config.json` and `networkstore` is that I thought that in the future, we would be able to store information about other network info (e.g.  `ip adress`,`hostname`, and so on) currently stored in the label as @apostasie commented.

However, if @AkihiroSuda is changing it for a different reason, I think this correction is fine.

AkihiroSuda:
The current type definition `[]cni.PortMapping` doesn't seem extensible to support non-port config fields?
It has to be changed to a new struct

apostasie:
> The current type definition `[]cni.PortMapping` doesn't seem extensible to support non-port config fields? It has to be changed to a new struct



> The current type definition `[]cni.PortMapping` doesn't seem extensible to support non-port config fields? It has to be changed to a new struct

@haytok do you want to go for it and change the data structure?

apostasie:
Note: failures here are logs related, which have been plaguing us.

Now, while debugging healthcheck failures, I now believe we might have an underlying problem with containerd cio.
See #4377 for a case where (apparently) running a task fails to capture the output.
^ either we do not implement task running correctly, or something is borked inside containerd.

haytok:
I apologize for the confusing comments and implementation.

I'll create a separate `struct` so that information other than port mapping can be written to `network-config.json`.

This change will be carried over to this PR!

Issue #4373, opened 2025-06-25
[Flaky] TestComposeExecWithIndex


apostasie:
### Description

Apparently, in certain circumstances, `ips := helpers.Capture(append(cmds, "ip", "addr", "show", "dev", "eth0")...)` fails to return meaningful data.

Comment says: `// try 5 times to ensure that results are stable`

So, this implies either:
- that after a `compose up -d`, a container might be started "enough" to be OK with an exec, but not ready enough for `ip addr` to return anything?
- or something is broken that makes it so the interface is not available in certain conditions

One way or the other, more debugging output is needed.

```
=== RUN   TestComposeExecWithIndex/1
=== PAUSE TestComposeExecWithIndex/1
=== CONT  TestComposeExecWithIndex/1
    compose_exec_linux_test.go:356: 
        
        +============================================================================================================+
        | 🏗       | "TestComposeExecWithIndex/1": setup                                                              |
        +============================================================================================================+
        
    compose_exec_linux_test.go:306: 
        
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl compose -f /tmp/TestComposeExecWithIndex4037248067/001/09b41bbf/compos |
        |         | e.yaml exec -i=false --no-TTY --index 1 svc0 cat /etc/hosts                                      |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🟢 # <nerdctl>                                                                                   |
        |         | 127.0.0.1  localhost localhost.localdomain                                                       |
        |         | ::1    localhost localhost.localdomain                                                           |
        |         | 10.4.1.2        svc0 svc0.09b41bbf_default 09b41bbf-svc0-2 09b41bbf-svc0-2.09b41bbf_default      |
        |         | 10.4.1.4        svc0 svc0.09b41bbf_default 09b41bbf-svc0-1 09b41bbf-svc0-1.09b41bbf_default      |
        |         | 10.4.1.3        svc0 svc0.09b41bbf_default 09b41bbf-svc0-3 09b41bbf-svc0-3.09b41bbf_default      |
        |         | # </nerdctl>                                                                                     |
        +------------------------------------------------------------------------------------------------------------+
        | 🌱      | XDG_SESSION_TYPE=tty                                                                             |
        |         | HOME=/home/rootless                                                                              |
        |         | XDG_SESSION_CLASS=user                                                                           |
        |         | XDG_SESSION_ID=4                                                                                 |
[...]

Issue #4364, opened 2025-06-24
`nerdctl commit` is not synchronous?


AkihiroSuda:
> Thanks, but this seems to be a bug of `nerdctl commit`?
It should be just synchronous?

_Originally posted by @AkihiroSuda in https://github.com/containerd/nerdctl/pull/4357#pullrequestreview-2952080469_

apostasie:
Context: https://github.com/containerd/nerdctl/issues/4106

PR #4360, opened 2025-06-23
[CI] Add delay for windows network test

apostasie:
Tentatively workaround #4322

Issue #4120, opened 2025-04-17
[CI]: Epic: fix everything


apostasie:
## Currently known flaky tests (could be a test issue or a nerdctl issue)

- #4119
- #4118
- #4106
- #4068
- #4066
- third-party: #4046
- nerdctl issue: #4043
- #3971
- #3727
- #3717
- #3524
- #3510
- #2690

## CI structural issues

- #3733 
- #3453

## Tests to add

- #3494 
- Expert: #3156
- #1989 
- #1957
- #178

## Enhancements: tigron

- #4029
- #4028
- #3448

## Enhancements: generic

- dockerfile specific: #4031
- dockerfile specific: #4021
- #3858
- #3845
- #3480
- #3303

## Performance tracking

- Latest major disruption: #3924 
- #3577

apostasie:
@AkihiroSuda could we label `area CI` and `help wanted`?

apostasie:
@AkihiroSuda the following are ready (in order), at your convenience.

- #4111 
- #4114
- #4116 
- #4125 
- #4081 <- this one is bigger

apostasie:
New round of CI PRs:
- [ ] #4250
- [ ] #4248 
- [ ] #4247 
- [ ] #4209 
- [ ] #4203

Shubhranshu153:
Can we add this issue to the epic: https://github.com/containerd/nerdctl/issues/4322

apostasie:
> Can we add this issue to the epic: [#4322](https://github.com/containerd/nerdctl/issues/4322)

Done.

PR #4378, opened 2025-06-26
Enable setting DNS options through global nerdctl config

swagatbora90:
This PR adds support for configuring DNS settings globally through the nerdctl configuration file (nerdctl.toml), allowing users to set default DNS servers, search domains, and options that apply to all containers unless explicitly overridden by command-line flags.

This is for a usecase where I want all  containers to use a specific dns server instead of system default. Currently, I have to use `--dns` flag with every container create/run commands. This functionality is available in docker which allows setting global dns config value in the daemon config file. 

#### Testing
Added a end to end test TestDNSWithGlobalConfig() that checks 4 scenarios:
  1. Global DNS settings used when no command-line options provided
  2. Command-line DNS options override global config (precedence test)
  3. Global DNS settings apply with host networking
  4. Global DNS settings apply with none networking

### DIFF
--- a/cmd/nerdctl/container/container_run_network_linux_test.go
+++ b/cmd/nerdctl/container/container_run_network_linux_test.go
@@ -996,3 +996,87 @@ func TestHostNetworkDnsConfigs(t *testing.T) {
 	}
 	testCase.Run(t)
 }
+
+func TestDNSWithGlobalConfig(t *testing.T) {
+	var configContent test.ConfigValue = `debug = false
+debug_full = false
+dns = ["10.10.10.10", "20.20.20.20"]
+dns_opts = ["ndots:2", "timeout:5"]
+dns_search = ["example.com", "test.local"]`
+
+	nerdtest.Setup()
+
+	testCase := &test.Case{
+		Config: test.WithConfig(nerdtest.NerdctlToml, configContent),
+		// NERDCTL_TOML not supported in Docker
+		Require: require.Not(nerdtest.Docker),
+		SubTests: []*test.Case{
+			{
+				Description: "Global DNS settings are used when command line options are not provided",
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					nerdctlTomlContent := string(helpers.Read(nerdtest.NerdctlToml))
+					helpers.T().Log("NERDCTL_TOML file content:\n%s", nerdctlTomlContent)
+					cmd := helpers.Command("run", "--rm", testutil.CommonImage, "cat", "/etc/resolv.conf")
+					return cmd
+				},
+				Expected: test.Expects(expect.ExitCodeSuccess, nil, expect.All(
+					expect.Contains("nameserver 10.10.10.10"),
+					expect.Contains("nameserver 20.20.20.20"),
+					expect.Contains("search example.com test.local"),
+					expect.Contains("options ndots:2 timeout:5"),
+				)),
+			},
+			{
+				Description: "Command line DNS options override global config",
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					nerdctlTomlContent := string(helpers.Read(nerdtest.NerdctlToml))
+					helpers.T().Log("NERDCTL_TOML file content:\n%s", nerdctlTomlContent)
+					cmd := helpers.Command("run", "--rm",
+						"--dns", "9.9.9.9",
+						"--dns-search", "override.com",
+						"--dns-opt", "ndots:3",
+						testutil.CommonImage, "cat", "/etc/resolv.conf")
+					return cmd
+				},
+				Expected: test.Expects(expect.ExitCodeSuccess, nil, expect.All(
+					expect.Contains("nameserver 9.9.9.9"),
+					expect.Contains("search override.com"),
+					expect.Contains("options ndots:3"),
+				)),
+			},
+			{
+				Description: "Global DNS settings should also apply when using host network",
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					nerdctlTomlContent := string(helpers.Read(nerdtest.NerdctlToml))
+					helpers.T().Log("NERDCTL_TOML file content:\n%s", nerdctlTomlContent)
+					cmd := helpers.Command("run", "--rm", "--network", "host",
+						testutil.CommonImage, "cat", "/etc/resolv.conf")
+					return cmd
+				},
+				Expected: test.Expects(expect.ExitCodeSuccess, nil, expect.All(
+					expect.Contains("nameserver 10.10.10.10"),
+					expect.Contains("nameserver 20.20.20.20"),
+					expect.Contains("search example.com test.local"),
+					expect.Contains("options ndots:2 timeout:5"),
+				)),
+			},
+			{
+				Description: "Global DNS settings should also apply when using none network",
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					nerdctlTomlContent := string(helpers.Read(nerdtest.NerdctlToml))
+					helpers.T().Log("NERDCTL_TOML file content:\n%s", nerdctlTomlContent)
+					cmd := helpers.Command("run", "--rm", "--network", "none",
+						testutil.CommonImage, "cat", "/etc/resolv.conf")
+					return cmd
+				},
+				Expected: test.Expects(expect.ExitCodeSuccess, nil, expect.All(
+					expect.Contains("nameserver 10.10.10.10"),
+					expect.Contains("nameserver 20.20.20.20"),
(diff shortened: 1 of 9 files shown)


apostasie:
Hey @swagatbora90 

Noting that:
```
docker run -ti --dns="" debian cat /etc/resolv.conf
nerdctl run -ti --dns="" debian cat /etc/resolv.conf
```

Different behaviors on this ^.
(also same difference on dns-search="")

Appreciate your PR is not about that, but maybe, if you wish, this PR could fix that discrepancy as well since it is touching that part?

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -996,3 +996,136 @@ func TestHostNetworkDnsConfigs(t *testing.T) {
...
+dns = ["10.10.10.10", "20.20.20.20"]
+dns_opts = ["ndots:2", "timeout:5"]
+dns_search = ["example.com", "test.local"]`
+	tomlPath := filepath.Join(t.TempDir(), "nerdctl.toml")
+	os.WriteFile(tomlPath, []byte(configContent), 0644)
+
+	nerdtest.Setup()
+
+	testCase := &test.Case{
+		Env: map[string]string{"NERDCTL_TOML": tomlPath},
```
apostasie: Tigron has a helper for that:

Just add this to your testCase instead of the Env:
```
			Config:      test.WithConfig(nerdtest.NerdctlToml, configContent),
```

(also takes care of writing the file in a temp location for you, and make it easier to retrieve)

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -996,3 +996,136 @@ func TestHostNetworkDnsConfigs(t *testing.T) {
...
+		// NERDCTL_TOML not supported in Docker
+		Require: require.Not(nerdtest.Docker),
+		SubTests: []*test.Case{
+			{
+				Description: "Global DNS settings are used when command line options are not provided",
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					nerdctlTomlPath := os.Getenv("NERDCTL_TOML")
```
apostasie: Is this going to work as you think?
os.Getenv points to global env, not the test / commands env.

If you use the config from above ^^^
```
helpers.Read(nerdtest.NerdctlToml)
```

Should give you what you want?

cmd/nerdctl/container/container_run_network_linux_test.go
```diff
@@ -996,3 +996,136 @@ func TestHostNetworkDnsConfigs(t *testing.T) {
...
+							helpers.T().Log("NERDCTL_TOML file content:\n%s", string(content))
+						} else {
+							helpers.T().Log("Failed to read NERDCTL_TOML file: %v", err)
+						}
+					}
+
+					cmd := helpers.Command("run", "--rm", testutil.CommonImage, "cat", "/etc/resolv.conf")
+					return cmd
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
```
apostasie: More compact, using the `test.Expects` helper and comparators instead of naked asserts (better debugging info):
```
Expected: test.Expects(expect.ExitCodeSuccess, nil, expect.All(
   expect.Contains("foo"),
   expect.Contains("bla"),
)),
```

apostasie:
@swagatbora90 left a few comments on tests.

Tigron testing has some (semblance of) documentation here: https://github.com/containerd/nerdctl/blob/main/docs/testing/tools.md

Very much work in progress, so, it is what it is... but then, am always happy to cosplay "testing-helper-AI-bot" until the doc and API are settled :-) (also, feedback on Tigron is always welcome)

cmd/nerdctl/container/container_run_network_linux_test.go
swagatbora90: If I read it correctly, the `Env` var should be passed to all Subtest. I need all the Subtest to use the global env, so it does work. 

However, your suggestion to use Config should work too. Let me change it like you suggested.

apostasie: Yes, `testCase.Env` is propagated to subtests, but this is _not_ the global os.Environ (otherwise, you would pollute other unrelated tests), so `os.Getenv` will not give you the test environment...

apostasie: Not completely sure I was clear, so, in a shell:
- do not rely on os.Getenv or os.Setenv - that _will_ cross pollute other tests and is specifically frowned-upon
- the testCase Env is private to the test, but it is inherited by subtests
- every test has a custom nerdctl.toml, regardless of the fact you want one explicitly or not (this is to guarantee that we are isolated from a possible, pre-existing, host `nerdctl.toml`)

swagatbora90:
> Hey @swagatbora90
> 
> Noting that:
> 
> ```
> docker run -ti --dns="" debian cat /etc/resolv.conf
> nerdctl run -ti --dns="" debian cat /etc/resolv.conf
> ```
> 
> Different behaviors on this ^. (also same difference on dns-search="")
> 
> Appreciate your PR is not about that, but maybe, if you wish, this PR could fix that discrepancy as well since it is touching that part?

Looks like there is no validation today on the DNS servers and domain names that are being passed. I can add a follow up PR to add validation, I would prefer to keep this PR only for adding the global config.

apostasie:
> Looks like there is no validation today on the DNS servers and domain names that are being passed. I can add a follow up PR to add validation, I would prefer to keep this PR only for adding the global config.

Fair.

cmd/nerdctl/container/container_run_network_linux_test.go
swagatbora90: Okay, I see your point. I did check that the NERDCTL_TOML config values are correct when running the test. Anyways, I have changed to use the `Config` attribute like you suggested.

apostasie:
- Windows failure is likely due to #4322 
I did workaround it for _some_ tests (#4360), but we probably want to do a full review all of network tests - or better, actually fix the issue, but I am not sure it is a nerdctl issue or an hcsshim one.
- Rootless failure is yet another variant of content digest not found
- TestComposeExec: #4380

@swagatbora90 re-run the CI? (eg: rebase or close/open)

apostasie:
Failures:
- soci (fixed upstream)
- TestLogs (possibly a similar issue to the recent IO fixes that got merged)
- 2x the infamous content digest not found (the good news part on this one is that it now shows a lot more, which should make it easier to repro)
- #4373 - somehow, `ip addr show dev eth0` returns nothing - could be a case of failing to wait on task IO

All unrelated to this PR ^

apostasie (approved):

apostasie:
Note: #4392 just got in - curious is this is going to help TestComposeExecWithIndex

swagatbora90:
New set of failing tests:
1. TestIPFSCompNoBuild - Is this known?
2. [TestKubeCommitSave](https://github.com/containerd/nerdctl/actions/runs/16011790007/job/45170887890?pr=4378) 
3. Almalinux  TestPush/plain_http_with_localhost - I have seen this consistently failing across many test runs
4. Rootful canary TestImagePullSoci - This is known and currently being worked on.

apostasie:
Number 4 yeah ack

Number 3 `Content digest not found` (which you have seen on TestPush) has a long history of pain.
In a shell: containerd does garbage collect layers it should not (either because nerdctl does not hold correctly on references, or because containerd GC has issues - very likely on `convert`, which is clearly wrong).

Number 2: #4106 

Number 1: is one of these legacy tests - we should start by rewriting the test with the new test toolkit and ensure the test itself is not broken - that being said, IPFS overall is not stable IMHO, and our compose implementation (shelling out) is a serious problem - so, I would not expect miracles out of the test rewrite. Also, IPFS might involve number 3 as well.

apostasie:
Content digest issue maybe starts here: #4062 (and then a bunch of linked issues)
We did fix a number of cases, but clearly there are more.
While there is still clearly a containerd issue on convert that is unfixed (PR upstream was not accepted), it seemed like we did have a workaround, but apparently not...

apostasie:
@swagatbora90 generally speaking, here is the central collection point for all CI evil: #4120

apostasie:
Correction: kube issue is #4364

Shubhranshu153:
@AkihiroSuda  This one too if we can add to milestone 2.1.3.
Thanks

Shubhranshu153 (approved):

swagatbora90:
The failed test does not appear to be related to the changes
```
=== Failed
=== FAIL: cmd/nerdctl/image TestPush/with_hosts_dir,_with_login (0.87s)
    image_push_linux_test.go:282: 
```
    
and is related to https://github.com/containerd/nerdctl/pull/4062

AkihiroSuda (approved):
Thanks

Issue #4397, opened 2025-07-03
cmd/image: save command usage lacks required IMAGE argument documentation


ChengyuZhu6:
### Description

The `nerdctl save` command usage documentation is incomplete and doesn't clearly indicate that `IMAGE` arguments are required, which can confuse users about the proper command syntax.
```
$ nerdctl save --help
The archive implements both Docker Image Spec v1.2 and OCI Image Spec v1.0.

Usage: nerdctl save [flags]

Flags:
      --all-platforms      Export content for all platforms
  -h, --help               help for save
  -o, --output string      Write to a file, instead of STDOUT
      --platform strings   Export content for a specific platform

See also 'nerdctl --help' for the global flags such as '--namespace', '--snapshotter', and '--cgroup-manager'.

$ nerdctl image save --help
The archive implements both Docker Image Spec v1.2 and OCI Image Spec v1.0.

Usage: nerdctl image save [flags]

Flags:
      --all-platforms      Export content for all platforms
  -h, --help               help for save
  -o, --output string      Write to a file, instead of STDOUT
      --platform strings   Export content for a specific platform

See also 'nerdctl --help' for the global flags such as '--namespace', '--snapshotter', and '--cgroup-manager'.
```

### Steps to reproduce the issue

1. nerdctl save --help
2. nerdctl image save --help

### Describe the results you received and expected

```
# nerdctl save --help
The archive implements both Docker Image Spec v1.2 and OCI Image Spec v1.0.

Usage: nerdctl save [flags] IMAGE [IMAGE...]

Flags:
      --all-platforms      Export content for all platforms
  -h, --help               help for save
  -o, --output string      Write to a file, instead of STDOUT
      --platform strings   Export content for a specific platform

See also 'nerdctl --help' for the global flags such as '--namespace', '--snapshotter', and '--cgroup-manager'.

# nerdctl image save --help
The archive implements both Docker Image Spec v1.2 and OCI Image Spec v1.0.

Usage: nerdctl image save [flags] IMAGE [IMAGE...]

Flags:
      --all-platforms      Export content for all platforms
  -h, --help               help for save
  -o, --output string      Write to a file, instead of STDOUT
      --platform strings   Export content for a specific platform

See also 'nerdctl --help' for the global flags such as '--namespace', '--snapshotter', and '--cgroup-manager'.
```

### What version of nerdctl are you using?

2.1.2

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #4398, opened 2025-07-03
cmd/image: update save command usage string

ChengyuZhu6:
Show proper syntax with flags and image arguments in help text.

Fixes: #4397


AkihiroSuda (approved):
Thanks

PR #4400, opened 2025-07-03
[WIP]: build tags

apostasie:
TBD


.github/workflows/job-build.yml
```diff
@@ -104,7 +104,12 @@ jobs:
...
           go mod vendor
-          rm -rf vendor/github.com/ipfs vendor/github.com/multiformats
-          BUILDTAGS=no_ipfs make binaries
+          rm -rf \
+            vendor/github.com/ipfs \
+            vendor/github.com/multiformats \
+            vendor/github.com/containerd/nydus-snapshotter \
+            vendor/github.com/containerd/stargz-snapshotter \
+            vendor/github.com/containerd/accelerated-container-image
+          BUILDTAGS=no_ipfs,no_nydus,no_esgz,no_obd make binaries
```
AkihiroSuda, line 115: Tags should be named `no_${SNAPSHOTTER_NAME}`
```suggestion
          BUILDTAGS=no_ipfs,no_nydus,no_stargz,no_overlaybd make binaries
```

.github/workflows/job-build.yml
```diff
@@ -104,7 +104,12 @@ jobs:
         name: "Run: make binaries with custom BUILDTAGS"
         run: |
           set -eux
-          # no_ipfs: make sure it does not incur any IPFS-related dependency
+          # Disabling all optional features and ensuring this does not incur any of the related dependencies
           go mod vendor
-          rm -rf vendor/github.com/ipfs vendor/github.com/multiformats
-          BUILDTAGS=no_ipfs make binaries
```
AkihiroSuda, line 110: This one should be still tested

.github/workflows/job-build.yml
apostasie: esgz -> stargz - I was wondering - we might want to do that generally (eg: on var and func names as well).

.github/workflows/job-build.yml
apostasie: They all are tested below, right?

apostasie:
Thanks @AkihiroSuda for the early feedback.
There is still a good amount of work TBD here before it can be considered for merge.

I will try to get that done tonight before I fly to EU for vacation tomorrow.

.github/workflows/job-build.yml
AkihiroSuda: No, we should still make sure that compilation of `no_ipfs` doesn't need explicitly specifying `no_stargz`, and vice versa.

.github/workflows/job-build.yml
AkihiroSuda: Maybe yes, but it can be discussed separately, as var names and func names are not visible to users and package maintainers.

.github/workflows/job-build.yml
apostasie: I see what you mean.
I'll separate them.

apostasie:
Not going to work on this anytime soon. Closing for now.

Issue #4119, opened 2025-04-17
[LEGACY] [CI]: stty output not always there (EL8 only?)


apostasie:
### Description

Sometimes, calling `stty` does not seem to output anything.
Witnessed on EL8.

Can stty fail silently?
Why would it return nothing?

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

```
=== RUN   TestExecTTY/stty_with_-t
=== PAUSE TestExecTTY/stty_with_-t
=== CONT  TestExecTTY/stty_with_-t
    container_exec_linux_test.go:110: 
        
        +============================================================================================================+
        | 🚀      | "TestExecTTY/stty_with_-t": starting test!                                                       |
        +============================================================================================================+
        | temp    | /tmp/TestExecTTYstty_with_-t24[778](https://github.com/containerd/nerdctl/actions/runs/14520021992/job/40738434866#step:10:779)13565/001                                                       |
        +============================================================================================================+
        | config  | {                                                                                                |
        |         |   "Namespace": "nerdctl-test",                                                                   |
        |         |   "PrivateMode": ""                                                                              |
        |         | }                                                                                                |
        +============================================================================================================+
        | data    | {                                                                                                |
        |         |   "container_name": "testexectty-3d80652a"                                                       |
        |         | }                                                                                                |
        +============================================================================================================+
        
    container_exec_linux_test.go:110: 
        
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl exec -t testexectty-3d80652a stty                                      |
        +------------------------------------------------------------------------------------------------------------+
        | 🌱      | XDG_SESSION_TYPE=tty                                                                             |
        |         | HOME=/home/rootless                                                                              |
        |         | XDG_SESSION_CLASS=user                                                                           |
        |         | XDG_SESSION_ID=4                                                                                 |
        |         | XDG_RUNTIME_DIR=/run/user/1001                                                                   |
        |         | PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin |
        |         | :/bin:/usr/games:/usr/local/games:/snap/bin                                                      |
        |         | DOCKER_CONFIG=/tmp/TestExecTTYstty_with_-t247[781](https://github.com/containerd/nerdctl/actions/runs/14520021992/job/40738434866#step:10:782)3565/001                                         |
[...]

PR #4410, opened 2025-07-11
fix: call wait before start

jonyhy96:
REF: https://github.com/containerd/containerd/blob/main/docs/getting-started.md#task-wait-and-start

> You always want to make sure you Wait before calling Start on a task. This makes sure that you do not encounter any races if the task has a simple program like /bin/true that exits promptly after calling start.

### DIFF
--- a/cmd/nerdctl/container/container_run.go
+++ b/cmd/nerdctl/container/container_run.go
@@ -435,6 +435,12 @@ func runAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
How to test this?

apostasie:
@AkihiroSuda I am not sure how we could test that - but this might be the reason we are seeing non-sensical failures on slow environments (like EL) where a command being ran seemingly returns nothing (for example #4119)?

Unless something is objectionable here, I would vote for merging this even without tests.

2 cents.

AkihiroSuda (approved):
Thanks

PR #4420, opened 2025-07-23
[CI]: pass github token to tasks querying the API

apostasie:
We are hitting Github throttling on some runs.

https://github.com/containerd/nerdctl/actions/runs/16476818588/job/46580573877?pr=4419#step:3:26

This PR does ensure we pass the token to tasks that query the API.


AkihiroSuda (approved):
Thanks

PR #4428, opened 2025-07-28
Limit registry connections to prevent 503 errors during push

Rozzii:
Add HTTP client configuration with connection limits to prevent registry overload when pushing images. This addresses 503 errors that occur when too many parallel requests are made to registries.

Changes:
- Add MaxConnsPerHost (default: 5) and MaxIdleConns (default: 50) options
- Add RequestTimeout configuration for large uploads
- Apply limits to both HTTPS and HTTP fallback scenarios
- Maintain backward compatibility with existing configurations


Rozzii:
PR is not relevant anymore since the new containerd image transfer system is used.

Issue #4436, opened 2025-08-01
Track development: nerdctl manifest annotate


ChengyuZhu6:
This issue is created to track the development progress of `nerdctl manifest annotate`.

PR #4442, opened 2025-08-04
manifest: support nerdctl manifest annotate command 

ChengyuZhu6:
Fixes: https://github.com/containerd/nerdctl/issues/4436

### DIFF
--- a/cmd/nerdctl/manifest/manifest.go
+++ b/cmd/nerdctl/manifest/manifest.go
@@ -35,6 +35,7 @@ func Command() *cobra.Command {
 	cmd.AddCommand(
 		InspectCommand(),
 		CreateCommand(),
+		AnnotateCommand(),
 	)
 
 	return cmd
--- a/cmd/nerdctl/manifest/manifest_annotate.go
+++ b/cmd/nerdctl/manifest/manifest_annotate.go
@@ -1,0 +1,98 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


pkg/manifestutil/manifestutils.go
```diff
@@ -34,6 +34,49 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/referenceutil"
 )
 
+type osArch struct {
+	os   string
+	arch string
+}
+
+// List of valid os/arch values
+// From: https://github.com/docker/cli/blob/master/cli/command/manifest/util.go#L22-L51
```
AkihiroSuda:
```suggestion
// From: https://github.com/docker/cli/blob/<THE ACTUAL VERSION YOU REFERRED>/cli/command/manifest/util.go#L22-L51
```

AkihiroSuda: Probably this platform list should be in https://github.com/containerd/platforms

AkihiroSuda: Or just remove the validation

ChengyuZhu6: Done.

ChengyuZhu6: > Or just remove the validation

Yes, I think we can remove the validation.

Do you think it would be appropriate to export the os and arch validation in `containerd/platform`? So that we can call the validation function from platform package?

ChengyuZhu6: > Or just remove the validation

I've removed the validation logic.

AkihiroSuda (approved):
Thanks

PR #4443, opened 2025-08-05
test:add ci for cleanup fifos

ningmingxiao:
add some ci for https://github.com/containerd/nerdctl/pull/4316

### DIFF
--- a/cmd/nerdctl/container/container_stop_linux_test.go
+++ b/cmd/nerdctl/container/container_stop_linux_test.go
@@ -199,3 +199,23 @@ func TestStopWithTimeout(t *testing.T) {
 	// The container should get the SIGKILL before the 10s default timeout
 	assert.Assert(t, elapsed < 10*time.Second, "Container did not respect --timeout flag")
 }
+func TestStopCleanupFIFOs(t *testing.T) {
+	if rootlessutil.IsRootless() {
+		t.Skip("/run/containerd/fifo/ doesn't exist on rootless")
+	}
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	testContainerName := testutil.Identifier(t)
+	oldNumFifos, err := countFIFOFiles("/run/containerd/fifo/")
+	assert.NilError(t, err)
+	// Stop the container after 2 seconds
+	go func() {
+		time.Sleep(2 * time.Second)
+		base.Cmd("stop", testContainerName).AssertOK()
+		newNumFifos, err := countFIFOFiles("/run/containerd/fifo/")
+		assert.NilError(t, err)
+		assert.Equal(t, oldNumFifos, newNumFifos)
+	}()
+	// Start a container that is automatically removed after it exits
+	base.Cmd("run", "--rm", "--name", testContainerName, testutil.NginxAlpineImage).AssertOK()
+}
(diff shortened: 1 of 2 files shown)


fahedouch:
is it possible to fix on the same PR https://github.com/containerd/nerdctl/pull/4316 ?

ningmingxiao:
> is it possible to fix on the same PR #4316 ?
pr 4316 long time noby answer, you can mege it and then merge my pr.
and i think we should call io.Close() after    @fahedouch 
```
case status := <-exitCh:

```

fahedouch:
Some suggestions:

- Close IO even when task is already Stopped:

  - In `Stop`, for the early return case `case containerd.Stopped`, consider closing IO before returning to prevent leaving FIFOs up until another path cleans them:
    * Example change:
       * Before `return nil` for `containerd.Stopped`, do:
           * `if io := task.IO(); io != nil { if err := io.Close(); err != nil { log.Warnf(...) } }`
   - Some shims/plugins may leave IO present until an explicit close; proactively closing avoids FIFO linger in this rare path.

- Log levels/messages:

   - The warning on IO close failure is good. Optionally include more context (namespace, task pid if available) for easier troubleshooting.

fahedouch:
CI is failing

ningmingxiao:
done

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -199,3 +199,23 @@ func TestStopWithTimeout(t *testing.T) {
...
+	if rootlessutil.IsRootless() {
+		t.Skip("/run/containerd/fifo/ doesn't exist on rootless")
+	}
+	testutil.DockerIncompatible(t)
+	base := testutil.NewBase(t)
+	testContainerName := testutil.Identifier(t)
+	oldNumFifos, err := countFIFOFiles("/run/containerd/fifo/")
+	assert.NilError(t, err)
+	// Stop the container after 2 seconds
+	go func() {
```
fahedouch, line 212: is it possible to avoid goroutine and make sequential check to avoid synchronisation issues and then flaky tests ?

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -199,3 +199,23 @@ func TestStopWithTimeout(t *testing.T) {
...
+		time.Sleep(2 * time.Second)
+		base.Cmd("stop", testContainerName).AssertOK()
+		newNumFifos, err := countFIFOFiles("/run/containerd/fifo/")
```
fahedouch, line 215: The test compares global counts in `/run/containerd/fifo/`. If other tests run in parallel and create/cleanup tasks, the global count may fluctuate, making the test flaky:

option: Prefer filtering by container/task ID in the FIFO names, if possible. Otherwise, add a short stabilization poll or mark the test as non-parallel and document the global dependency.

pkg/containerutil/containerutil.go
```diff
@@ -397,6 +403,13 @@ func Stop(ctx context.Context, container containerd.Container, timeout *time.Dur
 		return err
 	}
 
+	// signal will be sent once resume is finished
+	if paused {
+		if err := task.Resume(ctx); err != nil {
+			log.G(ctx).Errorf("Cannot unpause container %s: %s", container.ID(), err)
```
fahedouch:
```suggestion
			log.G(ctx).Errorf("cannot unpause container %s: %s", container.ID(), err)
```

cmd/nerdctl/container/container_stop_linux_test.go
ningmingxiao: if don't use t.Parallel()  the test cases will execute serially. if this testcase is running  other test case with  t.Parallel()  will not run.

Issue #4466, opened 2025-08-18
Manifest: manifest push should error on cross-registry manifest lists


ChengyuZhu6:
When creating a manifest list from images sourced from multiple registries and then pushing it with nerdctl (`nerdctl manifest push`), the push can fail with an error that does not clearly indicate a cross-registry issue. The expected behavior is for nerdctl to reject cross-registry sources either at manifest creation or at push time, with a clear, Docker-like error message.

docker: 
```
# docker manifest create quay.io/chengyuzhu6/test-image:test123 ghcr.io/stargz-containers/alpine@sha256:e103c1b4bf019dc290bcc7aca538dc2bf7a9d0fc836e186f5fa34945c5168310
Created manifest list quay.io/chengyuzhu6/test-image:test123
# docker manifest push quay.io/chengyuzhu6/test-image:test123
cannot use source images from a different registry than the target image: ghcr.io != quay.io"
```
nerdctl: 
```
# nerdctl manifest create quay.io/chengyuzhu6/test-image:test123 ghcr.io/stargz-containers/alpine@sha256:e103c1b4bf019dc290bcc7aca538dc2bf7a9d0fc836e186f5fa34945c5168310
Created manifest list quay.io/chengyuzhu6/test-image:test123
# nerdctl manifest push quay.io/chengyuzhu6/test-image:test123
FATA[0001] failed to push individual manifests: failed to push manifest quay.io/stargz-containers/alpine@sha256:e103c1b4bf019dc290bcc7aca538dc2bf7a9d0fc836e186f5fa34945c5168310: failed to create content writer: unexpected status from HEAD request to https://quay.io/v2/stargz-containers/alpine/manifests/sha256:e103c1b4bf019dc290bcc7aca538dc2bf7a9d0fc836e186f5fa34945c5168310: 401 UNAUTHORIZED
```

PR #4467, opened 2025-08-18
manifest: fix manifest push error on cross-registry

ChengyuZhu6:
fix manifest push error on cross-registry
Fixes: #4466


AkihiroSuda (approved):
Thanks

PR #4481, opened 2025-08-22
ci: Load br_netfilter module in linux runners

swagatbora90:
Related: https://github.com/containerd/nerdctl/pull/4311#issuecomment-3212186169

While testing out a CNI firewall plugin feature, I found that the tests for the feature were failing since because the CI ubuntu runners do not have br-netfilter module loaded, which is required for this feature to work. We need the bridged traffic to go through netfilter.

This updates the runner to load the br_netfilter module.


.github/workflows/job-test-in-container.yml
```diff
@@ -70,6 +70,12 @@ jobs:
       - name: "Init: expose GitHub Runtime variables for gha"
         uses: crazy-max/ghaction-github-runtime@3cb05d89e1f492524af3d41a1c98c83bc3025124  # v3.1.0
 
+      - name: "Init: install br-netfilter"
+        run: |
+          # This ensures that bridged traffic goes through netfilter
+          sudo apt-get update -qq
+          sudo apt-get install -qq nftables
```
AkihiroSuda: How is nftables relevant ?

swagatbora90: Yeah not relevant, so I have removed it. This was more for debugging to check if the issue was due to having different iptables backend.

.github/workflows/job-test-in-container.yml
```diff
@@ -70,6 +70,11 @@ jobs:
       - name: "Init: expose GitHub Runtime variables for gha"
         uses: crazy-max/ghaction-github-runtime@3cb05d89e1f492524af3d41a1c98c83bc3025124  # v3.1.0
 
+      - name: "Init: install br-netfilter"
+        run: |
+          # This ensures that bridged traffic goes through netfilter
+          sudo apt-get update -qq
```
AkihiroSuda: this apt-get is not needed

PR #4486, opened 2025-08-27
ci: install br_netfilter for test-in-lima jobs

swagatbora90:
In #4481, we added loading br_netfilter modules for jobs running in containers and host. However, I missed adding it for the jobs that run inside a lima VM causing certain tests (flaky almalinux) to fail.

cc @AkihiroSuda

### DIFF
--- a/.github/workflows/job-test-in-lima.yml
+++ b/.github/workflows/job-test-in-lima.yml
@@ -75,6 +75,10 @@ jobs:
           docker info
           docker version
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

swagatbora90:
I am not sure why my local tests did not catch this earlier, but somehow the enable_icc tests were getting skippes/not run. Leading me to think that this is only for the flaky tests.  It is only in a recent run, I found the tests failing. Hopefully this covers all the cases.

PR #4500, opened 2025-09-02
support all-platforms flag with soci convert

swagatbora90:
Add support for  `--all-platforms` flag with `soci cconvert`. 

`--all-platforms` flag is already supported in `nerdctl image convert` but the same was not being passed while invoking convert option with soci.


swagatbora90:
PTAL @Shubhranshu153

AkihiroSuda (approved):
Thanks

PR #4517, opened 2025-09-16
fix:ci panic because res=nil

ningmingxiao:


pkg/testutil/testutil.go
```diff
@@ -397,6 +397,7 @@ func (c *Cmd) AssertFail() {
 func (c *Cmd) AssertExitCode(exitCode int) {
 	c.Base.T.Helper()
 	res := c.runIfNecessary()
+	assert.Equal(c.Base.T, res == nil, false)
```
djdongjin, line 400:
```suggestion
	assert.NotNil(c.Base.T, res)
```

djdongjin (changes requested):

pkg/testutil/testutil.go
ningmingxiao: I find assert dosen't have func NotNil @djdongjin

djdongjin (approved):

pkg/testutil/testutil.go
AkihiroSuda: https://pkg.go.dev/gotest.tools/v3/assert#Assert seems the right choice here

Issue #4525, opened 2025-09-25
Regression: `<DATAROOT>/<ADDRHASH>/names/<NAMESPACE>` is accidentally created as `<DATAROOT>/<ADDRHASH>/<NAMESPACE>`


AkihiroSuda:
Starting with v2.0, `<DATAROOT>/<ADDRHASH>/names/<NAMESPACE>` is accidentally created as `<DATAROOT>/<ADDRHASH>/<NAMESPACE>`.

e.g.,
```
/var/lib/nerdctl/1935db59/default/foo
```

- - -

Caused by https://github.com/containerd/nerdctl/commit/bf89c084719237ea3f7aca1fa0d64bcb9f2f7df4#diff-7e8241ea95018a415ccbb4d0697f2842179f62d2a4b968671d4ebdb962f2e11aL32

Fixing this regression can be also a breaking change for existing containers, so this regression should not be fixed in a patch release?

AkihiroSuda:
> > ### Breaking Change Warning
> > This fix introduces a **breaking change** because existing containers will no longer be recognized in the corrected path. Affected users must:
> > 
> > * Migrate existing name data to the new path
> > 
> > Due to the breaking fix, this may not be included in a patch release.
> 
> 
> This change actually does not seem so breaking, and migration does not seem necessary.
> 
> The namestore is basically only used for detecting name collisions, and a dangling old namestore directory doesn't waste so much disk space.
> 

_Originally posted by @AkihiroSuda in https://github.com/containerd/nerdctl/issues/4535#issuecomment-3473395319_

PR #4535, opened 2025-10-09
Fix: namestore directory regression - restore 'names' subdirectory in path

yankay:
## Description

fix: #4525

This PR addresses a critical regression in the namestore directory creation logic, ensuring that container names are stored in the correct subdirectory structure.

### Problem

Starting with v2.0, the namestore directory has been incorrectly created as `<DATAROOT>/<ADDRHASH>/<NAMESPACE>` (e.g., `/var/lib/nerdctl/1935db59/default/foo`) instead of the intended `<DATAROOT>/<ADDRHASH>/names/<NAMESPACE>`. This was caused by [commit bf89c08](https://github.com/containerd/nerdctl/commit/bf89c08#diff-7e8241ea95018a415ccbb4d0697f2842179f62d2a4b968671d4ebdb962f2e11aL32), where the "names" subdirectory was inadvertently omitted from the path.

### Breaking Change Warning

This fix introduces a **breaking change** because existing containers will no longer be recognized in the corrected path. Affected users must:
- Migrate existing name data to the new path

Due to the breaking fix, this may not be included in a patch release.

### DIFF
--- a/pkg/namestore/namestore.go
+++ b/pkg/namestore/namestore.go
@@ -40,7 +40,7 @@ func New(stateDir, namespace string) (NameStore, error) {
 		return nil, errors.Join(ErrNameStore, store.ErrInvalidArgument)
 	}
 
-	st, err := store.New(filepath.Join(stateDir, namespace), 0, 0)
+	st, err := store.New(filepath.Join(stateDir, "names", namespace), 0, 0)
 	if err != nil {
 		return nil, errors.Join(ErrNameStore, err)
 	}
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
> ### Breaking Change Warning
> This fix introduces a **breaking change** because existing containers will no longer be recognized in the corrected path. Affected users must:
> 
> * Migrate existing name data to the new path
> 
> Due to the breaking fix, this may not be included in a patch release.


This change actually does not seem so breaking, and migration does not seem necessary.

The namestore is basically only used for detecting name collisions, and a dangling old namestore directory doesn't waste so much disk space.

PR #4542, opened 2025-10-15
healthcheck: fix path issues and add default config values

coderbirju:
This fixes a couple of issues we saw with the healthcheck automation
- Fixed PATH resolution errors by providing an explicit path to the nerdctl
binary in systemd service files, eliminating `'nerdctl' not found in PATH` errors
during healthcheck execution.
- Added default values for unspecified healthcheck flags to prevent silent
failures and ensure consistent behavior regardless of user-provided flags

### DIFF
--- a/pkg/healthcheck/healthcheck_manager_linux.go
+++ b/pkg/healthcheck/healthcheck_manager_linux.go
@@ -56,7 +56,13 @@ func CreateTimer(ctx context.Context, container containerd.Container, cfg *confi
 	// Always use health-interval for timer frequency
 	cmdOpts = append(cmdOpts, "--unit", containerID, "--on-unit-inactive="+hc.Interval.String(), "--timer-property=AccuracySec=1s")
 
-	cmdOpts = append(cmdOpts, "nerdctl", "container", "healthcheck", containerID)
+	// Get the full path to the current nerdctl binary
+	nerdctlPath, err := os.Executable()
+	if err != nil {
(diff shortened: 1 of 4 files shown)


coderbirju:
@AkihiroSuda can we add this to milestone [v2.1.7](https://github.com/containerd/nerdctl/milestone/57)?

pkg/healthcheck/healthcheck_manager_linux.go
```diff
@@ -56,7 +56,14 @@ func CreateTimer(ctx context.Context, container containerd.Container, cfg *confi
 	// Always use health-interval for timer frequency
 	cmdOpts = append(cmdOpts, "--unit", containerID, "--on-unit-inactive="+hc.Interval.String(), "--timer-property=AccuracySec=1s")
 
-	cmdOpts = append(cmdOpts, "nerdctl", "container", "healthcheck", containerID)
+	// Get the full path to the current nerdctl binary
+	nerdctlPath, err := os.Executable()
+	if err != nil {
+		log.G(ctx).Warnf("Could not determine nerdctl executable path: %v, using 'nerdctl'", err)
```
AkihiroSuda: How does this happen? Why not return an error?

coderbirju: Fixed.

AkihiroSuda (approved):
Thanks

PR #4575, opened 2025-10-31
add global options to healthcheck command

coderbirju:
Currently the global flags passed to nerdclt is not passed to the healthcheck command, this results in container not found errors while running healthchecks in a namespaced environement

This PR passes the global options to the healthcheck command as well

### DIFF
--- a/pkg/healthcheck/healthcheck_manager_linux.go
+++ b/pkg/healthcheck/healthcheck_manager_linux.go
@@ -48,9 +48,18 @@ func CreateTimer(ctx context.Context, container containerd.Container, cfg *confi
 	containerID := container.ID()
 	log.G(ctx).Debugf("Creating healthcheck timer unit: %s", containerID)
 
+	// Set all environment variables so that they are available for the nerdctl commands run via the systemd service file
 	cmdOpts := []string{}
 	if path := os.Getenv("PATH"); path != "" {
 		cmdOpts = append(cmdOpts, "--setenv=PATH="+path)
+	}
+
+	if nerdctlToml := os.Getenv("NERDCTL_TOML"); nerdctlToml != "" {
+		cmdOpts = append(cmdOpts, "--setenv=NERDCTL_TOML="+nerdctlToml)
+	}
+
+	if buildKitHost := os.Getenv("BUILDKIT_HOST"); buildKitHost != "" {
+		cmdOpts = append(cmdOpts, "--setenv=BUILDKIT_HOST="+buildKitHost)
 	}
 
 	// Always use health-interval for timer frequency


pkg/healthcheck/healthcheck_manager_linux.go
```diff
@@ -62,7 +62,11 @@ func CreateTimer(ctx context.Context, container containerd.Container, cfg *confi
 		return fmt.Errorf("could not determine nerdctl executable path: %v", err)
 	}
 
-	cmdOpts = append(cmdOpts, nerdctlPath, "container", "healthcheck", containerID)
+	cmdOpts = append(cmdOpts, nerdctlPath,
```
Shubhranshu153: what about --userns-remap (this seems this can potentially elevate privilege without the option )
And also lets do a check for all other global flags that are needed

https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#global-flags

pkg/healthcheck/healthcheck_manager_linux.go
AkihiroSuda, line 1: The pattern used in Compose should be followed:

https://github.com/containerd/nerdctl/blob/804a8cd0c38bf11188d35265f09c7e5a4e00a8ef/cmd/nerdctl/compose/compose.go#L73

pkg/healthcheck/healthcheck_manager_linux.go
```diff
@@ -62,7 +62,47 @@ func CreateTimer(ctx context.Context, container containerd.Container, cfg *confi
...
+
+	cmdOpts = append(cmdOpts, "container", "healthcheck", containerID)
+
+	if log.G(ctx).Logger.IsLevelEnabled(log.DebugLevel) {
+		cmdOpts = append(cmdOpts, "--debug")
+	}
+
 	if log.G(ctx).Logger.IsLevelEnabled(log.DebugLevel) {
 		cmdOpts = append(cmdOpts, "--debug")
 	}
```
AkihiroSuda: --debug seems added twice

pkg/healthcheck/healthcheck_manager_linux.go
coderbirju: Thanks @AkihiroSuda  for the suggestion, I have made chages as requested. We now pass down the options from the CLI during health-check creation.
Most changed files are for passing these new values to the function

PTAL
Can we also add this to [v2.2.0](https://github.com/containerd/nerdctl/milestone/57)

pkg/healthcheck/healthcheck_manager_linux.go
```diff
@@ -56,18 +56,11 @@ func CreateTimer(ctx context.Context, container containerd.Container, cfg *confi
...
-	cmdOpts = append(cmdOpts, nerdctlPath, "container", "healthcheck", containerID)
-	if log.G(ctx).Logger.IsLevelEnabled(log.DebugLevel) {
-		cmdOpts = append(cmdOpts, "--debug")
-	}
+	cmdOpts = append(cmdOpts, nerdctlCmd)
+	cmdOpts = append(cmdOpts, nerdctlArgs...)
+	cmdOpts = append(cmdOpts, "container", "healthcheck", containerID)
 
-	log.G(ctx).Debugf("creating healthcheck timer with: systemd-run %s", strings.Join(cmdOpts, " "))
+	log.G(ctx).Infof("creating healthcheck timer with: systemd-run %s", strings.Join(cmdOpts, " "))
```
AkihiroSuda: ?

cmd/nerdctl/helpers/cobra.go
```diff
@@ -156,7 +156,11 @@ func GlobalFlags(cmd *cobra.Command) (string, []string) {
 	flagSet.VisitAll(func(f *pflag.Flag) {
 		key := f.Name
 		val := f.Value.String()
-		if f.Changed {
+		// Include flag if:
+		// 1. It was explicitly changed via CLI (highest priority), OR
+		// 2. It has a non-default value (from TOML config)
+		// This ensures both CLI flags and TOML config values are propagated
+		if f.Changed || (val != f.DefValue && val != "") {
```
Shubhranshu153, line 163: This condition i think should not be there, the function seems to return all global flags, irrespective of changed or not changed. Not sure if this condition is required.

@AkihiroSuda Do you happen to know why we have the condition for compose.

coderbirju:
Rebased and added logic to pass through any config file env variables to the systemd commands. This should enable clients like finch to utilize the nerdctl healthchecks.

Shubhranshu153:
https://github.com/containerd/nerdctl/blob/2165e30feb950635f1d22becacfa5a1121412749/docs/config.md?plain=1#L13
Should we

PR #4583, opened 2025-11-05
Refactor image management to use transfer service

### LINKED ISSUE #4573: Refactor the image management code to use the Transfer service
Since we have dropped the support for nerdctl v1.6, we should be now able to refactor the image management cod to use the Transfer service introduced in containerd v1.7

Similar to:
- https://github.com/containerd/containerd/pull/8515

ChengyuZhu6:
Fixes: #4573

### DIFF
--- a/Dockerfile.d/etc_containerd_config.toml
+++ b/Dockerfile.d/etc_containerd_config.toml
@@ -5,3 +5,12 @@ version = 2
   [proxy_plugins.stargz]
     type = "snapshot"
     address = "/run/containerd-stargz-grpc/containerd-stargz-grpc.sock"
+  [proxy_plugins.stargz.exports]
+    root = "/var/lib/containerd-stargz-grpc/"
+    enable_remote_snapshot_annotations = "true"
+[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+  platform = "linux"
+  snapshotter = "overlayfs"
+[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+  platform = "linux"
+  snapshotter = "stargz"
--- a/extras/rootless/containerd-rootless-setuptool.sh
+++ b/extras/rootless/containerd-rootless-setuptool.sh
@@ -404,6 +404,15 @@ cmd_entrypoint_install_fuse_overlayfs() {
 		  [proxy_plugins."fuse-overlayfs"]
 		    type = "snapshot"
 		    address = "${XDG_RUNTIME_DIR}/containerd-fuse-overlayfs.sock"
+		  [proxy_plugins."fuse-overlayfs".exports]
+		    root = "${XDG_DATA_HOME}/containerd-fuse-overlayfs/"
+			enable_remote_snapshot_annotations = "true"
+		[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+			platform = "linux"
+			snapshotter = "fuse-overlayfs"
+		[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+			platform = "linux"
+			snapshotter = "overlayfs"
 		###  END  ###
 	EOT
 	INFO "Set \`export CONTAINERD_SNAPSHOTTER=\"fuse-overlayfs\"\` to use the fuse-overlayfs snapshotter."
@@ -449,6 +458,15 @@ cmd_entrypoint_install_stargz() {
 		  [proxy_plugins."stargz"]
 		    type = "snapshot"
 		    address = "${XDG_RUNTIME_DIR}/containerd-stargz-grpc/containerd-stargz-grpc.sock"
+		  [proxy_plugins.stargz.exports]
+		    root = "${XDG_DATA_HOME}/containerd-stargz-grpc/"
+		    enable_remote_snapshot_annotations = "true"
+		[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+			platform = "linux"
+			snapshotter = "stargz"
+		[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+			platform = "linux"
+			snapshotter = "overlayfs"
 		###  END  ###
 	EOT
 	INFO "Set \`export CONTAINERD_SNAPSHOTTER=\"stargz\"\` to use the stargz snapshotter."
--- a/pkg/containerdutil/version.go
+++ b/pkg/containerdutil/version.go
@@ -1,0 +1,53 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package containerdutil
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
+
+	containerd "github.com/containerd/containerd/v2/client"
+)
+
+func ServerSemVer(ctx context.Context, client *containerd.Client) (*semver.Version, error) {
+	v, err := client.Version(ctx)
+	if err != nil {
+		return nil, err
+	}
+	sv, err := semver.NewVersion(v.Version)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse the containerd version %q: %w", v.Version, err)
+	}
+	return sv, nil
+}
+
+// SupportsFullTransferService checks if the containerd version fully supports the Transfer service.
+// While containerd 1.7 has Transfer service, full support is only available in 2.0+.
+// The following features are missing in containerd 1.7:
+//   - Non-distributable artifacts support
+//   - Registry configuration options: WithHostDir(), WithDefaultScheme() etc.
+func SupportsFullTransferService(ctx context.Context, client *containerd.Client) bool {
+	sv, err := ServerSemVer(ctx, client)
+	if err != nil {
+		// If we can't determine version, assume it's an older version for safety
+		return false
+	}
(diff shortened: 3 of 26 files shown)


cmd/nerdctl/image/image_import.go
```diff
@@ -45,6 +45,7 @@ func ImportCommand() *cobra.Command {
 
 	cmd.Flags().StringP("message", "m", "", "Set commit message for imported image")
 	cmd.Flags().String("platform", "", "Set platform for imported image (e.g., linux/amd64)")
+	cmd.Flags().Bool("local", false, "Import content locally rather than using transfer service")
```
AkihiroSuda: For simplicity can we just remove the support for local mode?

ChengyuZhu6: Ok, but this means that containerd versions <= 1.6 will be unsupported starting with the next release.

AkihiroSuda: Already EOL

ChengyuZhu6: ok

ChengyuZhu6:
Most image management operations now use the transfer service:
- Standard image pull/push operations
- Multi-platform image handling  
- Progress reporting and streaming
- Basic registry authentication
- Image import/export (load/save)

### Limitations
Due to current transfer API constraints, the following features **continue to use the legacy implementation**:

#### 1. IPFS Registry and Nondistributable Artifacts

**Reason:** The transfer service API does not support custom registry resolver implementations.

**Affected features:**
- IPFS registry functionality (`ipfs://` protocol)
- Nondistributable artifact handling (e.g., Windows base layers)

**Status:** These features remain on the legacy resolver-based approach until the transfer API supports custom resolvers.

#### 2. Containerd 1.7.x Compatibility Issue

**Issue:** When using containerd 1.7.x daemons, registry options like `WithDefaultScheme("http")` and `WithHostDir()` are silently ignored.

**Technical details:**
- nerdctl uses containerd v2.2 client library which supports these options
- Options are properly serialized into protobuf messages
- However, containerd 1.7.x's `UnmarshalAny()` implementation doesn't process these fields
- Only `authorizer` and `headers` are handled in 1.7.x

**Impact:**
- Localhost registry operations may fail with: `"http: server gave HTTP response to HTTPS client"`
- Custom hosts.toml configurations are not applied on 1.7.x

ChengyuZhu6:
For limitation 2, there are two potential solutions:
1. Implement containerd version detection in nerdctl to use different code paths;
2. Backport to containerd 1.7.x 

For limitation 1, there are two potential solutions:
1. IPFS Registry and Nondistributable Artifacts remain on the legacy resolver-based approach.
2. containerd transfer API supports custom resolvers

ChengyuZhu6:
@AkihiroSuda @ktock

AkihiroSuda:
Thanks @ChengyuZhu6 

> For limitation 2, there are two potential solutions:
> 
> 1. Implement containerd version detection in nerdctl to use different code paths;
> 2. Backport to containerd 1.7.x

1.

> For limitation 1, there are two potential solutions:
> 
> 1. IPFS Registry and Nondistributable Artifacts remain on the legacy resolver-based approach.
> 2. containerd transfer API supports custom resolvers

2 in the long term, 1 in the short term

Dockerfile.d/etc_containerd_config.toml
```diff
@@ -5,3 +5,12 @@ version = 2
   [proxy_plugins.stargz]
     type = "snapshot"
     address = "/run/containerd-stargz-grpc/containerd-stargz-grpc.sock"
+  [proxy_plugins.stargz.exports]
+    root = "/var/lib/containerd-stargz-grpc/"
+    enable_remote_snapshot_annotations = "true"
+[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+  platform = "linux"
+  snapshotter = "overlayfs"
```
AkihiroSuda, line 13: What happens if this config is not specified?
i.e., what happens with the default config?

pkg/transferutil/progress.go
```diff
@@ -0,0 +1,231 @@
...
+	"strings"
+	"time"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"github.com/containerd/containerd/v2/core/transfer"
+	"github.com/containerd/containerd/v2/pkg/progress"
+)
+
+// From https://github.com/containerd/containerd/blob/v2.2.0-rc.0/cmd/ctr/commands/image/pull.go#L240-L473
```
AkihiroSuda, line 32: Can't we import the package from containerd/containerd to reduce code clones?

pkg/infoutil/infoutil.go
```diff
@@ -146,18 +145,6 @@ func ServerVersion(ctx context.Context, client *containerd.Client) (*dockercompa
 	return v, nil
 }
 
-func ServerSemVer(ctx context.Context, client *containerd.Client) (*semver.Version, error) {
```
AkihiroSuda, line 149: ?

pkg/containerdutil/version.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	sv, err := semver.NewVersion(v.Version)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse the containerd version %q: %w", v.Version, err)
+	}
+	return sv, nil
+}
+
+// SupportsFullTransferService checks if the containerd version fully supports the Transfer service.
+// While containerd 1.7 has Transfer service, full support is only available in 2.0+.
```
AkihiroSuda, line 41:
```suggestion
// While containerd 1.7 has Transfer service, full support is only available in 2.0+.
// The following features are missing in containerd 1.7:
// - [ADD MISSING FEATURES]
```

extras/rootless/containerd-rootless-setuptool.sh
```diff
@@ -404,6 +404,15 @@ cmd_entrypoint_install_fuse_overlayfs() {
...
 		    address = "${XDG_RUNTIME_DIR}/containerd-fuse-overlayfs.sock"
+		  [proxy_plugins."fuse-overlayfs".exports]
+		    root = "${XDG_DATA_HOME}/containerd-fuse-overlayfs/"
+			enable_remote_snapshot_annotations = "true"
+		[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+			platform = "linux"
+			snapshotter = "fuse-overlayfs"
+		[[plugins."io.containerd.transfer.v1.local".unpack_config]]
+			platform = "linux"
+			snapshotter = "overlayfs"
```
AkihiroSuda, line 415: Not a fan of complicating the config.
What happens if these lines are missing?
ctr doesn't need them AFAICS?

pkg/infoutil/infoutil.go
ChengyuZhu6: I initially wanted to put `SupportsFullTransferService()` in infoutil.go to consolidate the duplicate version detection logic from push.go and imgutil.go . However, this caused a circular dependency:
```
package github.com/containerd/nerdctl/v2/pkg/cmd/image
	imports github.com/containerd/nerdctl/v2/pkg/containerutil from remove.go
	imports github.com/containerd/nerdctl/v2/pkg/taskutil from containerutil.go
	imports github.com/containerd/nerdctl/v2/pkg/infoutil from taskutil.go
	imports github.com/containerd/nerdctl/v2/pkg/inspecttypes/dockercompat from infoutil.go
	imports github.com/containerd/nerdctl/v2/pkg/imgutil from dockercompat.go
	imports github.com/containerd/nerdctl/v2/pkg/infoutil from imgutil.go: import cycle not allowed
```
Created `containerdutil/version.go` as a base utility package to break the cycle. Moved both `ServerSemVer` and `SupportsFullTransferService` there since the latter depends on the former.

pkg/transferutil/progress.go
ChengyuZhu6: The progress handling code is in `cmd/ctr/commands/images`, not in `pkg/` or `core/`, so it cannot be imported directly.
And I removed `displayName()` function that splits names by '-':
```
func displayName(name string) string {
	parts := strings.Split(name, "-")
	for i := range parts {
		parts[i] = shortenName(parts[i])
	}
	return strings.Join(parts, " ")
}
```
which would break display for image names containing - (e.g., ubuntu-base )

extras/rootless/containerd-rootless-setuptool.sh
ChengyuZhu6: By default, it would be 
```
[[plugins."io.containerd.transfer.v1.local".unpack_config]]
  platform = "linux/amd64"(matches the host's OS/Architecture)
  snapshotter = "overlayfs" 
  differ = "walking"     
```

Dockerfile.d/etc_containerd_config.toml
ChengyuZhu6: By default, it would be 
```
[[plugins."io.containerd.transfer.v1.local".unpack_config]]
  platform = "linux/amd64"(matches the host's OS/Architecture)
  snapshotter = "overlayfs" 
  differ = "walking"     
```

extras/rootless/containerd-rootless-setuptool.sh
ChengyuZhu6: This config is required to specify which snapshotter the transfer service should use when unpacking images. We define multiple unpack targets:
- overlayfs for standard OCI images
- stargz / soci for lazy-pulling support

Without this, containerd defaults to basic unpacking and won't leverage stargz/soci snapshotters needed for lazy-pulling.

And if we only configure stargz in unpack_config, then all images would be forced to use the stargz snapshotter during transfer/unpacking, even standard OCI images that don't support lazy-pulling.

AkihiroSuda: Can you add that as comment lines, and update the relevant docs?
What about soci, nydus, etc?

ChengyuZhu6: Sure.

ChengyuZhu6: Done.

pkg/containerdutil/version.go
ChengyuZhu6: Done.

AkihiroSuda (approved):
Thanks

PR #4603, opened 2025-11-13
fix: Ensure logs are not interrupted when disk is full

ningmingxiao:
fix   /var/lib/nerdctl dir is full and container's log can't continue even space is freed up. @AkihiroSuda 

containerd loguri=file https://github.com/containerd/containerd/pull/12488 also have this problem

### DIFF
--- a/pkg/logging/jsonfile/jsonfile.go
+++ b/pkg/logging/jsonfile/jsonfile.go
@@ -18,12 +18,14 @@ package jsonfile
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	timetypes "github.com/docker/docker/api/types/time"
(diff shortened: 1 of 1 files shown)


pkg/logging/jsonfile/jsonfile.go
```diff
@@ -43,8 +45,25 @@ func Path(dataStore, ns, id string) string {
...
 }
 
+type discardWriter struct {
+	writer io.Writer
+}
+
+func (pw *discardWriter) Write(p []byte) (int, error) {
+	n, err := pw.writer.Write(p)
+	if err != nil && errors.As(err, new(syscall.Errno)) {
+		return len(p), nil
```
AkihiroSuda, line 55: What are you trying to do in this PR?
Doesn't seem correct to discard all syscall errors

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/4603/files#r2533213209

pkg/logging/jsonfile/jsonfile.go
ningmingxiao: sorry, I make a mistack. @AkihiroSuda

pkg/logging/jsonfile/jsonfile.go
```diff
@@ -43,8 +45,26 @@ func Path(dataStore, ns, id string) string {
 	return filepath.Join(dataStore, "containers", ns, id, id+"-json.log")
 }
 
+type discardWriter struct {
+	writer io.Writer
+}
+
+func (pw *discardWriter) Write(p []byte) (int, error) {
+	n, err := pw.writer.Write(p)
+	if err != nil && errors.Is(err, syscall.ENOSPC) {
```
AkihiroSuda, line 54: In the case of ENOSPC, the write should be retried with sleep?

ningmingxiao: I afraid that if long term retry failed the container main process may hang, because pipe is full.

AkihiroSuda: No need to retry forever

ningmingxiao: how long do you think is appropriate? @AkihiroSuda

ningmingxiao: If retry with some sleep I afraid that the consumer's consumption of container logs is slower than the producer's production of logs.

PR #4612, opened 2025-11-23
fix: split else-if to avoid identical-branches lint error

opjt:
## Greetings

This is my first contribution to the project. Thank you for maintaining such a high-quality codebase and for the opportunity to participate. I hope to continue contributing to other areas of the project going forward.

## Summary

While running make fix, I encountered a lint issue reported by revive regarding duplicated conditional branches in getUserAndGroup. To address this, I refactored the conditional logic to eliminate the duplicated else if branch without changing the function's behavior.

## What was changed

Split the duplicated else if into an independent if statement.

No functional change; only improves readability and resolves the lint warning.

### Thank you

Thank you in advance for reviewing this change, and please let me know if any adjustments are needed.

### DIFF
--- a/pkg/cmd/container/create_userns_opts_linux.go
+++ b/pkg/cmd/container/create_userns_opts_linux.go
@@ -324,7 +324,8 @@ func getUserAndGroup(spec string) (user.User, user.Group, error) {
 	parts := strings.Split(spec, ":")
 	if len(parts) > 2 {
 		return user.User{}, user.Group{}, fmt.Errorf("invalid identity mapping format: %s", spec)
-	} else if len(parts) == 2 && (parts[0] == "" || parts[1] == "") {
+	}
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
> lint

Can we enable it in https://github.com/containerd/nerdctl/blob/main/.golangci.yml

opjt:
@AkihiroSuda 
> > lint
> 
> Can we enable it in https://github.com/containerd/nerdctl/blob/main/.golangci.yml

Sorry about that.
It turns out the issue was caused by me using a newer version of golangci-lint.
After switching to the version specified in the Makefile (golangci-lint 2.4.0), the lint warnings disappeared.

``` bash
# my golangci-lint version
golangci-lint version                                                                                                                                                      
golangci-lint has version 2.6.2 built with go1.25.4 from dc16cf4 on 2025-11-14T02:47:46Z
```

PR #4615, opened 2025-11-24
Default net.ipv4.ip_unprivileged_port_start to 0 inside containers

### LINKED ISSUE #4595: Default the sysctl `net.ipv4.ip_unprivileged_port_start` to `0`

This FAQ entry:

https://github.com/containerd/nerdctl/blob/2165e30feb950635f1d22becacfa5a1121412749/docs/faq.md?plain=1#L319-L323

There is also [this somewhat related issue regarding `--cap-add NET_BIND_SERVICE`](https://github.com/containerd/nerdctl/issues/2439#issuecomment-1865495690) with extra details (_I chimed in Dec 2023, which would be containerd 1.x_).

The sysctl is namespaced, affecting binding of a port within the container, not the port on the host. It's generally considered a safe default to lower this down to `0` for containers.

Default `net.ipv4.ip_unprivileged_port_start` to `0`.

It is considered safe:
- Podman and Docker already lower this to `0` by default and have for some time now.
- Containerd allows it since 2.0, but only via the CRI plugin which `nerdctl` does not use.
- Kubernetes does not yet default to it, but it is considered a safe sysctl. Other tools in the k8s ecosystem (_such as `k3s`, `kind`, `minikube`_) do however default enable this setting.

### Additional context

`nerdctl` was one of the few [I came across that hasn't adopted this change](https://github.com/coredns/coredns/pull/7435#issuecomment-3162622921) - yet has a FAQ entry about it (_and only one report where a user encountered an issue_).

It's mostly a convenience, but it does help avoid a practice of images relying on `setcap` to enforce the associated capability for non-root processes to use. Which as my prior link details, when such images make a capability mandatory it prevents dropping it for security reasons (_even when that capability would not be used within the container_).

yashkukrecha:
This PR makes nerdctl default the container's `net.ipv4.ip_unprivileged_port_start` sysctl to `0`, unless the user has explicitly set this sysctl via `--sysctl`.

Key changes:

- Adds a new helper `withDefaultUnprivilegedPortSysctl` in `pkg/cmd/container/container.go`.
- Applies this helper during container creation, after user-supplied sysctls are parsed.
- If the user passes a `--sysctl` for `net.ipv4.ip_unprivileged_port_start`, nerdctl does not override it.

Note: Host-wide sysctl configuration and containerd-rootless-setuptool.sh were intentionally left unchanged in this PR to keep the scope focused on the container namespace default requested in the issue.

Fixes #4595

### DIFF
--- a/pkg/cmd/container/create.go
+++ b/pkg/cmd/container/create.go
@@ -27,6 +27,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -330,6 +331,10 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
 	}
 	opts = append(opts, umaskOpts...)
 
+	if !isHostNetwork(netLabelOpts) {
+		opts = append(opts, withDefaultUnprivilegedPortSysctl())
+	}
+
 	rtCOpts, err := generateRuntimeCOpts(options.GOptions.CgroupManager, options.Runtime)
 	if err != nil {
 		return nil, generateRemoveOrphanedDirsFunc(ctx, id, dataStore, internalLabels), err
@@ -561,6 +566,31 @@ func GenerateLogURI(dataStore string) (*url.URL, error) {
 	}
 
 	return cio.LogURIGenerator("binary", selfExe, args)
+}
+
+func isHostNetwork(netOpts types.NetworkOptions) bool {
+	return slices.Contains(netOpts.NetworkSlice, "host")
+}
+
+// withDefaultUnprivilegedPortSysctl ensures that containers can bind to
+// privileged ports (<1024) without requiring CAP_NET_BIND_SERVICE inside
+// the container by defaulting net.ipv4.ip_unprivileged_port_start to 0
+// in the container's network namespace.
+func withDefaultUnprivilegedPortSysctl() oci.SpecOpts {
+	const key = "net.ipv4.ip_unprivileged_port_start"
+	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *oci.Spec) error {
+		if s.Linux == nil {
+			s.Linux = &specs.Linux{}
+		}
+		if s.Linux.Sysctl == nil {
+			s.Linux.Sysctl = make(map[string]string)
+		}
+
+		if _, exists := s.Linux.Sysctl[key]; !exists {
+			s.Linux.Sysctl[key] = "0"
+		}
+		return nil
+	}
 }
 
 func withNerdctlOCIHook(cmd string, args []string) (oci.SpecOpts, error) {
--- a/cmd/nerdctl/container/container_inspect_linux_test.go
+++ b/cmd/nerdctl/container/container_inspect_linux_test.go
@@ -362,8 +362,12 @@ func TestContainerInspectHostConfigDefaults(t *testing.T) {
 	assert.Equal(t, "", inspect.HostConfig.UTSMode)
 	assert.Equal(t, hc.ShmSize, inspect.HostConfig.ShmSize)
 	assert.Equal(t, hc.Runtime, inspect.HostConfig.Runtime)
-	assert.Equal(t, 0, len(inspect.HostConfig.Sysctls))
 	assert.Equal(t, 0, len(inspect.HostConfig.Devices))
+	// Sysctls can be empty or contain "net.ipv4.ip_unprivileged_port_start" depending on the environment.
+	got := len(inspect.HostConfig.Sysctls)
+	if got != 0 && got != 1 {
+		t.Fatalf("unexpected number of Sysctls entries: %d (want 0 or 1)", got)
+	}
 }
 
 func TestContainerInspectHostConfigDNS(t *testing.T) {
--- a/cmd/nerdctl/container/container_run_runtime_linux_test.go
+++ b/cmd/nerdctl/container/container_run_runtime_linux_test.go
@@ -27,3 +27,31 @@ func TestRunSysctl(t *testing.T) {
 	base := testutil.NewBase(t)
 	base.Cmd("run", "--rm", "--sysctl", "net.ipv4.ip_forward=1", testutil.AlpineImage, "cat", "/proc/sys/net/ipv4/ip_forward").AssertOutExactly("1\n")
 }
+
+func TestRunSysctl_DefaultUnprivilegedPortStart(t *testing.T) {
+	t.Parallel()
+	base := testutil.NewBase(t)
+
+	// No --sysctl flags, default network mode (non-host).
+	// We expect net.ipv4.ip_unprivileged_port_start=0 inside the container,
+	// because withDefaultUnprivilegedPortSysctl should apply the default.
+	base.Cmd(
+		"run", "--rm",
+		testutil.AlpineImage,
+		"cat", "/proc/sys/net/ipv4/ip_unprivileged_port_start",
+	).AssertOutExactly("0\n")
+}
+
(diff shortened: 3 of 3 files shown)


docs/faq.md
```diff
@@ -322,6 +322,8 @@ Set sysctl value `net.ipv4.ip_unprivileged_port_start=0` .
 
 See https://rootlesscontaine.rs/getting-started/common/sysctl/#optional-allowing-listening-on-tcp--udp-ports-below-1024
 
+`nerdctl` now defaults `net.ipv4.ip_unprivileged_port_start=0` inside the container unless you override it with `--sysctl`. This does not change the host-side sysctl, which may still need to be configured for rootless port publishing.
```
AkihiroSuda: This topic is irrelevant to rootless and should not be documented here

pkg/cmd/container/create.go
```diff
@@ -330,6 +330,10 @@ func Create(ctx context.Context, client *containerd.Client, args []string, netMa
 	}
 	opts = append(opts, umaskOpts...)
 
+	if !isHostNetwork(netLabelOpts) {
+		opts = append(opts, withDefaultUnprivilegedPortSysctl(options.Sysctl))
```
AkihiroSuda: conflicts with: 

https://github.com/containerd/nerdctl/blob/1d69ed99ff4ff6b7555e292a10daf1d0f9d7e44c/pkg/cmd/container/run_linux.go#L100

pkg/cmd/container/create.go
```diff
@@ -563,6 +567,40 @@ func GenerateLogURI(dataStore string) (*url.URL, error) {
...
+			return true
+		}
+	}
+	return false
+}
+
+func withDefaultUnprivilegedPortSysctl(sysctls []string) oci.SpecOpts {
+	const key = "net.ipv4.ip_unprivileged_port_start"
+	for _, kv := range sysctls {
+		if strings.HasPrefix(kv, key) {
```
AkihiroSuda: Why not check the exact string?

pkg/cmd/container/create.go
```diff
@@ -563,6 +567,40 @@ func GenerateLogURI(dataStore string) (*url.URL, error) {
...
+	for _, n := range netOpts.NetworkSlice {
+		if n == "host" {
+			return true
+		}
+	}
+	return false
+}
+
+func withDefaultUnprivilegedPortSysctl(sysctls []string) oci.SpecOpts {
+	const key = "net.ipv4.ip_unprivileged_port_start"
```
AkihiroSuda, line 580: Needs a comment to explain the reason

pkg/cmd/container/create.go
```diff
@@ -563,6 +567,40 @@ func GenerateLogURI(dataStore string) (*url.URL, error) {
 	return cio.LogURIGenerator("binary", selfExe, args)
 }
 
+func isHostNetwork(netOpts types.NetworkOptions) bool {
+	for _, n := range netOpts.NetworkSlice {
```
AkihiroSuda: `slices.Contains` should be used

pkg/cmd/container/create_unprivileged_sysctl_test.go
AkihiroSuda, line 1: Needs to be integration tests
https://github.com/containerd/nerdctl/blob/1d69ed99ff4ff6b7555e292a10daf1d0f9d7e44c/cmd/nerdctl/container/container_run_runtime_linux_test.go#L25

pkg/cmd/container/create.go
yashkukrecha: I added the defaulting for `net.ipv4.ip_unprivileged_port_start` in `create.go`, building on top of the behavior in `run_linux.go`. Does this still conflict?

pkg/cmd/container/create.go
```diff
@@ -563,6 +568,31 @@ func GenerateLogURI(dataStore string) (*url.URL, error) {
 	return cio.LogURIGenerator("binary", selfExe, args)
 }
 
+func isHostNetwork(netOpts types.NetworkOptions) bool {
+	return slices.Contains(netOpts.NetworkSlice, "host")
+}
+
+// NOTE: withDefaultUnprivilegedPortSysctl ensures that containers can bind to
```
AkihiroSuda:
```suggestion
// withDefaultUnprivilegedPortSysctl ensures that containers can bind to
```

AkihiroSuda:
Please fix the lint errors, squash the commits, and sign off the DCO

yashkukrecha:
Hi @AkihiroSuda, should all the tests be passing?

haytok:
The tests for this update has failed.

```bash
=== Failed
=== FAIL: cmd/nerdctl/container TestContainerInspectHostConfigDefaults (0.67s)
    container_inspect_linux_test.go:343: HostConfig in TestContainerInspectHostConfigDefaults: &{ContainerIDFile: LogConfig:{Driver:json-file Opts:map[] LogURI: Address:/run/containerd/containerd.sock} PortBindings:map[] CgroupnsMode:private DNS:[] DNSOptions:[] DNSSearch:[] ExtraHosts:[] GroupAdd:[1 2 3 4 6 10 11 20 26 27] IpcMode:private OomScoreAdj:0 PidMode: ReadonlyRootfs:false Tmpfs:map[] UTSMode: ShmSize:0 Sysctls:map[net.ipv4.ip_unprivileged_port_start:0] Runtime:io.containerd.runc.v2 CPUSetMems: CPUSetCPUs: CPUQuota:0 CPUShares:0 CPUPeriod:0 CPURealtimePeriod:0 CPURealtimeRuntime:0 Memory:0 MemorySwap:0 OomKillDisable:false Devices:[] BlkioSettings:{BlkioWeight:0 BlkioWeightDevice:[] BlkioDeviceReadBps:[] BlkioDeviceWriteBps:[] BlkioDeviceReadIOps:[] BlkioDeviceWriteIOps:[]}}
    container_inspect_linux_test.go:365: assertion failed: 0 (int) != 1 (int)
```

Therefore, at least the following fixes are required:

- https://github.com/containerd/nerdctl/blob/f1591d9bd3b86b9fac9fc59dcb6069cd67142033/cmd/nerdctl/container/container_inspect_linux_test.go#L365

yashkukrecha:
Hi, the only failing job is in-host - Windows. Looking at the logs, all the `cmd/nerdctl/container` tests are failing because nerdctl run returns a non-zero exit code with:

`failed to create shim task: kernel: 'C:\Program Files\Linux Containers\kernel' not found`

This seems like an environment issue in the Windows runner rather than something caused by my PR. Does this test need to be resolved before merging my PR?

AkihiroSuda:
Can you try rebasing with the current main branch ?

AkihiroSuda:
Doesn't look rebased well
https://lima-vm.io/docs/dev/git/#rebasing-onto-upstream-master

yashkukrecha:
I rebased and it resulted in some failing flaky tests but also some failures unrelated to my PR, such as logging and binding to host port 5000.

AkihiroSuda (approved):
Not sure why Windows CI is still failing.
Maybe GHA is now deterministically picking a broken runner instance by the PR number?

pkg/cmd/container/create.go
```diff
@@ -563,6 +568,31 @@ func GenerateLogURI(dataStore string) (*url.URL, error) {
...
+		}
+		if s.Linux.Sysctl == nil {
+			s.Linux.Sysctl = make(map[string]string)
+		}
+
+		if _, exists := s.Linux.Sysctl[key]; !exists {
+			s.Linux.Sysctl[key] = "0"
+		}
+		return nil
+	}
```
AkihiroSuda, line 593: This PR seems breaking the Windows CI:
- #4664 

Probably this function shouldn't be called on non-Linux

PR #4618, opened 2025-11-25
feat: Improve namespace validation logic and enhanced inspect output

opjt:
related issue #3851 

@AkihiroSuda 

### fixes
- Added validation to require the --label flag for the update command.
- Added namespace existence checks to both the update and inspect commands.
  - (inspect) If non-existent namespaces are included, a warning log is now emitted. 
    - this logic is based on the behavior`volume-inspect`

### features
- Enhanced namespace inspect output:
  - Added Containers information (count, ids)
  - Added Images information (count, ids)
  - Added Volumes information (count, names)
- The inspect command now provides the same level of information available in namespace list.

### DIFF
--- a/cmd/nerdctl/namespace/namespace_inspect_test.go
+++ b/cmd/nerdctl/namespace/namespace_inspect_test.go
@@ -1,0 +1,156 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package namespace
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/require"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+	"github.com/containerd/nerdctl/mod/tigron/tig"
+
+	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/native"
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
+)
+
+func TestNamespaceInspect(t *testing.T) {
+	testCase := nerdtest.Setup()
+	testCase.Require = require.Not(nerdtest.Docker)
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		helpers.Ensure("namespace", "create", data.Identifier("first"))
+		helpers.Ensure("namespace", "create", "--label", "foo=fooval", "--label", "bar=barval", data.Identifier("second"))
+
+		// Create some resources in the namespaces
+		helpers.Ensure("--namespace", data.Identifier("first"), "run", "-d", "--name", data.Identifier("container1"), testutil.CommonImage)
+		helpers.Ensure("--namespace", data.Identifier("first"), "run", "-d", "--name", data.Identifier("container2"), testutil.CommonImage)
+		helpers.Ensure("--namespace", data.Identifier("second"), "run", "-d", "--name", data.Identifier("container3"), testutil.CommonImage)
+		// Create a volume
+		helpers.Ensure("--namespace", data.Identifier("first"), "volume", "create", data.Identifier("volume1"))
+
+		data.Labels().Set("ns1", data.Identifier("first"))
+		data.Labels().Set("ns2", data.Identifier("second"))
+		data.Labels().Set("container1", data.Identifier("container1"))
+		data.Labels().Set("container2", data.Identifier("container2"))
+		data.Labels().Set("container3", data.Identifier("container3"))
+		data.Labels().Set("volume1", data.Identifier("volume1"))
+	}
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("--namespace", data.Identifier("first"), "image", "rm", "-f", testutil.CommonImage)
+		helpers.Anyhow("--namespace", data.Identifier("second"), "image", "rm", "-f", testutil.CommonImage)
+		helpers.Anyhow("--namespace", data.Identifier("first"), "rm", "-f", data.Identifier("container1"))
+		helpers.Anyhow("--namespace", data.Identifier("first"), "rm", "-f", data.Identifier("container2"))
+		helpers.Anyhow("--namespace", data.Identifier("second"), "rm", "-f", data.Identifier("container3"))
+		helpers.Anyhow("--namespace", data.Identifier("first"), "volume", "rm", "-f", data.Identifier("volume1"))
+		helpers.Anyhow("namespace", "remove", data.Identifier("first"))
+		helpers.Anyhow("namespace", "remove", data.Identifier("second"))
+	}
+
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "arg missing should fail",
+			Command:     test.Command("namespace", "inspect"),
+			Expected:    test.Expects(1, []error{errors.New("requires at least 1 arg")}, nil),
+		},
+		{
+			Description: "non existent namespace returns empty array",
+			Command:     test.Command("namespace", "inspect", "doesnotexist"),
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					Output: expect.All(
+						expect.JSON([]native.Namespace{}, func(dc []native.Namespace, t tig.T) {
+							assert.Assert(t, len(dc) == 0, "expected empty array")
+						}),
+					),
+				}
+			},
+		},
+		{
+			Description: "inspect labels",
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("namespace", "inspect", data.Labels().Get("ns2"))
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					Output: expect.All(
+						expect.Contains(data.Labels().Get("ns2")),
+						expect.JSON([]native.Namespace{}, func(dc []native.Namespace, t tig.T) {
+							labels := *dc[0].Labels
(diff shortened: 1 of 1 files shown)


Shubhranshu153:
NIT: Squash the commits. (Probably 2 commits, one for the feature and one for the fix)

pkg/cmd/namespace/inspect.go
```diff
@@ -18,28 +18,134 @@ package namespace
...
+	if err != nil {
+		return info, err
+	}
+
+	info.Count = len(images)
+	ids := make([]string, info.Count)
+
+	info.IDs = ids
+	for idx, img := range images {
+		ids[idx] = strings.Split(img.Target.Digest.String(), ":")[1]
```
Shubhranshu153: NIT: Check for validity before accessing the ptr.
Although the digest regex validates the need for an tag, seems safer to do a check.

pkg/cmd/namespace/inspect.go
```diff
@@ -18,28 +18,134 @@ package namespace
...
+		result = append(result, nsInspect)
+	}
+	if err := formatter.FormatSlice(options.Format, options.Stdout, result); err != nil {
+		return err
+	}
+	for _, warn := range warns {
+		log.G(ctx).Warn(warn)
+	}
+
+	if len(warns) != 0 {
```
Shubhranshu153: should we throw an error on an warning, my understanding is we should log it and return nil or else cases lets say if there are no volumes it will throw an error.

pkg/cmd/namespace/inspect.go
```diff
@@ -18,28 +18,134 @@ package namespace
...
+			warns = append(warns, err)
+			continue
+		}
+
+		labels, err := namespaceService.Labels(ctx, ns)
 		if err != nil {
 			return err
 		}
+
+		containerInfo, err := containerInfo(ctx, client)
```
Shubhranshu153, line 51: the client and ctx wont have context of the namespaces that we are inspecting as such i dont think this logic is correct.

Can we add some integration tests to make sure we are able to inspect in a namespace that is not the default one which will be picked up. Without tests for edge cases is hard to validate the logic works properly.

pkg/cmd/namespace/inspect.go
```diff
@@ -18,28 +18,134 @@ package namespace
...
+		if err != nil {
+			warns = append(warns, err)
+		}
+
+		imageInfo, err := imageInfo(ctx, client)
```
Shubhranshu153, line 56: same as above for imageInfo.

pkg/cmd/namespace/inspect.go
opjt: like this ?
``` go
for idx, img := range images {
        digestStr := img.Target.Digest.String()
        if strings.Contains(digestStr, ":") {
	        ids[idx] = img.Target.Digest.Encoded()
        } else {
	        ids[idx] = digestStr
        }
}
```

pkg/cmd/namespace/inspect.go
opjt: I see we have the same pattern in volume inspect (https://github.com/containerd/nerdctl/blob/main/pkg/cmd/volume/inspect.go#L53). Should we keep the error return for consistency, or do you think both should be changed?

Shubhranshu153: There is a slight nuance there, the user is trying to get a particular volume so its correct to return error there.
Here we are trying to inspect the namespace so its ok to not have a volume in that namespace.

pkg/cmd/namespace/inspect.go
Shubhranshu153: we want the image ID

```
Digest is of the format:
sha256:7173b809ca12ec5dee4506cd86be934c4596dd234ee82c0662eac04a8c2c71dc

splitStr := strings.Split(img.Target.Digest.String(), ":")
if len(splitStr) >= 2 {
    ids[idx] = strings.Split(img.Target.Digest.String(), ":")[1][:12] <short hand is usually 12 characters, this is easy to read>
} else {
     warn <no digest found>
}
```

pkg/cmd/namespace/inspect.go
opjt: I'll work on adding a test case

opjt: Added test code

pkg/cmd/namespace/inspect.go
```diff
@@ -18,28 +18,135 @@ package namespace
...
+	}
+	for _, warn := range warns {
+		log.G(ctx).Warn(warn)
+	}
+
+	return nil
+}
+func containerInfo(ctx context.Context, client *containerd.Client) (native.ContainerInfo, error) {
+	info := native.ContainerInfo{}
+	containers, err := client.Containers(ctx)
```
Shubhranshu153, line 86: This is still not taking namespace as input (the context and client passed need to be created using the requested namsepace. Same is required for imageInfo

cmd/nerdctl/namespace/namespace_inspect_test.go
```diff
@@ -0,0 +1,156 @@
...
+				return helpers.Command("namespace", "inspect", data.Labels().Get("ns1"), data.Labels().Get("ns2"))
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					Output: expect.All(
+						expect.Contains(data.Labels().Get("ns1")),
+						expect.JSON([]native.Namespace{}, func(dc []native.Namespace, t tig.T) {
+							assert.Assert(t, len(dc[0].Volumes.Names) == 1, fmt.Sprintf("expected 1 volume (was %d)", len(dc[0].Volumes.Names)))
+							assert.Assert(t, dc[0].Containers.Count == 2, fmt.Sprintf("expected 2 container (was %d)", dc[0].Containers.Count))
+							assert.Assert(t, len(dc[0].Images.IDs) == 1, fmt.Sprintf("expected 1 image (was %d)", dc[0].Images.Count))
```
Shubhranshu153: These Test from what i understand is not correct, you entered 2 namespace so it should show both the the IDS, same for above.

The logic below is not right so my guess is the tests are needs to be edited as well.

pkg/cmd/namespace/inspect.go
opjt: I seem to be misunderstanding your concern. My recent manual testing of the current implementation shows that the logic correctly filters resources by namespace:

- `nerdctl namespace inspect test1 default` returns accurate counts (e.g., test1: 1 container, default: 2 containers), demonstrating that `containerInfo` and `imageInfo` are correctly scoping results using the `context.Context`.

Could you clarify if you believe the line 
> ctx = namespaces.WithNamespace(ctx, ns)

 is incorrectly used or if `containerd`'s `client.Containers(ctx)` method is known to ignore this context flag?
could you provide an example of a failure scenario or the incorrect output you are observing?

I want to ensure the code is robust against potential future changes, but based on current behavior, the filtering appears to be working as intended by propagating the namespace via the context.

Thanks again for your feedback!

Shubhranshu153: The client and context for the namespace is created as
```
client, ctx, cancel, err := clientutil.NewClient(cmd.Context(), options.GOptions.Namespace, options.GOptions.Address)
```
in file: https://github.com/containerd/nerdctl/blob/8a5888eaad22007d3f27f43296a6b045172d7c17/cmd/nerdctl/namespace/namespace_inspect.go#L4

so this is taking the global namespace and not the namespace requested in the command.
So for each namespace we are inspecting we need to specify the correct namespace we want to check.

Shubhranshu153: For the test i would rather trust the id rather than the numbers.
Also run it locally and check you see the different ids correctly.
if it works then your logic is correct.

opjt: it works
``` bash
pjt@lima-default:/Users/pjt/Projects/PJT/fork-git/nerdctl$ nerdctl namespace inspect test1 default
[
    {
        "Name": "test1",
        "Labels": {
            "tes": "3",
            "test": "1"
        },
        "Containers": {
            "count": 1,
            "ids": [
                "2e7b2558d94acc26ab4f4d8f40d6ffb5460a80a57cc1b0174c92c170b9773d27"
            ]
        },
        "Images": {
            "count": 2,
            "ids": [
                "4b7ce07002c6",
                "ec14c7992a97"
            ]
        },
        "Volumes": {
            "count": 0,
            "names": []
        }
    },
    {
        "Name": "default",
        "Labels": {
            "default": "true"
        },
        "Containers": {
            "count": 2,
            "ids": [
                "973a7436d32452fe754231fd1d68250522ac902108df4b40c8bb78139d04e83a",
                "b07920264d8e9254fb084bef494329e0eb2894893bffdb2e91eef2e6e261df46"
            ]
        },
        "Images": {
            "count": 2,
            "ids": [
                "4b7ce07002c6",
                "f7931603f70e"
            ]
        },
        "Volumes": {
            "count": 1,
            "names": [
                "12"
            ]
        }
    }
]

pjt@lima-default:/Users/pjt/Projects/PJT/fork-git/nerdctl$ nerdctl image ls --namespace test1
REPOSITORY                          TAG         IMAGE ID        CREATED         PLATFORM       SIZE       BLOB SIZE
ghcr.io/stargz-containers/alpine    3.13-org    ec14c7992a97    23 hours ago    linux/amd64    6.218MB    2.814MB
alpine                              latest      4b7ce07002c6    30 hours ago    linux/arm64    9.175MB    4.14MB
pjt@lima-default:/Users/pjt/Projects/PJT/fork-git/nerdctl$ nerdctl image ls --namespace default
REPOSITORY     TAG       IMAGE ID        CREATED       PLATFORM       SIZE       BLOB SIZE
hello-world    latest    f7931603f70e    7 days ago    linux/arm64    12.29kB    4.622kB
alpine         latest    4b7ce07002c6    7 days ago    linux/arm64    9.175MB    4.14MB
```
Is there anything I'm doing wrong?

Verifying by IDs instead of just the count is indeed a better testing. Thank you for the feedback

Shubhranshu153: its correct, i was missing this line

```
ctx = namespaces.WithNamespace(ctx, ns)
```
the context is updated.

opjt:
@AkihiroSuda When you have time, could you please take a look at this PR?

pkg/inspecttypes/native/namespace.go
AkihiroSuda, line 1: Why include IDs?

pkg/cmd/namespace/inspect.go
AkihiroSuda, line 1: This change will make inspect slower, no?

AkihiroSuda:
> features

This seems overloaded to this PR recently?
I suggest splitting this to another PR for further discussion.
By default, the namespace operations should remain lightweight.

opjt:
Thank you for the suggestion.

Regarding `namespace inspect`, I agree that it should contain more details, aligning with the discussion in [3851](https://github.com/containerd/nerdctl/issues/3851) . Is it acceptable to omit the IDs information in the output?

I also agree that this PR is currently overloaded with a mix of fixes and features, which makes it hard to review. I will split the changes into separate, more focused PRs for better reviewability and resubmit them shortly.

Issue #4622, opened 2025-11-27
canary: `TestImageConvert/soci_with_all-platforms` is failing after the release of soci-snapshotter v0.12.0: `a ztoc entry was found in the artifact store but not in the content store`


AkihiroSuda:
`TestImageConvert/soci_with_all-platforms` is failing after the release of soci-snapshotter v0.12.0 https://github.com/awslabs/soci-snapshotter/releases/tag/v0.12.0

https://github.com/containerd/nerdctl/blob/3696771649d1334b8fd7cbe9fd904c7fb32aa702/cmd/nerdctl/image/image_convert_linux_test.go#L111-L128

```
PASS cmd/nerdctl/image.TestImageConvert/soci (0.47s)
=== RUN   TestImageConvert/soci_with_all-platforms
    requirements.go:265: 
        
        +------------------------------------------------------------------------------------------------------------+
        | ➡️      | ⚙️ /usr/local/bin/nerdctl info --format {{ json . }}                                             |
        +------------------------------------------------------------------------------------------------------------+
        |         | 🟢 {"ID":"3e6f6531-15ba-4e7e-bbca-4d2b2f3a17be","Driver":"overlayfs","Plugins":{"Log":["fluentd" |
        |         | ,"journald","json-file","none","syslog"],"Storage":["native","overlayfs","stargz","soci"]},"Memo |
        |         | ryLimit":true,"SwapLimit":true,"CpuCfsPeriod":true,"CpuCfsQuota":true,"CPUShares":true,"CPUSet": |
        |         | true,"CPURealtime":false,"PidsLimit":true,"IPv4Forwarding":true,"BridgeNfIptables":true,"BridgeN |
        |         | fIp6tables":true,"OomKillDisable":false,"SystemTime":"2025-11-27T03:17:39.636389732Z","LoggingDr |
        |         | iver":"json-file","CgroupDriver":"systemd","CgroupVersion":"2","KernelVersion":"6.11.0-1018-azur |
        |         | e","OperatingSystem":"Ubuntu 24.04.3 LTS","OSType":"linux","Architecture":"x86_64","NCPU":4,"Mem |
        |         | Total":16772575232,"Name":"94342a7ca953","ServerVersion":"v2.2.0","SecurityOptions":["name=appar |
        |         | mor","name=seccomp,profile=builtin","name=cgroupns"],"Warnings":null}                            |
        +------------------------------------------------------------------------------------------------------------+
[...]

PR #4623, opened 2025-11-27
Fix TestImageConvert/soci* tests for v0.12.0 release

henry118:
Fixes: https://github.com/containerd/nerdctl/issues/4622

Temporarily fix the soci image conversion regression introduced in soci v0.12.0 release.

It looks like proper fix would require a soci code change.

This patch simply fix the nerdctl canary.

### DIFF
--- a/pkg/snapshotterutil/sociutil.go
+++ b/pkg/snapshotterutil/sociutil.go
@@ -117,6 +117,13 @@ func ConvertSociIndexV2(ctx context.Context, client *client.Client, srcRef strin
 
 	sociCmd.Args = append(sociCmd.Args, "convert")
 
+	// The following option temporarily fix the image conversion regression in SOCI v0.12.0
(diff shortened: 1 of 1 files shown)


pkg/snapshotterutil/sociutil.go
```diff
@@ -117,6 +117,12 @@ func ConvertSociIndexV2(ctx context.Context, client *client.Client, srcRef strin
 
 	sociCmd.Args = append(sociCmd.Args, "convert")
 
+	// The following option temporarily fix the image conversion issue in 0.12.0
```
AkihiroSuda: Could you put the issue URL in the comment

henry118: Done.

ChengyuZhu6 (approved):
LGTM

PR #4652, opened 2025-12-12
docs: add additional nerdtest `Requirement`

haytok:


haytok:
Although we can infer the available requirements by reading the code, having them documented makes it easier for developers—and tools (including AI)—to discover and use them.

AkihiroSuda (approved):
Thanks

Issue #4651, opened 2025-12-12
Should allow localhost DNS servers when using host network


zhangyoufu:
### Description

When a container is using host network, we should respect host's `/etc/resolv.conf` and allow DNS servers serving on loopback addresses, instead of override them with Google Public DNS. Docker did what I expect. nerdctl always override DNS servers when it found loopback addresses.

https://github.com/containerd/nerdctl/blob/246b6c18dcab05c80e391230edd68d5e367a5368/pkg/resolvconf/resolvconf.go#L180-L208

host `/etc/resolv.conf`:
```
# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8).
# Do not edit.
#
# This file might be symlinked as /etc/resolv.conf. If you're looking at
# /etc/resolv.conf and seeing this text, you have followed the symlink.
#
# This is a dynamic resolv.conf file for connecting local clients to the
# internal DNS stub resolver of systemd-resolved. This file lists all
# configured search domains.
#
# Run "resolvectl status" to see details about the uplink DNS servers
# currently in use.
#
# Third party programs should typically not access this file directly, but only
# through the symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a
# different way, replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 127.0.0.53
options edns0 trust-ad
search .
```

* `docker run --network host --rm alpine cat /etc/resolv.conf`:
```
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.

nameserver 127.0.0.53
search .
options edns0 trust-ad

# Based on host file: '/etc/resolv.conf'
# Overrides: []
```

* `nerdctl run --network host --rm alpine cat /etc/resolv.conf`:
```
INFO[0000] No non-localhost DNS nameservers are left in resolv.conf. Using default external servers: [nameserver 8.8.8.8 nameserver 8.8.4.4]
INFO[0000] IPv6 enabled; Adding default IPv6 external servers: [nameserver 2001:4860:4860::8888 nameserver 2001:4860:4860::8844]
nameserver 8.8.8.8
nameserver 8.8.4.4
nameserver 2001:4860:4860::8888
nameserver 2001:4860:4860::8844
```

### Steps to reproduce the issue

`nerdctl run --network host --rm alpine cat /etc/resolv.conf`

[...]

PR #4653, opened 2025-12-12
fix: allow localhost DNS servers when using host network

zhangyoufu:
(This change is vibed by Claude Haiku 4.5, reviewed by human.)

This commit addresses the issue where nerdctl was unconditionally stripping localhost DNS servers from /etc/resolv.conf when containers used --network=host.

Changes made:
1. Added `resolvconf.FilterResolvDNSWithLocalhostOption()`, like `resolvconf.FilterResolvDNS()` but allows controlling whether localhost nameservers are preserved.
2. Added allowLocalhostDNS parameter to fetchDNSResolverConfig()
3. Updated hostNetworkManager to use allowLocalhostDNS=true to preserve host DNS
4. Added comprehensive test coverage for the new functionality

The fix ensures:
- Host network mode respects host's /etc/resolv.conf including localhost resolvers
- Isolated networks (bridge, CNI) continue to have fallback Google DNS
- Backward compatible - all existing behavior unchanged
- Docker-compatible behavior

Fixes: #4651

### DIFF
--- a/pkg/resolvconf/resolvconf_linux_test.go
+++ b/pkg/resolvconf/resolvconf_linux_test.go
@@ -320,3 +320,100 @@ func TestFilterResolvDns(t *testing.T) {
 		}
 	}
 }
+
+func TestFilterResolvDnsWithLocalhostOption(t *testing.T) {
+	testCases := []struct {
+		name              string
+		input             string
+		allowLocalhostDNS bool
+		ipv6Enabled       bool
+		expected          string
+	}{
+		{
+			name:              "filter_disallow_localhost_ipv6_disabled",
+			input:             "nameserver 127.0.0.53\nnameserver 192.88.99.1\nnameserver ::1\nnameserver 2001:db8::1\n",
+			allowLocalhostDNS: false,
+			ipv6Enabled:       false,
+			expected:          "nameserver 192.88.99.1\n",
+		},
+		{
+			name:              "filter_allow_localhost_ipv6_disabled",
+			input:             "nameserver 127.0.0.53\nnameserver 192.88.99.1\nnameserver ::1\nnameserver 2001:db8::1\n",
+			allowLocalhostDNS: true,
+			ipv6Enabled:       false,
+			expected:          "nameserver 127.0.0.53\nnameserver 192.88.99.1\n",
+		},
+		{
+			name:              "filter_disallow_localhost_ipv6_enabled",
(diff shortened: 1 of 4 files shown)


pkg/resolvconf/resolvconf_linux_test.go
```diff
@@ -320,3 +320,96 @@ func TestFilterResolvDns(t *testing.T) {
 		}
 	}
 }
+
+func TestFilterResolvDnsWithLocalhostOption(t *testing.T) {
+	// Test 1: allowLocalhostDNS=false should strip localhost (original behavior)
```
AkihiroSuda: These should be subtests? (`t.Run`)

pkg/resolvconf/resolvconf_linux_test.go
AkihiroSuda, line 1: Can we have some integration tests too?

https://github.com/containerd/nerdctl/blob/20a3eeb4f7a49bd2c22f276189bee50b57601f0d/cmd/nerdctl/container/container_run_network_linux_test.go#L935-L978

zhangyoufu: I have implemented some simple integration tests. I don't think we should modify host `/etc/resolv.conf` content on-the-fly.

pkg/resolvconf/resolvconf_linux_test.go
zhangyoufu: done

pkg/resolvconf/resolvconf_linux_test.go
zhangyoufu: I'm investigating CI failure in rootless setup.

zhangyoufu: Previous CI failure due to:
1. Some rootless CI job make use of slirp, which provides a different DNS server than the host. The test case is modified to handle this.
2. `pkg/resolvconf` has special handling to retrieve the real DNS servers behind systemd-resolved when it sees 127.0.0.53 (systemd-resolved stub). When using host network, this special handling is unwanted. IMHO, this special handling is problematic. systemd-resolved can freely choose DNS servers from other links when netif comes and goes. We really should not stick on the "current" state of which DNS servers are choosen by systemd-resolved.

ChengyuZhu6 (approved):
LGTM

ChengyuZhu6:
Let's wait for @AkihiroSuda's review to see if he has any comments on the latest changes.

PR #4671, opened 2025-12-21
chore: improve the visibility of orphaned containers in logs

haytok:
When executing the command on compose files like the one below, orphaned containers are displayed, but the current implementation does not show the IDs of these orphaned containers, making them difficult to identify.

```
$ cat compose-full.yaml
services:
  test:
    image: docker.io/library/busybox:latest
    command: ["sleep", "infinity"]
  orphan:
    image: docker.io/library/busybox:latest
    command: ["sleep", "infinity"]

$ cat compose-orphan.yaml
services:
  test:
    image: docker.io/library/busybox:latest
    command: ["sleep", "infinity"]

$ sudo nerdctl compose -f compose-full.yaml up -d
...

$ sudo nerdctl compose -f compose-orphan.yaml down -v
...
WARN[0010] found 1 orphaned containers: [0x4000340000], you can run this command with the --remove-orphans flag to clean it up
...
```

Therefore, this commit modifies to display the IDs of orphaned containers.

Additionally, since there was other logic that performed similar displays, this commit also modifies it in the same manner.

### DIFF
--- a/pkg/composer/orphans.go
+++ b/pkg/composer/orphans.go
@@ -55,3 +55,11 @@ func (c *Composer) getOrphanContainers(ctx context.Context, parsedServices []*se
 
 	return orphanContainers, nil
 }
+
+func containerShortIDs(containers []containerd.Container) []string {
+	names := make([]string, 0, len(containers))
+	for _, c := range containers {
+		names = append(names, c.ID()[:12])
+	}
+	return names
+}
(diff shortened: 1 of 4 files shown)


pkg/composer/orphans.go
```diff
@@ -55,3 +55,11 @@ func (c *Composer) getOrphanContainers(ctx context.Context, parsedServices []*se
 
 	return orphanContainers, nil
 }
+
+func getContainerIDs(containers []containerd.Container) []string {
```
AkihiroSuda:
```suggestion
func containerShortIDs(containers []containerd.Container) []string {
```

See https://go.dev/doc/effective_go#Getters

haytok: > it's neither idiomatic nor necessary to put Get into the getter's name. (from https://go.dev/doc/effective_go#Getters)

Thanks!!! Fixed.

(I'll re-read Effective Go.)

PR #4686, opened 2026-01-10
fix: return error for invalid --pull option in nerdctl compose create

haytok:
In the current implementation, specifying an invalid string for the --pull option of the nerdctl compose create command causes it to fall back to missing, allowing the image to be pulled as shown below.

```
$ sudo nerdctl compose create --pull foo
WARN[0000] Ignoring: service svc0: pull_policy: "foo"
INFO[0000] Ensuring image alpine
...
INFO[0004] Creating container fix-compose-pull-policy-with-invalid-option-svc0-1
```

On the other hand, docker compose returns the following error in a similar situation.

```
$ docker compose create --pull foo
invalid --pull option "foo"
```

This commit fixes it to be compatible with docker compose.

### DIFF
--- a/cmd/nerdctl/compose/compose_create_linux_test.go
+++ b/cmd/nerdctl/compose/compose_create_linux_test.go
@@ -17,6 +17,7 @@
 package compose
 
 import (
+	"errors"
 	"fmt"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


haytok:
Thanks!!!

Issue #4613, opened 2025-11-24
Rewrite tests using `testutil.NewBase` to use `nerdtest.Setup` (Tigron)


AkihiroSuda:
All the tests using `testutil.NewBase` should be rewritten to use `nerdtest.Setup` (i.e., Tigron framework)

https://github.com/search?q=repo%3Acontainerd%2Fnerdctl%20testutil.NewBase&type=code
https://github.com/search?q=repo%3Acontainerd%2Fnerdctl+nerdtest.Setup&type=code

Example occurrence:
https://github.com/containerd/nerdctl/blob/1d69ed99ff4ff6b7555e292a10daf1d0f9d7e44c/cmd/nerdctl/container/container_list_test.go#L27-L62

Example PR:
- https://github.com/containerd/nerdctl/pull/4641

Similar PRs:
- https://github.com/containerd/nerdctl/pull/4145

This is a "good first issue", but needs a bunch of PRs.

- - -

> [!IMPORTANT]
> It is highly recommended to begin with updating a single **short** file.

<details open><summary>Files (46)</summary>

- [x] https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/compose/compose_create_linux_test.go https://github.com/containerd/nerdctl/pull/4670
- [x] https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/compose/compose_down_linux_test.go https://github.com/containerd/nerdctl/pull/4672
- [x] https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/compose/compose_kill_linux_test.go https://github.com/containerd/nerdctl/pull/4676
- [x] https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/compose/compose_pause_linux_test.go https://github.com/containerd/nerdctl/pull/4688
- [x] https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/compose/compose_port_linux_test.go https://github.com/containerd/nerdctl/pull/4713
- [x] https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/compose/compose_ps_linux_test.go https://github.com/containerd/nerdctl/pull/4790
- [x] https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/compose/compose_pull_linux_test.go https://github.com/containerd/nerdctl/pull/4796
[...]

PR #4688, opened 2026-01-11
test: refactor compose_pause_linux_test.go to use Tigron

haytok:
- related : #4613


ChengyuZhu6 (commented):
It looks like these steps are tightly coupled. We might want to combine them into one.

haytok:
Thanks for comments!
I've consolidated the subtests into one, so could you re-review?

PR #4704, opened 2026-01-16
feat: add support for container cp with tarballs

### LINKED ISSUE #4691: Feature Request: Support stdin/stdout streaming (-) in `nerdctl cp`
## Summary

The `nerdctl cp` command does not support using `-` (dash) as source or destination path for streaming tar archives via stdin/stdout. This feature is available in Docker CLI and is essential for efficient file transfer operations in development tooling.

## Docker Documentation

From [Docker cp reference](https://docs.docker.com/reference/cli/docker/container/cp/):

```bash
# Stream a tar archive into a container
cat archive.tar | docker cp - container:/path/
```

```bash
# Stream container contents to stdout
docker cp container:/var/log/app.log - | tar x -O | grep "ERROR"
```

## Podman Compatibility

**Podman fully supports stdin/stdout streaming:**
- Using `-` as `src_path` streams STDIN as a tar archive
- Using `-` as `dest_path` streams the resource to STDOUT

Example from Podman docs:
```bash
podman cp - containerID:/myfiles.tar.gz < myfiles.tar.gz
```

Reference: [Podman cp documentation](https://docs.podman.io/en/latest/markdown/podman-cp.1.html)

## Use Cases

1. **IDE/Editor Extensions**: VS Code Docker extension uses stdin streaming to efficiently write file contents directly to containers without creating temporary files on the host system.

2. **Backup/Restore Operations**: Streaming allows piping container data directly to compression tools or remote storage without intermediate files.

3. **CI/CD Pipelines**: Build systems often need to inject configuration or extract artifacts using piped operations for efficiency.

4. **Memory-Constrained Environments**: Streaming avoids the need to buffer entire file contents in temporary files, which is important for large files.

## Current Workaround

For VS Code Docker extension users, a workaround has been implemented in [PR #327](https://github.com/microsoft/vscode-docker-extensibility/pull/327) that:
- Uses shell commands with `tar` and `mktemp` for reading files
- Creates temporary files for writing operations
- Documents the platform assumptions (requires `/bin/sh`, `tar`, `mktemp`)

```typescript
// Current workaround for readFile
const command = `/bin/sh -c '${escapedCommandName} cp ${escapedContainer}:${escapedPath} - | tar -xOf -'`;

// Current workaround for writeFile (uses temp file)
const command = `/bin/sh -c 'TMPFILE=$(mktemp) && cat > "$TMPFILE" && ${escapedCommandName} cp "$TMPFILE" ${escapedContainer}:${escapedPath} && rm -f "$TMPFILE"'`;
```

## Current Behavior

```bash
# This doesn't work as expected
echo "content" | nerdctl cp - container:/path/file

# This also doesn't work
nerdctl cp container:/path/file -
```

## Expected Behavior

```bash
# Stream content into container
echo "file content" | tar cf - --files-from=/dev/stdin | nerdctl cp - mycontainer:/app/

# Stream content from container
nerdctl cp mycontainer:/app/config.json - | tar xf - -O
```

## Additional Context

This feature gap was identified while implementing Finch support for the VS Code Docker extension. Stdin/stdout streaming is fundamental to many Docker workflows and tooling integrations. Full Docker CLI compatibility for `cp` streaming would significantly improve the developer experience for users choosing Finch/nerdctl as their container runtime.

sondavidb:
Closes #4691 

This PR should allow support for the following in `container cp`:
- Specifying `-` in src to allow streaming the contents of a tarball into a container.
- Specifying `-` in dst to allow streaming the contents of a container into a tarball that is written to stdout

### DIFF
--- a/cmd/nerdctl/container/container_cp_linux_test.go
+++ b/cmd/nerdctl/container/container_cp_linux_test.go
@@ -19,6 +19,7 @@ package container
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"syscall"
@@ -29,6 +30,7 @@ import (
 
 	"github.com/containerd/nerdctl/v2/pkg/containerutil"
 	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
+	"github.com/containerd/nerdctl/v2/pkg/tarutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 )
@@ -51,7 +53,9 @@ const (
 	pathIsADirAbsolute       = string(os.PathSeparator) + "is-a-dir" + complexify
 	pathIsAVolumeMount       = string(os.PathSeparator) + "is-a-volume-mount" + complexify
 
-	srcFileName = "test-file" + complexify
+	srcFileName  = "test-file" + complexify
+	tarballName  = "test-tar" + complexify
+	cpFolderName = "nerdctl-cp-test"
 
 	// Since nerdctl cp must NOT obey container wd, but instead resolve paths against the root, we set this
 	// explicitly to ensure we do the right thing wrt that.
@@ -400,6 +404,49 @@ func TestCopyToContainer(t *testing.T) {
 				},
 			},
 		},
+		{
+			description:   "Copying to container, SRC_PATH is stdin",
+			sourceSpec:    "-",
+			sourceIsAFile: true,
+			toContainer:   true,
+			testCases: []testcases{
+				{
+					description:     "DEST_PATH is a directory, relative",
+					destinationSpec: pathIsADirRelative,
+					catFile:         filepath.Join(pathIsADirRelative, srcFileName),
+					setup: func(base *testutil.Base, container string, destPath string) {
+						base.Cmd("exec", container, "mkdir", "-p", destPath).AssertOK()
+					},
+				},
+				{
+					description:     "DEST_PATH is a directory, absolute",
+					destinationSpec: pathIsADirAbsolute,
+					catFile:         filepath.Join(pathIsADirAbsolute, srcFileName),
+					setup: func(base *testutil.Base, container string, destPath string) {
+						base.Cmd("exec", container, "mkdir", "-p", destPath).AssertOK()
+					},
+				},
+				{
+					description:     "DEST_PATH is stdout",
+					destinationSpec: "-",
+					expect: icmd.Expected{
+						ExitCode: 1,
+						Err:      "one of src or dest must be a container file specification",
+					},
+				},
+				{
+					description:     "DEST_PATH is a file",
+					destinationSpec: pathIsAFileAbsolute,
+					setup: func(base *testutil.Base, container string, destPath string) {
+						base.Cmd("exec", container, "touch", destPath).AssertOK()
+					},
+					expect: icmd.Expected{
+						ExitCode: 1,
+						Err:      containerutil.ErrCannotCopyDirToFile.Error(),
+					},
+				},
+			},
+		},
 	}
 
 	for _, tg := range testGroups {
@@ -532,6 +579,19 @@ func TestCopyFromContainer(t *testing.T) {
 					description:     "DEST_PATH is a directory, absolute, ending with a path separator",
 					destinationSpec: pathIsADirAbsolute + string(os.PathSeparator),
 					catFile:         filepath.Join(pathIsADirAbsolute, srcFileName),
+					expect: icmd.Expected{
+						ExitCode: 0,
+					},
+					setup: func(base *testutil.Base, container string, destPath string) {
+						err := os.MkdirAll(destPath, dirPerm)
+						assert.NilError(t, err)
+					},
+				},
+				{
+					description:     "DEST_PATH is stdout",
+					destinationSpec: "-",
+					// Extra dir to account for folder created from extracted tar file
+					catFile: filepath.Join(pathIsADirAbsolute, filepath.Base(srcDirName), srcFileName),
 					expect: icmd.Expected{
 						ExitCode: 0,
 					},
@@ -682,6 +742,19 @@ func TestCopyFromContainer(t *testing.T) {
 						assert.NilError(t, err)
 					},
 				},
+				{
+					description:     "DEST_PATH is stdout",
+					destinationSpec: "-",
+					catFile:         filepath.Join(pathIsADirAbsolute, srcDirName, srcFileName),
+					expect: icmd.Expected{
+						ExitCode: 0,
+					},
+					setup: func(base *testutil.Base, container string, destPath string) {
+						// Don't make the topmost dir as this is where the tarball must extract
+						err := os.MkdirAll(filepath.Dir(destPath), dirPerm)
+						assert.NilError(t, err)
+					},
+				},
 			},
 		},
 
@@ -704,6 +777,18 @@ func TestCopyFromContainer(t *testing.T) {
 				{
 					description:     "DEST_PATH is a directory, absolute",
 					destinationSpec: pathIsADirAbsolute,
+					catFile:         filepath.Join(pathIsADirAbsolute, srcFileName),
+					expect: icmd.Expected{
+						ExitCode: 0,
+					},
+					setup: func(base *testutil.Base, container string, destPath string) {
+						err := os.MkdirAll(destPath, dirPerm)
+						assert.NilError(t, err)
+					},
+				},
+				{
+					description:     "DEST_PATH is stdout",
+					destinationSpec: "-",
 					catFile:         filepath.Join(pathIsADirAbsolute, srcFileName),
 					expect: icmd.Expected{
 						ExitCode: 0,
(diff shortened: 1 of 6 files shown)


docs/command-reference.md
```diff
@@ -494,6 +494,8 @@ Usage:
 - `nerdctl cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-`
 - `nerdctl cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH`
 
+Using `-` as the `SRC_PATH` streams the contents of `STDIN` as a tar archive. The command extracts the content of the tar to the `DEST_PATH` in container's filesystem. In this case, `DEST_PATH` must specify a directory. Using `-` as the `DEST_PATH` streams the contents of the resource as a tar archive to `STDOUT`.
```
sondavidb, line 497: FYI, copied this exact wording from https://docs.docker.com/reference/cli/docker/container/cp/#corner-cases. LMK if this is a problem in any way.

sondavidb:
Will get to adding testing tomorrow. Having some trouble adding tests to the existing testing suite for now.

sondavidb:
I was very optimistic about the testing suite, stuffing the test cases in there took a lot of effort. This suite could certainly split up some functions (the amount of if statements in here is hard to keep track of, to say the least...)

In any case, tests passed locally so hoping this should be good to go 🙂

sondavidb:
Test failures look unrelated to changes?

pkg/containerutil/cp_resolve_linux.go
```diff
@@ -134,6 +142,14 @@ func getPathSpecFromHost(originalPath string) (*pathSpecifier, error) {
 
 // getPathSpecFromHost builds a pathSpecifier from a container location
```
Shubhranshu153: not related but can we change this to getPathSpecFromContainer

pkg/containerutil/cp_linux.go
```diff
@@ -237,21 +238,31 @@ func CopyFiles(ctx context.Context, client *containerd.Client, container contain
...
 
 	tarXDir := destinationSpec.resolvedPath
 	if !sourceSpec.isADir && !destinationSpec.endsWithSeparator && !(destinationSpec.exists && destinationSpec.isADir) {
 		tarXDir = filepath.Dir(destinationSpec.resolvedPath)
 	}
-	tarX := []string{tarBinary, "-x"}
-	if options.Container2Host && isGNUTar {
-		tarX = append(tarX, "--no-same-owner")
+	var tarX []string
+	if destinationSpec.originalPath == "-" {
```
Shubhranshu153: nit: can we add this to the containerCP option somthing like option.Stdin and option.Stdout and use that as the source of truth instead of doing this comparison for clarity.

pkg/containerutil/cp_resolve_linux.go
```diff
@@ -134,6 +142,14 @@ func getPathSpecFromHost(originalPath string) (*pathSpecifier, error) {
...
 func getPathSpecFromContainer(originalPath string, conSpec *oci.Spec, containerHostRoot string) (*pathSpecifier, error) {
+	if originalPath == "-" {
```
Shubhranshu153: this condition here is not intuitve, can we skip calling them when we have stdin and stdout options are true for their respective validations.

pkg/containerutil/cp_resolve_linux.go
```diff
@@ -60,6 +60,14 @@ type pathSpecifier struct {
 // getPathSpecFromHost builds a pathSpecifier from a host location
 // errors with errDoesNotExist, errIsNotADir, "EvalSymlinks: too many links", or other hard filesystem errors from lstat/stat
 func getPathSpecFromHost(originalPath string) (*pathSpecifier, error) {
+	if originalPath == "-" {
```
Shubhranshu153: same response as below as this case is unintuitive especially for isADir case.

cmd/nerdctl/container/container_cp_linux_test.go
```diff
@@ -400,6 +404,47 @@ func TestCopyToContainer(t *testing.T) {
...
+					catFile:         filepath.Join(pathIsADirAbsolute, srcFileName),
+					setup: func(base *testutil.Base, container string, destPath string) {
+						base.Cmd("exec", container, "mkdir", "-p", destPath).AssertOK()
+					},
+				},
+				{
+					description:     "DEST_PATH is tarball",
+					destinationSpec: "-",
+					expect: icmd.Expected{
+						ExitCode: 1,
```
Shubhranshu153, line 433: need error type, otherwise test is not clear why it should fail. may be also adding a bit of description on the error would help

cmd/nerdctl/container/container_cp_linux_test.go
```diff
@@ -540,6 +585,19 @@ func TestCopyFromContainer(t *testing.T) {
 						assert.NilError(t, err)
 					},
 				},
+				{
+					description:     "DEST_PATH is a tarball",
```
Shubhranshu153: all description are DEST_PATH is a tarball? not sure i understand this test. my understanding is we copy from container into stdout if we have "-" in destination spec.

sondavidb: Yeah, changed to `DEST_PATH is stdout`. I used tarball since that was the format of the output but what you said is more correct I think. Made changes across testing suite for this.

cmd/nerdctl/container/container_cp_linux_test.go
sondavidb: Added errors

pkg/containerutil/cp_linux.go
sondavidb: Sure, added options `FromStdin` and `ToStdout` in `ContainerCpOptions` + added `fromStdin` and `toStdout` in `pathSpecifier`

Shubhranshu153 (approved):

sondavidb:
Do we think CI failure is unrelated? it seems so to me but seems we are having trouble getting it to pass at all..

sondavidb:
cc @ChengyuZhu6 @AkihiroSuda (since I see you two were trying to re-run the test)

AkihiroSuda:
```
=== Failed
=== FAIL: cmd/nerdctl/container TestRunWithSystemdTrueEnabled/waits_for_systemd_to_become_ready_and_lists_systemd_jobs (0.04s)
    container_run_systemd_linux_test.go:120: 
        
        +============================================================================================================+
        | 🚀      | "TestRunWithSystemdTrueEnabled/waits_for_systemd_to_become_ready_and_lists_systemd_jobs": starti |
        |         | ng test!                                                                                         |
        +============================================================================================================+
        | ⏳      | /tmp/TestRunWithSystemdTrueEnabledwaits_for_systemd_to_become_ready_208478251/001/aab6ed55       |
        +============================================================================================================+
        |         |                                                                                                  |
        +============================================================================================================+
        | config  | {                                                                                                |
        |         |   "Namespace": "nerdctl-test",                                                                   |
        |         |   "PrivateMode": ""                                                                              |
        |         | }                                                                                                |
        +============================================================================================================+
        | labels  | {}                                                                                               |
        +============================================================================================================+
        
    container_run_systemd_linux_test.go:120: 
        
        | ➡️      | ⚙️ /usr/local/bin/nerdctl exec nerdctl-testrunwithsystemdtrueenabled sh -c -- tries=0            |
        |         |                                                                                                  |
        |         |       until systemctl is-system-running >/dev/null 2>&1; do                                      |
        |         |                                                                                                  |
        |         |         >&2 printf "Waiting for systemd to come up...\n"                                         |
        |         |         sleep 1s                                                                                 |
        |         |         tries=$(( tries + 1))                                                                    |
        |         |         [ $tries -lt 10 ] || {                                                                   |
        |         |           >&2 printf "systemd failed to come up in a reasonable amount of time\n"                |
        |         |           exit 1                                                                                 |
        |         |         }                                                                                        |
        |         |       done                                                                                       |
        |         |       systemctl list-jobs                                                                        |
        |         | 🚫 command returned a non-zero exit code                                                         |
        |         | ⚠️ 1                                                                                             |
        |         | 🟠 time="2026-02-15T06:17:31Z" level=fatal msg="OCI runtime exec failed: exec failed: unable to  |
        |         | start container process: error starting setns process: exec: already started"                    |
        | 🌱      | HOME=/root                                                                                       |
        |         | PATH=/usr/local/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin |
        |         | :/bin                                                                                            |
        |         | DOCKER_CONFIG=/tmp/TestRunWithSystemdTrueEnabledwaits_for_systemd_to_become_ready_208478251/003  |
        |         | NERDCTL_TOML=/tmp/TestRunWithSystemdTrueEnabledwaits_for_systemd_to_become_ready_208478251/003/n |
        |         | erdctl.toml                                                                                      |
        | ⏰      | <1s (limit: 3m0s)                                                                                |
        | 📁      | /tmp/TestRunWithSystemdTrueEnabledwaits_for_systemd_to_become_ready_208478251/002                |
        
    container_run_systemd_linux_test.go:120: 	
        
        <<<<<<<<<<<<<<<<<<<<
        	🖊️ Command should succeed
        	👀 testing:		`command returned a non-zero exit code`
        	❌ FAILED!		is `<nil>`
        >>>>>>>>>>>>>>>>>>>>
```

https://github.com/containerd/nerdctl/actions/runs/21416963179/job/63655402439?pr=4704

Probably unrelated to this PR. Let's merge this

PR #4714, opened 2026-01-21
test: refactor container_inspect_windows_test.go to use Tigron

JoonsooWon:
related : https://github.com/containerd/nerdctl/issues/4613

### DIFF
--- a/cmd/nerdctl/container/container_inspect_windows_test.go
+++ b/cmd/nerdctl/container/container_inspect_windows_test.go
@@ -17,50 +17,100 @@
 package container
 
 import (
+	"encoding/json"
 	"testing"
 
 	"gotest.tools/v3/assert"
 
+	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+	"github.com/containerd/nerdctl/mod/tigron/tig"
+
+	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/dockercompat"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 )
 
 func TestInspectProcessContainerContainsLabel(t *testing.T) {
-	testContainer := testutil.Identifier(t)
+	testCase := nerdtest.Setup()
 
-	base := testutil.NewBase(t)
-	defer base.Cmd("rm", "-f", testContainer).Run()
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		containerName := testutil.Identifier(t)
+		data.Labels().Set("containerName", containerName)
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/container/container_inspect_windows_test.go
```diff
@@ -21,46 +21,83 @@ import (
...
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: expect.ExitCodeSuccess,
+			Output: func(stdout string, t tig.T) {
+				containerName := data.Labels().Get("containerName")
+				inspect := nerdtest.InspectContainer(helpers, containerName)
+				lbs := inspect.Config.Labels
+				assert.Equal(t, "foo", lbs["foo"])
+				assert.Equal(t, "bar", lbs["bar"])
+			},
```
haytok, line 66: Writing `nerdctl inspect containerName` in the Command section means testing based on the stdout results, which I think is better. WDYT?

cmd/nerdctl/container/container_inspect_windows_test.go
```diff
@@ -21,46 +21,83 @@ import (
...
```
haytok, line 111: Same as before.

cmd/nerdctl/container/container_inspect_windows_test.go
```diff
@@ -21,46 +21,83 @@ import (
...
-	lbs := inspect.Config.Labels
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		containerName := data.Labels().Get("containerName")
+		return helpers.Command("inspect", containerName)
+	}
 
-	//check with HCS if the container is ineed a VM
-	isHypervContainer, err := testutil.HyperVContainer(inspect)
-	if err != nil {
-		t.Fatalf("unable to list HCS containers: %s", err)
```
haytok, line 60: Equivalent processing also needs to be written in Tigron.

cmd/nerdctl/container/container_inspect_windows_test.go
JoonsooWon: I also considered that approach, but I initially kept the previous implementation because of the parsing complexity.
I agree with your point, and I've fixed to use stdout results.
Thanks for the suggestion.

cmd/nerdctl/container/container_inspect_windows_test.go
JoonsooWon: I've fixed it
Thanks

haytok:
CC: @ChengyuZhu6 @AkihiroSuda 

When you have time, could you retry the CI and review 🙏

ChengyuZhu6:
Please rebase on main branch to pickup the changes.

JoonsooWon:
Thanks!
I've rebased on main branch.

PR #4728, opened 2026-01-30
Detect vendor before crafting cdiDeviceIDs for --gpus

shiv-tyagi:
This detects the GPU vendor from the CDI spec files while generating devicesIDs corresponding to the values passed to `--gpus` option. With this, the users can also use AMD gpus if a corresponding CDI spec is present.

Here is a similar PR for ctr https://github.com/containerd/containerd/pull/12839

### DIFF
--- a/pkg/cmd/container/run_cdi.go
+++ b/pkg/cmd/container/run_cdi.go
@@ -24,23 +24,68 @@ import (
 	"github.com/containerd/containerd/v2/core/containers"
 	cdispec "github.com/containerd/containerd/v2/pkg/cdi"
 	"github.com/containerd/containerd/v2/pkg/oci"
+	"github.com/containerd/log"
 )
 
+// detectGPUVendorFromCDI detects the first available GPU vendor from CDI cache.
+// Returns empty string if no known vendor is found.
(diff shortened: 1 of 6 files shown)


shiv-tyagi:
Can someone help me explain these test failures? These don't seem to be related to the changes in the PR.

pkg/cmd/container/run_cdi.go
AkihiroSuda, line 1: https://github.com/containerd/nerdctl/blob/main/docs/gpu.md should be updated too?

shiv-tyagi: Can we do that in a follow up?

Once this is merged, we might organise our docs better and put the latest links here to help users install the AMD CDI specs through `amd-ctk`.

shiv-tyagi:
I have added the `TestRunGPU` and `TestRunGPUWithOtherCDIDevices` integration tests as well to ensure the behavior is as expected.

pkg/cmd/container/run_cdi.go
shiv-tyagi: I have updated the documentation in this PR itself. PTAL whenever you get time.

cc @elezar

PR #4731, opened 2026-02-01
fix: add missing --ipc options in help text and shell completion

haytok:
The docs already documented
`--ipc=(host|private|shareable|container:<container>)`.

- https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-run

However, the implementation only had `host` and `private` in help text and completion.

This adds the missing options and smart completion for `container:<name>`


ChengyuZhu6 (approved):
LGTM

PR #4732, opened 2026-02-02
optimize nerdctl ps

ningmingxiao:
speed up nerdctl ps  we don't need call get container spec again.

### DIFF
--- a/pkg/cmd/container/list.go
+++ b/pkg/cmd/container/list.go
@@ -47,7 +47,7 @@ func List(ctx context.Context, client *containerd.Client, options types.Containe
 	if err != nil {
 		return nil, err
 	}
-	return prepareContainers(ctx, client, containers, cMap, options)
+	return prepareContainers(ctx, client, &containers, cMap, options)
 }
 
 // filterContainers returns containers matching the filters.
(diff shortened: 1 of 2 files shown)


pkg/formatter/formatter.go
```diff
@@ -73,6 +74,14 @@ func ContainerStatus(ctx context.Context, c containerd.Container) string {
 	}
 }
 
+func GetCommandFromSpec(spec typeurl.Any, trunc, quote bool) (string, error) {
```
AkihiroSuda, line 77: This code does not need to be a separate function, as it is called only once

pkg/cmd/container/list.go
```diff
@@ -47,7 +47,7 @@ func List(ctx context.Context, client *containerd.Client, options types.Containe
 	if err != nil {
 		return nil, err
 	}
-	return prepareContainers(ctx, client, containers, cMap, options)
+	return prepareContainers(ctx, client, &containers, cMap, options)
```
AkihiroSuda, line 50: why?

pkg/cmd/container/list.go
```diff
@@ -201,6 +198,7 @@ func prepareContainers(ctx context.Context, client *containerd.Client, container
 			}
 			li.Size = containerSize
 		}
+		(*containers)[i] = nil
```
AkihiroSuda, line 201: ?

AkihiroSuda:
> speed up

Any benchmark results?

ningmingxiao:
1000 container speedup 1-2s @AkihiroSuda

ningmingxiao:
I find nerdctl ps  use about 1g memory  when container spec size is too big  there are many env in spec (because user use enableServiceLinks:true and k8s has many services )
see :https://github.com/containerd/containerd/pull/12846
image

pkg/cmd/container/list.go
ningmingxiao: here I want to let gc recycle the memory  quickly

pkg/formatter/formatter.go
ningmingxiao: since spec have much data,  If I use it into a function, it may cause quickly gc (gc may happen after the function return) 
if I use it into a for loop (gc may happen after every  loop).  @AkihiroSuda

AkihiroSuda: You can just create a scope with `{}`

Also please add the code comments about GC

pkg/cmd/container/list.go
AkihiroSuda: That behavior has to be explained in the code comments in the function declaration

ningmingxiao:
we can wait https://github.com/containerd/containerd/pull/12846 or https://github.com/containerd/containerd/pull/12862 to be merged.  @AkihiroSuda

PR #4758, opened 2026-02-27
fix: per-network static IP assignment for multi-network containers

tariromukute:
## Summary

When using `nerdctl compose` with services connected to multiple networks that each have a static IPv4 address (`ipv4_address`), all static IPs were passed to every CNI plugin via a shared `CNI_ARGS` (`IP=<addr>`). This caused failures because each bridge plugin would attempt to allocate an IP address that belongs to a different network's subnet.

This PR fixes multi-network static IP assignment by:
1. Collecting per-network IPs and passing them as a JSON-encoded annotation (`nerdctl/ip-per-network`) instead of the single-value `--ip=` flag (which can only hold one IP).
2. Using `cnilibrary` directly (instead of go-cni's `Setup()`) to set up each network individually with:
   - The correct container-side interface name (`eth0`, `eth1`, `eth2`, ...) — go-cni's `WithConfListBytes` always assigned `eth0` to separate CNI instances.
   - Per-network `CNI_ARGS` containing only the IP for that specific network.
3. Maintaining backward compatibility: single-network containers with a static IP still use the legacy `--ip=` flag and the existing go-cni `Setup()` path.

## Changelog [KeepA](https://keepachangelog.com/)

[Fixed]:
- Fixed `nerdctl compose up` failing for services with static IPs (`ipv4_address`) on multiple networks due to all IPs being passed to every CNI plugin via shared `CNI_ARGS`.
- Fixed container-side interface name collision (`eth0` already exists) when setting up multiple networks individually, by using `cnilibrary.AddNetworkList`/`DelNetworkList` directly with correct per-network `IfName`.

[Added]:
- New `nerdctl/ip-per-network` annotation (JSON map of network name → static IPv4 address) for multi-network static IP propagation from compose to the OCI hook.
- Direct `cnilibrary` based per-network CNI setup and teardown functions (`perNetworkAdd`, `perNetworkDel`) with correct interface naming.

## Steps to Reproduce

1. Create a compose file with a service attached to two networks, each with a static IP:

```yaml
version: '3.8'
services:
  web:
    image: nginx:alpine
    ports:
      - "8080:80"
    networks:
      public_net_access:
        ipv4_address: 192.168.72.130
      public_net:
        ipv4_address: 192.168.70.130
    environment:
      - NGINX_HOST=web.local
      - NGINX_PORT=80

networks:
  public_net:
    driver: bridge
    name: demo-oai-public-net
    ipam:
      config:
        - subnet: 192.168.70.128/26
  public_net_access:
    name: oai-public-access
    ipam:
      config:
        - subnet: 192.168.72.128/26
```

2. Run `nerdctl compose up -d`

## Expected behaviour

Each network should only receive `IP=` and interface in it's corresponding network. Creation should be successful.

## Actual behaviour

Both networks receive request to create the same interface and IP. See error below

```
332164/cid -l=com.docker.compose.project=examples -l=com.docker.compose.service=web -l=com.docker.compose.config-hash=b43a23c3c83aac9237234aee90ad8aa99bbf51c418d20d8f62941a24dc5a21d3 -d --name=examples-web-1 --pull=never -e=NGINX_HOST=web.local -e=NGINX_PORT=80 --net=demo-oai-public-net --ip=192.168.70.130 --net=oai-public-access --ip=192.168.72.130 --hostname=web -p=8080:80/tcp --restart=no nginx:alpine] 
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running createRuntime hook #0: exit status 1, stdout: , stderr: time="2026-02-27T10:02:07+01:00" level=warning msg="Container failed starting. Removing allocated network configuration."
time="2026-02-27T10:02:07+01:00" level=fatal msg="failed to call cni.Setup: plugin type=\"bridge\" failed (add): failed to allocate all requested IPs: 192.168.72.130" 
FATA[0000] error while creating container examples-web-1: error while creating container examples-web-1: exit status 1 
```


AkihiroSuda:
Thanks, but please add DCO sign
https://wiki.linuxfoundation.org/dco

tariromukute:
> Thanks, but please add DCO sign https://wiki.linuxfoundation.org/dco

I have done so, thanks.

pkg/ocihook/ocihook.go
```diff
@@ -530,17 +603,18 @@ func applyNetworkSettings(opts *handlerOpts) (err error) {
...
+	commonOpts := []cni.NamespaceOpts{}
+	commonOpts = append(commonOpts, portMapOpts...)
+	commonOpts = append(commonOpts, macAddressOpts...)
+	commonOpts = append(commonOpts, ip6AddressOpts...)
+	commonOpts = append(commonOpts,
 		cni.WithLabels(map[string]string{
 			"IgnoreUnknown": "1",
 		}),
 		cni.WithArgs("NERDCTL_CNI_DHCP_HOSTNAME", opts.state.Annotations[labels.Hostname]),
 	)
```
haytok: NIT: 
`commonOpts` is used for `opts.cni.Setup`, so these can be defined directly under `// Legacy path: single IP (or no IP) shared across all networks`.

pkg/ocihook/ocihook.go
```diff
@@ -460,6 +475,64 @@ func portReserverPidFilePath(opts *handlerOpts) string {
...
+		ContainerID:    opts.fullID,
+		NetNS:          nsPath,
+		IfName:         perNetworkIfName(networkIndex),
+		Args:           extraArgs,
+		CapabilityArgs: make(map[string]interface{}),
+	}
+	if len(portMappings) > 0 {
+		rt.CapabilityArgs["portMappings"] = portMappings
+	}
+	result, err := cniConfig.AddNetworkList(ctx, confList, rt)
```
haytok, line 529: In the new method, since `AddNetworkList` is called instead of `Setup`, the `ips` capability (for IPv6 static addresses) is not added to `rt.CapabilityArgs`, so it seems necessary to add them.

- https://www.cni.dev/docs/conventions/#well-known-capabilities

> ips | Dynamically allocate IPs for container interface. Runtime which has the ability of address allocation can pass these to plugins. | ips | A list of IP (string entries). [ “10.10.0.1/24”, “3ffe:ffff:0:01ff::1/64” ] | none | CNI static plugin

haytok (commented):
Thanks for creating this PR! Would it be possible to have e2e tests using [Tigron](https://github.com/containerd/nerdctl/tree/main/docs/testing) for this feature ?

AkihiroSuda:
@tariromukute Could you take a look at the review comments?

tariromukute:
> @tariromukute Could you take a look at the review comments?

Thanks @haytok for taking time to review the PR and for the comments. I will look address them this coming week.

AkihiroSuda:
Needs rebase

pkg/ocihook/ocihook.go
tariromukute: Thanks. Moved this.

pkg/ocihook/ocihook.go
tariromukute: Thanks for the feedback. Addressed this in commit https://github.com/containerd/nerdctl/commit/7255bf150b96a8277f8074b88df28e40317e0277

Issue #4746, opened 2026-02-18
Flaky test: TestRunWithSystemdTrueEnabled


AkihiroSuda:
> ```
> === Failed
> === FAIL: cmd/nerdctl/container TestRunWithSystemdTrueEnabled/waits_for_systemd_to_become_ready_and_lists_systemd_jobs (0.04s)
>     container_run_systemd_linux_test.go:120: 
>         
>         +============================================================================================================+
>         | 🚀      | "TestRunWithSystemdTrueEnabled/waits_for_systemd_to_become_ready_and_lists_systemd_jobs": starti |
>         |         | ng test!                                                                                         |
>         +============================================================================================================+
>         | ⏳      | /tmp/TestRunWithSystemdTrueEnabledwaits_for_systemd_to_become_ready_208478251/001/aab6ed55       |
>         +============================================================================================================+
>         |         |                                                                                                  |
>         +============================================================================================================+
>         | config  | {                                                                                                |
>         |         |   "Namespace": "nerdctl-test",                                                                   |
>         |         |   "PrivateMode": ""                                                                              |
>         |         | }                                                                                                |
>         +============================================================================================================+
>         | labels  | {}                                                                                               |
>         +============================================================================================================+
>         
>     container_run_systemd_linux_test.go:120: 
>         
>         +------------------------------------------------------------------------------------------------------------+
>         | ➡️      | ⚙️ /usr/local/bin/nerdctl exec nerdctl-testrunwithsystemdtrueenabled sh -c -- tries=0            |
>         |         |                                                                                                  |
>         |         |       until systemctl is-system-running >/dev/null 2>&1; do                                      |
>         |         |                                                                                                  |
>         |         |         >&2 printf "Waiting for systemd to come up...\n"                                         |
>         |         |         sleep 1s                                                                                 |
>         |         |         tries=$(( tries + 1))                                                                    |
>         |         |         [ $tries -lt 10 ] || {                                                                   |
>         |         |           >&2 printf "systemd failed to come up in a reasonable amount of time\n"                |
>         |         |           exit 1                                                                                 |
>         |         |         }                                                                                        |
>         |         |       done                                                                                       |
>         |         |       systemctl list-jobs                                                                        |
>         +------------------------------------------------------------------------------------------------------------+
>         |         | 🚫 command returned a non-zero exit code                                                         |
>         +------------------------------------------------------------------------------------------------------------+
>         |         | ⚠️ 1                                                                                             |
[...]

PR #4766, opened 2026-03-03
fix: Add delay for systemd to startup before running exec

Shubhranshu153:
Attempting to fix: https://github.com/containerd/nerdctl/issues/4746

TestRunWithSystemdTrueEnabled seems to pass now after a few runs.


Shubhranshu153:
Well TestRunWithSystemdTrueEnabled in Canary is still failing with the same error.

PR #4796, opened 2026-03-16
test: refactor compose_pull_linux_test.go to use nerdtest/tigron

juannio:
Update tests to use nerdtest.Setup and the Tigron testing framework as per issue [#4613](https://github.com/containerd/nerdctl/issues/4613)

### DIFF
--- a/cmd/nerdctl/compose/compose_pull_linux_test.go
+++ b/cmd/nerdctl/compose/compose_pull_linux_test.go
@@ -18,14 +18,21 @@ package compose
 
 import (
 	"fmt"
+	"path/filepath"
 	"testing"
 
+	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 )
 
 func TestComposePullWithService(t *testing.T) {
-	base := testutil.NewBase(t)
-	var dockerComposeYAML = fmt.Sprintf(`
+	testCase := nerdtest.Setup()
+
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/compose/compose_pull_linux_test.go
```diff
@@ -53,10 +59,26 @@ volumes:
   db:
 `, testutil.WordpressImage, testutil.MariaDBImage)
 
-	comp := testutil.NewComposeDir(t, dockerComposeYAML)
-	defer comp.CleanUp()
-	projectName := comp.ProjectName()
-	t.Logf("projectName=%q", projectName)
```
haytok, line 59: Please display `projectName` in the log.

cmd/nerdctl/compose/compose_pull_linux_test.go
```diff
@@ -53,10 +59,26 @@ volumes:
...
+		return &test.Expected{
+			ExitCode: 0,
+			Output:   expect.DoesNotContain("wordpress"),
+		}
+	}
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		if data.Labels().Get("composeYAML") != "" {
+			helpers.Anyhow("rmi", data.Labels().Get("mariaDBImage"))
+		}
+	}
```
haytok: Which part of the original implementation does this process correspond to?

cmd/nerdctl/compose/compose_pull_linux_test.go
juannio: FIxed, thanks!

cmd/nerdctl/compose/compose_pull_linux_test.go
juannio: Good catch, this wasn't in the original implementation. I added it based on patterns I saw in other migrated tests, but since this is a straightforward migration I'll keep it faithful to the original. Removed, thanks!

Issue #4786, opened 2026-03-10
Regression, in v2.1.3 unable to bind containers on different IPs to same ports


figaw:
### Description

Using version 2.1.3 of nerdctl, it incorrectly checks for conflicting ports when starting containers.

Using versions:
```
containerd_version: "2.1.6"
runc_version: "1.3.4"
cni_version: "1.9.0"
nerdctl_version: "2.1.2"
```

I'm successfully able to execute (and access) the following compose spec:

```yaml
services:
  nginx-primary:
    image: nginx:latest
    container_name: nginx_primary
    ports:
      - "192.168.1.141:80:80"
      - "192.168.1.142:80:80"
    restart: always

  nginx-sec:
    image: nginx:latest
    container_name: nginx_sec
    ports:
      - "192.168.1.143:80:80"
    restart: always
```

But any version later than 2.1.2, (even the latest 2.1.6, or 2.2.1) fails.

---

> NB: I tested with all sorts of different combinations of containerd versions, runc versions, CNI plugin versions. But whenever I make the change to 2.1.3 or later, it begins to fail.

### Steps to reproduce the issue

1. Update nerdctl to version 2.1.3
2. Attempt to start multiple containers, binding to different IPs but on the same port
3. The "second" container is not allowed to start

### Describe the results you received and expected

```shell
INFO[0000] Creating network test_default
INFO[0000] Ensuring image nginx:latest
INFO[0000] Ensuring image nginx:latest
INFO[0000] Creating container nginx_sec
INFO[0000] Running [/usr/local/bin/nerdctl run --cidfile=/tmp/compose-539193217/cid -l=com.docker.compose.project=test -l=com.docker.compose.service=nginx-sec -d --name=nginx_sec --pull=never --net=test_default --hostname=nginx-sec -p=192.168.1.143:80:80/tcp --restart=always nginx:latest]
FATA[0000] failed to load networking flags: bind for 192.168.1.143:80 failed: port is already allocated
FATA[0000] error while creating container nginx_sec: error while creating container nginx_sec: exit status 1
```

I expect the containers to just start.

### What version of nerdctl are you using?

Any version later than 2.1.2. 2.1.3 or even the latest 2.1.6, or 2.2.1.

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

### Host information

```
Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v2.1.6
 Storage Driver: overlayfs
[...]

PR #4800, opened 2026-03-19
fix: allow binding containers on different IPs to the same port

yankay:
## Summary

Fixes https://github.com/containerd/nerdctl/issues/4786

The port conflict check introduced in PR #4097 (commit 04f836d7) incorrectly reports "port is already allocated" when binding containers to different host IPs but the same port number. For example:

```yaml
services:
  nginx-primary:
    ports:
      - "192.168.1.141:80:80"
      - "192.168.1.142:80:80"
  nginx-sec:
    ports:
      - "192.168.1.143:80:80"   # fails: port is already allocated
```

### Root cause

1. `ReadIPTables` only read the `CNI-HOSTPORT-DNAT` parent chain, which dispatches by port but does **not** include `-d <ip>` destination IP filtering. The actual IP filtering is in `CNI-DN-*` sub-chains.
2. `ParseIPTableRules` only extracted port numbers from `--dports` and completely ignored destination IPs.

This made every iptables-managed port appear globally allocated regardless of which host IP it was bound to.

### Fix

- **`ReadIPTables`**: Read `CNI-DN-*` sub-chains (which contain DNAT rules with both `-d <ip>` and `--dport`), falling back to the parent chain if no sub-chains exist.
- **`ParseIPTableRules`**: Return `PortRule{IP, Port}` instead of just port numbers. Parse both `--dport` (sub-chains) and `--dports` (parent chain), and extract `-d <ip>` destination.
- **`getUsedPorts`**: Filter iptables ports by IP overlap — wildcard addresses (`""`, `0.0.0.0`, `::`) conflict with everything; specific IPs only conflict with the same IP or wildcards.

## Test plan

- [x] Unit tests pass (`go test ./pkg/portutil/...`)
- [x] Manual test: three containers on `127.0.0.1/127.0.0.2/127.0.0.3:8880` all start successfully
- [x] Manual test: duplicate IP+port correctly rejected
- [x] Manual test: `0.0.0.0` wildcard correctly conflicts with existing bindings
- [x] Manual test: all containers accessible via `curl` (HTTP 200)

### DIFF
--- a/pkg/portutil/iptable/iptables.go
+++ b/pkg/portutil/iptable/iptables.go
@@ -22,26 +22,46 @@ import (
 	"strings"
 )
 
-// ParseIPTableRules takes a slice of iptables rules as input and returns a slice of
-// uint64 containing the parsed destination port numbers from the rules.
-func ParseIPTableRules(rules []string) []uint64 {
-	ports := []uint64{}
+type PortRule struct {
+	IP   string
+	Port uint64
+}
 
-	// Regex to match the '--dports' option followed by the port number
-	dportRegex := regexp.MustCompile(`--dports ((,?\d+)+)`)
+// ParseIPTableRules takes a slice of iptables rules as input and returns a
+// slice of PortRule containing the parsed destination IP and port from the
+// rules. When a rule has no -d flag, IP is empty (meaning the rule applies to
+// all addresses).
+func ParseIPTableRules(rules []string) []PortRule {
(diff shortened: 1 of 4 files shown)


ChengyuZhu6 (approved):
LGTM

Shubhranshu153 (approved):

Shubhranshu153:
LGTM

PR #4842, opened 2026-04-09
defaults: use user-writable data root on darwin

66Ton99:
## Summary

On Darwin, `defaults.DataRoot()` currently always returns `/var/lib/nerdctl`.
For regular users this can fail immediately with:

`mkdir /var/lib/nerdctl: permission denied`

This change makes Darwin use a user-writable default data root.

## Change

In `pkg/defaults/defaults_darwin.go`:

- `DataRoot()` now returns `$HOME/.local/share/nerdctl` when the home directory is available
- falls back to `/var/lib/nerdctl` if home cannot be resolved

## Why

This aligns the Darwin default with typical non-root usage and avoids immediate startup failures for users who run `nerdctl` without elevated privileges.

## Scope

- Darwin only
- no Linux/Windows behavior changes

## Validation

- `go test ./pkg/defaults ./pkg/config`

## Review request

A review from maintainers familiar with `nerdctl` default path semantics across platforms would be appreciated.

### DIFF
--- a/pkg/defaults/defaults_darwin.go
+++ b/pkg/defaults/defaults_darwin.go
@@ -16,11 +16,12 @@
 
 // This is a dummy file to allow usage of library functions
 // on Darwin-based systems.
-// All functions and variables are empty/no-ops
+// Most functions and variables are stubs/no-ops
 
 package defaults
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
@@ -38,18 +39,24 @@ func CNIPath() string {
 }
 
 func CNIRuntimeDir() (string, error) {
+	if os.Geteuid() != 0 {
+		return filepath.Join(xdgRuntimeDir(), "cni"), nil
+	}
 	return "/var/run/cni", nil
 }
 
 func CNINetConfPath() string {
+	if os.Geteuid() != 0 {
+		return filepath.Join(xdgConfigHome(), "cni", "net.d")
+	}
 	return gocni.DefaultNetDir
 }
 
 func DataRoot() string {
-	if home, err := os.UserHomeDir(); err == nil && home != "" {
-		return filepath.Join(home, ".local", "share", "nerdctl")
+	if os.Geteuid() == 0 {
+		return "/var/lib/nerdctl"
 	}
-	return "/var/lib/nerdctl"
+	return filepath.Join(xdgDataHome(), "nerdctl")
 }
 
 func CgroupManager() string {
@@ -61,11 +68,21 @@ func CgroupnsMode() string {
 }
 
 func NerdctlTOML() string {
+	if os.Geteuid() != 0 {
+		return filepath.Join(xdgConfigHome(), "nerdctl", "nerdctl.toml")
+	}
 	return "/etc/nerdctl/nerdctl.toml"
 }
 
 func HostsDirs() []string {
-	return []string{}
+	if os.Geteuid() != 0 {
+		xch := xdgConfigHome()
+		return []string{
+			filepath.Join(xch, "containerd", "certs.d"),
+			filepath.Join(xch, "docker", "certs.d"),
+		}
+	}
+	return []string{"/etc/containerd/certs.d", "/etc/docker/certs.d"}
 }
 
 func HostGatewayIP() string {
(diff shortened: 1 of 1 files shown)


66Ton99:
Related downstream packaging PR:
- NixOS/nixpkgs: https://github.com/NixOS/nixpkgs/pull/508463

pkg/defaults/defaults_darwin.go
```diff
@@ -41,6 +46,9 @@ func CNINetConfPath() string {
 }
 
 func DataRoot() string {
+	if home, err := os.UserHomeDir(); err == nil && home != "" {
+		return filepath.Join(home, ".local", "share", "nerdctl")
+	}
 	return "/var/lib/nerdctl"
```
AkihiroSuda: This should remain the default when EUID is 0

pkg/defaults/defaults_darwin.go
```diff
@@ -41,6 +46,9 @@ func CNINetConfPath() string {
 }
 
 func DataRoot() string {
+	if home, err := os.UserHomeDir(); err == nil && home != "" {
+		return filepath.Join(home, ".local", "share", "nerdctl")
```
AkihiroSuda: `$XDG_DATA_HOME` should be respected

pkg/defaults/defaults_darwin.go
AkihiroSuda, line 1: All the paths should be handled consistently, not just `DataRoot`

pkg/defaults/defaults_darwin.go
AkihiroSuda: The convention should be consistent with nerdbox.
Let me cc @containerd/nerdbox-committers @containerd/nerdbox-reviewers

pkg/defaults/defaults_darwin.go
66Ton99: Fixed in a5cdb0d5. DataRoot now keeps /var/lib/nerdctl when EUID == 0.

pkg/defaults/defaults_darwin.go
66Ton99: Fixed in a5cdb0d5. For non-root on Darwin, DataRoot now respects XDG_DATA_HOME (fallback: ~/.local/share).

pkg/defaults/defaults_darwin.go
66Ton99: Fixed in a5cdb0d5. I applied the same root/non-root + XDG convention consistently to DataRoot, CNINetConfPath, CNIRuntimeDir, NerdctlTOML, HostsDirs, and CDISpecDirs.

pkg/defaults/defaults_darwin.go
66Ton99: Addressed in a5cdb0d5. Updated Darwin defaults to follow the same user-writable/XDG convention consistently across paths for non-root, while keeping system defaults for root.

pkg/defaults/defaults_darwin.go
```diff
@@ -53,17 +68,66 @@ func CgroupnsMode() string {
...
+	if os.Geteuid() != 0 {
+		return []string{
+			filepath.Join(xdgConfigHome(), "cdi"),
+			filepath.Join(xdgRuntimeDir(), "cdi"),
+		}
+	}
 	return []string{"/etc/cdi", "/var/run/cdi"}
 }
+
+func xdgConfigHome() string {
```
AkihiroSuda, line 102: Duplicated code: https://github.com/containerd/nerdctl/blob/main/pkg/rootlessutil/xdg_linux.go

Should be consolidated to a single package (but not a single file).

pkg/defaults/defaults_darwin.go
```diff
@@ -53,17 +68,66 @@ func CgroupnsMode() string {
...
+		return filepath.Join(home, ".local", "share")
+	}
+	return "/var/lib"
+}
+
+func xdgRuntimeDir() string {
+	if xdr := os.Getenv("XDG_RUNTIME_DIR"); xdr != "" {
+		return xdr
+	}
+	return fmt.Sprintf("/run/user/%d", os.Geteuid())
```
AkihiroSuda, line 132: Inexistent on macOS

pkg/defaults/defaults_darwin.go
AkihiroSuda, line 1: Probably (most of) `defaults_darwin.go` and `defaults_freebsd.go` can be consolidated into a single `defaults_unix.go`

PR #4850, opened 2026-04-16
feat: add missing HostConfig fields to dockercompat inspect response

ayush-panta:
This PR populates missing `HostConfig` fields in the dockercompat inspect response, bringing `nerdctl inspect --mode=dockercompat` closer to what consumers expect from Docker's `GET /containers/{id}/json`.

All data comes from sources already available in `ContainerFromNative()` via the OCI spec and container labels.

**OCI spec fields:** `PidsLimit`, `Ulimits`, `MemoryReservation`, `MemorySwappiness`, and `Annotations`. Ulimits are converted from OCI `RLIMIT_*` format to Docker's lowercase format (inverse of `run_ulimit_linux.go`). Annotations are filtered to exclude `nerdctl/` prefixed internal entries.

**Label-based fields:** `NetworkMode` (from `nerdctl/networks`), `RestartPolicy` (parsed from `containerd.io/restart.policy` via `restart.NewPolicy()`), and `AutoRemove` (from `nerdctl/auto-remove`). These labels already exist — we're just populating the fields.

**Privileged:** The OCI spec doesn't store a "privileged" flag; it just applies all caps and device access. Inferring that from the spec is fragile, so we store a `nerdctl/privileged` label at creation time and read it back at inspect. This required minor additions to `create.go` and `labels.go`.

**CapAdd/CapDrop:** nerdctl doesn't persist the original `--cap-add`/`--cap-drop` flags; only the resulting OCI bounding set survives. We reconstruct by diffing the bounding set against containerd's 14 default capabilities ([containerd source](https://github.com/containerd/containerd/blob/main/pkg/oci/spec.go), [moby source](https://github.com/moby/moby/blob/master/daemon/pkg/oci/caps/defaults.go) — same set). Caps in bounding but not in defaults → CapAdd. Defaults not in bounding → CapDrop. There's a slight difference from Docker when --cap-add ALL is used: Docker stores and returns the literal ["ALL"], while we show the expanded list since we only have the OCI spec. This is functionally equivalent. 

**Testing:**
- Unit: `TestGetCapabilitiesFromNative` (7 cases), `TestGetUlimitsFromNative` (4 cases), extended "container from nerdctl" in `TestContainerFromNative` with all new fields.
- Integration: extended `TestContainerInspectHostConfig` with resource and label flags + assertions, extended `TestContainerInspectHostConfigDefaults` with zero-value checks. Added `TestContainerInspectHostConfigPrivileged` (rootful only) and `TestContainerInspectHostConfigCapabilities` as separate tests since privileged requires root and conflicts with cap flags.

### DIFF
--- a/cmd/nerdctl/container/container_inspect_linux_test.go
+++ b/cmd/nerdctl/container/container_inspect_linux_test.go
@@ -391,10 +391,16 @@ func TestContainerInspectHostConfig(t *testing.T) {
 			"--add-host", "host2:10.0.0.2",
 			"--ipc", "host",
 			"--memory", "512m",
+			"--memory-reservation", "200m",
+			"--memory-swappiness", "60",
+			"--pids-limit", "100",
+			"--ulimit", "nofile=1024:65536",
 			"--read-only",
 			"--shm-size", "256m",
 			"--uts", "host",
+			"--restart", "on-failure:3",
 			"--runtime", "io.containerd.runc.v2",
+			"--annotation", "com.example.key=test-val",
 			testutil.AlpineImage, "sleep", nerdtest.Infinity)
 		nerdtest.EnsureContainerStarted(helpers, data.Identifier())
 	}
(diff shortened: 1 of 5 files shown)


cmd/nerdctl/container/container_inspect_linux_test.go
```diff
@@ -881,3 +906,73 @@ type hostConfigValues struct {
 	GroupAddSize int
 	Runtime      string
 }
+
+func TestContainerInspectHostConfigPrivileged(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Require = nerdtest.Rootful
```
AkihiroSuda: ?

ayush-panta: I've removed the `nerdtest.Rootful` filter for this test case as a privileged container can be created regardless of rootful/rootless context (please correct me if I'm wrong).

ayush-panta:
CI is green on relevant tests; remaining failures seem to be due to flakiness/are unrelated to my changes. Ready for review.

ayush-panta:
The only remaining failure is in `TestRunBindMountBind`. Unrelated to PR.

AkihiroSuda (approved):
Thanks

PR #4856, opened 2026-04-18
fix: support human-readable sizes for fluentd-buffer-limit log option

shouhei:
## Problem

`fluentd-buffer-limit` currently accepts only raw integer values in `nerdctl`.  
In Docker-compatible workflows, users may specify human-readable values such as:

```
... --log-opt fluentd-buffer-limit=16M ...
```

## Fix

Use github.com/docker/go-units to parse fluentd-buffer-limit.  
This allows nerdctl to accept human-readable values such as 16M and 16MiB in addition to raw byte values.

### DIFF
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -365,7 +365,7 @@ Logging flags:
     - The `fluentd` logging driver supports the following logging options:
       - :whale: `--log-opt=fluentd-address=<ADDRESS>`: The address of the `fluentd` daemon, tcp(default) and unix sockets are supported..
       - :whale: `--log-opt=fluentd-async=<true|false>`: Enable async mode for fluentd. The default value is false.
-      - :whale: `--log-opt=fluentd-buffer-limit=<LIMIT>`: The buffer limit for fluentd. If the buffer is full, the call to record logs will fail. The default is 8192. (<https://github.com/fluent/fluent-logger-golang/tree/master#bufferlimit>)
+      - :whale: `--log-opt=fluentd-buffer-limit=<LIMIT>`: The buffer limit for fluentd. If the buffer is full, the call to record logs will fail. The default is 1MiB. Accepts human-readable sizes (e.g., `1KiB`, `1MiB`, `1GiB`) or raw byte values. (<https://github.com/fluent/fluent-logger-golang/tree/master#bufferlimit>)
       - :whale: `--log-opt=fluentd-retry-wait=<1s|1ms>`: The time to wait before retrying to send logs to fluentd. The default value is 1s.
       - :whale: `--log-opt=fluentd-max-retries=<1>`: The maximum number of retries to send logs to fluentd. The default value is MaxInt32.
       - :whale: `--log-opt=fluentd-sub-second-precision=<true|false>`: Enable sub-second precision for fluentd. The default value is false.
--- a/pkg/logging/fluentd_logger.go
+++ b/pkg/logging/fluentd_logger.go
@@ -27,6 +27,7 @@ import (
 	"sync"
 	"time"
 
+	units "github.com/docker/go-units"
 	"github.com/fluent/fluent-logger-golang/fluent"
 
 	"github.com/containerd/containerd/v2/core/runtime/v2/logging"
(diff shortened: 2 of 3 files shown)


pkg/logging/fluentd_logger.go
```diff
@@ -223,10 +224,11 @@ func parseFluentdConfig(config map[string]string) (fluent.Config, error) {
 	}
 	bufferLimit := defaultBufferLimit
 	if config[fluentdBufferLimit] != "" {
-		bufferLimit, err = strconv.Atoi(config[fluentdBufferLimit])
+		parsedBufferLimit, err := units.RAMInBytes(config[fluentdBufferLimit])
 		if err != nil {
 			return result, fmt.Errorf("error occurs %w,invalid buffer limit (%s)", err, config[fluentdBufferLimit])
 		}
+		bufferLimit = int(parsedBufferLimit)
```
utafrali, line 234: `units.RAMInBytes` returns `int64`, but the result is silently narrowed to `int`. On 32-bit platforms `int` is 32 bits, so any value above ~2 GiB would wrap silently without error. First check whether `fluent.Config.BufferLimit` actually requires `int` or already accepts `int64` (in which case you can drop the cast entirely). If the cast is unavoidable, add an explicit bounds check:
```go
if parsedBufferLimit > int64(math.MaxInt) {
    return result, fmt.Errorf("invalid buffer limit: value %d overflows int", parsedBufferLimit)
}
bufferLimit = int(parsedBufferLimit)
```

pkg/logging/fluentd_logger_test.go
```diff
@@ -229,6 +229,22 @@ func TestParseFluentdConfig(t *testing.T) {
...
+				FluentHost:             defaultHost,
+				FluentNetwork:          defaultProtocol,
+				FluentSocketPath:       "",
+				BufferLimit:            defaultBufferLimit,
+				RetryWait:              int(defaultRetryWait),
+				MaxRetry:               defaultMaxRetries,
+				Async:                  false,
+				AsyncReconnectInterval: 0,
+				SubSecondPrecision:     false,
+				RequestAck:             false}, false},
```
utafrali, line 247: The new test only covers the happy path. The rejection behavior has changed from `strconv.Atoi` to `units.RAMInBytes`, so add at least one `wantErr: true` case with an invalid value (e.g. `"1X"` or `"not-a-size"`) to confirm that malformed input still returns an error after the parser swap.

docs/command-reference.md
```diff
@@ -365,7 +365,7 @@ Logging flags:
     - The `fluentd` logging driver supports the following logging options:
       - :whale: `--log-opt=fluentd-address=<ADDRESS>`: The address of the `fluentd` daemon, tcp(default) and unix sockets are supported..
       - :whale: `--log-opt=fluentd-async=<true|false>`: Enable async mode for fluentd. The default value is false.
-      - :whale: `--log-opt=fluentd-buffer-limit=<LIMIT>`: The buffer limit for fluentd. If the buffer is full, the call to record logs will fail. The default is 8192. (<https://github.com/fluent/fluent-logger-golang/tree/master#bufferlimit>)
+      - :whale: `--log-opt=fluentd-buffer-limit=<LIMIT>`: The buffer limit for fluentd. If the buffer is full, the call to record logs will fail. The default is 1MiB. Accepts human-readable sizes (e.g., `1k`, `1m`, `1g`) or raw byte values. (<https://github.com/fluent/fluent-logger-golang/tree/master#bufferlimit>)
```
utafrali: The examples `1k`, `1m`, `1g` use `RAMInBytes` (binary units), so `1m` = 1,048,576 bytes, not 1,000,000. This surprises users coming from SI conventions. Consider using IEC suffixes in the examples (`1KiB`, `1MiB`, `1GiB`) or adding a short note clarifying that the binary map is used, which is consistent with how Docker documents this option.

utafrali (commented):
The core change is correct and backward-compatible: swapping `strconv.Atoi` for `units.RAMInBytes` is the right approach and the existing test suite validates the happy path. The main issues are a silent `int64`-to-`int` narrowing cast and missing error-path test coverage after the parser change.

shouhei:
@utafrali Thanks for the review.

I updated the PR to follow your suggestions:
- added the explicit overflow check before converting `int64` to `int`
- added an invalid-value test case for `fluentd-buffer-limit`
- updated the documentation examples to use IEC suffixes

haytok (approved):
LGTM, Thanks!

AkihiroSuda (approved):
Thanks

PR #4857, opened 2026-04-18
fix: wait for logger to finish before reading logs of stopped containers

haytok:
`TestLogsFollowNoExtraneousLineFeed` sometimes fails on CI.

- https://github.com/containerd/nerdctl/actions/runs/24197307638/job/71240723501?pr=4835
- https://github.com/containerd/nerdctl/actions/runs/24197307638/job/71215370984?pr=4835

```bash
        | ➡️      | ⚙️ /usr/local/bin/nerdctl.gomodjail run --name testlogsfollownoextraneouslinefeed-f5d64e9b ghcr. |
        |         | io/stargz-containers/alpine:3.13-org sh -c printf 'Hello without newline'                        |
        |         | 🟢 Hello without newline
...
    container_logs_test.go:562: 	🔗

        <<<<<<<<<<<<<<<<<<<<
        	🖊️ Inspecting output (equals)
        	👀 testing:		``
        	❌ FAILED!		= `Hello without newline`
        >>>>>>>>>>>>>>>>>>>>
```

My Investigation revealed that the flakiness of
`TestLogsFollowNoExtraneousLineFeed` is caused by `nerdctl logs -f` not waiting for the logger to finish writing when the container has already stopped.

Therefore, this commit fixes `nerdctl logs -f` to wait for the logger to finish writing when the container has already stopped.

This commit should fix the flakiness of `TestLogsFollowNoExtraneousLineFeed`.

After applying this fix, running the test **1000 times** showed no flakiness.

```bash
$ sudo go test -count=1000 -run '^TestLogsFollowNoExtraneousLineFeed$'
test target: "nerdctl"
PASS
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/container	419.653s
```

### DIFF
--- a/pkg/cmd/container/logs.go
+++ b/pkg/cmd/container/logs.go
@@ -105,6 +105,14 @@ func Logs(ctx context.Context, client *containerd.Client, container string, opti
 				}
 			}
 
+			// When follow was requested but the task has already stopped,
+			// wait for the logger to finish writing before reading the log file.
+			if !follow && options.Follow {
+				if err := logging.WaitForLogger(dataStore, l[labels.Namespace], found.Container.ID()); err != nil {
+					log.G(ctx).WithError(err).Warn("failed to wait for logger shutdown")
+				}
+			}
+
 			var detailPrefix string
 			if options.Details {
 				if logConfigJSON, ok := l["nerdctl/log-config"]; ok {


haytok:
## My Investigation

Timing-dependent bugs often pass in a single test run but can be detected by increasing the number of iterations.

```bash
sudo go test -count=100 -run '^TestLogsFollowNoExtraneousLineFeed$'
```

Running this command confirmed that TestLogsFollowNoExtraneousLineFeed fails intermittently for the same reason CI failed.

So I then added the following debug code and reproduced the failure:

```golang
	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
		return &test.Expected{
			ExitCode: 0,
			Output: func(stdout string, info tig.T) {
				info.Helper()
				if stdout != "Hello without newline" {
					logPath := helpers.Capture("inspect", "--format", "{{.LogPath}}", data.Identifier())
					content, err := os.ReadFile(strings.TrimSpace(logPath))
					if err != nil {
						info.Log(fmt.Sprintf("=== failed to read JSON log: %v ===", err))
					} else {
						info.Log(fmt.Sprintf("=== JSON log content (stdout mismatch) ===\n%s", content))
					}
				}
				expect.Equals("Hello without newline")(stdout, info)
			},
		}
	}
```

The result showed that the JSON log file contained the expected data, but the output of `nerdctl logs -f` did not reflect it.

To summarize the analysis:

- The JSON log file already contains the expected output at the time the debug message is logged in the comparator.
- When `helpers.Command("logs", "-f", data.Identifier())` is executed, `helpers.Ensure("run", "--name", data.Identifier(), "sh", "-c", "printf 'Hello without newline'")` has already completed writing to stdout, but the logger process has not finished writing to the JSON log file.

Based on this, analyzing the logging code confirms that when a container has already stopped, `nerdctl logs -f` does not wait for the logger to finish processing the remaining logs:

https://github.com/containerd/nerdctl/blob/4e5cc62bf19ae7f86e46fe577b3a381b0a9574bc/pkg/cmd/container/logs.go#L86-L104

Therefore, this commit should fix the flakiness of `TestLogsFollowNoExtraneousLineFeed`.

AkihiroSuda (approved):
Thanks

PR #4862, opened 2026-04-22
fix: free reserved ports in rootful mode

### LINKED ISSUE #4843: Host reserved ports are not freed when container stops in rootful mode
### Description

In rootful mode, port mappings will spawn a `sleep infinity` process on the host to reserve the ports.

When the container stops, the iptables rules are cleaned up as expected. However, the reserved process remains and continues to stay in the LISTEN state.

This behavior can be confusing in some scenarios, as the port appears to still be in use even though the container has already exited.

Related change in v2.1.6:
    rootful: reserve the ports on the host (#4526 )

1. In rootful mode, root user
2. `nerdctl run -d  -p  7000:7000  -name test-container test-image` 
3. `nerdctl stop test-container`
4. `ss -ntlp | grep ":7000"`  (or  `netstat -ntlp` )

## Received:
After the container stops, `ss -ntlp` still shows a `sleep infinity` process in the LISTEN state.

## Expected:
Would it be possible to add cleanup logic for this process when the container stops?

Furthermore, is it possible to support a configuration similar to Docker:
```
# /etc/docker/daemon.json
{
  "userland-proxy": false
}
```

v2.1.6

None

### Host information

nerdctl-full-2.1.6-linux-amd64.tar.gz

unsuman:
Fixes #4843

This PR reuses the existing clean-up logic in `pkg/ocihook/ocihook.go` on `nerdctl stop` to free up reserved ports in rootful mode.

### DIFF
--- a/pkg/ocihook/ocihook.go
+++ b/pkg/ocihook/ocihook.go
@@ -456,8 +456,24 @@ func reserveSocket(protocol, hostAddr string) (*os.File, error) {
 }
 
 // portReserverPidFilePath returns /run/nerdctl/<namespace>/<id>/port-reserver.pid
-func portReserverPidFilePath(opts *handlerOpts) string {
-	return filepath.Join("/run/nerdctl/", opts.state.Annotations[labels.Namespace], opts.state.ID, "port-reserver.pid")
+func portReserverPidFilePath(namespace, id string) string {
+	return filepath.Join("/run/nerdctl/", namespace, id, "port-reserver.pid")
+}
+
+func CleanupPortReserverProcess(namespace, id string) error {
+	// In rootless mode, port-reserver is handled by Rootlesskit, so no cleanup is needed.
+	if rootlessutil.IsRootlessChild() {
+		return nil
+	}
+
+	pidFile := portReserverPidFilePath(namespace, id)
+	if err := killProcessByPidFile(pidFile); err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


pkg/ocihook/ocihook.go
```diff
@@ -503,10 +519,9 @@ func applyNetworkSettings(opts *handlerOpts) (err error) {
 			if err != nil {
 				log.L.Debugf("killing the port reserver process (pid=%d)", reserverCmdPid)
 				_ = reserverCmd.Process.Kill()
-				_ = os.RemoveAll(filepath.Dir(portReserverPidFilePath(opts)))
 			}
 		}()
-		if err := writePidFile(portReserverPidFilePath(opts), reserverCmdPid); err != nil {
+		if err := writePidFile(portReserverPidFilePath(opts.state.Annotations[labels.Namespace], opts.state.ID), reserverCmdPid); err != nil {
```
unsuman, line 525: Interesting! The code is already in the desired form, right?

```
Run . ./hack/github/action-helpers.sh
go: downloading github.com/miekg/pkcs11 v1.1.1
# github.com/containerd/nerdctl/v2/pkg/ocihook
Error: pkg/ocihook/ocihook.go:522:59: not enough arguments in call to portReserverPidFilePath
	have (*handlerOpts)
	want (string, string)
Error: Process completed with exit code 1.
```
**[EDIT]**: My branch was not up to date with the latest main 🙂

AkihiroSuda (approved):
Thanks

PR #4863, opened 2026-04-22
test: refactor container_stop_linux.go to use Tigron

opjt:
- related issue : #4613

### DIFF
--- a/cmd/nerdctl/container/container_stop_linux_test.go
+++ b/cmd/nerdctl/container/container_stop_linux_test.go
@@ -19,6 +19,7 @@ package container
 import (
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -27,30 +28,22 @@ import (
 	"gotest.tools/v3/assert"
 
 	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/require"
 	"github.com/containerd/nerdctl/mod/tigron/test"
+	"github.com/containerd/nerdctl/mod/tigron/tig"
 
-	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 	iptablesutil "github.com/containerd/nerdctl/v2/pkg/testutil/iptables"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nettestutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/portlock"
 )
 
 func TestStopStart(t *testing.T) {
-	const (
-		hostPort = 8080
-	)
-	testContainerName := testutil.Identifier(t)
-	base := testutil.NewBase(t)
-	defer base.Cmd("rm", "-f", testContainerName).Run()
+	testCase := nerdtest.Setup()
 
-	base.Cmd("run", "-d",
-		"--restart=no",
-		"--name", testContainerName,
-		"-p", fmt.Sprintf("127.0.0.1:%d:80", hostPort),
-		testutil.NginxAlpineImage).AssertOK()
-
-	check := func(httpGetRetry int) error {
+	httpCheck := func(data test.Data, httpGetRetry int) error {
+		hostPort, _ := strconv.Atoi(data.Labels().Get("hostPort"))
 		resp, err := nettestutil.HTTPGet(fmt.Sprintf("http://127.0.0.1:%d", hostPort), httpGetRetry, false)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -27,53 +27,71 @@ import (
...
-			return err
-		}
-		respBody, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return err
-		}
-		if !strings.Contains(string(respBody), testutil.NginxAlpineIndexHTMLSnippet) {
-			return fmt.Errorf("expected contain %q, got %q",
-				testutil.NginxAlpineIndexHTMLSnippet, string(respBody))
+	const hostPort = 8080
```
haytok: We should use `portlock package` to acquire and release ports:

https://github.com/containerd/nerdctl/blob/main/pkg/testutil/portlock/portlock.go

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -91,131 +109,166 @@ func TestStopWithStopSignal(t *testing.T) {
...
-	} else {
-		redirectChain := "CNI-HOSTPORT-DNAT"
-		chain = iptablesutil.GetRedirectedChain(t, ipt, redirectChain, testutil.Namespace, containerID)
-	}
-	assert.Equal(t, iptablesutil.ForwardExists(t, ipt, chain, containerIP, hostPort), true)
-
-	base.Cmd("stop", testContainerName).AssertOK()
-	assert.Equal(t, iptablesutil.ForwardExists(t, ipt, chain, containerIP, hostPort), false)
-}
+	const hostPort = 9999
```
haytok: We should use portlock package to acquire and release ports.

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -27,53 +27,71 @@ import (
...
+
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		helpers.Ensure("run", "-d",
+			"--restart=no",
+			"--name", data.Identifier(),
+			"-p", fmt.Sprintf("127.0.0.1:%d:80", hostPort),
+			testutil.NginxAlpineImage)
+	}
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
```
haytok: Could you split test scenarios into SubTests with NoParallel?

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -27,53 +27,71 @@ import (
...
-		t.Fatal("expected to get an error")
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: 0,
+			Output: func(stdout string, _ tig.T) {
+				resp, err := nettestutil.HTTPGet(fmt.Sprintf("http://127.0.0.1:%d", hostPort), 5, false)
+				assert.NilError(t, err)
+				respBody, err := io.ReadAll(resp.Body)
+				assert.NilError(t, err)
+				assert.Assert(t, strings.Contains(string(respBody), testutil.NginxAlpineIndexHTMLSnippet))
```
haytok: It would be better to enable the use of the check function and to use it.

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -91,131 +109,166 @@ func TestStopWithStopSignal(t *testing.T) {
...
 	}
 
-	t.Cleanup(tearDown)
-	tearDown()
-	setup()
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		ipt, err := iptables.New()
+		assert.NilError(t, err)
+		chain := data.Labels().Get("chain")
+		containerIP := data.Labels().Get("containerIP")
```
haytok: NIT: better to move iptables.New() and the data.Labels().Get(...) calls into the Output callback.

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -91,131 +109,166 @@ func TestStopWithStopSignal(t *testing.T) {
...
-	}()
-	// Start a container that is automatically removed after it exits
-	base.Cmd("run", "--rm", "--name", testContainerName, testutil.NginxAlpineImage).AssertOK()
+		assert.NilError(helpers.T(), err)
+		assert.Equal(helpers.T(), oldNumFifos, newNumFifos)
+
+		return cmd
+	}
+
+	testCase.Expected = test.Expects(expect.ExitCodeNoCheck, nil, nil)
```
haytok: The FIFO count belongs in Setup and the comparison in Expected.Output — Command should just return helpers.Command("stop", ...) since stop is what's being tested. WDYT?

opjt: Agreed. I also felt a bit off about this part

opjt:
@haytok thanks for your review! Could you take another look?

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -66,14 +82,41 @@ func TestStopStart(t *testing.T) {
...
+			NoParallel:  true,
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("stop", data.Labels().Get("containerName"))
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: expect.ExitCodeSuccess,
+					Output: func(stdout string, _ tig.T) {
+						helpers.Fail("exec", data.Labels().Get("containerName"), "ps")
+						assert.Assert(t, httpCheck(data, 1) != nil, "expected HTTP to fail after stop")
```
haytok: If you use the outer `t` in an assert within a SubTest callback, Go testing will issue a warning stating `test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test` when the test fails, so we should fix this.

```suggestion
					Output: func(stdout string, t tig.T) {
						helpers.Fail("exec", data.Labels().Get("containerName"), "ps")
						assert.Assert(t, httpCheck(data, 1) != nil, "expected HTTP to fail after stop")
```

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -66,14 +82,41 @@ func TestStopStart(t *testing.T) {
...
+		{
+			Description: "container is restarted",
+			NoParallel:  true,
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("start", data.Labels().Get("containerName"))
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: expect.ExitCodeSuccess,
+					Output: func(stdout string, _ tig.T) {
```
haytok: Same as above:

```suggestion
					Output: func(stdout string, t tig.T) {
```

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -91,131 +134,183 @@ func TestStopWithStopSignal(t *testing.T) {
...
-	start := time.Now()
-	base.Cmd("stop", "--time=3", testContainerName).AssertOK()
-	elapsed := time.Since(start)
+	testCase.Setup = func(data test.Data, helpers test.Helpers) {
+		// Start a container that sleeps forever
+		helpers.Ensure("run", "-d", "--name", data.Identifier(), testutil.CommonImage, "sleep", nerdtest.Infinity)
+	}
 
-	// The container should get the SIGKILL before the 10s default timeout
-	assert.Assert(t, elapsed < 10*time.Second, "Container did not respect --timeout flag")
```
haytok, line 200: Why did you delete this test?

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -91,131 +134,183 @@ func TestStopWithStopSignal(t *testing.T) {
...
+	}
 
-	// Stop the container with a 5-second timeout and SIGKILL
-	start := time.Now()
-	base.Cmd("stop", "--time=5", "--signal", "SIGKILL", testContainerName).AssertOK()
-	elapsed := time.Since(start)
+	testCase.Expected = test.Expects(expect.ExitCodeSuccess, nil, nil)
 
-	// The container should be stopped almost immediately, well before the 5-second timeout
-	assert.Assert(t, elapsed < 5*time.Second, "Container wasn't stopped immediately with SIGKILL")
```
haytok, line 182: Why did you delete this test?

cmd/nerdctl/container/container_stop_linux_test.go
```diff
@@ -66,14 +82,41 @@ func TestStopStart(t *testing.T) {
 		return nil
 	}
 
-	assert.NilError(t, check(5))
```
haytok, line 69: Has this been deleted?

cmd/nerdctl/container/container_stop_linux_test.go
haytok, line 99: Shouldn't this be split into a separate subtest?

opjt: Thanks for the review, Will fix it

cmd/nerdctl/container/container_stop_linux_test.go
opjt: Will add it back..

cmd/nerdctl/container/container_stop_linux_test.go
opjt: The elapsed time assertion was replaced with cmd.WithTimeout(5s), which is more direct — if the stop command doesn't complete within 5 seconds, the test fails immediately

cmd/nerdctl/container/container_stop_linux_test.go
opjt: I will fix it

cmd/nerdctl/container/container_stop_linux_test.go
opjt: I think keeping them in the same subtest is fine as a verification that stop was successful.

cmd/nerdctl/container/container_stop_linux_test.go
opjt: Added it back in testCase.Setup

cmd/nerdctl/container/container_stop_linux_test.go
opjt: replaced `cmd.WithTimeout(10 * time.Second)`

haytok (approved):
LGTM, Thanks! Let's see if CI can be on green.

opjt:
@AkihiroSuda Could you please re-run the CI when you have a moment? Thank you!

AkihiroSuda (approved):
Thanks

Issue #4709, opened 2026-01-19
nerdctl compose should support djocker compose service health check


guangdashao:
### What is the problem you're trying to solve

nerdctl compose  supprt docker compose service health check 

### Describe the solution you'd like

nerdctl compose  supprt docker compose service health check

PR #4869, opened 2026-04-27
feat: support healthcheck field in service config

opjt:
- related issue : https://github.com/containerd/nerdctl/issues/4709

 ## Summary

`services.<SERVICE>.healthcheck` was listed as unimplemented — this PR wires it up.

compose-go already parses the `healthcheck` block, so this is mostly connecting
it to the existing `--health-*` flags in serviceparser.

**Limitations:**
- `CMD` exec form is converted to `CMD-SHELL` since `--health-cmd` always stores the command as `CMD-SHELL` (see `withHealthcheck` in `pkg/cmd/container/create.go`)
- `start_interval` is not supported — I plan to work on this after this PR is merged.

 ## Test plan

  ```yaml
  services:
    web:
      image: nginx:alpine
      healthcheck:
        test: ["CMD-SHELL", "curl -f http://localhost"]
        interval: 10s
        timeout: 5s
        retries: 3
        start_period: 2s
```
```bash
  nerdctl compose up -d
  nerdctl inspect <container-name> | grep health
```

### DIFF
--- a/cmd/nerdctl/compose/compose_up_linux_test.go
+++ b/cmd/nerdctl/compose/compose_up_linux_test.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/docker/go-connections/nat"
 	"gotest.tools/v3/assert"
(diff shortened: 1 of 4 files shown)


opjt:
Hi, just following up on this PR. Thanks!

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -747,6 +764,44 @@ func newContainer(project *types.Project, parsed *Service, i int) (*Container, e
 		c.RunArgs = append(c.RunArgs, "-w="+svc.WorkingDir)
 	}
 
+	if svc.HealthCheck != nil {
+		hc := svc.HealthCheck
+		if hc.Disable {
+			c.RunArgs = append(c.RunArgs, "--no-healthcheck")
+		} else {
+			if len(hc.Test) > 0 {
+				switch hc.Test[0] {
```
haytok: The `default` statement is missing, so please add it and modify the code to output a warning.

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -747,6 +764,44 @@ func newContainer(project *types.Project, parsed *Service, i int) (*Container, e
...
+					// CMD exec form is converted to CMD-SHELL because --health-cmd always stores
+					// the command as CMD-SHELL (see pkg/cmd/container/create.go: withHealthcheck).
+					// This means the command will be executed via /bin/sh -c instead of exec directly.
+					if len(hc.Test) >= 2 {
+						log.L.Warnf("service %s: healthcheck: CMD exec form is not supported, converting to CMD-SHELL", svc.Name)
+						c.RunArgs = append(c.RunArgs, fmt.Sprintf("--health-cmd=%s", strings.Join(hc.Test[1:], " ")))
+					}
+				}
+			}
+			if hc.Interval != nil {
```
haytok, line 792: Although flags such as `--health-interval` are added to the command when specifying `test: ["NONE"]`, they should not be included.

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -747,6 +764,44 @@ func newContainer(project *types.Project, parsed *Service, i int) (*Container, e
...
+		hc := svc.HealthCheck
+		if hc.Disable {
+			c.RunArgs = append(c.RunArgs, "--no-healthcheck")
+		} else {
+			if len(hc.Test) > 0 {
+				switch hc.Test[0] {
+				case healthcheck.CmdNone:
+					c.RunArgs = append(c.RunArgs, "--no-healthcheck")
+				case healthcheck.CmdShell:
+					if len(hc.Test) >= 2 {
```
haytok: Warning messages should be added `if len(hc.Test) < 2`. WDYT?.

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -747,6 +764,44 @@ func newContainer(project *types.Project, parsed *Service, i int) (*Container, e
...
+						c.RunArgs = append(c.RunArgs, fmt.Sprintf("--health-cmd=%s", hc.Test[1]))
+					}
+				case healthcheck.Cmd:
+					// CMD exec form is converted to CMD-SHELL because --health-cmd always stores
+					// the command as CMD-SHELL (see pkg/cmd/container/create.go: withHealthcheck).
+					// This means the command will be executed via /bin/sh -c instead of exec directly.
+					if len(hc.Test) >= 2 {
```
haytok: Warning messages should be added `if len(hc.Test) < 2`. WDYT?.

cmd/nerdctl/compose/compose_up_linux_test.go
```diff
@@ -1314,3 +1315,61 @@ services:
 
 	testCase.Run(t)
 }
+
+func TestComposeUpHealthcheck(t *testing.T) {
```
haytok, line 1319: IMO, tests in case of `test: ["NONE"]` and `disable: true` should be added.

- https://github.com/compose-spec/compose-spec/blob/main/spec.md#healthcheck

pkg/composer/serviceparser/serviceparser_test.go
```diff
@@ -623,3 +623,64 @@ services:
...
+	project, err := testutil.LoadProject(comp.YAMLFullPath(), comp.ProjectName(), nil)
+	assert.NilError(t, err)
+
+	getContainersFromService := func(svcName string) []Container {
+		svcConfig, err := project.GetService(svcName)
+		assert.NilError(t, err)
+		svc, err := Parse(project, svcConfig)
+		assert.NilError(t, err)
+		return svc.Containers
+	}
```
haytok: This function can be used in `TestParseRestartPolicy`, so that should be defined in outer scope.

pkg/composer/serviceparser/serviceparser.go
opjt: I'll add the `default` case with a warning log.

pkg/composer/serviceparser/serviceparser_test.go
opjt: I'll move it to the outer scope. thanks

cmd/nerdctl/compose/compose_up_linux_test.go
opjt: Since `serviceparser_test` already covers both `test: ["NONE"]` and `disable: true` mapping to `--no-healthcheck`, I think adding just one of them in the test is sufficient. WDYT?

pkg/composer/serviceparser/serviceparser.go
opjt: I will fix it to avoid appending unnecessary flags.

pkg/composer/serviceparser/serviceparser.go
opjt: The `default` case is unreachable since `compose-go` already validates the healthcheck test type.
I'll remove it

<https://github.com/compose-spec/compose-go/blob/v2.10.2/loader/validate.go#L80>

pkg/composer/serviceparser/serviceparser.go
opjt: Do we really need to warn the user about what is simply a user error?
Docker silently ignores this case, so I'm not sure which direction we should go in this project.

haytok: Since nothing was displayed in Docker when `len(hc.Test) == 1`, it’s best to leave it as is in nerdctl as well. Thanks!

pkg/composer/serviceparser/serviceparser_test.go
```diff
@@ -601,25 +610,69 @@ services:
...
+	assert.Assert(t, in(c.RunArgs, "--health-retries=3"))
+	assert.Assert(t, in(c.RunArgs, "--health-start-period=5s"))
+
+	c = getContainersFromService(t, project, "cmd_exec")[0]
+	assert.Assert(t, in(c.RunArgs, "--health-cmd=curl -f http://localhost"))
+	assert.Assert(t, in(c.RunArgs, "--health-interval=1m0s"))
+
+	c = getContainersFromService(t, project, "disabled_flag")[0]
+	assert.Assert(t, in(c.RunArgs, "--no-healthcheck"))
+	assert.Assert(t, !in(c.RunArgs, "--health-cmd="))
```
haytok: `in` is `strutil.InStringSlice` (exact element match), so `in(c.RunArgs, "--health-cmd=")` never matches the actual element (`--health-cmd=<value>`). Shouldn't we check whether any of the argument elements have a prefix of `--health-cmd=`?

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -747,6 +764,46 @@ func newContainer(project *types.Project, parsed *Service, i int) (*Container, e
 		c.RunArgs = append(c.RunArgs, "-w="+svc.WorkingDir)
 	}
 
+	if svc.HealthCheck != nil {
```
haytok, line 767: When both `disable: true` and `test: [...]` are specified like below, the switch still appends `--health-cmd=...` before the if disabled check, so nerdctl run ends up with both `--health-cmd=...` and `--no-healthcheck`.

```yaml
services:
  web:
    image: nginx:alpine
    healthcheck:
      test: ["CMD-SHELL", "curl -f http://localhost"]
      interval: 1m
      timeout: 5s
      retries: 3
      start_period: 2s
      disable: true
```

In this case, `test` should be ignored.

pkg/composer/serviceparser/serviceparser.go
```diff
@@ -121,6 +123,21 @@ func warnUnknownFields(svc types.ServiceConfig) {
...
+		if unknown := reflectutil.UnknownNonEmptyFields(svc.HealthCheck,
+			"Test",
+			"Timeout",
+			"Interval",
+			"Retries",
+			"StartPeriod",
+			"Disable",
+			"Extensions",
+			// TODO: add support 'startinterval'
```
haytok: NIT:
```suggestion
			// TODO: add support 'StartInterval'
```

pkg/composer/serviceparser/serviceparser.go
opjt: I'll fix this and update the test case

haytok (approved):
LGTM, Thanks!

AkihiroSuda (approved):
Thanks

PR #4639, opened 2025-12-06
feature: support selinux

ningmingxiao:
ping @AkihiroSuda

Issue #3867, opened 2025-02-06
Update the list of the unimplemented Docker features (and implement them)


AkihiroSuda:
The list of the unimplemented Docker features is outdated:
https://github.com/containerd/nerdctl/blob/21590cd689246dd8ec4b3b59b843e6f04b9317e9/docs/command-reference.md?plain=1#L414-L418

- [x] Make RO mounts RRO by default https://github.com/containerd/nerdctl/issues/2651
  - Docker-side PR: https://github.com/moby/moby/pull/45278 https://github.com/moby/moby/pull/46037
- [ ] `--mount type=image` https://github.com/containerd/nerdctl/pull/4993
  - Docker-side PR: https://github.com/moby/moby/pull/48798
- [ ] `--security-opt writable-cgroups`
  - Docker-side PR: https://github.com/moby/moby/pull/48828
- [ ] buildx debugger
- [X] CDI https://github.com/containerd/nerdctl/pull/4170
- [X] mount `:z` and `:Z` https://github.com/containerd/nerdctl/issues/371 https://github.com/containerd/nerdctl/pull/4639
- [X] userns-remap mode https://github.com/containerd/nerdctl/pull/3941
- (And more, see **Sub-issues** below)

apostasie:
"Good first issue"?

AkihiroSuda:
> "Good first issue"?

"Good first issue, only for those who have good knowledge and experiences with Docker" 😅

apostasie:
"Good first issue, only for those who have good knowledge and experiences with Docker and also who have the time and desire to indulge in ungrateful grunt documentation work instead of being snarky on github issues"

IstvanCsVarga:
Took a stab at this, opened #4809 with some initial fixes. `--health-start-interval` was incorrectly marked as implemented, there was a pull/push typo in the compose section, and the unimplemented compose commands list was missing a bunch of V2 subcommands. Happy to adjust.

ogulcanaydogan:
I'd like to pick up remaining items after #4809. Will audit the current unimplemented list against recent merged PRs and submit a targeted doc update for features that have since been implemented.

jiwahn:
I would like to give --mount type=image support a shot. Is this still relevant / up for grabs?

AkihiroSuda:
> I would like to give --mount type=image support a shot. Is this still relevant / up for grabs?

Yes, please, thanks

mujii88:
[...]

PR #4882, opened 2026-05-01
docs: fix missing list prefix for --security-opt label entry

ogulcanaydogan:
The `--security-opt label=<selinuxlabel>` entry was added in #4639 but
was indented with two spaces as a continuation of the line above rather
than as its own Markdown list item. This causes it to render as
sub-content of the `--security-opt apparmor` bullet instead of a
sibling entry.

Change the leading `  ` to `- ` to match all other entries in the
Security flags list.

Partial fix for #3867.


yankay:
LGTM, thanks!

AkihiroSuda (approved):

PR #4890, opened 2026-05-10
fix: handle CNI config stat errors without panic

immanuwell:
Small fix for a CNI config path footgun.

If `--cni-netconfpath` / `$NETCONFPATH` points at a regular file, `fsExists` gets a stat error and then derefs a nil `FileInfo`. Boom: panic. This can happen from a bad local path, no weird infra limit needed.

Repro:
1. Add `TestFSExistsPropagatesStatError` from this PR on `main`.
2. Run `go test ./pkg/netutil -run TestFSExistsPropagatesStatError -count=1`.
3. It panics at `pkg/netutil/store.go:56`.

This returns the stat error instead, and keeps missing-file behavior intact.

Tests:
- `go test ./pkg/netutil -run TestFSExistsPropagatesStatError -count=1`
- `go test ./pkg/netutil -count=1`
- `go test ./pkg/...`

### DIFF
--- a/pkg/netutil/netutil_test.go
+++ b/pkg/netutil/netutil_test.go
@@ -366,3 +366,16 @@ func TestNetworkWithDefaultNameAlreadyExists(t *testing.T) {
 	assert.Assert(t, len(defaultNamedNetworksFileDefinitions) == 1)
 	assert.Assert(t, defaultNamedNetworksFileDefinitions[0] == testConfFile)
 }
+
+func TestFSExistsPropagatesStatError(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "cni-conf-root")
+	assert.NilError(t, filesystem.WriteFile(path, nil, 0600))
+
+	cniEnv := CNIEnv{
+		NetconfPath: path,
+	}
+
+	exists, err := fsExists(&cniEnv, DefaultNetworkName)
+	assert.Assert(t, !exists)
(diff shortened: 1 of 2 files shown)


haytok (approved):
Not sure if the scenario covered by `TestFSExistsPropagatesStatError` can happen via `--cni-netconfpath` from the CLI, but if `os.Stat` returns a non-NotExist error the current code panics on the `nil`, and returning the error instead is obviously safe. LGTM, Thanks!

sathiraumesh (approved):

pkg/netutil/netutil_test.go
```diff
@@ -366,3 +366,16 @@ func TestNetworkWithDefaultNameAlreadyExists(t *testing.T) {
...
+	path := filepath.Join(t.TempDir(), "cni-conf-root")
+	assert.NilError(t, filesystem.WriteFile(path, nil, 0600))
+
+	cniEnv := CNIEnv{
+		NetconfPath: path,
+	}
+
+	exists, err := fsExists(&cniEnv, DefaultNetworkName)
+	assert.Assert(t, !exists)
+	assert.Assert(t, err != nil)
```
AkihiroSuda, line 380: nit: https://pkg.go.dev/gotest.tools/v3/assert#Error could be used

AkihiroSuda (approved):

PR #4897, opened 2026-05-14
refactor: migrate multi_platform_linux_test.go to nerdtest.Setup

ogulcanaydogan:
Part of #4613.

Migrates `cmd/nerdctl/container/multi_platform_linux_test.go` from `testutil.NewBase` to the Tigron `nerdtest.Setup` framework, following the pattern established in #4641.

Changes:
- `testutil.RequireExecPlatform` replaced with a custom `requireMultiPlatformExec` requirement backed by `platformutil.CanExecProbably`
- `testregistry.NewWithNoAuth` replaced with `nerdtest.RegistryWithNoAuth`
- Dockerfiles and compose files written via `data.Temp().Save()`
- `TestMultiPlatformRun` converted to use `SubTests` for per-platform isolation
- `testMultiPlatformRun` helper updated to use `test.Helpers`
- Builder cache prune registered in `Cleanup` callbacks
- `testutil.DockerIncompatible` replaced with `require.Not(nerdtest.Docker)`
- `testutil.RequiresBuild` replaced with `nerdtest.Build`

PR #4898, opened 2026-05-14
refactor: migrate container_run_log_driver_syslog_test.go to nerdtest.Setup

ogulcanaydogan:
Closes part of #4613.

## What

Migrates `cmd/nerdctl/container/container_run_log_driver_syslog_test.go` from the legacy `testutil.Base` / `testutil.NewBase` pattern to the `nerdtest.Setup` / `test.Case` framework, following the same approach used in #4897 (`multi_platform_linux_test.go`) and #4641.

## Design notes

The three tests (TestSyslogNetwork, TestSyslogFacilities, TestSyslogFormat) each expand a cross-product of parameters into independent sub-cases via `buildSyslogSubTests`:

- Each sub-case manages its own syslog server lifecycle (start in `Setup`, receive in `Cleanup`) using closure-captured variables (`addr`, `done`, `closer`, `containerName`, `tag`, `msg`), which is necessary because the channel-based frame validation cannot be expressed through `data.Labels()`.
- CA and cert are shared from the outer fixture to sub-cases via pointer-to-pointer (`**testca.CA`, `**testca.Cert`) populated in the outer `testCase.Setup`.
- `testca.New` requires `testing.TB`; it is passed as `*testing.T` through the `newSyslogTestCase(t *testing.T)` helper rather than `helpers.T()`, which only implements the narrower `tig.T` interface.
- Network skip logic respects the rootless case with a distinct message per the existing convention in the repo.

Validator functions (`rfc5424Validator`, `rfc3164Validator`, `emptyFormatValidator`) are extracted as package-level helpers; their logic is identical to the originals.

## Test plan

- `go vet ./cmd/nerdctl/container/...` passes with no errors.
- Logic is a mechanical translation of the existing test; no behaviour changes.

### DIFF
--- a/cmd/nerdctl/container/container_run_log_driver_syslog_test.go
+++ b/cmd/nerdctl/container/container_run_log_driver_syslog_test.go
@@ -18,8 +18,8 @@ package container
 
 import (
 	"fmt"
+	"io"
 	"os"
-	"runtime"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
```
=== Failing tests ===
TestLoadStdinFromPipe
TestSyslogNetwork
TestSyslogNetwork/udp_1_rfc5424
TestSyslogNetwork/udp_user_rfc5424
TestSyslogNetwork/unixgram_1_rfc5424
TestSyslogNetwork/unixgram_user_rfc5424
```

https://github.com/containerd/nerdctl/actions/runs/25862295284/job/75995732134?pr=4898

ogulcanaydogan:
Thanks for catching this. The root cause: `runPacketSyslog` has a 4x100ms read window (~400ms total). Running the subtests in parallel (Tigron default) puts the arm runner under load, pushing container startup past that window so the server drains before the log entry arrives. Fixed by setting `NoParallel: true` on each case, matching the original sequential `t.Run` behaviour.

ogulcanaydogan:
Thanks for catching this. The root cause is in `runPacketSyslog`: UDP and unixgram use a deadline-based read loop with a 4x100ms (400ms) hard cap. On slow ARM runners, container startup plus syslog driver initialization exceeds that budget, so the server closes the socket and sends an empty string on the channel before any datagram arrives.

Fixed in 66c23586 by increasing the pre-packet retry count from 3 to 20 (2s total). Once the first datagram arrives, it resets to 3 retries (400ms drain) to avoid lingering unnecessarily.

cmd/nerdctl/container/container_run_log_driver_syslog_test.go
```diff
@@ -18,111 +18,195 @@ package container
...
 		for rFK, rFV := range syslogFacilities {
+			rFK := rFK
 			fPriV := rFV
-			// test both string and number facility
 			for _, fPriK := range []string{rFK, strconv.Itoa(int(fPriV) >> 3)} {
+				fPriK := fPriK
 				for fmtK, fmtValidFunc := range fmtValidFuncs {
+					fmtK := fmtK
+					fmtValidFunc := fmtValidFunc
+
```
sathiraumesh: we can just create a combination structure first and then later iterate over and run over them. I think its much readable and better
```
type params struct {
		network      string
		fPriK        string
		fPriV        syslog.Priority
		fmtK         string
		fmtValidFunc func(string, string, string, string, syslog.Priority, bool) error
	}
	var cobintaions []params
	for _, network := range networks {
		for rFK, rFV := range syslogFacilities {
			for _, fPriK := range []string{rFK, strconv.Itoa(int(rFV) >> 3)} {
				for fmtK, fmtValidFunc := range fmtValidFuncs {
					combos = append(combos, params{network, fPriK, rFV, fmtK, fmtValidFunc})
				}
			}
		}
	}
```

cmd/nerdctl/container/container_run_log_driver_syslog_test.go
```diff
@@ -18,111 +18,195 @@ package container
...
-						defer closer.Close()
-						defer close(done)
-						select {
-						case rcvd := <-done:
-							if err := fmtValidFunc(rcvd, msg, tag, hostname, fPriV, network == "tcp+tls"); err != nil {
-								t.Error(err)
+							args = append(args, testutil.CommonImage, "echo", msg)
+							return helpers.Command(args...)
+						},
+						Expected: test.Expects(0, nil, nil),
```
sathiraumesh: Can use exit code constants from the` tigron` framework

sathiraumesh (changes requested):

ogulcanaydogan:
Addressed both points:
1. Refactored `buildSyslogSubTests` to pre-build a `[]syslogCombination` slice before constructing test cases, removing the nested loop variable shadowing (`x := x` pattern).
2. Replaced `test.Expects(0, nil, nil)` with `test.Expects(expect.ExitCodeSuccess, nil, nil)`.

ogulcanaydogan:
Thanks for the review! Regarding the params struct suggestion: the current implementation already uses this exact pattern — `syslogCombination` struct collects all combinations first, then `buildSyslogSubTests` iterates over them to build the Tigron sub-cases. The exit code at line 148 already uses `expect.ExitCodeSuccess` from the Tigron framework. Please let me know if anything else needs adjusting.

AkihiroSuda (approved):
Thanks

PR #4901, opened 2026-05-14
Compose down to accept a list of services

ysedira:
runing `compose down <service-name>` with a list of service names fails with an error `unknown command <service-name> for down`.

However, this work with docker-compose without an issue, and also work for `compose up`.

This PR address the issue by passing on args to downAction and if present, only specified services will be stoped and removed.


AkihiroSuda (approved):
Thanks

ChengyuZhu6 (approved):
LGTM

PR #4904, opened 2026-05-15
docs: remove stray quote from interactive flag

MukundaKatta:
## Summary
- remove an extra trailing quote from the `nerdctl run -i, --interactive` command reference entry

## Testing
- git diff --check


AkihiroSuda:
Needs DCO signing

MukundaKatta:
Pushed 22ebf23 to fix the DCO failure.

Root cause: the single PR commit was missing the required Signed-off-by trailer. I amended the commit with signoff and force-pushed with lease. The docs diff itself is unchanged.

Validation run locally:
- git diff --check
- verified the amended commit contains Signed-off-by

AkihiroSuda (approved):

PR #4911, opened 2026-05-17
fix(healthcheck): release exec process resources after probe

haytok:
Each invocation of a health check runs the user-defined command inside the container via containerd's `task.Exec()` API.

We can verify this in the `probeHealthCheck` section of the healthcheck package as follows:

```golang
func probeHealthCheck(ctx context.Context, task containerd.Task, hc *Healthcheck, processSpec *specs.Process) (*HealthcheckResult, error) {

...

	process, err := task.Exec(ctx, execID, processSpec, cio.NewCreator(
		cio.WithStreams(nil, outputBuf, outputBuf),
	))

...

	if err := process.Start(ctx); err != nil {
		log.G(ctx).Debugf("failed to start health check: %v", err)
		return nil, fmt.Errorf("start error: %w", err)
	}

	exitStatusC, err := process.Wait(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to wait for health check: %w", err)
	}
```

However, the current implementation does not release the resources allocated during the health check once the check is complete.

As a result, for example, pipes that should normally be deleted are not removed and continue to accumulate over time.

We can verify this behavior using the following command:

```bash
$ sudo nerdctl run -d --name=health --health-cmd="curl -f http://localhost" --health-interval=1s --health-timeout=1m0s --health-retries=3 --health-start-period=2s nginx:alpine
f2fdd8346a546bc6ef446980efdce1132a436fa63bd64a8ce6756c6197e7ec3f

$ TASK_PID=$(sudo nerdctl inspect health --format '{{.State.Pid}}')

$ SHIM_PID=$(ps -o ppid= -p $TASK_PID | tr -d ' ')

$ sudo ls -la /proc/$SHIM_PID/fd | grep pipe | head -3
lr-x------ 1 root root  64 May 17 20:51 100 -> pipe:[1093131]
lr-x------ 1 root root  64 May 17 20:52 101 -> pipe:[1092248]
lr-x------ 1 root root  64 May 17 20:52 102 -> pipe:[1092228]

$ sudo ls -la /proc/$SHIM_PID/fd | grep pipe | wc -l
16

$ sleep 10

$ sudo ls -la /proc/$SHIM_PID/fd | grep pipe | wc -l
26
```

So, this change adds a deferred `process.Delete()` after `task.Exec()` so that the allocated resources are released when the exec process completes.

### DIFF
--- a/pkg/healthcheck/executor.go
+++ b/pkg/healthcheck/executor.go
@@ -75,6 +75,11 @@ func probeHealthCheck(ctx context.Context, task containerd.Task, hc *Healthcheck
 		log.G(ctx).Debugf("failed to exec health check: %v", err)
 		return nil, fmt.Errorf("exec error: %w", err)
 	}
+	defer func() {
+		if _, err := process.Delete(ctx); err != nil {
+			log.G(ctx).WithError(err).Debug("failed to delete exec process")
+		}
+	}()
 
 	if err := process.Start(ctx); err != nil {
 		log.G(ctx).Debugf("failed to start health check: %v", err)
--- a/cmd/nerdctl/container/container_health_check_linux_test.go
+++ b/cmd/nerdctl/container/container_health_check_linux_test.go
@@ -20,6 +20,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


pkg/healthcheck/executor.go
```diff
@@ -75,6 +75,11 @@ func probeHealthCheck(ctx context.Context, task containerd.Task, hc *Healthcheck
 		log.G(ctx).Debugf("failed to exec health check: %v", err)
 		return nil, fmt.Errorf("exec error: %w", err)
 	}
+	defer func() {
+		if _, err := process.Delete(ctx); err != nil {
```
sathiraumesh, line 79: I think its better to do gracefull shut down of the process if succeed if `task.Kill(ctx, syscall.SIGTERM) ` or change a logic a bit because i see in timeout there is a non graceful shutdown logic in the process as weel

sathiraumesh (changes requested):

pkg/healthcheck/executor.go
haytok: Since the `exec` process has either been killed or has exited by the time the `defer` block is executed, a graceful shutdown is irrelevant here.

https://github.com/containerd/nerdctl/blob/ad00cb4c5bcdda6f0f1cc85e99e0d912c8d1809b/pkg/healthcheck/executor.go#L91

https://github.com/containerd/nerdctl/blob/ad00cb4c5bcdda6f0f1cc85e99e0d912c8d1809b/pkg/healthcheck/executor.go#L108

sathiraumesh: yes make sense in this context

sathiraumesh (approved):

AkihiroSuda:
Needs rebase

haytok:
Thanks for checking!! I've resolved this conflict.

ChengyuZhu6 (approved):
LGTM

PR #4913, opened 2026-05-18
ci: add timestamp for command execution

ningmingxiao:
Some test cases are time-sensitive, so timestamp information is added.

### DIFF
--- a/mod/tigron/test/command.go
+++ b/mod/tigron/test/command.go
@@ -180,7 +180,7 @@ func (gc *GenericCommand) Run(expect *Expected) {
 	}
 
 	debug = append(debug,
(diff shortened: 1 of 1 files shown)


mod/tigron/test/command.go
```diff
@@ -180,7 +180,7 @@ func (gc *GenericCommand) Run(expect *Expected) {
 	}
 
 	debug = append(debug,
-		[]any{"➡️", commandDecorator + " " + gc.cmd.Binary + " " + strings.Join(gc.cmd.Args, " ")},
+		[]any{"➡️", commandDecorator + " " + gc.cmd.Binary + " " + strings.Join(gc.cmd.Args, " ") + " " + time.Now().Format("2006-01-02 15:04:05.000")},
```
sathiraumesh, line 183: Not sure here it is needed but its always good idea to use utc time stamps instead of local time

sathiraumesh (approved):

PR #4916, opened 2026-05-18
test: refactor container_run_restart_linux_test.go to use Tigron

opjt:
- releated issue: #4613 

### 1. `killDaemon` / `ensureDaemonActive` / `dumpDaemonLogs` location

These helpers were previously methods on `testutil.Base` and are used only by `TestRunRestart` in this file.  
I've kept them inline (package-private functions) in the test file

if the project expects more daemon-resilience tests in the future, moving them to `pkg/testutil/nerdtest/` (e.g., daemon_linux.go) as exported helpers make sense. 

### 2. `nerdtest.ContainerdPlugin` plugin type

The migrated tests keep `io.containerd.internal.v1` to match original behavior. In newer containerd (verified on v2.2.1) this plugin has moved:  

```bash
$ sudo ctr plugins ls | grep restart
io.containerd.monitor.container.v1   restart   -   ok
```

As a result, the 4 tests skip on the v2.x CI jobs but still run on the legacy jobs.
Updating the type (or supporting both) is left as a follow-up.

### DIFF
--- a/cmd/nerdctl/container/container_run_restart_linux_test.go
+++ b/cmd/nerdctl/container/container_run_restart_linux_test.go
@@ -19,44 +19,118 @@ package container
 import (
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
 
 	"gotest.tools/v3/assert"
-	"gotest.tools/v3/poll"
 
 	"github.com/containerd/containerd/v2/core/runtime/restart"
+	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+	"github.com/containerd/nerdctl/mod/tigron/tig"
 
+	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/dockercompat"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nettestutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/portlock"
 )
 
-func TestRunRestart(t *testing.T) {
+func daemonSystemctlTarget() string {
+	if nerdtest.IsDocker() {
+		return "docker.service"
+	}
+	return "containerd.service"
+}
+
+func daemonSystemctlArgs() []string {
+	if os.Geteuid() != 0 {
+		return []string{"--user"}
+	}
+	return nil
+}
+
+func killDaemon(t tig.T) {
+	t.Helper()
+	target := daemonSystemctlTarget()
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/container/container_run_restart_linux_test.go
```diff
@@ -19,42 +19,117 @@ package container
...
 	)
-	testContainerName := testutil.Identifier(t)
+	for i := 0; i < maxRetry; i++ {
+		cmd := exec.Command("systemctl", append(daemonSystemctlArgs(), "is-active", target)...)
+		out, err := cmd.CombinedOutput()
+		t.Log(fmt.Sprintf("(retry=%d) %s", i, string(out)))
+		if err == nil {
+			// The daemon is now running, but the daemon may still refuse connections to containerd.sock
+			t.Log(fmt.Sprintf("daemon %q is now running, checking whether the daemon can handle requests", target))
+			infoOut, infoErr := exec.Command(testutil.GetTarget(), "info").CombinedOutput()
```
ChengyuZhu6, line 84: Why not use `helpers.Command`?

opjt: The loop needs to know whether `info` succeeded without failing the test on intermediate attempts.
As far as I could find, **Tigron** doesn't offer a way to do both:
 
 - `Run` returns nothing
 - `Ensure/Capture` fatal on non-zero exit, and `Anyhow` doesn't expose success.

`exec.Command` was the simplest fit. Open to suggestions if there's a pattern I missed.

pkg/testutil/nerdtest/requirements.go
```diff
@@ -196,6 +197,44 @@ var RootlessWithoutDetachNetNS = require.All(Rootless, require.Not(RootlessWithD
 // Rootful marks a test as suitable only for rootful env
 var Rootful = require.Not(Rootless)
 
+// ContainerdPlugin requires that the given containerd plugin (and capabilities) is available.
+var ContainerdPlugin = func(requiredType, requiredID string, requiredCaps []string) *test.Requirement {
+	return &test.Requirement{
+		Check: func(data test.Data, helpers test.Helpers) (bool, string) {
+			if IsDocker() {
+				return true, ""
```
AkihiroSuda: Why true?

opjt: Docker target doesn't use containerd plugins, so the check isn't applicable. 
Returning true lets the test still run on Docker, matching the original 
```
if !nerdtest.IsDocker() {
		testutil.RequireContainerdPlugin(base, "io.containerd.internal.v1", "restart", []string{"unless-stopped"})
	}
```
behavior.

AkihiroSuda: `ContainerdPlugin()` should not be called for Docker.
When accidentally called it should return an error.

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/4916/changes#r3332179674

AkihiroSuda (approved):
Thanks

PR #4932, opened 2026-05-23
chore: remove go.uber.org/mock dependency

### LINKED ISSUE #3325: Remove dependency on `go.uber.org/mock`

Metbcy:
Fixes #3325.

Replaces the generated gomock mocks with small hand-rolled fakes that use function fields, and drops `go.uber.org/mock` from `go.mod` / `go.sum`.

Changes:
- `pkg/infoutil/infoutilmock/infoutil_mock.go`: hand-rolled `FakeWindowsInfoUtil` with `RtlGetVersionFunc`, `GetRegistryStringValueFunc`, `GetRegistryIntValueFunc` fields.
- `pkg/infoutil/infoutil_windows_test.go`: switched to the fake.
- `pkg/cmd/builder/build_test.go`: replaced the inline gomock-based `MockParse` with a `fakePlatformParser` using function fields.
- `go.mod` / `go.sum`: drop `go.uber.org/mock v0.6.0`.

I went with in-line fakes per the issue text rather than relocating mockgen to a separate package (the alternative @fahedouch raised on the thread); happy to refactor toward that approach instead if preferred.

Verified locally:
- `go build ./...`
- `go vet ./...`
- `go test ./pkg/cmd/builder/...` passes.
- The Windows test file compiles cleanly under `GOOS=windows`.
- `grep -r go.uber.org/mock` returns no source references.


AkihiroSuda (approved):
Thanks

PR #4934, opened 2026-05-24
fix: handle long fractional Unix timestamps

immanuwell:
Tiny parser fix. Before this, a Unix timestamp with a long fractional part got validated before truncation, so it could overflow int64 and fail. Kinda annoying, but real: logs --since/--until feed user timestamp strings through this path.

Repro before the fix:
```
go test ./pkg/timestamp
```
with `1136073600.123456789123456789123`, it errors with `value out of range` instead of truncating to 9 ns digits.

Now the parser truncates first, then parses. Existing short fractions still work.

Checked:
```
go test ./pkg/timestamp
go test ./pkg/...
git diff --check origin/main..HEAD
```

### DIFF
--- a/pkg/timestamp/timestamp.go
+++ b/pkg/timestamp/timestamp.go
@@ -25,7 +25,6 @@ package timestamp
 
 import (
 	"fmt"
-	"math"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


sathiraumesh (approved):

AkihiroSuda (approved):
Thanks

PR #4938, opened 2026-05-27
formatter: make ellipsis unicode safe

immanuwell:
Fixes `formatter.Ellipsis` so it truncates by rune, not byte. Tiny one, but real: unicode text can show up in commands and registry descriptions.

Repro before this patch:

```console
$ go test ./pkg/formatter -run TestEllipsis -count=1
--- FAIL: TestEllipsis/unicode_truncated_to_single_rune
    formatter_test.go:238: assertion failed: é != Ã
```

Tested:

```console
go test ./pkg/formatter ./pkg/cmd/search -count=1
```

No related issue found with gh search, so no closes line here.

### DIFF
--- a/pkg/formatter/formatter_test.go
+++ b/pkg/formatter/formatter_test.go
@@ -19,6 +19,7 @@ package formatter
 import (
 	"testing"
 	"time"
+	"unicode/utf8"
 
 	"gotest.tools/v3/assert"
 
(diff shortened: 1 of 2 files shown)


sathiraumesh (approved):

pkg/formatter/formatter_test.go
```diff
@@ -192,3 +193,49 @@ func TestFormatPorts(t *testing.T) {
...
+			expected:        "é",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			result := Ellipsis(tt.input, tt.maxDisplayWidth)
+			assert.Assert(t, utf8.ValidString(result), "expected valid UTF-8, got %q", result)
```
sathiraumesh: I don't think this line is needed since we already check the expected and the result which is already specified

sathiraumesh (changes requested):

pkg/formatter/formatter_test.go
immanuwell: @sathiraumesh done in the latest commit here

AkihiroSuda (approved):
Thanks

PR #4945, opened 2026-05-29
fix(container): chunk mounts metadata to prevent max label size crash

### LINKED ISSUE #4914: nerdctl/mounts label too long
### Description

`sudo ~/Downloads/nerdctl-2.2.0-linux-amd64/nerdctl run --namespace moby --privileged -it --rm --entrypoint bash IMAGE` shows this

```
FATA[0000] create container failed validation: containers.Labels: label key and value length (5005 bytes) greater than maximum size (4096 bytes), key: nerdctl/mounts: invalid argument
```

1.
2.
3.

The container should be started

2.2.2/1.6.0/0.19.0

None

### Host information

Debian 13

```
$ cat /proc/cmdline 
ro SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1 systemd.unified_cgroup_hierarchy=0 split_lock_detect=off kvm.enable_virt_at_load=0 quiet splash
```

Skywalkr-dev:
When a container has too many volume mounts, the marshaled JSON string exceeds the 4096-byte protocol buffer limit enforced by containerd, causing creation to fail. 

This patch fixes the issue by breaking the mounts metadata string into smaller 4000-byte pieces (using 'nerdctl/mounts/chunk-%d' labels). The container volume extraction path has been updated to stitch these chunks back together dynamically before parsing. 

Unit tests have been added to verify that chunked reassembly works exactly like the legacy payload behavior.

FIXES : #4914

### DIFF
--- a/pkg/cmd/container/create.go
+++ b/pkg/cmd/container/create.go
@@ -825,11 +825,13 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
 
 	if len(internalLabels.mountPoints) > 0 {
 		mounts := dockercompatMounts(internalLabels.mountPoints)
-		mountPointsJSON, err := json.Marshal(mounts)
+		jsonMountBytes, err := json.Marshal(mounts)
 		if err != nil {
+			return nil, fmt.Errorf("failed to marshal mounts: %w", err)
+		}
+		if err := labels.SetMount(m, jsonMountBytes); err != nil {
 			return nil, err
 		}
-		m[labels.Mounts] = string(mountPointsJSON)
 	}
 
 	if internalLabels.macAddress != "" {
--- a/pkg/cmd/container/run_mount.go
+++ b/pkg/cmd/container/run_mount.go
@@ -335,8 +335,10 @@ func generateMountOpts(ctx context.Context, client *containerd.Client, ensuredIm
 					return nil, nil, nil, err
 				}
 			}
-			if m, found := ls[labels.Mounts]; found {
-				err = json.Unmarshal([]byte(m), &vfMountPoints)
+
+			nerdctlMounts := labels.GetMount(ls)
+			if nerdctlMounts != "" {
+				err = json.Unmarshal([]byte(nerdctlMounts), &vfMountPoints)
 				if err != nil {
 					return nil, nil, nil, err
 				}
--- a/pkg/cmd/volume/rm.go
+++ b/pkg/cmd/volume/rm.go
@@ -92,8 +92,9 @@ func usedVolumes(ctx context.Context, containers []containerd.Container) (map[st
 			}
 			return nil, err
 		}
-		mountsJSON, ok := l[labels.Mounts]
-		if !ok {
+
+		mountsJSON := labels.GetMount(l)
+		if mountsJSON == "" {
 			continue
 		}
 
--- a/pkg/containerutil/containerutil.go
+++ b/pkg/containerutil/containerutil.go
@@ -584,8 +584,15 @@ func GetContainerVolumes(containerLabels map[string]string) []*ContainerVolume {
 	var vols []*ContainerVolume
 	volLabels := []string{labels.AnonymousVolumes, labels.Mounts}
 	for _, volLabel := range volLabels {
-		names, ok := containerLabels[volLabel]
-		if !ok {
+		var names string
+
+		if volLabel == labels.Mounts {
+			names = labels.GetMount(containerLabels)
+		} else {
+			names = containerLabels[volLabel]
+		}
+
+		if names == "" {
 			continue
 		}
 		var (
--- a/pkg/containerutil/containerutil_test.go
+++ b/pkg/containerutil/containerutil_test.go
@@ -19,6 +19,8 @@ package containerutil
 import (
 	"reflect"
 	"testing"
+
+	"github.com/containerd/nerdctl/v2/pkg/labels"
 )
 
 func TestParseExtraHosts(t *testing.T) {
@@ -79,5 +81,42 @@ func TestParseExtraHosts(t *testing.T) {
 				t.Fatalf("expected %v, actual %v", test.expected, extraHosts)
 			}
 		})
+	}
+}
+
+func TestGetContainerVolumes_Indexed(t *testing.T) {
+	m0 := `{"Type":"volume","Name":"vol-0","Source":"/var/lib/vol-0","Destination":"/mnt/vol-0"}`
+	m1 := `{"Type":"volume","Name":"vol-1","Source":"/var/lib/vol-1","Destination":"/mnt/vol-1"}`
+	m2 := `{"Type":"volume","Name":"vol-2","Source":"/var/lib/vol-2","Destination":"/mnt/vol-2"}`
+
+	rawJSON := "[" + m0 + "," + m1 + "," + m2 + "]"
+
+	indexedLabels := map[string]string{
+		"nerdctl/mounts.0": m0,
+		"nerdctl/mounts.1": m1,
+		"nerdctl/mounts.2": m2,
+	}
+
+	legacyLabels := map[string]string{
+		labels.Mounts: rawJSON,
+	}
+
+	indexedResult := GetContainerVolumes(indexedLabels)
+	legacyResult := GetContainerVolumes(legacyLabels)
+
+	if len(indexedResult) == 0 {
+		t.Fatal("Expected to extract volumes from indexed labels, but got 0 results")
+	}
+
+	if len(indexedResult) != len(legacyResult) {
+		t.Errorf("Mismatched output! Indexed found %d volumes, Legacy found %d volumes.",
+			len(indexedResult), len(legacyResult))
+	}
+
+	if indexedResult[0].Name != "vol-0" {
+		t.Errorf("Expected first volume to be named 'vol-0', got '%s'", indexedResult[0].Name)
+	}
+	if len(indexedResult) > 2 && indexedResult[2].Name != "vol-2" {
+		t.Errorf("Expected third volume to be named 'vol-2', got '%s'", indexedResult[2].Name)
 	}
 }
--- a/pkg/inspecttypes/dockercompat/dockercompat.go
+++ b/pkg/inspecttypes/dockercompat/dockercompat.go
@@ -400,7 +400,7 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
 	}
 
 	c.HostConfig.Tmpfs = make(map[string]string)
-	if nerdctlMounts := n.Labels[labels.Mounts]; nerdctlMounts != "" {
+	if nerdctlMounts := labels.GetMount(n.Labels); nerdctlMounts != "" {
 		mounts, err := parseMounts(nerdctlMounts)
 		if err != nil {
 			return nil, err
--- a/pkg/labels/mount.go
+++ b/pkg/labels/mount.go
@@ -1,0 +1,70 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package labels
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// SetMount parses a JSON array of mounts and stores each individual mount object as an indexed label.
+func SetMount(m map[string]string, jsonMountBytes []byte) error {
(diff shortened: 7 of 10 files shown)


pkg/cmd/volume/rm.go
```diff
@@ -92,8 +92,24 @@ func usedVolumes(ctx context.Context, containers []containerd.Container) (map[st
 			}
 			return nil, err
 		}
-		mountsJSON, ok := l[labels.Mounts]
-		if !ok {
+
+		var mountsJSON string
+		if legacyMount, ok := l[labels.Mounts]; ok && legacyMount != "" {
```
sathiraumesh: I believe this logic can be refactored into `pkg/labels/mount.go  ` as hepler function like `labels.SetMount(...) ` and l`abels.GetMount()` (this logic is duplicated in 4 other files)

pkg/cmd/volume/rm.go
```diff
@@ -92,8 +92,24 @@ func usedVolumes(ctx context.Context, containers []containerd.Container) (map[st
...
+			mountsJSON = legacyMount
+		} else {
+			counter := 0
+			for {
+				key := fmt.Sprintf("nerdctl/mounts/chunk-%d", counter)
```
sathiraumesh: these can be refactored too instead of hard coding something like 
```

const (
    maxLabelValueSize = 4096 (what you did aprady)
    chunkKeyFormat    = Prefix + "mounts/chunk-%d" ... 

```

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -400,11 +400,29 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
...
+				counter++
+			} else {
+				break
+			}
+		}
+	}
+
+	if nerdctlMounts != "" {
 		mounts, err := parseMounts(nerdctlMounts)
+
```
sathiraumesh: remove these non needed changes

sathiraumesh (changes requested):

Skywalkr-dev:
@sathiraumesh is it fine now ?

pkg/containerutil/containerutil_test.go
```diff
@@ -81,3 +83,41 @@ func TestParseExtraHosts(t *testing.T) {
...
+	part1 := `{"Type":"volume","Name":"vol-1","Source":"/var/lib/vol-1","Destination":"/mnt/vol-1"},`
+	part2 := `{"Type":"volume","Name":"vol-2","Source":"/var/lib/vol-2","Destination":"/mnt/vol-2"}]`
+
+	rawJSON := part0 + part1 + part2
+
+	chunkedLabels := map[string]string{
+		"nerdctl/mounts/chunk-0": part0,
+		"nerdctl/mounts/chunk-1": part1,
+		"nerdctl/mounts/chunk-2": part2,
+	}
```
AkihiroSuda, line 98: Do we really need to create chunks?
Why not just have a single label for a single `Mount`?
```
"nerdctl/mounts.0": `{"Type":"volume","Name":"vol-0","Source":"/var/lib/vol-0","Destination":"/mnt/vol-0"}`,
"nerdctl/mounts.1": `{"Type":"volume","Name":"vol-1","Source":"/var/lib/vol-1","Destination":"/mnt/vol-1"}`,
"nerdctl/mounts.2": `{"Type":"volume","Name":"vol-2","Source":"/var/lib/vol-2","Destination":"/mnt/vol-2"}`,
```

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/4945/changes#r3328980164

pkg/containerutil/containerutil_test.go
Skywalkr-dev: > Do we really need to create chunks? Why not just have a single label for a single `Mount`?
> 
> ```
> "nerdctl/mounts.0": `{"Type":"volume","Name":"vol-0","Source":"/var/lib/vol-0","Destination":"/mnt/vol-0"}`,
> "nerdctl/mounts.1": `{"Type":"volume","Name":"vol-1","Source":"/var/lib/vol-1","Destination":"/mnt/vol-1"}`,
> "nerdctl/mounts.2": `{"Type":"volume","Name":"vol-2","Source":"/var/lib/vol-2","Destination":"/mnt/vol-2"}`,
> ```

@AkihiroSuda fair enough, I considered the buffer cap too far. Now all mounts will have their individual labels rather than having chunks split by the buffer size.

pkg/cmd/container/create.go
```diff
@@ -824,12 +824,15 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
...
-		mountPointsJSON, err := json.Marshal(mounts)
-		if err != nil {
-			return nil, err
+		if len(internalLabels.mountPoints) > 0 {
+			mounts := dockercompatMounts(internalLabels.mountPoints)
+			mountPointsJSON, err := json.Marshal(mounts)
+			if err != nil {
+				return nil, err
+			}
+			// helper from labels.go
```
sathiraumesh: no need this comment its straight forward what you do

pkg/cmd/container/create.go
```diff
@@ -824,12 +824,15 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
 	}
 
 	if len(internalLabels.mountPoints) > 0 {
-		mounts := dockercompatMounts(internalLabels.mountPoints)
-		mountPointsJSON, err := json.Marshal(mounts)
-		if err != nil {
-			return nil, err
+		if len(internalLabels.mountPoints) > 0 {
+			mounts := dockercompatMounts(internalLabels.mountPoints)
+			mountPointsJSON, err := json.Marshal(mounts)
```
sathiraumesh: much cleaner if you can move this to the setMounts(..., mounts) and handle things there and return error

pkg/cmd/container/create.go
```diff
@@ -824,12 +824,15 @@ func withInternalLabels(internalLabels internalLabels) (containerd.NewContainerO
 	}
 
 	if len(internalLabels.mountPoints) > 0 {
-		mounts := dockercompatMounts(internalLabels.mountPoints)
-		mountPointsJSON, err := json.Marshal(mounts)
-		if err != nil {
-			return nil, err
+		if len(internalLabels.mountPoints) > 0 {
```
sathiraumesh: why do duplicate ifs i think you should refactor

pkg/cmd/container/run_mount.go
```diff
@@ -335,8 +335,10 @@ func generateMountOpts(ctx context.Context, client *containerd.Client, ensuredIm
 					return nil, nil, nil, err
 				}
 			}
-			if m, found := ls[labels.Mounts]; found {
-				err = json.Unmarshal([]byte(m), &vfMountPoints)
+
+			nerdctlMounts, _ := labels.GetMount(ls)
```
sathiraumesh: we don't even need the boolean return since it is not doing anything at all only return sting since you only check aganist it

pkg/cmd/volume/rm.go
```diff
@@ -92,8 +92,9 @@ func usedVolumes(ctx context.Context, containers []containerd.Container) (map[st
 			}
 			return nil, err
 		}
-		mountsJSON, ok := l[labels.Mounts]
-		if !ok {
+
+		mountsJSON, _ := labels.GetMount(l)
+		if mountsJSON == "" { //no legacy or chunk labels found
```
sathiraumesh: no need the comment here as well

pkg/containerutil/containerutil.go
```diff
@@ -584,7 +584,15 @@ func GetContainerVolumes(containerLabels map[string]string) []*ContainerVolume {
 	var vols []*ContainerVolume
 	volLabels := []string{labels.AnonymousVolumes, labels.Mounts}
 	for _, volLabel := range volLabels {
-		names, ok := containerLabels[volLabel]
+		var names string
+		var ok bool
+
+		if volLabel == labels.Mounts {
+			names, ok = labels.GetMount(containerLabels) // helper from labels.go
```
sathiraumesh: can we remove those unwanted comments

sathiraumesh (changes requested):

Skywalkr-dev:
@AkihiroSuda @sathiraumesh does the current changes look fine ?

pkg/labels/mount.go
```diff
@@ -0,0 +1,77 @@
...
+package labels
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// SetMount centralizes the logic for marshaling and chunking mount metadata into a label map.
+func SetMount(m map[string]string, mounts any) error {
```
sathiraumesh: please make sure the the type of mounts is a given concrete type this can be buggy to have the interface type here to accept any thing

sathiraumesh (changes requested):

pkg/labels/mount.go
Skywalkr-dev: @sathiraumesh  yup noted, tho directly passing MountPoint as the type would cause import cycle error. 
Cuz of this, i marshaled the MountPoints as a JSON string to avoid errors during build. 
Does the current changes work

Skywalkr-dev:
@AkihiroSuda @sathiraumesh does the current change work ????

pkg/labels/mount.go
```diff
@@ -0,0 +1,69 @@
...
+
+// GetMount extracts and reassembles indexed mount metadata back into a single global JSON array string
+func GetMount(containerLabels map[string]string) string {
+	// Try legacy label first for backward compatibility
+	if legacyMount, ok := containerLabels[Mounts]; ok && legacyMount != "" {
+		return legacyMount
+	}
+
+	var rawMounts []string
+	for i := 0; ; i++ {
```
sathiraumesh: formatting and probably good idea to put a limit here or some sort of condition to exit

pkg/labels/mount.go
```diff
@@ -0,0 +1,69 @@
...
+		key := fmt.Sprintf(MountsKeyFormat, i)
+		if chunk, found := containerLabels[key]; found && chunk != "" {
```
sathiraumesh: can a chunk be ` ""`

pkg/labels/mount.go
```diff
@@ -0,0 +1,69 @@
...
+
+package labels
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// SetMount centralizes the logic for marshaling and chunking mount metadata into a label map.
```
sathiraumesh: formatting

sathiraumesh (changes requested):

pkg/labels/mount.go
Skywalkr-dev: It is more or less like a fail safe where certain downstream API tools set a mount's value to ` "" ` instead of entirely deleting the key itself. 
It'll lead to the entire JSON string being improperly formatted.

pkg/labels/mount.go
Skywalkr-dev: yup noted, ill add length of `containerLabels` as a boundary then. 
thanks

pkg/labels/mount.go
Skywalkr-dev: @sathiraumesh done, thanks

Skywalkr-dev:
@sathiraumesh are the current changes fine ?

pkg/inspecttypes/dockercompat/dockercompat.go
AkihiroSuda: ping @Skywalkr-dev

AkihiroSuda (changes requested):
https://github.com/containerd/nerdctl/pull/4945/changes#r3369720875

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -400,11 +400,13 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
...
 
 	c.HostConfig.Tmpfs = make(map[string]string)
-	if nerdctlMounts := n.Labels[labels.Mounts]; nerdctlMounts != "" {
+	nerdctlMounts := labels.GetMount(n.Labels)
+	if nerdctlMounts != "" {
 		mounts, err := parseMounts(nerdctlMounts)
 		if err != nil {
 			return nil, err
 		}
+
```
AkihiroSuda: ?

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -400,11 +400,13 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
 	}
 
 	c.HostConfig.Tmpfs = make(map[string]string)
-	if nerdctlMounts := n.Labels[labels.Mounts]; nerdctlMounts != "" {
+	nerdctlMounts := labels.GetMount(n.Labels)
+	if nerdctlMounts != "" {
```
AkihiroSuda:
```suggestion
	if nerdctlMounts := ...; nerdctlMounts != "" {
```

pkg/inspecttypes/dockercompat/dockercompat.go
Skywalkr-dev: sorry about that, unformatted buffer from nvim sneaked its way in.

pkg/inspecttypes/dockercompat/dockercompat.go
Skywalkr-dev: @AkihiroSuda does it seem fine now ?

AkihiroSuda (approved):
Thanks

PR #4953, opened 2026-06-08
fix(network): allow inspecting pseudo networks

immanuwell:
refs #3005

`nerdctl network ls` shows `host` and `none`, but `nerdctl network inspect host` and `nerdctl network inspect none` still fail. the inspect path asks to allow pseudo networks, but lookup only checked on-disk CNI configs, so these names fell through. kinda awkward.

this returns a small synthetic config for pseudo networks in that lookup path, so inspect can render them fine.

repro:
1. run `nerdctl network ls`
2. see `host` and `none`
3. run `nerdctl network inspect host`
4. before: fails with `no network found matching: host`
5. after: returns JSON with `"Name": "host"`

checks:
- `go test ./pkg/netutil -count=1`
- `go test ./pkg/inspecttypes/dockercompat -count=1`
- `go test ./cmd/nerdctl/network -run ^ -count=1`

### DIFF
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -60,8 +60,17 @@ func (e *CNIEnv) ListNetworksMatch(reqs []string, allowPseudoNetwork bool) (list
 
 	list = make(map[string][]*NetworkConfig)
 	for _, req := range reqs {
-		if !allowPseudoNetwork && (req == "host" || req == "none") {
-			errs = append(errs, fmt.Errorf("pseudo network not allowed: %s", req))
+		if req == "host" || req == "none" {
+			if !allowPseudoNetwork {
+				errs = append(errs, fmt.Errorf("pseudo network not allowed: %s", req))
+				continue
+			}
+			cfg, err := newPseudoNetworkConfig(req)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			list[req] = []*NetworkConfig{cfg}
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/netutil/netutil.go
```diff
@@ -60,10 +60,19 @@ func (e *CNIEnv) ListNetworksMatch(reqs []string, allowPseudoNetwork bool) (list
 
 	list = make(map[string][]*NetworkConfig)
 	for _, req := range reqs {
-		if !allowPseudoNetwork && (req == "host" || req == "none") {
+		if !allowPseudoNetwork && isPseudoNetwork(req) {
```
sathiraumesh: probably can do in a cleaner way i beleive and remove extra function declration and two calls 

```
if name == "host" || name == "none" {
  if !allowPseudoNetwork {
        errs = append(errs, fmt.Errorf("pseudo network not allowed: %s", req))
        continue
    }
    cfg, err := newPseudoNetworkConfig(req)
    if err != nil {
        errs = append(errs, err)
        continue
    }
    list[req] = []*NetworkConfig{cfg}
    continue
}
```

sathiraumesh (changes requested):

pkg/netutil/netutil.go
immanuwell: @sathiraumesh got it, done

AkihiroSuda:
```
 FAIL - does not have a valid DCO
```
https://github.com/containerd/nerdctl/actions/runs/27420363683/job/81044018040?pr=4953

pkg/netutil/netutil.go
```diff
@@ -88,6 +97,38 @@ func (e *CNIEnv) ListNetworksMatch(reqs []string, allowPseudoNetwork bool) (list
 	return list, errs
 }
 
+func newPseudoNetworkConfig(name string) (*NetworkConfig, error) {
+	confJSON, err := json.Marshal(struct {
```
AkihiroSuda: Isn't this struct already defined in somewhere else?

immanuwell: good point
I switched newPseudoNetworkConfig to reuse the existing cniNetworkConfig, and made nerdctlID / nerdctlLabels omitempty so the synthetic pseudo-network config stays minimal
PTAL

pkg/netutil/netutil.go
```diff
@@ -60,8 +60,17 @@ func (e *CNIEnv) ListNetworksMatch(reqs []string, allowPseudoNetwork bool) (list
...
+				continue
+			}
+			cfg, err := newPseudoNetworkConfig(req)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			list[req] = []*NetworkConfig{cfg}
 			continue
 		}
```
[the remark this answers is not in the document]
AkihiroSuda: > which can shadow a real on-disk CNI network config named "host" or "none".

We do not support such config.

AkihiroSuda:
`test / in-host / docker` failed 7 times
https://github.com/containerd/nerdctl/actions/runs/27458993998/job/81517770971?pr=4953
```
2026-06-15T20:35:11.1401312Z === Failing tests ===
2026-06-15T20:35:11.1401617Z TestRunSeccompCapSysPtrace
2026-06-15T20:35:11.1401881Z TestUpdateRestartPolicy
2026-06-15T20:35:11.1402103Z =====================
```

AkihiroSuda (approved):
Thanks

PR #4954, opened 2026-06-08
fix: reject `/` as the `-v` destination in nerdctl run for Docker com…

haytok:
…patibility

In the current implementation, when specifying `/` as the destination of the `-v` option in the nerdctl run command, the following error occurs but the container is created.

```bash
> sudo nerdctl run -d --name nginx -v ./:/ nginx
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: unable to apply apparmor profile: apparmor failed to apply profile: open /proc/thread-self/attr/exec: no such file or directory

> sudo nerdctl ps -a
CONTAINER ID    IMAGE                             COMMAND                   CREATED          STATUS     PORTS    NAMES
0f880e0e68f8    docker.io/library/nginx:latest    "/docker-entrypoint.…"    5 seconds ago    Created             nginx
```

However, in the same situation, Docker fails to create the container.

```bash
$ docker run -d --name stop -v ./:/ nginx
docker: Error response from daemon: invalid volume specification: '/Users/haytok/workspace:/': invalid mount config for type "bind": invalid specification: destination can't be '/'

$ docker ps -a
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```

Therefore, this commit fixes the behavior so that the container creation fails when `/` is specified as the destination of the `-v` option for compatibility with Docker.

### DIFF
--- a/cmd/nerdctl/container/container_run_mount_linux_test.go
+++ b/cmd/nerdctl/container/container_run_mount_linux_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/containerd/nerdctl/mod/tigron/tig"
 
 	"github.com/containerd/nerdctl/v2/cmd/nerdctl/helpers"
+	"github.com/containerd/nerdctl/v2/pkg/mountutil"
 	"github.com/containerd/nerdctl/v2/pkg/rootlessutil"
 	"github.com/containerd/nerdctl/v2/pkg/testutil"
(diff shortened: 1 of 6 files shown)


sathiraumesh (approved):

AkihiroSuda (approved):
Thanks

PR #4972, opened 2026-06-18
fix: remove duplicated defaults from help output

immanuwell:
A few help entries print the default value twice right now, which looks a bit goofy.

Before:
`go run ./cmd/nerdctl run --help | rg 'memory-swappiness|sig-proxy|systemd'`
`go run ./cmd/nerdctl image history --help | rg 'human|default'`

You can see lines like:
`--memory-swappiness ... (default -1) (default -1)`
`--sig-proxy ... (default true) (default true)`
`--systemd ... (default: false) (default "false")`
`--human ... (default true) (default true)`

This just drops the hardcoded default text where Cobra already prints it for us, and adds small tests so it does not drift back.

Checks:
`go test ./cmd/nerdctl/container -run TestRunHelpDoesNotDuplicateDefaults -count=1`
`go test ./cmd/nerdctl/image -run TestHistoryHelpDoesNotDuplicateDefaults -count=1`

### DIFF
--- a/cmd/nerdctl/container/container_run.go
+++ b/cmd/nerdctl/container/container_run.go
@@ -89,7 +89,7 @@ func setCreateFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool("help", false, "show help")
 
 	cmd.Flags().BoolP("tty", "t", false, "Allocate a pseudo-TTY")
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):
Thanks

PR #4986, opened 2026-06-19
fix: honor --workdir in compose run

immanuwell:
Fix `nerdctl compose run --workdir` so it actually sets the container cwd.

Right now `compose run --workdir /tmp` writes `/tmp` into the project working dir, not the service working dir. So `pwd` still does not run from `/tmp`, and relative host paths can get a bit weird too.

Repro
1. Create a compose file with:
```yaml
services:
  app:
    image: alpine
    entrypoint:
      - pwd
```
2. Run `nerdctl compose run --workdir /tmp app`
3. Before this patch, it does not print `/tmp`

This patch writes `ro.WorkDir` to `targetSvc.WorkingDir`, which is what `serviceparser` turns into `-w=...`. Added a regression test too, pretty small and to the point.

### DIFF
--- a/cmd/nerdctl/compose/compose_run_linux_test.go
+++ b/cmd/nerdctl/compose/compose_run_linux_test.go
@@ -326,6 +326,51 @@ services:
 	testCase.Run(t)
 }
 
+func TestComposeRunWithWorkdir(t *testing.T) {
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #4987, opened 2026-06-19
fix: clarify healthcheck help defaults

immanuwell:
small help text bug.

`nerdctl run --help` says `--health-interval`, `--health-timeout`, and `--health-retries` default to `30s` or `3`.
That is a bit off, tbh. `0` there means not set, then nerdctl uses the image value, or falls back later.

repro:
1. `GOOS=linux make binaries`
2. `./_output/nerdctl run --help | rg "health-(interval|timeout|retries)"`

before this patch, the help text claims fixed defaults.
after this patch, it says what `0` really means. small thing, but less confusing and more honest.

### DIFF
--- a/cmd/nerdctl/container/container_run_help_test.go
+++ b/cmd/nerdctl/container/container_run_help_test.go
@@ -41,4 +41,7 @@ func TestRunHelpDoesNotDuplicateDefaults(t *testing.T) {
 	assert.Assert(t, !strings.Contains(help, "Tune container memory swappiness (0 to 100) (default -1) (default -1)"))
 	assert.Assert(t, strings.Contains(help, "Allow running systemd in this container (default \"false\")"))
 	assert.Assert(t, !strings.Contains(help, "Allow running systemd in this container (default: false) (default \"false\")"))
+	assert.Assert(t, strings.Contains(help, "Time between running the check; 0 uses the image value or 30s when unset there too"))
+	assert.Assert(t, strings.Contains(help, "Maximum time to allow one check to run; 0 uses the image value or 30s when unset there too"))
+	assert.Assert(t, strings.Contains(help, "Consecutive failures needed to report unhealthy; 0 uses the image value or 3 when unset there too"))
 }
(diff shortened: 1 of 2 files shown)


cmd/nerdctl/container/container_run_help_test.go
```diff
@@ -41,4 +41,10 @@ func TestRunHelpDoesNotDuplicateDefaults(t *testing.T) {
 	assert.Assert(t, !strings.Contains(help, "Tune container memory swappiness (0 to 100) (default -1) (default -1)"))
 	assert.Assert(t, strings.Contains(help, "Allow running systemd in this container (default \"false\")"))
 	assert.Assert(t, !strings.Contains(help, "Allow running systemd in this container (default: false) (default \"false\")"))
+	assert.Assert(t, strings.Contains(help, "Time between running the check; 0 uses the image value or 30s when unset there too"))
+	assert.Assert(t, strings.Contains(help, "Maximum time to allow one check to run; 0 uses the image value or 30s when unset there too"))
+	assert.Assert(t, strings.Contains(help, "Consecutive failures needed to report unhealthy; 0 uses the image value or 3 when unset there too"))
+	assert.Assert(t, !strings.Contains(help, "Time between running the check (default: 30s)"))
+	assert.Assert(t, !strings.Contains(help, "Maximum time to allow one check to run (default: 30s)"))
+	assert.Assert(t, !strings.Contains(help, "Consecutive failures needed to report unhealthy (default: 3)"))
```
haytok: IMO these assertions aren't necessary. The positive checks above already cover the wording change.

immanuwell:
@haytok done, removed the redundant negative assertions and kept the positive checks only
PTAL

cmd/nerdctl/container/container_run_help_test.go
immanuwell: thanks, removed these 3 assertions

haytok (approved):
LGTM, Thanks!

ChengyuZhu6 (approved):
LGTM

PR #4990, opened 2026-06-22
feat(mount): support --mount type=image

mayur-tolexo:
Implements `--mount type=image` (read-only), part of #3867.

Mounts an image's filesystem into a container, matching Docker (moby#48798):
`nerdctl run --mount type=image,source=<image>,destination=<path> ...`

## What it does
- New mount type `image` parsed/validated in `ProcessFlagMount` (source + destination required; read-only; `subpath` and read-write are rejected with clear errors).
- `setupImageMount` ensures + unpacks the source image and creates a read-only, GC-rooted snapshot view of its rootfs, mounted at the destination. The same image can be mounted at multiple destinations (moby#50268).
- The view's snapshot key is stored in a `nerdctl/image-mount-snapshots` label and removed when the container is deleted (using the container's own snapshotter), mirroring the anonymous-volume cleanup pattern. Views created during a failed `create` are also cleaned up, so snapshots don't leak.

## Out of scope
- Read-write image mounts and `subpath` (rejected with clear errors; left as follow-ups).
- Linux only (the type is only recognized in the Linux mount parser).

## Tests
`cmd/nerdctl/container/container_run_mount_image_linux_test.go`:
- image filesystem readable at the target
- mount is read-only
- same image to two destinations
- errors: missing source, unsupported subpath

Plus a parser unit test in `pkg/mountutil/mountutil_linux_test.go`. Docs updated in `command-reference.md`.

Verified in a Linux + containerd sandbox: gofmt, go vet, build, unit + integration tests pass; the view is created, survives while the container runs, and is removed on deletion; a failed create leaves no leaked snapshots.

### DIFF
--- a/cmd/nerdctl/container/container_run_mount_image_linux_test.go
+++ b/cmd/nerdctl/container/container_run_mount_image_linux_test.go
@@ -1,0 +1,137 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package container
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/containerd/nerdctl/mod/tigron/expect"
+	"github.com/containerd/nerdctl/mod/tigron/require"
+	"github.com/containerd/nerdctl/mod/tigron/test"
+
+	"github.com/containerd/nerdctl/v2/pkg/testutil"
+	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
+)
+
+// TestRunMountTypeImage verifies that `--mount type=image` mounts the source
+// image's filesystem into the container so its files are readable at the target.
+func TestRunMountTypeImage(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		return helpers.Command("run", "--rm",
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img", testutil.CommonImage),
+			testutil.CommonImage, "cat", "/mnt/img/etc/os-release")
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: expect.ExitCodeSuccess,
+			Output:   expect.Contains("Alpine"),
+		}
+	}
+
+	testCase.Run(t)
+}
+
+// TestRunMountTypeImageMultipleDestinations verifies the same image can be
+// mounted at two destinations in one container.
+func TestRunMountTypeImageMultipleDestinations(t *testing.T) {
+	testCase := nerdtest.Setup()
+	// nerdctl-only: Docker keys an image mount by its source image and rejects
+	// mounting the same image twice ("mount already exists with name").
+	testCase.Require = require.Not(nerdtest.Docker)
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		return helpers.Command("run", "--rm",
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/a", testutil.CommonImage),
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/b", testutil.CommonImage),
+			testutil.CommonImage, "cat", "/mnt/a/etc/os-release", "/mnt/b/etc/os-release")
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: expect.ExitCodeSuccess,
+			Output:   expect.Contains("Alpine"),
+		}
+	}
+
+	testCase.Run(t)
+}
+
+// TestRunMountTypeImageReadOnly verifies an image mount is read-only so writing
+// fails. This matches Docker, which also mounts images read-only.
+func TestRunMountTypeImageReadOnly(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		return helpers.Command("run", "--rm",
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img", testutil.CommonImage),
+			testutil.CommonImage, "touch", "/mnt/img/should-fail")
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: expect.ExitCodeGenericFail,
+			Errors:   []error{fmt.Errorf("Read-only file system")},
+		}
+	}
+
+	testCase.Run(t)
+}
+
+// TestRunMountTypeImageErrors verifies that an image mount missing its source,
+// or using the not-yet-supported image-subpath option, is rejected. Docker
+// implements image-subpath, so that case diverges and the test is not run
+// against Docker.
+func TestRunMountTypeImageErrors(t *testing.T) {
+	testCase := nerdtest.Setup()
+	testCase.Require = require.Not(nerdtest.Docker)
+
+	testCase.SubTests = []*test.Case{
+		{
+			Description: "missing source",
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("run", "--rm", "--mount", "type=image,destination=/mnt/img",
+					testutil.CommonImage, "true")
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: expect.ExitCodeGenericFail,
+					Errors:   []error{fmt.Errorf("source")},
+				}
+			},
+		},
+		{
+			Description: "image-subpath not supported",
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("run", "--rm",
+					"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img,image-subpath=etc", testutil.CommonImage),
+					testutil.CommonImage, "true")
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: expect.ExitCodeGenericFail,
+					Errors:   []error{fmt.Errorf("image-subpath")},
+				}
+			},
+		},
+	}
+
+	testCase.Run(t)
+}
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -298,10 +298,9 @@ Volume flags:
   Consists of multiple key-value pairs, separated by commas and each
   consisting of a `<key>=<value>` tuple.
   e.g., `-- mount type=bind,source=/src,target=/app,bind-propagation=shared`.
-  - :whale: `type`: Current supported mount types are `bind`, `volume`, `tmpfs`.
+  - :whale: `type`: Current supported mount types are `bind`, `volume`, `tmpfs`, `image`.
     The default type will be set to `volume` if not specified.
     i.e., `--mount src=vol-1,dst=/app,readonly` equals `--mount type=volume,src=vol-1,dst=/app,readonly`
-    - unimplemented type: `image`
   - Common Options:
     - :whale: `src`, `source`: Mount source spec for bind and volume. Mandatory for bind.
     - :whale: `dst`, `destination`, `target`: Mount destination spec.
@@ -326,6 +325,10 @@ Volume flags:
       Defaults to `1777` or world-writable.
   - Options specific to `volume`:
     - unimplemented options: `volume-nocopy`, `volume-label`, `volume-driver`, `volume-opt`
+  - Options specific to `image`:
+    - :whale: `src`, `source`: image reference (mandatory).
+    - :whale: Currently, the image filesystem is mounted read-only.
+    - unimplemented options: `image-subpath`
 - :whale: `--volumes-from`: Mount volumes from the specified container(s), e.g. "--volumes-from my-container".
 
 Rootfs flags:
--- a/pkg/cmd/container/remove.go
+++ b/pkg/cmd/container/remove.go
@@ -109,6 +109,13 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions 
 		return err
 	}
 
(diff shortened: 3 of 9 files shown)


mayur-tolexo:
### Local test results

Linux + containerd dev sandbox (containerd 1.7.24, runc 1.1.15, go 1.26; overlayfs snapshotter).

```
vet: clean
build: ok
--- unit ---
ok  	github.com/containerd/nerdctl/v2/pkg/mountutil	0.004s
--- integration ---
--- PASS: TestRunMountTypeImageErrors (0.00s)
--- PASS: TestRunMountTypeImageReadOnly (0.35s)
--- PASS: TestRunMountTypeImageMultipleDestinations (0.43s)
--- PASS: TestRunMountTypeImage (0.53s)
```

Also manually verified: image FS readable at the target; `touch` into the mount fails (read-only); same image mounted at two destinations; the snapshot view is created at `run`, survives while the container runs, and is removed on a clean `rm` (count 1→0); a `create` that fails after a view is created leaves **0** leaked snapshots.

pkg/mountutil/mountutil_linux.go
```diff
@@ -369,6 +375,13 @@ func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, e
 			if trueValue {
 				rwOption = key
 			}
+			// A writable request is either rw/rro=true or readonly/ro=false.
```
AkihiroSuda: Why rro means writable here?

docs/command-reference.md
```diff
@@ -313,6 +312,9 @@ Volume flags:
       Defaults to `1777` or world-writable.
   - Options specific to `volume`:
     - unimplemented options: `volume-nocopy`, `volume-label`, `volume-driver`, `volume-opt`
+  - Options specific to `image`:
+    - :nerd_face: `src`, `source`: image reference (mandatory). The image filesystem is mounted read-only.
```
AkihiroSuda: 🐳 for Docker-compatible options, 🤓 for other ones

pkg/mountutil/mountutil_linux.go
mayur-tolexo: Thanks for the quick review, It shouldn't. Let me fix it.

pkg/mountutil/mountutil_linux.go
```diff
@@ -394,6 +407,28 @@ func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, e
...
+			return nil, fmt.Errorf("type=image mounts are read-only")
+		}
+		return &Processed{
+			Type: Image,
+			Mount: specs.Mount{
+				Type:        Image,
+				Source:      src,
+				Destination: cleanMount(dst),
+			},
+		}, nil
```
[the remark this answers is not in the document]
mayur-tolexo: Done. Set Mode: "ro" on the image Processed so inspect/labels report the mount read-only.

pkg/cmd/container/remove.go
```diff
@@ -275,6 +282,16 @@ func RemoveContainer(ctx context.Context, c containerd.Container, globalOptions
...
 		}
+
+		// Remove the read-only views backing type=image mounts - soft failure.
+		if snapshotsJSON, ok := containerLabels[labels.ImageMountSnapshots]; ok {
+			var keys []string
+			if err = json.Unmarshal([]byte(snapshotsJSON), &keys); err != nil {
+				log.G(ctx).WithError(err).Warnf("failed to unmarshall image-mount snapshots for container %q", id)
+			} else {
+				removeImageMountViews(ctx, client.SnapshotService(imageMountSnapshotter), keys)
+			}
```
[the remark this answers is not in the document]
mayur-tolexo: fixed

pkg/mountutil/mountutil_linux.go
```diff
@@ -369,6 +375,13 @@ func ProcessFlagMount(s string, volStore volumestore.VolumeStore) (*Processed, e
 			if trueValue {
 				rwOption = key
 			}
```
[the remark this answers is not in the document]
mayur-tolexo: Out of scope for this PR. That if trueValue { rwOption = key } branch is pre-existing and shared by bind/volume; rw=false not mapping to read-only is an existing behavior unrelated to type=image. Image mounts handle this separately (any explicit write request, including rw/readonly=false, is rejected since image mounts are read-only). 

Happy to fix the general rw=false case in a separate PR.

docs/command-reference.md
```diff
@@ -313,6 +312,9 @@ Volume flags:
       Defaults to `1777` or world-writable.
   - Options specific to `volume`:
     - unimplemented options: `volume-nocopy`, `volume-label`, `volume-driver`, `volume-opt`
+  - Options specific to `image`:
+    - :whale: `src`, `source`: image reference (mandatory). The image filesystem is mounted read-only.
+    - unimplemented options: `subpath`
```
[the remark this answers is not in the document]
mayur-tolexo: Kept the wording consistent with the existing volume section's unimplemented options style.

mayur-tolexo:
Tested locally against containerd, using `alpine` as both the source and the container image.

Reading a file from the mounted image:

```console
$ nerdctl run --rm --mount type=image,source=alpine,destination=/mnt/img alpine cat /mnt/img/etc/alpine-release
3.24.1
```

The mount is read-only:

```console
$ nerdctl run --rm --mount type=image,source=alpine,destination=/mnt/img alpine touch /mnt/img/x
touch: /mnt/img/x: Read-only file system
```

Same image mounted at two destinations:

```console
$ nerdctl run --rm --mount type=image,source=alpine,destination=/a --mount type=image,source=alpine,destination=/b alpine sh -c 'cat /a/etc/alpine-release; cat /b/etc/alpine-release'
3.24.1
3.24.1
```

Invalid specs are rejected:

```console
$ nerdctl run --rm --mount type=image,source=alpine,destination=/mnt/img,readonly=false alpine true
FATA[0000] type=image mounts are read-only

$ nerdctl run --rm --mount type=image,destination=/mnt/img alpine true
FATA[0000] type=image requires a source (the image reference)

$ nerdctl run --rm --mount type=image,source=alpine,destination=/mnt/img,subpath=etc alpine true
FATA[0000] mount option "subpath" is not yet supported
```

`inspect` reports it read-only, and the snapshot view is cleaned up on removal:

```console
$ nerdctl run --name demo --mount type=image,source=alpine,destination=/mnt/img alpine true
$ nerdctl inspect --format '{{range .Mounts}}{{.Type}} {{.Source}} -> {{.Destination}} RW={{.RW}} Mode={{.Mode}}{{end}}' demo
image alpine -> /mnt/img RW=false Mode=ro

# image-mount snapshot exists while the container is around:
$ ctr snapshot ls | grep -c image-mount
1
$ nerdctl rm demo
demo
$ ctr snapshot ls | grep -c image-mount
0
```

pkg/mountutil/mountutil_linux.go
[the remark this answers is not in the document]
mayur-tolexo: Create a new issue to fix: https://github.com/containerd/nerdctl/issues/4991

pkg/cmd/container/run_mount.go
```diff
@@ -122,17 +123,83 @@ func parseMountFlags(volStore volumestore.VolumeStore, options types.ContainerCr
...
+	if !strutil.InStringSlice(opts, "ro") {
+		opts = append(opts, "ro")
+	}
+	return specs.Mount{
+		Type:        m.Type,
+		Source:      m.Source,
+		Destination: destination,
+		Options:     opts,
+	}, snapshotKey, nil
+}
```
jiwahn, line 175: If you plan to add `image-subpath` later, I think the mount setup should probably first materialize the image rootfs on the host, then bind-mount the selected path into the container.

cmd/nerdctl/container/container_run_mount_image_linux_test.go
```diff
@@ -0,0 +1,134 @@
...
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		return helpers.Command("run", "--rm",
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/a", testutil.CommonImage),
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/b", testutil.CommonImage),
+			testutil.CommonImage, "cat", "/mnt/a/etc/os-release", "/mnt/b/etc/os-release")
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: 0,
```
sathiraumesh: you can use the Tigron frameworks exit codes

sathiraumesh (changes requested):

pkg/cmd/container/run_mount.go
mayur-tolexo: Agreed. The current overlay-as-OCI-mount approach is whole-rootfs only, `subpath` would need the rootfs materialized on the host and the selected path bind-mounted in.
 
 
 I kept it minimal here to avoid a daemonless host-mount that must be unmounted on `rm`.

mayur-tolexo: I'd add the materialize-then-bind structure in the subpath PR. Let me know if you'd rather I lay that groundwork now.

cmd/nerdctl/container/container_run_mount_image_linux_test.go
mayur-tolexo: done

pkg/cmd/container/run_mount.go
mayur-tolexo: https://github.com/containerd/nerdctl/pull/4993

AkihiroSuda:
Needs rebase

AkihiroSuda:
CI failing

```
--- FAIL: TestProcessFlagMountImage (0.00s)
    --- FAIL: TestProcessFlagMountImage/type=image,source=alpine:latest,destination=/mnt/img,rw (0.00s)
        mountutil_linux_test.go:521: assertion failed: expected error to contain "read-only", got "invalid field 'rw' must be a key=value pair"
FAIL
FAIL	github.com/containerd/nerdctl/v2/pkg/mountutil	0.007s
```

AkihiroSuda:
Needs rebase

cmd/nerdctl/container/container_run_mount_image_linux_test.go
```diff
@@ -0,0 +1,137 @@
...
+			ExitCode: expect.ExitCodeSuccess,
+			Output:   expect.Contains("Alpine"),
+		}
+	}
+
+	testCase.Run(t)
+}
+
+// TestRunMountTypeImageReadOnly verifies an image mount is read-only (writing
+// fails). nerdctl-only: Docker mounts images read-write by default.
```
AkihiroSuda: Why can't we adopt Docker's behavior?

docs/command-reference.md
```diff
@@ -326,6 +325,9 @@ Volume flags:
       Defaults to `1777` or world-writable.
   - Options specific to `volume`:
     - unimplemented options: `volume-nocopy`, `volume-label`, `volume-driver`, `volume-opt`
+  - Options specific to `image`:
+    - :whale: `src`, `source`: image reference (mandatory). The image filesystem is mounted read-only.
```
AkihiroSuda:
```suggestion
    - :whale: `src`, `source`: image reference (mandatory).
    - :nerd_face: Currently, the image filesystem is mounted read-only.
```

mayur-tolexo:
Ran the same commands against Docker 29.4.0 and nerdctl with this PR to confirm they line up. Both mount the image read-only, both accept `readonly=false` but keep it read-only, and both reject a missing source. The one gap is `image-subpath`, which Docker implements and nerdctl doesn't yet.

Docker:

```
$ docker --version
Docker version 29.4.0, build 9d7ad9f

$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine:latest cat /mnt/img/etc/os-release
WARNING: Image mount is an experimental feature
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.24.1

$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine:latest touch /mnt/img/x
WARNING: Image mount is an experimental feature
touch: /mnt/img/x: Read-only file system

$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,readonly=false alpine:latest touch /mnt/img/x
WARNING: Image mount is an experimental feature
touch: /mnt/img/x: Read-only file system

$ docker run --rm --mount type=image,destination=/mnt/img alpine:latest true
docker: Error response from daemon: invalid mount config for type "image": field Source must not be empty
```

nerdctl with this PR:

```
$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine:latest cat /mnt/img/etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.24.1

$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine:latest touch /mnt/img/x
touch: /mnt/img/x: Read-only file system

$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,readonly=false alpine:latest touch /mnt/img/x
touch: /mnt/img/x: Read-only file system

$ nerdctl run --rm --mount type=image,destination=/mnt/img alpine:latest true
time="2026-07-13T06:21:21Z" level=fatal msg="type=image requires a source (the image reference)"
```

`image-subpath` is the one difference for now — Docker mounts the subdirectory, nerdctl rejects it:

```
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=etc alpine:latest ls /mnt/img
WARNING: Image mount is an experimental feature
alpine-release
apk
busybox-paths.d
...

$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=etc alpine:latest ls /mnt/img
time="2026-07-13T06:21:21Z" level=fatal msg="mount option \"image-subpath\" is not yet supported"
```

Also verified the snapshot view is cleaned up on `rm` (no leak) and ran it end to end on both containerd 1.7 and 2.2.

cmd/nerdctl/container/container_run_mount_image_linux_test.go
mayur-tolexo: You're right, and honestly the comment I had on that test was just wrong. Docker mounts image mounts read-only too (checked on 29.4.0), not read-write, so there is nothing nerdctl-specific here. I dropped the `require.Not(nerdtest.Docker)` so the test now runs against Docker as well, and made `readonly=false` a no-op instead of an error since that is what Docker does (it silently keeps the mount read-only).

docs/command-reference.md
mayur-tolexo: Applied. I also renamed the unimplemented `subpath` on the next line to `image-subpath` to match Docker's actual option name.

docs/command-reference.md
```diff
@@ -326,6 +325,10 @@ Volume flags:
       Defaults to `1777` or world-writable.
   - Options specific to `volume`:
     - unimplemented options: `volume-nocopy`, `volume-label`, `volume-driver`, `volume-opt`
+  - Options specific to `image`:
+    - :whale: `src`, `source`: image reference (mandatory).
+    - :nerd_face: Currently, the image filesystem is mounted read-only.
```
AkihiroSuda: `:nerd_face:` emoji is not needed if the behavior is same as Docker

mayur-tolexo: Done, switched it to :whale: to match the other bullets since the read-only behavior is the same as Docker.

AkihiroSuda (approved):
Thanks

Issue #2651, opened 2023-11-21
recursive read-only (RRO) mount: support the Docker v25 form


AkihiroSuda:
https://github.com/containerd/nerdctl/blob/4299debf881b375226fdf2fbb2a1f818f1d456f1/README.md?plain=1#L211-L215

ref:
- https://github.com/docker/cli/pull/4316
- https://github.com/docker/cli/pull/4606

AkihiroSuda:
May depend on:
- https://github.com/containerd/nerdctl/issues/1390

thaJeztah:
Hope I didn't make life hard on you, removing the boolean format ❤️  (I was looking at it again, and thought "now" is the moment to still change it; and it's easier to add back later, than to ship it and remove later).

AkihiroSuda:
https://github.com/containerd/nerdctl/blob/460875345229737b441a3a8618679dd0522e3d17/pkg/mountutil/mountutil_linux.go#L383-L402

It also has to support `bind-recursive=(readonly|writeonly)`

PR #4992, opened 2026-06-22
fix(mount): remove the non-Docker `rw` option from --mount

### LINKED ISSUE #4991: --mount ...,rw=false is silently ignored (mount stays writable)
### Description

`nerdctl run --mount` accepts boolean values for the `rw`/`ro`/`readonly`/`rro` options (e.g. `readonly=true`). But `rw=false` (and `rro=false`) is silently ignored — the mount remains **writable** instead of being treated as read-only.

### Steps to reproduce

```console
$ mkdir -p /tmp/bindsrc
$ nerdctl run --rm --mount type=bind,source=/tmp/bindsrc,target=/mnt,rw=false alpine touch /mnt/x
$ ls /tmp/bindsrc
x      # the write succeeded -> rw=false had no effect
```

Control (works as expected):

```console
$ nerdctl run --rm --mount type=bind,source=/tmp/bindsrc,target=/mnt,readonly alpine touch /mnt/x
touch: /mnt/x: Read-only file system
```

### Expected

`rw=false` means "not read-write", so the mount should be read-only and the write should fail (same as `readonly`/`ro`).

### Actual

The write succeeds; `rw=false` is a no-op.

### Root cause

In `pkg/mountutil/mountutil_linux.go`, `ProcessFlagMount` only records the option when the parsed value is `true`:

```go
case "readonly", "ro", "rw", "rro":
    trueValue, _ := strconv.ParseBool(value)
    if trueValue {
        rwOption = key
    }
```

So `rw=false` leaves `rwOption` empty (the writable default). It should map `rw=false` / `rro=false` to read-only (e.g. `ro`).

### Affected

`--mount type=bind` and `type=volume`.

(Spotted during review of #4990; `type=image` is unaffected — it validates the read-only/writable intent separately.)

### Environment

nerdctl: main; containerd 1.7; Linux.

mayur-tolexo:
Fixes #4991

Docker's `--mount` has no `rw` option — only `readonly`/`ro`:

```console
$ docker run --mount type=bind,source=/tmp/x,target=/mnt,rw=false ...
invalid argument ... unknown option 'rw' in 'rw=false'
```

In nerdctl, `--mount` accepted `rw`, but it was redundant (read-write is the default) and `rw=false` was **silently ignored**, leaving the mount writable instead of read-only.

## Change
- Remove `rw` from the `--mount` option parser (`ProcessFlagMount`), so `rw`/`rw=false` now return an error, matching Docker.
- `ro`/`readonly`/`rro` are unchanged.
- The `-v src:dst:rw` syntax is **unaffected** — it's handled by a separate parser (`parseVolumeOptions`), and Docker supports `:rw` for `-v`.
- Docs: `readonly`/`ro` marked Docker-compatible (🐳); `rro` marked nerdctl-specific (🤓, recursive read-only).

## Behavioral note
`--mount ...,rw` now errors instead of being a no-op. Since `rw` was the writable default (and `rw=false` was silently broken), this only affects users explicitly passing a non-standard option.

## Tests
`TestProcessFlagMountRW`: `rw`/`rw=true`/`rw=false` are rejected (asserting the stable error fragments), and `ro`/`rro` still parse to their read-only options.

Verified in a Linux + containerd sandbox: gofmt, `go vet`, and the mountutil tests pass; manually, `--mount ...,rw` and `rw=false` error, while `--mount ...,ro` and `...,rro` still enforce read-only.

### DIFF
--- a/pkg/mountutil/mountutil_linux_test.go
+++ b/pkg/mountutil/mountutil_linux_test.go
@@ -352,3 +352,66 @@ func TestProcessFlagVAnonymousVolumes(t *testing.T) {
 		})
 	}
 }
+
+// TestProcessFlagMountRW verifies that the non-Docker `rw` option is no longer
+// accepted by --mount, while ro/readonly/rro remain valid read-only flags.
+func TestProcessFlagMountRW(t *testing.T) {
+	// rw is no longer a valid --mount option.
+	rejected := []struct {
+		spec string
+		want string
+	}{
+		{"type=bind,source=/foo,target=/bar,rw", "must be a key=value pair"},
+		{"type=bind,source=/foo,target=/bar,rw=true", "unexpected key 'rw'"},
+		{"type=bind,source=/foo,target=/bar,rw=false", "unexpected key 'rw'"},
+	}
+	for _, tt := range rejected {
+		t.Run(tt.spec, func(t *testing.T) {
+			_, err := ProcessFlagMount(tt.spec, nil)
+			assert.ErrorContains(t, err, tt.want)
+		})
+	}
+
+	// ro/rro still parse into a complete read-only bind mount.
+	src := t.TempDir()
+	accepted := []struct {
+		spec  string
+		wants *Processed
+	}{
+		{
+			spec: "type=bind,source=" + src + ",target=/bar,ro",
+			wants: &Processed{
+				Type: Bind,
+				Mount: specs.Mount{
+					Type:        "bind",
+					Source:      src,
+					Destination: "/bar",
+					Options:     []string{"rbind", "ro", "rprivate"},
+				},
+			},
+		},
+		{
+			spec: "type=bind,source=" + src + ",target=/bar,rro",
+			wants: &Processed{
+				Type: Bind,
+				Mount: specs.Mount{
+					Type:        "bind",
+					Source:      src,
+					Destination: "/bar",
+					Options:     []string{"rbind", "ro", "rro", "rprivate"},
+				},
+			},
+		},
+	}
+	for _, tt := range accepted {
+		t.Run(tt.spec, func(t *testing.T) {
+			got, err := ProcessFlagMount(tt.spec, nil)
+			assert.NilError(t, err)
+			assert.Equal(t, got.Type, tt.wants.Type)
+			assert.Equal(t, got.Mount.Type, tt.wants.Mount.Type)
+			assert.Equal(t, got.Mount.Source, tt.wants.Mount.Source)
+			assert.Equal(t, got.Mount.Destination, tt.wants.Mount.Destination)
+			assert.DeepEqual(t, got.Mount.Options, tt.wants.Mount.Options)
+		})
+	}
+}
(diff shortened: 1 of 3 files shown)


mayur-tolexo:
Verified the `rw` behavior matches Docker, and confirmed `rro` (nerdctl-specific recursive read-only) still works.

### Docker (29.4.0) — `--mount` rejects `rw`, `-v :rw` works

```console
$ docker run --rm --mount type=bind,source=/tmp/bindsrc,target=/mnt,rw alpine true
invalid argument "type=bind,source=/tmp/bindsrc,target=/mnt,rw" for "--mount" flag: invalid field 'rw' must be a key=value pair

$ docker run --rm --mount type=bind,source=/tmp/bindsrc,target=/mnt,rw=false alpine true
invalid argument "type=bind,source=/tmp/bindsrc,target=/mnt,rw=false" for "--mount" flag: unknown option 'rw' in 'rw=false'

$ docker run --rm -v /tmp/bindsrc:/mnt:rw alpine sh -c 'touch /mnt/ok && echo writable-ok'
writable-ok
```

### nerdctl (this PR) — same behavior for `rw`

```console
$ nerdctl run --rm --mount type=bind,source=/tmp/b,target=/mnt,rw alpine true
FATA[0000] invalid field 'rw' must be a key=value pair

$ nerdctl run --rm --mount type=bind,source=/tmp/b,target=/mnt,rw=false alpine true
FATA[0000] unexpected key 'rw' in 'rw=false'

$ nerdctl run --rm -v /tmp/b:/mnt:rw alpine sh -c 'touch /mnt/ok && echo writable-ok'
writable-ok
```

The `--mount ...,rw` errors line up with Docker's (the bare-`rw` message is identical); `-v ...:rw` keeps working on both.

### `rro` works as expected (nerdctl-specific, recursive read-only)

```console
$ nerdctl run --rm --mount type=bind,source=/tmp/b,target=/mnt,rro alpine touch /mnt/x
touch: /mnt/x: Read-only file system
```

Docker has no `rro` for `--mount` (it's a nerdctl/runc feature), so Docker rejects it — expected divergence:

```console
$ docker run --rm --mount type=bind,source=/tmp/bindsrc,target=/mnt,rro alpine true
invalid argument "...,rro" for "--mount" flag: invalid field 'rro' must be a key=value pair
```

AkihiroSuda:
> Docker has no rro for --mount (it's a nerdctl/runc feature)

The feature exists in Docker too, but the option name is different. (Out of the scope of this PR)
https://github.com/docker/cli/blob/v29.6.0/opts/mount.go#L90-L104
- https://github.com/containerd/nerdctl/issues/2651

pkg/mountutil/mountutil_linux_test.go
```diff
@@ -352,3 +352,40 @@ func TestProcessFlagVAnonymousVolumes(t *testing.T) {
...
+			_, err := ProcessFlagMount(tt.spec, nil)
+			assert.ErrorContains(t, err, tt.want)
+		})
+	}
+
+	// ro/rro still parse and yield the corresponding read-only option.
+	src := t.TempDir()
+	accepted := []struct {
+		spec string
+		want string
```
AkihiroSuda: This should want a complete struct, not just an option string

mayur-tolexo: added the Processed struct.

AkihiroSuda (approved):
Thanks

PR #4993, opened 2026-06-22
feat(mount): support --mount type=image with image-subpath

mayur-tolexo:
## What

Implements `--mount type=image` end-to-end, matching Docker's `--mount type=image` (moby/moby#48798), including the `image-subpath` option.

Part of #3867. Builds on #4990 (read-only whole-rootfs image mount); this squashes that MVP and the `image-subpath` work into a single commit.

- `--mount type=image,source=<image>,destination=<path>`: mounts the image rootfs read-only at the destination.
- `--mount type=image,source=<image>,destination=<path>,image-subpath=<rel/path>`: mounts only a subdirectory of the image rootfs.

Image mounts are read-only, matching Docker (verified against Docker 29.6.2: writes return `Read-only file system`).

## How

- **Parser** (`pkg/mountutil`): accepts `image-subpath` and normalizes it. Internal `..` is collapsed (`a/../b` → `b`); an absolute path or a `..` that escapes the rootfs is rejected; a value that normalizes to the rootfs itself (`.`, or `a/..`) mounts the whole rootfs, which is what Docker does. Presence is tracked separately from the value, so an explicit empty `image-subpath=` is rejected rather than silently read as "unset", and `image-subpath` is rejected on any type other than `image`.
- **Whole-rootfs** (no subpath): hands the snapshotter's read-only view straight to the OCI runtime, which owns the mount lifecycle. Unchanged from #4990.
- **Subpath**: an OCI overlay mount cannot select a subdirectory, so the read-only view is materialized on a host directory under the data root, and a read-only bind mount of the subdirectory is added. The subpath is opened with `os.OpenInRoot`, which requires it to exist and rejects absolute or escaping symlinks rather than following them against the host; anything left in the path is a relative symlink inside the rootfs, so the kernel resolves the bind source to the image's own target. The host materialization path is recorded on a container label and unmounted/removed on `nerdctl rm`, alongside the snapshot view.
- **Reserved labels**: labels in the internal `nerdctl/` namespace are now stripped from an image's config labels. `--label` already rejects that prefix, but image config labels bypassed it, which would let an image forge internal container state — including the image-mount host paths that `nerdctl rm` unmounts and deletes.

## Not included (parity notes vs moby#48798)

- Read-write image mounts: Docker presents image mounts read-only in practice, so this matches.
- Daemon-restart layer restoration / image-layer refcounting: daemon-specific. nerdctl is daemonless; the snapshot view is protected from GC via a `gc.root` label. A subpath mount's host materialization persists across container restart but not across a host reboot.

## Test

- Unit: `pkg/mountutil` parser cases (subpath normalization, escaping-traversal / absolute / empty / non-image rejection, root-normalizing → whole rootfs).
- Integration: `cmd/nerdctl/container/container_run_mount_image_linux_test.go` (subpath exposure, multiple subpaths, read-only, absolute-symlink rejection, relative-symlink resolution, error cases).
- Manual e2e against Docker 29.6.2 — logs posted in a comment below.

### DIFF
--- a/cmd/nerdctl/container/container_run_mount_image_linux_test.go
+++ b/cmd/nerdctl/container/container_run_mount_image_linux_test.go
@@ -95,10 +95,80 @@ func TestRunMountTypeImageReadOnly(t *testing.T) {
 	testCase.Run(t)
 }
 
+// TestRunMountTypeImageSubpath verifies that image-subpath exposes only the
+// selected directory of the image rootfs at the destination: the image's
+// /etc/os-release is reachable as <destination>/os-release.
+func TestRunMountTypeImageSubpath(t *testing.T) {
+	testCase := nerdtest.Setup()
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		return helpers.Command("run", "--rm",
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img,image-subpath=etc", testutil.CommonImage),
+			testutil.CommonImage, "cat", "/mnt/img/os-release")
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: expect.ExitCodeSuccess,
+			Output:   expect.Contains("Alpine"),
+		}
+	}
+
+	testCase.Run(t)
+}
+
+// TestRunMountTypeImageSubpathMultiple verifies that two image-subpath mounts of
+// the same image at different destinations each expose their own subdirectory,
+// exercising the multi-mount label round-trip and cleanup.
+func TestRunMountTypeImageSubpathMultiple(t *testing.T) {
+	testCase := nerdtest.Setup()
+	// nerdctl-only: Docker keys an image mount by its source image and rejects
+	// mounting the same image twice ("mount already exists with name").
+	testCase.Require = require.Not(nerdtest.Docker)
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		return helpers.Command("run", "--rm",
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/etc,image-subpath=etc", testutil.CommonImage),
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/bin,image-subpath=bin", testutil.CommonImage),
+			testutil.CommonImage, "ls", "/mnt/etc", "/mnt/bin")
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: expect.ExitCodeSuccess,
+		}
+	}
+
+	testCase.Run(t)
+}
+
+// TestRunMountTypeImageSubpathReadOnly verifies an image-subpath mount is
+// read-only so writing fails. nerdctl-only: Docker's image-subpath read-only
+// behavior is unverified, so this is not run against Docker.
+func TestRunMountTypeImageSubpathReadOnly(t *testing.T) {
+	testCase := nerdtest.Setup()
+	testCase.Require = require.Not(nerdtest.Docker)
+
+	testCase.Command = func(data test.Data, helpers test.Helpers) test.TestableCommand {
+		return helpers.Command("run", "--rm",
+			"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img,image-subpath=etc", testutil.CommonImage),
+			testutil.CommonImage, "touch", "/mnt/img/should-fail")
+	}
+
+	testCase.Expected = func(data test.Data, helpers test.Helpers) *test.Expected {
+		return &test.Expected{
+			ExitCode: expect.ExitCodeGenericFail,
+			Errors:   []error{fmt.Errorf("Read-only file system")},
+		}
+	}
+
+	testCase.Run(t)
+}
+
 // TestRunMountTypeImageErrors verifies that an image mount missing its source,
-// or using the not-yet-supported image-subpath option, is rejected. Docker
-// implements image-subpath, so that case diverges and the test is not run
-// against Docker.
+// or using the not-yet-supported subpath option, or an image-subpath that
+// escapes the rootfs, is rejected. These are nerdctl-specific behaviours here,
+// so the test is not run against Docker.
 func TestRunMountTypeImageErrors(t *testing.T) {
 	testCase := nerdtest.Setup()
 	testCase.Require = require.Not(nerdtest.Docker)
@@ -118,16 +188,44 @@ func TestRunMountTypeImageErrors(t *testing.T) {
 			},
 		},
 		{
-			Description: "image-subpath not supported",
+			Description: "subpath not supported",
 			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
 				return helpers.Command("run", "--rm",
-					"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img,image-subpath=etc", testutil.CommonImage),
+					"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img,subpath=etc", testutil.CommonImage),
 					testutil.CommonImage, "true")
 			},
 			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
 				return &test.Expected{
 					ExitCode: expect.ExitCodeGenericFail,
-					Errors:   []error{fmt.Errorf("image-subpath")},
+					Errors:   []error{fmt.Errorf("subpath")},
+				}
+			},
+		},
+		{
+			Description: "image-subpath parent traversal rejected",
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("run", "--rm",
+					"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img,image-subpath=../etc", testutil.CommonImage),
+					testutil.CommonImage, "true")
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: expect.ExitCodeGenericFail,
+					Errors:   []error{fmt.Errorf("escapes")},
+				}
+			},
+		},
+		{
+			Description: "image-subpath absolute rejected",
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("run", "--rm",
+					"--mount", fmt.Sprintf("type=image,source=%s,destination=/mnt/img,image-subpath=/etc", testutil.CommonImage),
+					testutil.CommonImage, "true")
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: expect.ExitCodeGenericFail,
+					Errors:   []error{fmt.Errorf("relative")},
 				}
 			},
 		},
--- a/pkg/cmd/container/run_mount.go
+++ b/pkg/cmd/container/run_mount.go
@@ -127,19 +127,24 @@ func parseMountFlags(volStore volumestore.VolumeStore, options types.ContainerCr
 const gcRootLabel = "containerd.io/gc.root"
 
 // setupImageMount ensures and unpacks ref, then creates a read-only GC-rooted
-// snapshot view of its rootfs. Image mounts are always read-only, matching
-// Docker. It returns the OCI mount for destination and the view's snapshot key.
-func setupImageMount(ctx context.Context, client *containerd.Client, options types.ContainerCreateOptions, ref, destination string) (specs.Mount, string, error) {
+// snapshot view of its rootfs. Without a subpath it returns the snapshotter's
+// own mount for destination (runc owns its lifecycle). With a subpath it
+// materializes the view on a host directory and returns a read-only bind mount
+// of the resolved subdirectory, because an OCI overlay mount cannot select a
+// subdir; that host directory is returned so it can be unmounted on removal.
+// It returns the OCI mount, the view's snapshot key, and the host materialization
+// path (empty when no subpath is used).
+func setupImageMount(ctx context.Context, client *containerd.Client, options types.ContainerCreateOptions, ref, destination, subpath string) (specs.Mount, string, string, error) {
 	ensured, err := imgutil.EnsureImage(ctx, client, ref, options.ImagePullOpt)
 	if err != nil {
-		return specs.Mount{}, "", fmt.Errorf("failed to ensure image %q for image mount: %w", ref, err)
+		return specs.Mount{}, "", "", fmt.Errorf("failed to ensure image %q for image mount: %w", ref, err)
 	}
 	if err := ensured.Image.Unpack(ctx, options.GOptions.Snapshotter); err != nil {
-		return specs.Mount{}, "", fmt.Errorf("failed to unpack image %q for image mount: %w", ref, err)
+		return specs.Mount{}, "", "", fmt.Errorf("failed to unpack image %q for image mount: %w", ref, err)
 	}
 	diffIDs, err := ensured.Image.RootFS(ctx)
 	if err != nil {
-		return specs.Mount{}, "", fmt.Errorf("failed to get rootfs of image %q for image mount: %w", ref, err)
+		return specs.Mount{}, "", "", fmt.Errorf("failed to get rootfs of image %q for image mount: %w", ref, err)
 	}
 	chainID := identity.ChainID(diffIDs).String()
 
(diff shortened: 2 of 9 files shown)


mayur-tolexo:
Did another round of manual testing against Docker 29.4.0 (nerdctl built from this branch) to check parity. Sharing the results.

### Whole-rootfs and subpath read the same as Docker

```console
$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine cat /mnt/img/etc/alpine-release
3.24.1
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine cat /mnt/img/etc/alpine-release
WARNING: Image mount is an experimental feature
3.24.1

$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=etc alpine cat /mnt/img/alpine-release
3.24.1
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=etc alpine cat /mnt/img/alpine-release
WARNING: Image mount is an experimental feature
3.24.1
```

Writes are rejected on both (read-only):

```console
$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine touch /mnt/img/x
touch: /mnt/img/x: Read-only file system
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img alpine touch /mnt/img/x
WARNING: Image mount is an experimental feature
touch: /mnt/img/x: Read-only file system
```

### Bad mount specs are rejected, same as Docker

nerdctl:

```console
$ nerdctl run --rm --mount type=image,destination=/mnt/img alpine true
time="2026-07-21T08:16:29Z" level=fatal msg="type=image requires a source (the image reference)"
$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,subpath=etc alpine true
time="2026-07-21T08:16:29Z" level=fatal msg="unexpected key 'subpath' in 'subpath=etc'"
$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=../etc alpine true
time="2026-07-21T08:16:29Z" level=fatal msg="image-subpath \"../etc\" escapes the image rootfs"
$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=/etc alpine true
time="2026-07-21T08:16:29Z" level=fatal msg="image-subpath must be relative to the image rootfs, got \"/etc\""
$ nerdctl run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=nope alpine true
time="2026-07-21T08:16:29Z" level=fatal msg="image-subpath \"nope\" does not exist in image \"alpine:latest\""
$ nerdctl run --rm --mount type=bind,source=/tmp,destination=/mnt,image-subpath=etc alpine true
time="2026-07-21T08:16:29Z" level=fatal msg="image-subpath is only supported for type=image"
```

Docker on the same specs:

```console
$ docker run --rm --mount type=image,destination=/mnt/img alpine true
docker: Error response from daemon: invalid mount config for type "image": field Source must not be empty
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,subpath=etc alpine true
invalid argument "type=image,source=alpine:latest,destination=/mnt/img,subpath=etc" for "--mount" flag: unknown option 'subpath' in 'subpath=etc'
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=../etc alpine true
docker: Error response from daemon: invalid mount config for type "image": subpath must be a relative path within the volume
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=/etc alpine true
docker: Error response from daemon: invalid mount config for type "image": subpath must be a relative path within the volume
$ docker run --rm --mount type=image,source=alpine:latest,destination=/mnt/img,image-subpath=nope alpine true
docker: Error response from daemon: cannot access path /var/lib/docker/rootfs/overlayfs/b59a45207c4d.../nope: lstat ...: no such file or directory
$ docker run --rm --mount type=bind,source=/tmp,destination=/mnt,image-subpath=etc alpine true
invalid argument "..." for "--mount" flag: cannot mix 'image-*' options with mount type 'bind'
```

### How the subpath mount is wired

nerdctl is daemonless, so for a subpath it materializes the image rootfs on the host (read-only) and bind-mounts the subdir into the container. For a container started with `image-subpath=etc`:

```console
$ grep image-mounts /proc/mounts
/dev/vda1 /var/lib/nerdctl/image-mounts/8b71320c4cd1215df46c5a7da54744badbe10192ad2801d95d7a11494d9bb918-image-mount ext4 ro,relatime,discard 0 0
```

The OCI mount injected for `/mnt/img` (from the runtime `config.json`):

```json
{
  "destination": "/mnt/img",
  "type": "bind",
  "source": "/var/lib/nerdctl/image-mounts/8b71320c4cd1215df46c5a7da54744badbe10192ad2801d95d7a11494d9bb918-image-mount/etc",
  "options": ["rbind", "ro"]
}
```

And inside the container:

```console
$ grep ' /mnt/img ' /proc/<pid>/mountinfo
1969 1959 254:1 /docker/volumes/nerdctl-cd-data/_data/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs/etc /mnt/img ro,relatime master:1 - ext4 /dev/vda1 rw,discard
```

The snapshot view and host path are recorded on the container so they can be cleaned up on removal:

```console
$ nerdctl inspect imgtest --format '{{json .Config.Labels}}' | tr ',' '\n' | grep image-mount
"nerdctl/image-mount-hostpaths":"[\"/var/lib/nerdctl/image-mounts/8b71320c4cd1215df46c5a7da54744badbe10192ad2801d95d7a11494d9bb918-image-mount\"]"
"nerdctl/image-mount-snapshots":"[\"8b71320c4cd1215df46c5a7da54744badbe10192ad2801d95d7a11494d9bb918-image-mount\"]"
```

On `nerdctl rm` both are torn down. With two subpath mounts:

```console
$ grep -c image-mounts /proc/mounts   # while running
2
$ grep -c image-mounts /proc/mounts   # after stop + rm
0
```

### One difference from Docker

`rmi` of an image used only as a mount source (not the container's base image) succeeds in nerdctl, and the running mount keeps working because the snapshot view is gc-rooted. Docker refuses unless forced. Not a safety issue, just a behavioral difference worth noting.

```console
$ nerdctl rmi busybox:latest   # while a mount-only container runs
Untagged: docker.io/library/busybox:latest@sha256:fd8d9aa63ba2...
Deleted: sha256:66cb17eae60e...
$ nerdctl exec mntonly ls /mnt/bb/bin/busybox
/mnt/bb/bin/busybox

$ docker rmi busybox:latest   # while a mount-only container runs
Error response from daemon: conflict: unable to delete busybox:latest (must be forced) - container bf6231dbb625 is using its referenced image fd8d9aa63ba2
```

AkihiroSuda:
> This builds on https://github.com/containerd/nerdctl/pull/4990 (the first commit here is that MVP: read-only whole-rootfs image mount). The second commit adds image-subpath, completing the feature.

I only see a single commit

AkihiroSuda:
If this PR is expected to be merged after #4990, please click `Convert to draft`

AkihiroSuda:
Is this still a draft?

mayur-tolexo:
> Is this still a draft?

Yeah its ready for review.

jiwahn:
I think images used by existing containers as image mounts should not be removed by rmi

pkg/mountutil/mountutil_linux.go
```diff
@@ -360,6 +361,30 @@ func ProcessFlagTmpfs(s string) (*Processed, error) {
...
+	clean := path.Clean(p)
+	// Clean collapses ".." segments; anything still leading with ".." escapes root.
+	if clean == ".." || strings.HasPrefix(clean, "../") {
+		return "", fmt.Errorf("image-subpath %q escapes the image rootfs", p)
+	}
+	// "." means the whole rootfs (e.g. from "a/.."); that is the no-subpath case,
+	// not a subdirectory selection, so reject it as a misuse of image-subpath.
+	if clean == "." {
+		return "", fmt.Errorf("image-subpath %q must select a subdirectory, not the image rootfs", p)
+	}
```
jiwahn: nit, but docker accepts this.

mayur-tolexo: Good catch, fixed. A subpath that resolves to the rootfs (`.`, or anything normalizing to it like `a/..`) now returns empty and mounts the whole image view like Docker, instead of erroring. Updated the two tests to match.

mayur-tolexo:
Agreed it shouldn't be silently removable. The view is already pinned with a `containerd.io/gc.root` label on the snapshot, which is independent of the image record, so `rmi` can't pull the layers out from under a running or stopped container — the mount keeps working. What's missing is Docker's up-front `rmi` conflict ("image is being used by container"), which needs the mount's image ref recorded on the container and checked in `rmi`. That's rmi-side refcounting, separate from the mount feature this PR targets, so I'd rather track it as a follow-up than widen this PR. Happy to open an issue for it.

pkg/mountutil/mountutil_linux.go
```diff
@@ -360,6 +361,31 @@ func ProcessFlagTmpfs(s string) (*Processed, error) {
 	return res, nil
 }
 
+// validateImageSubpath normalizes an image-subpath value and rejects paths that
+// are absolute or escape the image rootfs. A value that resolves to the rootfs
+// itself (empty, or one normalizing to ".", e.g. "a/..") returns empty, the
+// no-subpath case Docker also accepts. Image paths are always forward-slash, so
+// it uses path, not filepath.
+func validateImageSubpath(p string) (string, error) {
```
djdongjin, line 368: we can use `os.OpenInRoot` to make sure the subpath isn't escaped from the rootfs host path.

ref: image volume support in containerd cri

https://github.com/containerd/containerd/blob/f7302c343cfed1d4b773082272dd4ca884217e79/internal/cri/server/container_image_mount.go#L247

mayur-tolexo: Done — `resolveImageSubpath` now re-checks with `os.OpenInRoot` after `securejoin`, so the scoped kernel lookup confirms existence and rejects absolute symlinks/escapes. Verified an absolute in-image symlink (`bin/sh` -> `/bin/busybox`) is rejected while a safe relative one (`var/run` -> `../run`) resolves within the image; both match Docker. Thanks for the pointer.

_e2e/bin/nerdctl
AkihiroSuda, line 1: Seems unexpectedly committed

_e2e/probe/main.go
AkihiroSuda, line 1: ?

_e2e/bin/nerdctl
mayur-tolexo: Oops, those were local test scratch files, removed it.

pkg/cmd/container/run_mount.go
```diff
@@ -171,13 +181,93 @@ func setupImageMount(ctx context.Context, client *containerd.Client, options typ
...
+		}
+		os.RemoveAll(hostMountpoint)
+		removeView()
+	}
+
+	resolved, err := resolveImageSubpath(hostMountpoint, subpath)
+	if err != nil {
+		cleanup()
+		return specs.Mount{}, "", "", fmt.Errorf("image-subpath %q in image %q: %w", subpath, ref, err)
+	}
```
djdongjin, line 222: this duplicated the cleanup logic twice. You can have a named return like `retErr` and a defered func to clean up.

pkg/cmd/container/run_mount.go
```diff
@@ -171,13 +181,93 @@ func setupImageMount(ctx context.Context, client *containerd.Client, options typ
...
+	// re-anchoring it at rootfs. Docker rejects those subpaths too.
+	f, err := os.OpenInRoot(rootfs, subpath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return "", errors.New("does not exist in the image")
+		}
+		return "", err
+	}
+	f.Close()
+	return resolved, nil
```
djdongjin, line 253: you don't need to do both securejoin and openinroot.

this is enough

```golang
	file, err := os.OpenInRoot(rootfs, subpath)
	if err != nil {
		return "", err
	}
	defer file.Close()
    return file.Name(), nil
```

pkg/cmd/container/run_mount.go
```diff
@@ -149,16 +149,26 @@ func setupImageMount(ctx context.Context, client *containerd.Client, options typ
...
+	// removeView drops the snapshot view on any failure after it was created.
+	removeView := func() {
 		if rmErr := s.Remove(ctx, snapshotKey); rmErr != nil && !errdefs.IsNotFound(rmErr) {
 			log.G(ctx).WithError(rmErr).Warnf("failed to remove image-mount snapshot %q", snapshotKey)
 		}
-		return specs.Mount{}, "", fmt.Errorf("image mount expects exactly one mount from the snapshotter, got %d", len(mounts))
 	}
 
+	if subpath != "" {
+		return setupImageSubpathMount(ctx, options, ref, destination, subpath, snapshotKey, mounts, removeView)
```
djdongjin, line 162: why not just call `removeView` here if it returns an error?

djdongjin: and then you can just have a named return and a defer func which clean up `if retErr != nil`. So you don't need to call removeView explicitly everywhere.

pkg/mountutil/mountutil_linux.go
djdongjin: do you still need this function?

djdongjin (changes requested):

AkihiroSuda:
@mayur-tolexo Could you check the comments from djdongjin ? Thanks

pkg/cmd/container/run_mount.go
mayur-tolexo: Done. Both setup functions now use a named return and a single deferred cleanup, one for the view and one for the host dir, so there are no explicit unwind calls left.

pkg/cmd/container/run_mount.go
mayur-tolexo: Folded into the deferred cleanup above, the duplicate is gone.

pkg/cmd/container/run_mount.go
mayur-tolexo: Switched to OpenInRoot only and return the file name. Anything it accepts is a relative symlink that stays inside the rootfs, so the kernel resolves the bind source to the same target. Added an integration test for a subpath through a relative `..` symlink, and dropped the `var/run` test, which only passed because the 3.13 image has an absolute symlink there.

pkg/mountutil/mountutil_linux.go
mayur-tolexo: Kept it. It rejects absolute and escaping values at parse time, before the image is pulled, unpacked and materialized, like the other mount options, and it is what maps `.` or `a/..` to a whole-rootfs mount as Docker does. Happy to drop it if you would rather have the late error.

AkihiroSuda:
@djdongjin Was your concern resolved?

djdongjin:
> @djdongjin Was your concern resolved?

taking a look today

pkg/mountutil/mountutil_linux.go
djdongjin: we should just remove this.

> it is what maps . or a/.. to a whole-rootfs mount

isn't `os.OpenInRoot` itself already sufficient?

djdongjin (approved):

cmd/nerdctl/container/container_run_mount_image_linux_test.go
```diff
@@ -95,10 +95,131 @@ func TestRunMountTypeImageReadOnly(t *testing.T) {
...
+
+	testCase.Run(t)
+}
+
+// TestRunMountTypeImageSubpathAbsoluteSymlink verifies a subpath through an
+// absolute symlink (Alpine's /bin/sh -> /bin/busybox) is rejected, not followed
+// against the host. Docker rejects it too but with its own message.
+func TestRunMountTypeImageSubpathAbsoluteSymlink(t *testing.T) {
+	testCase := nerdtest.Setup()
+	testCase.Require = require.Not(nerdtest.Docker)
```
AkihiroSuda: You could keep the test for Docker, but switch `test.Expected.Errors` message for Docker

pkg/cmd/container/run_mount.go
```diff
@@ -171,13 +187,79 @@ func setupImageMount(ctx context.Context, client *containerd.Client, options typ
...
+	}
+	source, err := resolveImageSubpath(hostMountpoint, subpath)
+	if err != nil {
+		return specs.Mount{}, "", fmt.Errorf("image-subpath %q in image %q: %w", subpath, ref, err)
+	}
+	return specs.Mount{
+		Type:        "bind",
+		Source:      source,
+		Destination: destination,
+		Options:     []string{"rbind", "ro"},
```
AkihiroSuda: Is `rbind` necessary here?
`{"rbind", "ro"}` doesn't make the mount recursively read-only.

mayur-tolexo: Dropped to `bind`. Nothing is mounted under the host dir, so `rbind` never picked up anything, and as you say `ro` wouldn't have covered a submount if one showed up.

cmd/nerdctl/container/container_run_mount_image_linux_test.go
mayur-tolexo: Done, it runs against Docker now.

I had to change the fixture though. With Alpine's `/bin/sh -> /bin/busybox` Docker fails with `lstat /usr/bin/busybox: no such file or directory` — it resolves the absolute symlink against the daemon host and only errors because the target happens to be missing there, so the message carries a host path and depends on the host. Switched to a built image with `ln -s /etc /abs`: `/etc` exists everywhere, so Docker consistently gives `path concatenation escapes the base directory` and nerdctl gives `path escapes from parent`. The test now asserts the escape is refused rather than an ENOENT.

AkihiroSuda (approved):
Thanks

AkihiroSuda (approved):
Thanks

Issue #3185, opened 2024-07-05
Incorrect usage is displayed when running nerdctl without any command


narharim:
### Description

Without any command passed to client, it's display [flags] instead of [command] in usage

### Steps to reproduce the issue

1. Run `go build .` inside `cmd/nerdctl`
2. Run `./nerdctl`

### Describe the results you received and expected

Received 
Usage: nerdctl [flags]

Expected
Usage: nerdctl [command]

### What version of nerdctl are you using?

v2.0.0-rc.0

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

narharim:
https://github.com/containerd/nerdctl/blob/bc24aa95d8b2b0b2342b88fbf97d1a44a961aac2/cmd/nerdctl/main.go#L55
I think this line should be replaced with `if c != c.Root()`

fahedouch:
> @AkihiroSuda I would like to raise a PR for this.

please go ahead!

apostasie:
@narharim curious if you are still interested in working on this?

PR #4994, opened 2026-06-22
fix: skip rootless reexec for help-only command paths

immanuwell:
rootless help paths can die early right now.

`nerdctl` and parent commands like `nerdctl system` try to reexec into RootlessKit before showing help. if rootless containerd is not running, users get `rootless containerd not running?` instead of usage.

This skips that reexec when the command only needs to show subcommands. real runtime commands still keep the old path, so no funny business.

How to repro

1. On Linux as a non root user, make sure rootless containerd is not running.
2. Run `go build -o /tmp/nerdctl ./cmd/nerdctl`
3. Run `/tmp/nerdctl`
4. Run `/tmp/nerdctl system`

Before, both commands fail with `rootless containerd not running?`.
After, both print help. `system info` still follows the old path.

Kinda adjacent to #3185.

### DIFF
--- a/cmd/nerdctl/main_linux.go
+++ b/cmd/nerdctl/main_linux.go
@@ -35,6 +35,9 @@ func appNeedsRootlessParentMain(cmd *cobra.Command, args []string) bool {
 	if !rootlessutil.IsRootlessParent() {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


sathiraumesh (approved):

AkihiroSuda (approved):
Thanks

PR #4999, opened 2026-06-23
feat: label-based filtering added to `nerdctl events`

### LINKED ISSUE #4690: Feature Request: Support --since, --until, and label filtering in `nerdctl events`
## Summary

The `nerdctl events` command has limited support for event filtering compared to Docker CLI. Specifically, the `--since`, `--until`, and label-based filtering options either don't work as expected or are not implemented.

## Docker Documentation

From [Docker events reference](https://docs.docker.com/reference/cli/docker/system/events/):

**Supported formats:**
- Unix timestamps: `1483283804` or `1483283804.123456789`
- RFC3339: `2006-01-02T15:04:05`
- Go duration strings: `10m`, `1h30m`

```bash
# Events from the last 10 minutes
docker events --since '10m'

# Events in a time range
docker events --since '2024-01-05T00:35:30' --until '2024-01-05T00:36:05'
```

### Label filtering
```bash
# Filter events by container label
docker events --filter 'label=com.example.app=myapp'
```

## Podman Compatibility

**Podman fully supports these filters:**
- `--since`: All events created since the given timestamp
- `--until`: All events created until the given timestamp
- Accepts both RFC3339Nano timestamps and Go duration strings (e.g., `10m`, `5h`)

Reference: [Podman events documentation](https://docs.podman.io/en/latest/markdown/podman-events.1.html)

## Use Cases

1. **IDE/Editor Extensions**: VS Code Docker extension uses `--since` to fetch only recent events when reconnecting to the event stream, avoiding the overhead of processing historical events.

2. **CI/CD Pipelines**: Build systems often need to monitor events within specific time windows for deployment verification.

3. **Monitoring Tools**: Dashboard and alerting tools filter events by time range and labels for targeted monitoring.

4. **Development Debugging**: Developers frequently need to see events from a specific time period when troubleshooting issues.

## Current Workaround

For VS Code Docker extension users, a workaround has been implemented in [PR #327](https://github.com/microsoft/vscode-docker-extensibility/pull/327) that:
- Parses relative time strings (e.g., `1m`, `30s`) and applies client-side filtering
- Throws a clear error when label filtering is requested

## Current Behavior

```bash
# These options may be silently ignored or produce unexpected results
nerdctl events --since '10m'
nerdctl events --filter 'label=key=value'
```

## Expected Behavior

```bash
# Show events from the last 10 minutes
nerdctl events --since '10m'

# Show events in a time window
nerdctl events --since '2024-01-10T10:00:00' --until '2024-01-10T11:00:00'

# Filter by label
nerdctl events --filter 'label=com.docker.compose.project=myproject'
```

## Additional Context

This feature gap was identified while implementing Finch support for the VS Code Docker extension. The containerd event API provides timestamps, so implementing these filters should be feasible. Full Docker CLI compatibility for event filtering would improve tooling integration for users choosing Finch/nerdctl as their container runtime.

Mujib-Ahasan:
## Description
This PR adds support for label-based filtering in nerdctl events. Users can now filter the event stream using container labels, for example:
```
nerdctl events --filter label=com.example.app=myapp
```

Fixes: #4690 

The implementation resolves container metadata from incoming events and matches the configured label filter before displaying the event. This helps users focus on events related to a specific workload in environments with so many running containers.

- [x] I have tested this out in my local machine before pushing it.

### DIFF
--- a/pkg/cmd/system/events.go
+++ b/pkg/cmd/system/events.go
@@ -44,6 +44,7 @@ type EventOut struct {
 	Topic     string
 	Status    Status
 	Event     string
+	Labels    map[string]string
 }
 
 type Status string
@@ -89,6 +90,31 @@ func generateEventFilter(filter, filterValue string) (func(e *EventOut) bool, er
 			}
 
 			return strings.EqualFold(string(e.Status), filterValue)
+		}, nil
+	case "LABEL":
+		parts := strings.SplitN(filterValue, "=", 2)
+		key := parts[0]
+		if key == "" {
+			return nil, fmt.Errorf("%s is an invalid label filter", filterValue)
+		}
+		wantValue := len(parts) == 2
+		var value string
+		if wantValue {
+			value = parts[1]
+		}
+		return func(e *EventOut) bool {
+			if len(e.Labels) == 0 {
+				return false
+			}
+			got, ok := e.Labels[key]
+			if !ok {
+				return false
+			}
+
+			if !wantValue {
+				return true
+			}
+			return got == value
 		}, nil
 	}
 
@@ -161,6 +187,13 @@ func Events(ctx context.Context, client *containerd.Client, options types.System
 			return err
 		}
 	}
+	labelFilterEnabled := false
+	for _, f := range options.Filters {
+		if strings.HasPrefix(strings.ToLower(f), "label=") {
+			labelFilterEnabled = true
+			break
+		}
+	}
 	filterMap, err := generateEventFilters(options.Filters)
 	if err != nil {
 		return err
@@ -175,6 +208,7 @@ func Events(ctx context.Context, client *containerd.Client, options types.System
 		if e != nil {
 			var out []byte
 			var id string
+			labels := map[string]string{}
 			if e.Event != nil {
 				v, err := typeurl.UnmarshalAny(e.Event)
 				if err != nil {
@@ -194,11 +228,19 @@ func Events(ctx context.Context, client *containerd.Client, options types.System
 			} else {
 				_, ok := data["container_id"]
 				if ok {
-					id = data["container_id"].(string)
+					if containerID, ok := data["container_id"].(string); ok {
+						id = containerID
+					}
 				}
 			}
-
-			eOut := EventOut{e.Timestamp, id, e.Namespace, e.Topic, TopicToStatus(e.Topic), string(out)}
+			if labelFilterEnabled && id != "" {
+				if container, err := client.ContainerService().Get(ctx, id); err != nil {
+					log.G(ctx).WithError(err).WithField("containerID", id).Debug("failed to retrieve container labels")
+				} else {
+					labels = container.Labels
+				}
+			}
+			eOut := EventOut{e.Timestamp, id, e.Namespace, e.Topic, TopicToStatus(e.Topic), string(out), labels}
 			match := applyFilters(&eOut, filterMap)
 			if match {
 				if tmpl != nil {
--- a/cmd/nerdctl/system/system_events_linux_test.go
+++ b/cmd/nerdctl/system/system_events_linux_test.go
@@ -51,6 +51,27 @@ func testEventFilterExecutor(data test.Data, helpers test.Helpers) test.Testable
 	return cmd
 }
 
+func testEventLabelFilterExecutor(data test.Data, helpers test.Helpers) test.TestableCommand {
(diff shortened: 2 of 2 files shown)


pkg/cmd/system/events.go
```diff
@@ -197,8 +213,13 @@ func Events(ctx context.Context, client *containerd.Client, options types.System
 					id = data["container_id"].(string)
 				}
 			}
-
-			eOut := EventOut{e.Timestamp, id, e.Namespace, e.Topic, TopicToStatus(e.Topic), string(out)}
+			if id != "" {
+				container, err := client.ContainerService().Get(ctx, id)
```
sathiraumesh: probably can improve like 

```

if container, err := client.ConatinerService.Get(ctx, id); err == nil {
    labels = container.Labels
}
```

pkg/cmd/system/events.go
AkihiroSuda, line 1: Needs tests
https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/system/system_events_linux_test.go

Mujib-Ahasan:
@AkihiroSuda there are so many failed checks and after surfacing little bit I strongly feels they are flaky and not related to the changes in this PR.

pkg/cmd/system/events.go
```diff
@@ -197,8 +213,12 @@ func Events(ctx context.Context, client *containerd.Client, options types.System
 					id = data["container_id"].(string)
 				}
 			}
-
-			eOut := EventOut{e.Timestamp, id, e.Namespace, e.Topic, TopicToStatus(e.Topic), string(out)}
+			if id != "" {
+				if container, err := client.ContainerService().Get(ctx, id); err == nil {
```
sathiraumesh: the err is ignored here may be good idea to add a log

sathiraumesh (changes requested):

AkihiroSuda:
Please squash the commits and rebase with the current main branch

Mujib-Ahasan:
I squashed commits and rebased on main then force pushed.
Cc @AkihiroSuda

AkihiroSuda:
```
2026-06-29T19:48:08.7201314Z === Failing tests ===
2026-06-29T19:48:08.7201753Z TestEventFilters
2026-06-29T19:48:08.7202138Z TestEventFilters/LabelFilter
2026-06-29T19:48:08.7202617Z TestEventFilters/LabelKeyOnlyFilter
```

https://github.com/containerd/nerdctl/actions/runs/28356414001/job/84140963740?pr=4999

The tests are failing with Docker v28.

Please try rebasing after we merge:
- https://github.com/containerd/nerdctl/pull/5021 (still draft, will try to complete by tomorrow)

Mujib-Ahasan:
Copilot telling the reason is this:
> The failing event tests are a race: your test starts events in the background then runs a short-lived container (no -d / no long-lived command). the container often starts and exits so quickly that the events subscriber misses the "start" event (only sees "die"/"destroy"), causing the expectation "Status":"start" to fail. Fix the tests so the container stays running (or wait for events subscription readiness) and the test will reliably observe the "start" event. 

so what I did: $ git diff
```
-       helpers.Ensure("run", "--rm", "--label", data.Labels().Get("containerLabel"), testutil.CommonImage)
+       helpers.Ensure(
+               "run",
+               "-d",
+               "--name", data.Identifier(),
+               "--label", data.Labels().Get("containerLabel"),
+               testutil.CommonImage,
+               "tail", "-f", "/dev/null",
+       )
+       helpers.Ensure("rm", "-f", data.Identifier())
```

this gives container enough time to live. Run test, then cleanup.

@AkihiroSuda is this a right direction I am following?

AkihiroSuda:
> this gives container enough time to live. Run test, then cleanup.

Don't you need some `time.Sleep` between `run` and `rm`?

Mujib-Ahasan:
I think the test is failing because it iss running against Docker host rather than nerdctl. Would it make sense to add `Require: require.Not(nerdtest.Docker)` so this test only runs for nerdctl?

AkihiroSuda:
> I think the test is failing because it iss running against Docker host rather than nerdctl. Would it make sense to add `Require: require.Not(nerdtest.Docker)` so this test only runs for nerdctl?

Any reason not to make it Docker-compatible?

Mujib-Ahasan:
> Any reason not to make it Docker-compatible?

I apologize!😅

I think I managed to solve that particular test!

AkihiroSuda (approved):
Thanks

PR #5009, opened 2026-06-25
logging: write container output synchronously so final output  is not truncated

### LINKED ISSUE #5006: gomodjail CI failing (`TestLogs*`)
```
=== Failing tests ===
TestLoadStdinFromPipe
TestLogsAfterRestartingContainer
TestLogsAfterRestartingContainer/logs_-f_works
TestLogsAfterRestartingContainer/logs_-f_works_after_restart
TestLogsFollowNoExtraneousLineFeed
TestLogsWithoutNewlineOrEOF
```

Apparently since https://github.com/containerd/nerdctl/commit/2b47b0cac4631ec4018d58efde14e4b1a7c1d289

AkihiroSuda:
Logging tests such as TestLogsWithoutNewlineOrEOF and
TestLogsAfterRestartingContainer flake under gomodjail: a container's final
output, in particular a line emitted right before exit with no trailing
newline, is sometimes dropped from the log. Running

    nerdctl run --name c alpine printf "'Hello World!\nThere is no newline'"
    nerdctl logs -f c

would intermittently print only "'Hello World!" instead of the full output.

Two independent problems, both reproduced locally under gomodjail:

1. getContainerWait hung for short-lived containers. The logging process is
   started by containerd while it sets up the container's IO, before the task
   is created, so the first con.Task() returns NotFound and the code retried
   forever "waiting for the task to start". For a fast container the task can
   instead have already exited and been removed before the logger ever sees
   it, so it never appeared and the logger blocked forever holding the logger
   lock. It now concludes the container has exited once it is missing and the
   container has been observed producing output.

2. The container's final chunk was lost to teardown. On exit containerd closes
   the stdio FIFOs and tears the logging process down almost immediately. The
   old path read the FIFO, copied it through an io.Pipe and a bufio splitter,
   and handed each line to the driver over a buffered channel; a trailing
   chunk with no newline was held in the splitter until EOF and then raced the
   teardown across several goroutines, so it was frequently lost. The logger
   now reads each FIFO directly and, for drivers that can write synchronously
   (json-file, via the new SyncDriver interface), writes each entry inline from
   the reading goroutine and flushes a trailing no-newline fragment as soon as
   it is read. Streaming drivers keep using the buffered channel so a slow
   driver cannot block the container.

The viewer also does a final read of the JSON log file when it receives the
stop signal, so entries flushed just before exit are not missed.

Verified locally with the gomodjail-packed binary: 250+ iterations of the
failing printf case, the restart (doubled-output) case, multi-line output and
follow-on-running-container all pass with no truncation.

Fixes https://github.com/containerd/nerdctl/issues/5006

Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/pkg/logging/json_logger.go
+++ b/pkg/logging/json_logger.go
@@ -48,8 +48,9 @@ var JSONDriverLogOpts = []string{
 }
 
 type JSONLogger struct {
-	Opts   map[string]string
-	logger *logrotate.Logger
+	Opts    map[string]string
+	logger  *logrotate.Logger
+	encoder *jsonfile.SyncEncoder
 }
 
 func JSONFileLogOptsValidate(logOptMap map[string]string) error {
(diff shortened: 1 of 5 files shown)


ChengyuZhu6 (approved):
LGTM

PR #5010, opened 2026-06-25
fix(network): match each gateway to its subnet for dual-stack

### LINKED ISSUE #4951: `nerdctl network create` fails when creating dual-stack bridge network with explicit IPv4 and IPv6 gateways
### Description

`nerdctl network create` fails when creating dual-stack bridge network with explicit IPv4 and IPv6 gateways

Run the following command:

```
nerdctl network create \
    --driver bridge \
    --ipv6 \
    --subnet 172.20.0.0/16 \
    --subnet fd00:dead:beef:1::/64 \
    --gateway 172.20.0.1 \
    --gateway fd00:dead:beef:1::1 \
    mynet-dual
```

The command fails with:

`FATA[0000] no matching subnet "172.20.0.0/16" for gateway "fd00:dead:beef:1::1`

The network should be created successfully.

The IPv4 gateway should be matched with the IPv4 subnet:

`172.20.0.1 -> 172.20.0.0/16`

and the IPv6 gateway should be matched with the IPv6 subnet:

`fd00:dead:beef:1::1 -> fd00:dead:beef:1::/64`

This configuration is valid and represents a standard dual-stack bridge network.

`nerdctl version 2.3.1`

None

### Host information

_No response_

mayur-tolexo:
A single `--gateway` gets applied to every `--subnet`, so creating a dual-stack network and passing both gateways fails — the IPv6 gateway ends up checked against the IPv4 subnet:

```
$ nerdctl network create --driver bridge --ipv6 \
    --subnet 172.20.0.0/16 --subnet fd00:dead:beef:1::/64 \
    --gateway 172.20.0.1 --gateway fd00:dead:beef:1::1 mynet-dual
FATA[0000] no matching subnet "172.20.0.0/16" for gateway "fd00:dead:beef:1::1"
```

This makes `--gateway` repeatable (`--subnet` already is) and matches each gateway to the subnet it belongs to, so the v4 gateway goes with the v4 subnet and the v6 one with the v6 subnet. A gateway that isn't inside any subnet still errors out.

Same command after the change:

```
$ nerdctl network create --driver bridge --ipv6 \
    --subnet 172.20.0.0/16 --subnet fd00:dead:beef:1::/64 \
    --gateway 172.20.0.1 --gateway fd00:dead:beef:1::1 mynet-dual
ab70041865f6e3a9bddf26b62c398179a24d4ded8d6b7907c904bbc051f7a341

$ nerdctl network inspect mynet-dual | jq ".[0].IPAM.Config"
[
  {
    "Subnet": "172.20.0.0/16",
    "Gateway": "172.20.0.1"
  },
  {
    "Subnet": "fd00:dead:beef:1::/64",
    "Gateway": "fd00:dead:beef:1::1"
  }
]
```

`--ip-range` has the same single-value-per-network limitation on dual-stack (Docker takes it per subnet too). I kept this PR to `--gateway` to stay focused on #4951; happy to do `--ip-range` the same way as a follow-up.

Closes #4951.

### DIFF
--- a/cmd/nerdctl/network/network_create.go
+++ b/cmd/nerdctl/network/network_create.go
@@ -47,7 +47,7 @@ func createCommand() *cobra.Command {
 	cmd.RegisterFlagCompletionFunc("ipam-driver", completion.IPAMDrivers)
 	cmd.Flags().StringArray("ipam-opt", nil, "Set IPAM driver specific options")
 	cmd.Flags().StringArray("subnet", nil, `Subnet in CIDR format that represents a network segment, e.g. "10.5.0.0/16"`)
-	cmd.Flags().String("gateway", "", `Gateway for the master subnet`)
+	cmd.Flags().StringArray("gateway", nil, "IPv4 or IPv6 Gateway for the master subnet")
 	cmd.Flags().String("ip-range", "", `Allocate container ip from a sub-range`)
 	cmd.Flags().StringArray("label", nil, "Set metadata for a network")
 	cmd.Flags().Bool("ipv6", false, "Enable IPv6 networking")
@@ -84,7 +84,7 @@ func createAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	gatewayStr, err := cmd.Flags().GetString("gateway")
+	gateways, err := cmd.Flags().GetStringArray("gateway")
 	if err != nil {
 		return err
 	}
@@ -114,7 +114,7 @@ func createAction(cmd *cobra.Command, args []string) error {
 		IPAMDriver:  ipamDriver,
 		IPAMOptions: strutil.ConvertKVStringsToMap(ipamOpts),
 		Subnets:     subnets,
-		Gateway:     gatewayStr,
+		Gateway:     gateways,
 		IPRange:     ipRangeStr,
 		Labels:      labels,
 		IPv6:        ipv6,
--- a/cmd/nerdctl/network/network_create_linux_test.go
+++ b/cmd/nerdctl/network/network_create_linux_test.go
@@ -110,6 +110,41 @@ func TestNetworkCreate(t *testing.T) {
 			},
 		},
 		{
+			Description: "dual-stack with explicit gateways",
+			Require:     nerdtest.OnlyIPv6,
+			Setup: func(data test.Data, helpers test.Helpers) {
+				// Before the fix the IPv6 gateway was checked against the IPv4
+				// subnet and creation failed.
+				helpers.Ensure("network", "create", data.Identifier(),
+					"--ipv6",
+					"--subnet", "10.5.0.0/16",
+					"--subnet", "2001:db8:5::/64",
+					"--gateway", "10.5.0.1",
+					"--gateway", "2001:db8:5::1",
+				)
+			},
+			Cleanup: func(data test.Data, helpers test.Helpers) {
+				helpers.Anyhow("network", "rm", data.Identifier())
+			},
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("network", "inspect", data.Identifier())
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: expect.ExitCodeSuccess,
+					Output: func(stdout string, t tig.T) {
+						netw := nerdtest.InspectNetwork(helpers, data.Identifier())
+						gateways := map[string]string{}
+						for _, c := range netw.IPAM.Config {
+							gateways[c.Subnet] = c.Gateway
+						}
+						assert.Equal(t, gateways["10.5.0.0/16"], "10.5.0.1")
+						assert.Equal(t, gateways["2001:db8:5::/64"], "2001:db8:5::1")
+					},
+				}
+			},
+		},
+		{
 			Description: "internal enabled",
 			Setup: func(data test.Data, helpers test.Helpers) {
 				helpers.Ensure("network", "create", "--internal", data.Identifier())
--- a/pkg/netutil/netutil_unix.go
+++ b/pkg/netutil/netutil_unix.go
@@ -215,7 +215,7 @@ func (e *CNIEnv) generateCNIPlugins(driver string, name string, ipam map[string]
 	return plugins, nil
 }
 
-func (e *CNIEnv) generateIPAM(driver string, subnets []string, gatewayStr, ipRangeStr string, opts map[string]string, ipv6 bool, internal bool) (map[string]interface{}, error) {
+func (e *CNIEnv) generateIPAM(driver string, subnets []string, gateways []string, ipRangeStr string, opts map[string]string, ipv6 bool, internal bool) (map[string]interface{}, error) {
 	var ipamConfig interface{}
 	switch driver {
 	case "default", "host-local":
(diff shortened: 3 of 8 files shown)


mayur-tolexo:
Ran the integration test locally (containerd v1.7.24, linux/arm64). The new dual-stack subtest plus the existing `network create` cases pass:

```
$ go test ./cmd/nerdctl/network/ -run "^TestNetworkCreate$" -count=1 -v
=== RUN   TestNetworkCreate
=== RUN   TestNetworkCreate/vanilla
=== RUN   TestNetworkCreate/with_MTU
=== RUN   TestNetworkCreate/with_ipv6
=== RUN   TestNetworkCreate/dual-stack_with_explicit_gateways
=== RUN   TestNetworkCreate/internal_enabled
--- PASS: TestNetworkCreate (0.00s)
PASS
ok  	github.com/containerd/nerdctl/v2/cmd/nerdctl/network	1.139s
```

mayur-tolexo:
Checked this against Docker (29.4.0) to make sure the dual-stack behaviour lines up.

Docker:

```
$ docker network create --driver bridge --ipv6 \
    --subnet 10.83.0.0/16 --subnet fd00:cafe:1234::/64 \
    --gateway 10.83.0.1 --gateway fd00:cafe:1234::1 dt2
6d450fe9fbe6...
$ docker network inspect dt2 --format "{{json .IPAM.Config}}" | jq
[
  { "Subnet": "10.83.0.0/16", "Gateway": "10.83.0.1" },
  { "Subnet": "fd00:cafe:1234::/64", "Gateway": "fd00:cafe:1234::1" }
]
```

nerdctl with the same flags now gives the same result:

```
$ nerdctl network create --driver bridge --ipv6 \
    --subnet 10.83.0.0/16 --subnet fd00:cafe:1234::/64 \
    --gateway 10.83.0.1 --gateway fd00:cafe:1234::1 dt2
$ nerdctl network inspect dt2 --format "{{json .IPAM.Config}}" | jq
[
  { "Subnet": "10.83.0.0/16", "Gateway": "10.83.0.1" },
  { "Subnet": "fd00:cafe:1234::/64", "Gateway": "fd00:cafe:1234::1" }
]
```

One thing worth noting: Docker matches each gateway to whichever subnet contains it, not by position. Passing the gateways in reverse order still pairs them correctly, and this PR does the same (it is `subnet.Contains(gw)`, not index based):

```
# both docker and nerdctl, gateways given v6-first
... --subnet 10.84.0.0/16 --subnet fd00:cafe:5678::/64 \
    --gateway fd00:cafe:5678::1 --gateway 10.84.0.1
-> 10.84.0.0/16          : 10.84.0.1
   fd00:cafe:5678::/64   : fd00:cafe:5678::1
```

And a gateway that is not inside any subnet errors the same way on both:

```
docker:  no matching subnet for gateway 192.168.1.1
nerdctl: no matching subnet for gateway "192.168.1.1"
```

cmd/nerdctl/network/network_create.go
```diff
@@ -47,7 +47,7 @@ func createCommand() *cobra.Command {
 	cmd.RegisterFlagCompletionFunc("ipam-driver", completion.IPAMDrivers)
 	cmd.Flags().StringArray("ipam-opt", nil, "Set IPAM driver specific options")
 	cmd.Flags().StringArray("subnet", nil, `Subnet in CIDR format that represents a network segment, e.g. "10.5.0.0/16"`)
-	cmd.Flags().String("gateway", "", `Gateway for the master subnet`)
+	cmd.Flags().StringArray("gateway", nil, `Gateway for the subnet; repeat for a dual-stack network, e.g. "10.5.0.1" and "fd00::1"`)
```
AkihiroSuda: Is this compatible with Docker?
If yes, the description should correspond to it.

mayur-tolexo: Yes, it matches Docker. `docker network create --gateway` is repeatable and pairs each gateway with the subnet that contains it (verified on Docker 29.4.0, including reverse order and the no-matching-subnet error). I updated the flag help and the docs to Docker's wording, "IPv4 or IPv6 Gateway for the master subnet".

cmd/nerdctl/network/network_create_linux_test.go
```diff
@@ -109,6 +109,41 @@ func TestNetworkCreate(t *testing.T) {
...
+			},
+			Cleanup: func(data test.Data, helpers test.Helpers) {
+				helpers.Anyhow("network", "rm", data.Identifier())
+			},
+			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+				return helpers.Command("network", "inspect", data.Identifier())
+			},
+			Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+				return &test.Expected{
+					ExitCode: 0,
```
sathiraumesh: you can use the Tigron frameworks exit code constants

pkg/netutil/netutil_unix.go
```diff
@@ -303,12 +305,31 @@ func (e *CNIEnv) parseIPAMRanges(subnets []string, gateway, ipRange string, ipv6
 		if !findIPv4 && subnet.IP.To4() != nil {
 			findIPv4 = true
 		}
+		// Match each gateway to the subnet that contains it, so dual-stack
+		// pairs the v4 gateway with the v4 subnet and v6 with v6.
+		gateway := ""
+		for j, g := range gateways {
+			gw := net.ParseIP(g)
+			if gw == nil {
```
sathiraumesh: you could refactor the code like this for more readability

```
if gw := net.ParseIP(g); gw != nil && subnet.Contains(gw) {
    gateway, used[j] = g, true
				break
} else {
   return nil, findIPv4, fmt.Errorf("failed to parse gateway 
}
```

pkg/netutil/netutil_unix.go
```diff
@@ -303,12 +305,31 @@ func (e *CNIEnv) parseIPAMRanges(subnets []string, gateway, ipRange string, ipv6
...
+		}
 		ipamRange, err := parseIPAMRange(subnet, gateway, ipRange)
 		if err != nil {
 			return nil, findIPv4, err
 		}
 		ranges = append(ranges, []IPAMRange{*ipamRange})
 	}
+	// A gateway that matched no subnet is an error.
+	for j, g := range gateways {
+		if !used[j] {
```
sathiraumesh: why do we do this check here can't we kind of refactor this logic to the above group somehow i feel like this part is not kind of needed

pkg/netutil/netutil_windows.go
```diff
@@ -72,13 +72,19 @@ func (e *CNIEnv) generateCNIPlugins(driver string, name string, ipam map[string]
...
-func (e *CNIEnv) generateIPAM(driver string, subnets []string, gatewayStr, ipRangeStr string, opts map[string]string, ipv6 bool, internal bool) (map[string]interface{}, error) {
+func (e *CNIEnv) generateIPAM(driver string, subnets []string, gateways []string, ipRangeStr string, opts map[string]string, ipv6 bool, internal bool) (map[string]interface{}, error) {
 	switch driver {
 	case "default":
 	default:
 		return nil, fmt.Errorf("unsupported ipam driver %q", driver)
 	}
 
+	// Windows is single-subnet, so use at most one gateway.
+	gatewayStr := ""
```
sathiraumesh, line 83: is this used somewhere  ?

sathiraumesh (changes requested):

cmd/nerdctl/network/network_create_linux_test.go
mayur-tolexo: Done, using `expect.ExitCodeSuccess` now.

pkg/netutil/netutil_unix.go
mayur-tolexo: Pulled the parse out of the loop into a single up-front pass for readability. Kept it separate from the `Contains` check on purpose though: a valid gateway that is not in *this* subnet still belongs to another subnet on a dual-stack net, so an `else` error there would reject correct input.

pkg/netutil/netutil_unix.go
mayur-tolexo: This one has to stay outside the loop, a gateway only counts as "no matching subnet" once every subnet has been checked. Docker errors the same way, so I kept it and tightened the comment.

pkg/netutil/netutil_windows.go
mayur-tolexo: Yep, just below at `parseIPAMRange(subnet, gatewayStr, ipRangeStr)`. Windows is single-subnet so it takes at most the first gateway.

AkihiroSuda (approved):
Thanks

PR #5013, opened 2026-06-26
test: remove redundant test lock pre-check

immanuwell:
This pre-check bails out too early.

`testutil.M` already takes a blocking filesystem lock, but the extra `os.Stat` check can fail filtered package runs before any test even starts. It kinda defeats the lock we already have. Pretty easy to hit in practice because `go test ./cmd/nerdctl/...` runs package test binaries in parallel by default.

Repro:
`go test ./cmd/nerdctl/... -run TestNonExistent -count=0`

Before:
some `cmd/nerdctl/*` packages fail with `Another test binary is already running...`

After:
the same command passes, and concurrent invocations still serialize fine.

### DIFF
--- a/pkg/testutil/testutil.go
+++ b/pkg/testutil/testutil.go
@@ -18,7 +18,6 @@ package testutil
 
 import (
 	"encoding/json"
-	"errors"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #5014, opened 2026-06-26
refactor: migrate container_run_network_windows_test.go to nerdtest.Setup

takumi12311123:
Refactors the test cases in `cmd/nerdctl/container/container_run_network_windows_test.go` to use the Tigron framework (`nerdtest.Setup`).

`TestRunPort` is left as-is because it only delegates to `baseTestRunPort` (defined in `container_run_network_base_test.go`), which is shared with the Linux test and will be migrated together with that file.

Note: this is a Windows-only test file; verified locally with `gofmt`, `GOOS=windows go vet`, and `golangci-lint`, but test execution will be exercised by CI on a Windows runner.

related : https://github.com/containerd/nerdctl/issues/4613

### DIFF
--- a/cmd/nerdctl/container/container_run_network_windows_test.go
+++ b/cmd/nerdctl/container/container_run_network_windows_test.go
@@ -25,39 +25,24 @@ import (
 	"github.com/Microsoft/hcsshim"
 	"gotest.tools/v3/assert"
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #5019, opened 2026-06-28
fix: return inspect format errors

immanuwell:
this makes inspect return a real error when the format string is malformed or unsupported. before this, the command logged the formatter error but still exited 0, which is kinda sneaky in scripts.

repro
1. run `nerdctl image inspect -f '{{bad' alpine`
2. current main prints the template error but exits 0
3. this patch makes it exit non zero

the change only touches the final inspect formatting path. normal json output and valid templates stay the same.


ChengyuZhu6 (approved):
LGTM

Issue #5012, opened 2026-06-26
network create: option parity with docker network create


mayur-tolexo:
I was comparing `nerdctl network create` with `docker network create` (Docker 29.4.0) and figured it is worth tracking the gaps in one place. Docker reference for these flags: https://docs.docker.com/reference/cli/docker/network/create/

Most of it already lines up — `--driver`, `--opt`, `--ipam-driver`, `--ipam-opt`, `--subnet`, `--gateway`, `--label`, `--internal` and `--ipv6` all behave like Docker.

A few things differ or are missing:

- `--ip-range` only takes a single value and applies it to every subnet, whereas Docker takes it per subnet (it is repeatable). So on a dual-stack network the IPv4 range ends up being checked against the IPv6 subnet and creation fails. Same shape as the `--gateway` problem fixed in #5010, and the fix would be the same: make it repeatable and match each range to the subnet that contains it.
- There is no `--ipv4`. Docker uses it to disable IPv4 so you can create an IPv6-only network; nerdctl has no equivalent.
- `--aux-address` (per-subnet auxiliary addresses) is not implemented.

Happy to work on these - `--ip-range`, `--ipv4` and `--aux-address` - if they sound reasonable.

AkihiroSuda:
> Happy to work on these - --ip-range, --ipv4 and --aux-address - if they sound reasonable.

Yes, thanks.
v6-only networking may need some changes on CNI plugins too?

mayur-tolexo:
Thanks!

I went through `pkg/netutil` to see how much of v6-only is CNI vs nerdctl, and it looks mostly nerdctl-side — the standard plugins (bridge, host-local, portmap, firewall) already handle IPv6, so I don't think they need code changes:

- `generateIPAM` always appends a default IPv4 range when no v4 subnet is given (the `!findIPv4` branch), so today every bridge network gets IPv4. `--ipv4=false` means skipping that and requiring/defaulting a v6 subnet instead.
- The host-local default route is hardcoded to `0.0.0.0/0`; for a v6-only network it has to become `::/0`, otherwise host-local adds a v4 default route with no matching gateway.

The plugin side is more about runtime prerequisites than code — the host needs `net.ipv6.conf.all.forwarding=1`, and IPv6 masquerade/portmap only kick in with reasonably recent containernetworking/plugins (ip6tables). I'll check the end-to-end path against a current plugins build as part of this.

[...]

PR #5024, opened 2026-06-30
feat(network): add --ipv4 to allow IPv6-only networks

mayur-tolexo:
Part of #5012. Docker's `network create` has `--ipv4` (on by default) so you can pass `--ipv4=false` with `--ipv6` to get an IPv6-only network. nerdctl had no equivalent — `generateIPAM` always appends a default IPv4 range when no v4 subnet is given, so every bridge network ends up with IPv4.

This adds the flag. It's carried through as `DisableIPv4` (the inverse) so the zero value keeps IPv4 enabled and nothing changes for existing callers, including the default bridge network. When IPv4 is disabled the default v4 range is skipped, the host-local default route becomes `::/0` instead of `0.0.0.0/0`, and an IPv4 subnet is rejected. It requires `--ipv6` and an explicit IPv6 subnet, and is rejected on Windows where IPv6-only isn't supported.

One difference from Docker worth calling out: with `--ipv6 --ipv4=false` and no subnet, Docker auto-assigns a ULA /64. nerdctl doesn't allocate IPv6 subnets anywhere today — even a plain `--ipv6` without a subnet only gets the default IPv4 range — so this errors and asks for an explicit `--subnet` instead. Auto-allocating a ULA range is worth doing on its own since it would apply to the dual-stack case too.

Verified against Docker 29.4 locally: same addresses for an IPv6-only network, same rejection without `--ipv6`. Logs in a comment below.

### DIFF
--- a/cmd/nerdctl/network/network_create.go
+++ b/cmd/nerdctl/network/network_create.go
@@ -50,6 +50,7 @@ func createCommand() *cobra.Command {
 	cmd.Flags().StringArray("gateway", nil, "IPv4 or IPv6 Gateway for the master subnet")
 	cmd.Flags().StringArray("ip-range", nil, `Allocate container ip from a sub-range`)
 	cmd.Flags().StringArray("label", nil, "Set metadata for a network")
+	cmd.Flags().Bool("ipv4", true, "Enable IPv4 networking (set to false together with --ipv6 for an IPv6-only network)")
 	cmd.Flags().Bool("ipv6", false, "Enable IPv6 networking")
 	cmd.Flags().Bool("internal", false, "Restrict external access to the network")
 	return cmd
@@ -97,6 +98,10 @@ func createAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	labels = strutil.DedupeStrSlice(labels)
+	ipv4, err := cmd.Flags().GetBool("ipv4")
+	if err != nil {
+		return err
+	}
 	ipv6, err := cmd.Flags().GetBool("ipv6")
 	if err != nil {
 		return err
@@ -118,6 +123,7 @@ func createAction(cmd *cobra.Command, args []string) error {
 		IPRange:     ipRanges,
 		Labels:      labels,
 		IPv6:        ipv6,
+		IPv4:        &ipv4,
 		Internal:    internal,
 	}, cmd.OutOrStdout())
 }
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -1280,6 +1280,7 @@ Flags:
 - :whale: `--gateway`: IPv4 or IPv6 Gateway for the master subnet
 - :whale: `--ip-range`: Allocate container ip from a sub-range
 - :whale: `--label`: Set metadata on a network
+- :whale: `--ipv4`: Enable IPv4. Enabled by default; set to false with `--ipv6` and an IPv6 subnet for an IPv6-only network. `--ipv4=false` is not supported on Windows.
 - :whale: `--ipv6`: Enable IPv6. Should be used with a valid subnet.
 - :whale: `--internal`: Restrict external access to the network.
 
--- a/pkg/api/types/network_types.go
+++ b/pkg/api/types/network_types.go
@@ -35,7 +35,11 @@ type NetworkCreateOptions struct {
 	IPRange     []string
 	Labels      []string
 	IPv6        bool
-	Internal    bool
+	// IPv4 enables IPv4 on the network. A nil value defaults to enabled, so a
+	// directly-constructed NetworkCreateOptions keeps IPv4 on; setting it to
+	// false together with IPv6 yields an IPv6-only network.
+	IPv4     *bool
+	Internal bool
 }
 
 // NetworkInspectOptions specifies options for `nerdctl network inspect`.
--- a/pkg/cmd/network/create.go
+++ b/pkg/cmd/network/create.go
@@ -27,6 +27,19 @@ import (
 )
 
 func Create(options types.NetworkCreateOptions, stdout io.Writer) error {
+	// A nil IPv4 defaults to enabled.
+	ipv4 := options.IPv4 == nil || *options.IPv4
+	// At least one address family must be enabled, matching docker which
+	// rejects a network with both IPv4 and IPv6 turned off.
+	if !ipv4 && !options.IPv6 {
+		return fmt.Errorf("IPv4 or IPv6 must be enabled")
+	}
+	if !ipv4 && len(options.Subnets) == 0 {
+		// IPv6-only needs a concrete IPv6 subnet: unlike docker, nerdctl does
+		// not auto-allocate one, and the empty-subnet default below would pick
+		// an IPv4 range, contradicting the disabled IPv4.
+		return fmt.Errorf("IPv6-only network requires an IPv6 subnet, specify --subnet manually")
+	}
 	if len(options.Subnets) == 0 {
 		if len(options.Gateway) > 0 || len(options.IPRange) > 0 {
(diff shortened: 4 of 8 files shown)


mayur-tolexo:
Ran it side by side with Docker 29.4 to confirm the IPv6-only behaviour lines up. In both cases eth0 comes up with only an `inet6` address (no `inet`), out of the subnet that was passed, and disabling IPv4 without `--ipv6` is rejected.

Docker:

```
$ docker network create --ipv6 --ipv4=false --subnet fd00:dead::/64 v6demo
$ docker run --rm --net v6demo alpine ip -o addr show eth0
eth0    inet6 fd00:dead::2/64 scope global
eth0    inet6 fe80::eceb:f3ff:fe52:137c/64 scope link tentative

$ docker network create --ipv4=false noipv6
Error response from daemon: IPv4 or IPv6 must be enabled
```

nerdctl with this PR:

```
$ nerdctl network create --ipv6 --ipv4=false --subnet fd00:dead::/64 v6demo
$ nerdctl run --rm --net v6demo alpine ip -o addr show eth0
eth0    inet6 fd00:dead::2/64 scope global
eth0    inet6 fe80::4cf4:47ff:fecf:5beb/64 scope link

$ nerdctl network create --ipv4=false noipv6
IPv4 or IPv6 must be enabled
```

The host-local config nerdctl writes for the network is IPv6-only, with the v6 default route and no IPv4 range:

```
routes: [{"dst": "::/0"}]
ranges: [[{"gateway": "fd00:dead::1", "subnet": "fd00:dead::/64"}]]
```

And the other guard rails, matching what you'd expect:

```
$ nerdctl network create --ipv6 --ipv4=false noipsubnet
IPv6-only network requires an IPv6 subnet, specify --subnet manually

$ nerdctl network create --ipv6 --ipv4=false --subnet 10.9.0.0/16 v4subnet
--ipv4=false conflicts with an IPv4 subnet
```

cmd/nerdctl/network/network_create.go
```diff
@@ -118,6 +123,7 @@ func createAction(cmd *cobra.Command, args []string) error {
 		IPRange:     ipRangeStr,
 		Labels:      labels,
 		IPv6:        ipv6,
+		DisableIPv4: !ipv4,
```
sathiraumesh: lets not introduce it as `DisableIPv4` lets introduce it as `IPv4` matches the existing implementation

pkg/cmd/network/create.go
```diff
@@ -27,6 +27,17 @@ import (
 )
 
 func Create(options types.NetworkCreateOptions, stdout io.Writer) error {
+	if options.DisableIPv4 {
```
sathiraumesh: Then here check for mutual exclusion here with IPv4 and IPv6 flags

sathiraumesh (changes requested):

cmd/nerdctl/network/network_create.go
mayur-tolexo: Done. Had to set `IPv4: true` in `createDefaultNetworkConfig` too, else the default network's zero-value bool leaves IPv4 off.

pkg/cmd/network/create.go
mayur-tolexo: Moved it here and matched docker's `IPv4 or IPv6 must be enabled`.

pkg/api/types/network_types.go
```diff
@@ -35,7 +35,11 @@ type NetworkCreateOptions struct {
 	IPRange     string
 	Labels      []string
 	IPv6        bool
-	Internal    bool
+	// IPv4 enables IPv4 on the network. It defaults to on; setting it false
+	// together with IPv6 yields an IPv6-only network. Callers constructing this
+	// struct directly must set it, since the zero value means IPv4 disabled.
+	IPv4     bool
+	Internal bool
```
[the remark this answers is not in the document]
mayur-tolexo: Keeping it as `IPv4` to match the existing `IPv6` field (and the ask above). Reworded the comment so it's clear the default-on lives on the CLI flag, and the zero value is false for anyone building the struct directly.

pkg/cmd/network/create.go
```diff
@@ -27,6 +27,17 @@ import (
 )
 
 func Create(options types.NetworkCreateOptions, stdout io.Writer) error {
+	// At least one address family must be enabled, matching docker which
+	// rejects a network with both IPv4 and IPv6 turned off.
+	if !options.IPv4 && !options.IPv6 {
+		return fmt.Errorf("IPv4 or IPv6 must be enabled")
+	}
```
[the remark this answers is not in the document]
mayur-tolexo: Left as-is to match docker's exact string; it starts with the `IPv4` initialism so ST1005 doesn't flag it.

pkg/cmd/network/create.go
```diff
@@ -27,6 +27,17 @@ import (
...
+	if !options.IPv4 && len(options.Subnets) == 0 {
+		// IPv6-only needs a concrete IPv6 subnet: unlike docker, nerdctl does
+		// not auto-allocate one, and the empty-subnet default below would pick
+		// an IPv4 range, contradicting the disabled IPv4.
+		return fmt.Errorf("IPv6-only network requires an IPv6 subnet, specify --subnet manually")
+	}
```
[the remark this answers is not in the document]
mayur-tolexo: Same — starts with `IPv6`, so the lowercase-error rule is fine with it.

cmd/nerdctl/network/network_create.go
```diff
@@ -50,6 +50,7 @@ func createCommand() *cobra.Command {
 	cmd.Flags().StringArray("gateway", nil, "IPv4 or IPv6 Gateway for the master subnet")
 	cmd.Flags().String("ip-range", "", `Allocate container ip from a sub-range`)
 	cmd.Flags().StringArray("label", nil, "Set metadata for a network")
+	cmd.Flags().Bool("ipv4", true, "Enable IPv4 networking (set to false together with --ipv6 for an IPv6-only network)")
```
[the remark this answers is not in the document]
mayur-tolexo: Put the Windows note in the command reference instead; the help strings here stay terse and don't spell out per-OS caveats.

docs/command-reference.md
```diff
@@ -1268,6 +1268,7 @@ Flags:
 - :whale: `--gateway`: IPv4 or IPv6 Gateway for the master subnet
 - :whale: `--ip-range`: Allocate container ip from a sub-range
 - :whale: `--label`: Set metadata on a network
+- :whale: `--ipv4`: Enable IPv4. Enabled by default; set to false with `--ipv6` and an IPv6 subnet for an IPv6-only network.
```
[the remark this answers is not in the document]
mayur-tolexo: Done.

pkg/api/types/network_types.go
```diff
@@ -35,7 +35,11 @@ type NetworkCreateOptions struct {
 	IPRange     string
 	Labels      []string
 	IPv6        bool
-	Internal    bool
+	// IPv4 enables IPv4 on the network. The --ipv4 CLI flag defaults it to on;
+	// setting it false together with IPv6 yields an IPv6-only network. The zero
+	// value is false, so callers constructing this struct directly must set it.
+	IPv4     bool
```
AkihiroSuda: Alternatively you can make it `*bool` and say `nil` defaults to `true`

mayur-tolexo: Good call, switched it to `*bool`. nil now means enabled, so a directly-constructed NetworkCreateOptions keeps IPv4 on and only an explicit false turns it off. Dropped the `IPv4: true` I had on the default bridge network too since it is not needed anymore.

AkihiroSuda (approved):
Thanks

PR #5025, opened 2026-06-30
fix: hide internal alias flags from help output

immanuwell:
small help output papercut.

right now `nerdctl --help` shows internal alias shim flags like `-a, --a`, `-H, --H`, `-n, --n`. same thing in `nerdctl compose --help` with `-f, --f`. looks a bit weird, and not really what users should see.

this hides those shim flags from help, and shows aliases on the real flag line instead. alias parsing still works, so no behavior change there.

repro
```bash
go build ./cmd/nerdctl
./nerdctl --help
./nerdctl compose --help
```

before
```text
-a, --a string
-H, --H string
-n, --n string
-f, --f stringArray
```

after
```text
--address string ... (aliases: -a, -H, --host)
--namespace string ... (aliases: -n)
--file stringArray ... (aliases: -f)
```

checks
```bash
go test ./cmd/nerdctl -run 'TestRootHelpHidesAliasImplementationFlags|TestRootHiddenAliasesStillParse'
go test ./cmd/nerdctl/compose -run 'TestComposeHelpHidesAliasImplementationFlags|TestComposeHiddenFileAliasStillParses'
```

### DIFF
--- a/cmd/nerdctl/compose/compose_help_test.go
+++ b/cmd/nerdctl/compose/compose_help_test.go
@@ -1,0 +1,62 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

PR #5054, opened 2026-07-08
save: add --quiet/-q to suppress progress output

s3onghyun:
Fixes #4958.

`nerdctl save` always writes a progress bar to stderr, which clutters logs when it is used in scripts and CI. This adds a `--quiet`/`-q` flag that suppresses the progress output.

The implementation mirrors the existing `--quiet` flag on `nerdctl load`: the flag is threaded into `ImageSaveOptions.Quiet`, and when set, the transfer progress handler writes to `io.Discard` instead of `os.Stderr`. The archive itself is written exactly as before; only the progress output is suppressed.

- `cmd/nerdctl/image/image_save.go`: register `-q, --quiet` and pass it into the options
- `pkg/api/types/image_types.go`: add `Quiet` to `ImageSaveOptions`
- `pkg/cmd/image/save.go`: route progress to `io.Discard` when quiet
- `docs/command-reference.md`: document the flag
- `cmd/nerdctl/image/image_save_test.go`: add `TestSaveQuiet` verifying the archive is still produced correctly with `--quiet`

Manual check: `nerdctl save --help` now lists `-q, --quiet   Suppress the progress output`.

PR #5064, opened 2026-07-13
save: add --quiet/-q to suppress progress output

### LINKED ISSUE #4958: `nerdctl save` --quiet and/or --progress plain

When using `nerdctl save` in scripts the output logs get filled with progress bars.

Adding the same `-q` and/or `--progress=plain` that are available in at least (the build command)[https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-build].

Ideally this would be a globally available flag, but that's asking a lot more

### Additional context

_No response_

s3onghyun:
Fixes #4958.

> Re-opening [#5054](https://github.com/containerd/nerdctl/pull/5054), which I accidentally closed with a force-push. This is the same single squashed commit (`expect.ExitCodeSuccess` review suggestion from @sathiraumesh already applied). Thanks @AkihiroSuda.

`nerdctl save` always writes a progress bar to stderr, which clutters logs when it is used in scripts and CI. This adds a `--quiet`/`-q` flag that suppresses the progress output.

The implementation mirrors the existing `--quiet` flag on `nerdctl load`: the flag is threaded into `ImageSaveOptions.Quiet`, and when set, the transfer progress handler writes to `io.Discard` instead of `os.Stderr`. The archive itself is written exactly as before; only the progress output is suppressed.

- `cmd/nerdctl/image/image_save.go`: register `-q, --quiet` and pass it into the options
- `pkg/api/types/image_types.go`: add `Quiet` to `ImageSaveOptions`
- `pkg/cmd/image/save.go`: route progress to `io.Discard` when quiet
- `docs/command-reference.md`: document the flag
- `cmd/nerdctl/image/image_save_test.go`: add `TestSaveQuiet` verifying the archive is still produced correctly with `--quiet`

Manual check: `nerdctl save --help` now lists `-q, --quiet   Suppress the progress output`.

### DIFF
--- a/cmd/nerdctl/image/image_save.go
+++ b/cmd/nerdctl/image/image_save.go
@@ -45,6 +45,7 @@ func SaveCommand() *cobra.Command {
 		SilenceErrors:     true,
 	}
 	cmd.Flags().StringP("output", "o", "", "Write to a file, instead of STDOUT")
+	cmd.Flags().BoolP("quiet", "q", false, "Suppress the progress output")
 
 	// #region platform flags
 	// platform is defined as StringSlice, not StringArray, to allow specifying "--platform=amd64,arm64"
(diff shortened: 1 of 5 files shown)


AkihiroSuda (approved):
Thanks

PR #5067, opened 2026-07-13
update: fix --cpus writing cpuset instead of quota/period

### LINKED ISSUE #5066: nerdctl update --cpus does not apply the CPU limit (writes cpuset instead of quota/period)
While working on a [terraform provider for nerdctl][1] it seems claude AI found an error which I'll provide below. I'll let the AI tell the story but I did verify I can reproduce the error on my local test host running Debian 13.

## Description:

`nerdctl update --cpus <n> CONTAINER` exits 0 but the CPU limit is not changed: neither the persisted OCI spec nor the running task's resources reflect the new value. `--memory` and `--restart` in the same invocation work; `--cpu-quota`/`--cpu-period` also work.

## Reproduce (v2.3.4, also present on main):

```
nerdctl run -d --name test --cpus 0.25 nginx:alpine
nerdctl update --cpus 0.5 test
nerdctl inspect --format '{{.HostConfig.CPUQuota}}' test   # still 25000, expected 50000
```

## Cause:

In `cmd/nerdctl/container/container_update.go`, --cpus is converted to cpuQuota/cpuPeriod in processUpdateCommandFlag, but the spec-assembly section only writes CPU.Quota/CPU.Period under cmd.Flags().Changed("cpu-quota") / Changed("cpu-period") guards. The Changed("cpus") block instead assigns opts.CpusetCpus to spec.Linux.Resources.CPU.Cpus — an apparent copy-paste of the Changed("cpuset-cpus") block — so the computed quota is dropped.

#### Additional note

The surrounding guards compare pointers against local addresses (e.g. spec.Linux.Resources.CPU.Quota != &opts.CPUQuota), which is always true; presumably value comparisons were intended.

Workaround: pass --cpu-quota and --cpu-period directly as whole ints instead of floats.

[1]: https://github.com/kmaris/terraform-provider-nerdctl

kmaris:
The --cpus flag was incorrectly assigning opts.CpusetCpus to the CPU Cpus (cpuset) field. Persist the converted quota and period instead, matching how --cpu-quota and --cpu-period are applied.

Closes #5066

### DIFF
--- a/cmd/nerdctl/container/container_update_linux_test.go
+++ b/cmd/nerdctl/container/container_update_linux_test.go
@@ -30,7 +30,6 @@ import (
 
 func TestUpdateContainer(t *testing.T) {
 	testCase := nerdtest.Setup()
-	testCase.Require = require.Not(nerdtest.Docker)
 
 	testCase.Setup = func(data test.Data, helpers test.Helpers) {
 		containerName := testutil.Identifier(t)
@@ -48,6 +47,7 @@ func TestUpdateContainer(t *testing.T) {
 		{
 			Description: "should fail on unsupported restart policy value",
 			NoParallel:  true,
+			Require:     require.Not(nerdtest.Docker),
 			Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
 				containerName := data.Labels().Get("containerName")
 				return helpers.Command("update", "--memory", "999999999", "--restart", "123", containerName)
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
https://github.com/apps/dco

cmd/nerdctl/container/container_update_linux_test.go
AkihiroSuda: The test should not be skipped on Docker?

instead of `inspect --mode=native`, `run IMAGE cat /sys/fs/cgroup/FILE` could be used?

kmaris: I updated the subtests so that the cpu update will run and check the cpu.max cgroup file. Let me know if you'd like further changes, happy to help.

AkihiroSuda:
Please squash the commits, then LGTM

AkihiroSuda (approved):
Thanks

PR #5073, opened 2026-07-16
compose: guard x-nerdctl-* extension type assertions

arpitjain099:
The compose image verify, pull, and push paths pull `x-nerdctl-*` extension values out of the parsed compose file and assert them to `string` without checking the type first. The `ok` on those lookups only tells you the key is present, not that the value is a string.

compose-go accepts any value under an `x-*` key, so a malformed entry like this loads without error:

```yaml
services:
  app:
    image: alpine:latest
    x-nerdctl-verify: 123
```

and then `nerdctl compose up` panics on it:

```
panic: interface conversion: interface {} is int, not string
  .../pkg/cmd/compose/compose.go:166 imageVerifyOptionsFromCompose
```

So a typo in a compose file crashes the CLI with a stack trace instead of the value being ignored or reported. This is a robustness fix, not a security issue: it's the user crashing their own CLI on their own malformed file.

The change switches every extension read to the comma-ok form so a non-string value falls back to the default (`provider=none` for verify, skip the flag for the cosign fields) rather than panicking. Same treatment across all three sites:

- `pkg/cmd/compose/compose.go` (`imageVerifyOptionsFromCompose`, compose up), 6 fields
- `pkg/composer/pull.go` (compose pull), 6 fields
- `pkg/composer/push.go` (compose push), 2 fields

Added a regression test in `pkg/cmd/compose` that writes a compose file with an int-valued `x-nerdctl-verify`, loads it through the real compose-go loader (`testutil.LoadProject` -> `serviceparser.Parse`, the same path production uses), calls `imageVerifyOptionsFromCompose`, and asserts no panic with the provider defaulting to `none`. It panics at compose.go:166 before the fix and passes after.

```
go test ./pkg/cmd/compose/...
```

### DIFF
--- a/pkg/cmd/compose/compose.go
+++ b/pkg/cmd/compose/compose.go
@@ -162,27 +162,27 @@ func New(client *containerd.Client, globalOptions types.GlobalCommandOptions, op
 
 func imageVerifyOptionsFromCompose(ps *serviceparser.Service) types.ImageVerifyOptions {
 	var opt types.ImageVerifyOptions
-	if verifier, ok := ps.Unparsed.Extensions[serviceparser.ComposeVerify]; ok {
-		opt.Provider = verifier.(string)
+	if verifier, ok := ps.Unparsed.Extensions[serviceparser.ComposeVerify].(string); ok {
+		opt.Provider = verifier
 	} else {
 		opt.Provider = "none"
 	}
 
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):
Thanks

PR #5041, opened 2026-07-04
feat(network): support --aux-address on network create

mayur-tolexo:
Part of https://github.com/containerd/nerdctl/issues/5012.
Docker's network create reserves auxiliary addresses so IPAM never hands them out to containers. nerdctl had no equivalent.

host-local has no exclude list, but it allocates across every range in a set, so each reserved IP is carved out by splitting the subnet range around it. The reserved name=IP pairs are matched to the subnet that contains them (so dual-stack picks the right family), rejected when they hit the network or gateway address or fall outside every subnet, and recorded so network inspect reports AuxiliaryAddresses the same way Docker does.

PR #5078, opened 2026-07-17
fix(network): stop injecting ipRange into the CNI config

### LINKED ISSUE #5068: `nerdctl network create --ip-range` injects a non-existent property `ipRange` to CNI IPAM config
https://github.com/containerd/nerdctl/blob/cb8fa6c3de73a98324530cc20950ac5cdf048f7e/pkg/netutil/cni_plugin.go#L31-L37

`nerdctl network --ip-range` injects `ipRange` property in `.plugins[0].ipam.ranges[][]`, but this property does not seem supported by the CNI plugin.

https://github.com/containernetworking/plugins/blob/v1.9.1/plugins/ipam/host-local/backend/allocator/config.go#L68-L73

`ipRange` could be dynamically computed from `rangeStart` and `rangeEnd`.

- - -

Reproducer:

```console
$ nerdctl network create  --driver=bridge --subnet=172.28.0.0/16 --ip-range=172.28.5.0/24   --gateway=172.28.5.254 br0

$ cat /etc/cni/net.d/default/nerdctl-br0.conflist 
{
  "cniVersion": "1.0.0",
  "name": "br0",
  "nerdctlID": "0f1575ab177c63b4a9de71c81aaf76cd2680bc20fa9b0ba6444e14e5303803e8",
  "plugins": [
    {
      "type": "bridge",
      "bridge": "br-0f1575ab177c",
      "isGateway": true,
      "ipMasq": true,
      "hairpinMode": true,
      "ipam": {
        "ranges": [
          [
            {
              "gateway": "172.28.5.254",
              "ipRange": "172.28.5.0/24",
              "rangeEnd": "172.28.5.255",
              "rangeStart": "172.28.5.1",
              "subnet": "172.28.0.0/16"
            }
          ]
        ],
        "routes": [
          {
            "dst": "0.0.0.0/0"
          }
        ],
        "type": "host-local"
      }
    },
    {
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      }
    },
    {
      "type": "firewall",
      "ingressPolicy": "same-bridge"
    },
    {
      "type": "tuning"
    }
  ]
}
```

mayur-tolexo:
Fixes #5068.

`network create --ip-range` was writing an `ipRange` property into the host-local IPAM ranges on disk, but host-local has no such field — it only reads `subnet`/`rangeStart`/`rangeEnd`/`gateway` and silently ignores anything else. So the property never affected allocation; it was only being read back by `network inspect` (via case-insensitive JSON) to report Docker's `IPRange`.

This drops the field from the CNI config and instead recomputes the range CIDR from `rangeStart`/`rangeEnd` at inspect time. `FirstIPInSubnet` only bumps the network's last byte and `LastIPInSubnet` gives the broadcast, so the bounds invert cleanly back to the original CIDR. `network inspect` still reports the same `IPRange` as before and as Docker — the existing inspect e2e that asserts it is unchanged, plus a unit test covers the reconstruction across v4/v6 prefixes.

Same class of issue as `--aux-address` in #5041; there the reserved pairs aren't derivable so they move to a nerdctl label, whereas `ipRange` is fully derivable and needs no storage.

Keeping this a draft until #5041 settles, since both touch the same IPAM code and I'll rebase whichever lands second.

### DIFF
--- a/pkg/inspecttypes/dockercompat/dockercompat.go
+++ b/pkg/inspecttypes/dockercompat/dockercompat.go
@@ -49,6 +49,7 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/native"
 	"github.com/containerd/nerdctl/v2/pkg/ipcutil"
 	"github.com/containerd/nerdctl/v2/pkg/labels"
+	subnetutil "github.com/containerd/nerdctl/v2/pkg/netutil/subnet"
 	"github.com/containerd/nerdctl/v2/pkg/ocihook/state"
 )
 
@@ -1086,9 +1087,18 @@ type structuredCNI struct {
 	Name    string `json:"name"`
 	Plugins []struct {
 		Ipam struct {
-			Ranges [][]IPAMConfig `json:"ranges"`
+			Ranges [][]cniIPAMRange `json:"ranges"`
 		} `json:"ipam"`
 	} `json:"plugins"`
+}
+
+// cniIPAMRange is the on-disk host-local range. Its bounds let inspect recompute
+// the ip-range CIDR, which host-local has no field for.
+type cniIPAMRange struct {
+	Subnet     string `json:"subnet"`
+	Gateway    string `json:"gateway"`
+	RangeStart string `json:"rangeStart"`
+	RangeEnd   string `json:"rangeEnd"`
 }
 
 type MemorySetting struct {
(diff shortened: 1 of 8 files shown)


mayur-tolexo:
Re-verified inspect parity against Docker 29.4.0 with this branch built in a Linux/containerd sandbox. The fix drops `ipRange` from the on-disk CNI config and rebuilds it on inspect from the `rangeStart`/`rangeEnd` bounds host-local actually stores (`subnet.CIDRFromRange`, the inverse of `FirstIPInSubnet`/`LastIPInSubnet`).

nerdctl built off this branch:

```
$ nerdctl network create --subnet 172.28.0.0/16 --ip-range 172.28.5.0/24 --gateway 172.28.5.254 iprv4a
$ nerdctl network inspect iprv4a --format '{{json .IPAM.Config}}'
[{"Subnet":"172.28.0.0/16","Gateway":"172.28.5.254","IPRange":"172.28.5.0/24"}]

$ nerdctl network create --subnet 10.24.0.0/16 --ip-range 10.24.24.0/25 iprv4b
$ nerdctl network inspect iprv4b --format '{{json .IPAM.Config}}'
[{"Subnet":"10.24.0.0/16","Gateway":"10.24.0.1","IPRange":"10.24.24.0/25"}]

$ nerdctl network create --ipv6 --subnet 2001:db8::/48 --ip-range 2001:db8:0:5::/64 iprv6
$ nerdctl network inspect iprv6 --format '{{json .IPAM.Config}}'
[{"Subnet":"2001:db8::/48","Gateway":"2001:db8::1","IPRange":"2001:db8:0:5::/64"}, ...]
```

Docker 29.4.0, same inputs:

```
$ docker network inspect <net> --format '{{json .IPAM.Config}}'
[{"Subnet":"172.28.0.0/16","IPRange":"172.28.5.0/24","Gateway":"172.28.5.254"}]
[{"Subnet":"10.24.0.0/16","IPRange":"10.24.24.0/25","Gateway":"10.24.24.0"}]
[{"Subnet":"2001:db8::/48","IPRange":"2001:db8:0:5::/64","Gateway":"2001:db8:0:5::"}]
```

IPRange comes out identical to Docker in all three, including the /25 that doesn't align to the subnet prefix and the v6 /64. (The gateway still differs because nerdctl defaults it to .1 of the subnet rather than the range's network address, which is a pre-existing default unrelated to this change.)

And the actual point of the fix — the field is no longer written to the on-disk conflist, and inspect rebuilds it from the bounds host-local does store:

```
$ python3 -c 'import json;print(json.load(open("/etc/cni/net.d/nerdctl-iprv4a.conflist"))["plugins"][0]["ipam"]["ranges"])'
[[{'gateway': '172.28.5.254', 'rangeEnd': '172.28.5.255', 'rangeStart': '172.28.5.1', 'subnet': '172.28.0.0/16'}]]

$ grep -o ipRange /etc/cni/net.d/nerdctl-iprv4a.conflist
(no match)
```

rangeStart 172.28.5.1 / rangeEnd 172.28.5.255 invert back to 172.28.5.0/24, which is what inspect reports. The v6 case stores rangeEnd `2001:db8:0:5:ffff:ffff:ffff:ffff` (the full /64, nothing dropped) and reconstructs `2001:db8:0:5::/64`.

One limitation worth noting: a `/31` or `/127` ip-range collapses to a single stored address (network+1 equals the broadcast), indistinguishable from a `/32`/`/128`, so those recompute as `/32`/`/128`. There is a unit-test case pinning that; every other prefix round-trips exactly.

AkihiroSuda:
Needs rebase

AkihiroSuda (approved):
Thanks

PR #5088, opened 2026-07-20
pkg/identifiers: add tests for ValidateDockerCompat

nikolauspschuetz:
`pkg/identifiers` had no tests, even though `ValidateDockerCompat` validates container, volume, and network names throughout the codebase. This adds table-driven coverage: the accepted charset (alphanumeric plus `_.-`), the Docker-compatible two-character minimum, rejection of leading separators and invalid characters, and the empty-string case. Each error case also asserts the wrapped `errdefs.ErrInvalidArgument`.

Test-only, no behavior change.

### DIFF
--- a/pkg/identifiers/validate_test.go
+++ b/pkg/identifiers/validate_test.go
@@ -1,0 +1,58 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pkg/identifiers/validate_test.go
AkihiroSuda, line 1: The file name should be `<IMPLEMENTATION>_test.go`, i.e., `validate_test.go`

AkihiroSuda (approved):

Issue #5027, opened 2026-06-30
`nerdctl images`: adopt Docker v29 output


AkihiroSuda:
Docker v29.6.1:
```console
$ docker images -a
                                                                                                                                                                                                                                                                                                                                                                               i Info →   U  In Use
IMAGE                                                                                 ID             DISK USAGE   CONTENT SIZE   EXTRA
hello-world@sha256:09538a1f51d3ec5af0449a1640937dfdf79b0e9b8c4da5b8a883086d5c1492ef   09538a1f51d3       20.6kB         4.21kB        
moby/buildkit:buildx-stable-1                                                         0168606be231        348MB          105MB    U  
```

nerdctl v2.3.4:
```console
$ nerdctl images -a
REPOSITORY     TAG       IMAGE ID        CREATED           PLATFORM     SIZE       BLOB SIZE
hello-world    <none>    09538a1f51d3    26 seconds ago    linux/386    16.38kB    4.214kB
```

- https://github.com/docker/cli/pull/6566

PR #5092, opened 2026-07-21
feat(image): add --tree flag to image list

### LINKED ISSUE #5005: `nerdctl images` lacks `--tree`
The `--tree` flag  in  `docker images` (`docker image ls`)  is still not supported in nerdctl

https://docs.docker.com/reference/cli/docker/image/ls/

ekalinin:
Adds `--tree` to `nerdctl images`, expanding the Docker v29 view from #5093 with a row per
platform, like `docker image ls --tree`: same columns (`IMAGE`, `ID`, `DISK USAGE`, `CONTENT SIZE`,
`EXTRA`) and the same branch glyphs docker/cli uses.

- Every platform the index declares gets a row; the ones that were never pulled show zero sizes, as
  docker does. This is confined to the tree - the collapsed and legacy views, and
  `EnsureAllContent`, keep describing only what is in the store.
- A platform row carries its own manifest digest as the ID, its own sizes, and `U` when a container
  runs that platform. Matching is on the full platform form rather than on the displayed name, so
  several `windows/amd64` manifests differing only by OSVersion stay distinct; the container label
  is normalized first, since `platforms.DefaultString` does not (on arm64 it carries a variant).
- Conflicting flags are rejected the way docker's `shouldUseTree` does, extended to the
  nerdctl-only `--names`. Validated in `pkg/cmd/image` so library callers are covered too, and in
  the CLI so the error precedes the containerd connection.
- Image groups are not separated by a blank line: the rows share a tabwriter with the header, where
  a blank line would end the column block and misalign every following group. docker/cli sizes its
  columns itself and has no such constraint.

Also fixes the platform of a multi-platform image being read from the image config instead of the
index descriptor: alpine's `linux/arm/v6` and `linux/arm/v7` configs both declare a bare
`linux/arm`, which collapsed them onto one key and dropped a platform. **This affects the non-tree
view too**, whose aggregated sizes were missing it - happy to split it into its own PR if you
prefer.

Review feedback: the expected per-platform content sizes now live in `testutil` (`contentsize` in
`images.yaml` plus accessors) and the test asserts them for every pulled platform, while
`DISK USAGE` is only asserted to cover `CONTENT SIZE`, since what is unpacked depends on what the
rest of the suite did with the shared image store. The test is no longer skipped for Docker and
only diverges where docker does: `docker pull` has no `--all-platforms`, docker's `--digests`
conflict message names its internal flag, and `--names` does not exist there. It lives in a
`_linux_test.go` because the `testutil` image registry is Linux-only.

Known difference: with several tags on one target, nerdctl prints a tree per image record while
docker groups the names into a single node - the collapsed view from #5093 behaves the same way.

Closes #5005

### DIFF
--- a/pkg/cmd/image/list.go
+++ b/pkg/cmd/image/list.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"slices"
 	"sort"
 	"strings"
 	"text/tabwriter"
@@ -54,11 +55,42 @@ import (
 
 // ListCommandHandler `List` and print images matching filters in `options`.
 func ListCommandHandler(ctx context.Context, client *containerd.Client, options *types.ImageListOptions) error {
+	if err := ValidateListOptions(options); err != nil {
+		return err
+	}
 	imageList, err := List(ctx, client, options.Filters, options.NameAndRefFilter)
 	if err != nil {
 		return err
 	}
 	return printImages(ctx, client, imageList, options)
+}
+
+// ValidateListOptions rejects option combinations the list views cannot honor, mirroring the
+// checks docker/cli makes in shouldUseTree. Unlike the implicit choice between the default and the
+// legacy view, Tree is an explicit request, so silently falling back would be surprising.
+//
+// It is enforced here, where the options are actually consumed, so that library callers get the
+// error too. The CLI calls it as well, so that the error surfaces before a containerd connection
+// is attempted.
+func ValidateListOptions(options *types.ImageListOptions) error {
+	if !options.Tree {
+		return nil
+	}
+	for _, conflict := range []struct {
+		set  bool
+		flag string
+	}{
+		{options.Quiet, "--quiet"},
+		{options.NoTrunc, "--no-trunc"},
+		{options.Digests, "--digests"},
+		{options.Format != "", "--format"},
+		{options.Names, "--names"},
+	} {
+		if conflict.set {
+			return fmt.Errorf("%s is not yet supported with --tree", conflict.flag)
+		}
+	}
+	return nil
 }
 
 // List queries containerd client to get image list and only returns those matching given filters.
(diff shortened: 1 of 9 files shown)


pkg/cmd/image/list.go
```diff
@@ -429,3 +438,150 @@ func isAttestationManifestDescriptor(desc ocispec.Descriptor) bool {
...
 	return desc.Annotations[manifestReferenceType] == attestationManifest
 }
+
+// treeNode is one image record (a single name -> target mapping) rendered as a
+// top-level row plus one child row per platform, similar to `docker image ls --tree`.
+type treeNode struct {
+	name      string // familiar "repo:tag" or "<none>:<none>"
+	id        string // target (index or manifest) digest, truncated unless --no-trunc
+	diskUsage int64  // sum of the platforms' unpacked snapshot sizes
+	blobSize  int64  // sum of the platforms' content-store blob sizes
```
AkihiroSuda: Non-tree mode should use the new size definition too
- https://github.com/containerd/nerdctl/issues/5027

So I suggest working on it before `--tree`

cmd/nerdctl/image/image_list_test.go
```diff
@@ -135,6 +135,75 @@ func TestImages(t *testing.T) {
...
+			helpers.Ensure("pull", "--quiet", commonImage.String())
+		},
+		SubTests: []*test.Case{
+			{
+				Description: "tree layout",
+				Command:     test.Command("images", "--tree"),
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						Output: expect.All(
+							expect.Contains("IMAGE", "ID", "DISK USAGE", "CONTENT SIZE", "IN USE"),
```
AkihiroSuda: DISK USAGE and CONTENT SIZE should be tested.
The expected sizes should be defined in `testutil`.

cmd/nerdctl/image/image_list_test.go
```diff
@@ -135,6 +135,75 @@ func TestImages(t *testing.T) {
...
 
+func TestImagesTree(t *testing.T) {
+	nerdtest.Setup()
+
+	commonImage, _ := referenceutil.Parse(testutil.CommonImage)
+
+	testCase := &test.Case{
+		// `docker image ls --tree` is experimental and requires the containerd
+		// image store, so keep this test nerdctl-only.
+		Require: require.Not(nerdtest.Docker),
```
AkihiroSuda: the containerd image store is now enabled in Docker, so this test should not be skipped

AkihiroSuda:
Needs rebase

AkihiroSuda (approved):
Thanks

PR #5093, opened 2026-07-22
feat(image): adopt Docker v29 output for images list

### LINKED ISSUE #5027: `nerdctl images`: adopt Docker v29 output

ekalinin:
Make the default `nerdctl images` output match Docker v29: a collapsed view with
  `IMAGE`, `ID`, `DISK USAGE`, `CONTENT SIZE` and `EXTRA` columns, an "In Use" (`U`)
  indicator, and `<untagged>` for dangling images. Multi-platform images are collapsed
  into a single row with aggregated disk and content size.

  Details that follow Docker's behavior:
  - `DISK USAGE` is the content store blobs plus the unpacked snapshots (Docker's
    `Total`), `CONTENT SIZE` is the blobs alone. Both are formatted with 3 significant
    digits, like Docker.
  - In use is resolved by image target digest, the way Docker matches containers to
    images, so every reference to a used target is flagged, not only the one the
    container was created from.
  - The "In Use" legend is only printed when the output is a terminal, so piped and
    redirected output stays clean.
  - Rows are ordered by image reference with untagged images last, rather than by
    creation time.
  - The `ID` column is the OCI digest of the image target, which matches Docker's ID
    with the containerd image store (it differs only from the legacy graphdriver
    image ID). The help text and docs are updated accordingly.

  The new view is used only for the bare command. Passing `--format`, `--quiet`,
  `--no-trunc`, `--digests` or `--names` falls back to the legacy table (`REPOSITORY`,
  `TAG`, `IMAGE ID`, `CREATED`, `PLATFORM`, `SIZE`, `BLOB SIZE`), so existing scripts and
  templates keep working, including their creation-time ordering. This mirrors Docker's
  own `shouldUseTree` fallback.

  Since the default output now matches Docker, the images tests also run against the
  Docker target; only the nerdctl-specific `--names` subtest stays gated. Tests that
  assert on the default `images` output for untagged images (image prune/remove and
  build-without-tag) are updated to expect `<untagged>`.

  The expanded per-platform `--tree` view is left for a follow-up.

  Closes #5027

### DIFF
--- a/cmd/nerdctl/image/image_list.go
+++ b/cmd/nerdctl/image/image_list.go
@@ -33,11 +33,18 @@ func ImagesCommand() *cobra.Command {
 	shortHelp := "List images"
 	longHelp := shortHelp + `
 
-Properties:
+By default (Docker v29 compatible view) the following columns are shown:
+- IMAGE:        Image reference ("repository:tag", "repository@digest", or "<untagged>")
+- ID:           OCI digest of the image target (index/manifest), shared for multi-platform images. Matches Docker's ID with the containerd image store (differs from the legacy graphdriver image ID).
+- DISK USAGE:   Total on-disk size: content store blobs plus the unpacked snapshots
+- CONTENT SIZE: Size of the blobs (such as layer tarballs) in the content store
+- EXTRA:        Flags for the image; "U" means the image is in use by a container
+
+Passing --format, --quiet, --no-trunc, --digests or --names falls back to the legacy table:
 - REPOSITORY: Repository
 - TAG:        Tag
 - NAME:       Name of the image, --names for skip parsing as repository and tag.
-- IMAGE ID:   OCI Digest. Usually different from Docker image ID. Shared for multi-platform images.
+- IMAGE ID:   OCI digest of the image target (index/manifest), shared for multi-platform images. Matches Docker's ID with the containerd image store (differs from the legacy graphdriver image ID).
 - CREATED:    Created time
 - PLATFORM:   Platform
 - SIZE:       Size of the unpacked snapshots
--- a/cmd/nerdctl/image/image_list_test.go
+++ b/cmd/nerdctl/image/image_list_test.go
@@ -44,7 +44,6 @@ func TestImages(t *testing.T) {
 	commonImage, _ := referenceutil.Parse(testutil.CommonImage)
 
 	testCase := &test.Case{
-		Require: require.Not(nerdtest.Docker),
 		Setup: func(data test.Data, helpers test.Helpers) {
 			helpers.Ensure("pull", "--quiet", commonImage.String())
 			helpers.Ensure("pull", "--quiet", testutil.NginxAlpineImage)
@@ -58,18 +57,13 @@ func TestImages(t *testing.T) {
 						Output: func(stdout string, t tig.T) {
 							lines := strings.Split(strings.TrimSpace(stdout), "\n")
 							assert.Assert(t, len(lines) >= 2, "there should be at least two lines\n")
-							header := "REPOSITORY\tTAG\tIMAGE ID\tCREATED\tPLATFORM\tSIZE\tBLOB SIZE"
-							if nerdtest.IsDocker() {
-								header = "REPOSITORY\tTAG\tIMAGE ID\tCREATED\tSIZE"
-							}
-							tab := tabutil.NewReader(header)
+							tab := tabutil.NewReader("IMAGE\tID\tDISK USAGE\tCONTENT SIZE\tEXTRA")
 							err := tab.ParseHeader(lines[0])
 							assert.NilError(t, err, "ParseHeader should not fail\n")
 							found := false
 							for _, line := range lines[1:] {
-								repo, _ := tab.ReadRow(line, "REPOSITORY")
-								tag, _ := tab.ReadRow(line, "TAG")
-								if repo+":"+tag == commonImage.FamiliarName()+":"+commonImage.Tag {
+								image, _ := tab.ReadRow(line, "IMAGE")
+								if image == commonImage.FamiliarName()+":"+commonImage.Tag {
 									found = true
 									break
 								}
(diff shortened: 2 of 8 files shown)


cmd/nerdctl/image/image_list_test.go
AkihiroSuda: Should no longer be skipped.

Same for other tests too.

ekalinin: Fixed.

cmd/nerdctl/image/image_list.go
```diff
@@ -33,7 +33,14 @@ func ImagesCommand() *cobra.Command {
 	shortHelp := "List images"
 	longHelp := shortHelp + `
 
-Properties:
+By default (Docker v29 compatible view) the following columns are shown:
+- IMAGE:        Image reference ("repository:tag", "repository@digest", or "<untagged>")
+- ID:           OCI Digest. Usually different from Docker image ID. Shared for multi-platform images.
+- DISK USAGE:   Size of the unpacked snapshots
+- CONTENT SIZE: Size of the blobs (such as layer tarballs) in the content store
```
AkihiroSuda, line 40: DISK USAGE and CONTENT SIZE seem to differ from Docker
```console
$ docker images
                                                                                                                                                                                                                                i Info →   U  In Use
IMAGE           ID             DISK USAGE   CONTENT SIZE   EXTRA
alpine:latest   28bd5fe8b56d       13.6MB         4.27MB        

$ nerdctl images
                                                                                                                                                                                                                                i Info →   U  In Use
IMAGE            ID              DISK USAGE    CONTENT SIZE    EXTRA
alpine:latest    28bd5fe8b56d    9.31MB        4.185MB 
```

Perhaps not a bug though

cmd/nerdctl/image/image_list.go
```diff
@@ -33,7 +33,14 @@ func ImagesCommand() *cobra.Command {
 	shortHelp := "List images"
 	longHelp := shortHelp + `
 
-Properties:
+By default (Docker v29 compatible view) the following columns are shown:
+- IMAGE:        Image reference ("repository:tag", "repository@digest", or "<untagged>")
+- ID:           OCI Digest. Usually different from Docker image ID. Shared for multi-platform images.
```
AkihiroSuda: No longer different?

AkihiroSuda (approved):
Thanks

cmd/nerdctl/image/image_list.go
ekalinin: It's a bug. Thanks!

Per moby's [api/types/image.ManifestSummary.Size](https://github.com/moby/moby/blob/v28.5.2/api/types/image/manifest.go#L38), Docker's tree view uses `Total` for DISK USAGE and `Content` for CONTENT SIZE, where `Total = Content + Unpacked` (content-store blobs + unpacked snapshots). nerdctl was putting only the unpacked snapshot size in DISK USAGE, which is exactly why `9.31 + 4.27 ≈ 13.6`. Fixed: DISK USAGE now sums content + snapshot.

Fixed.

cmd/nerdctl/image/image_list.go
ekalinin: Right, no longer different. Fixed.

ekalinin:
Force-pushed. Both threads above are addressed (`DISK USAGE` is now `Total` = content + unpacked, and the `ID` note is updated). On top of that, two more Docker-compat fixes that I found while re-reading the diff:

**1. "In Use" is now resolved by image target digest instead of image name**

moby matches containers to images by digest ([`daemon/containerd/image_list.go#L338`](https://github.com/moby/moby/blob/v28.5.2/daemon/containerd/image_list.go#L338)):

```go
  if c.ImageManifest != nil && c.ImageManifest.Digest == target.Digest {
```

nerdctl was keying the in-use set on `container.Image().Name()`, so only the reference the container had been created from got the `U` flag. With `nerdctl tag alpine:latest myalias:v1` and a container running off `alpine:latest, the myalias:v1` row stayed empty even though it points at the same target. Now every reference to a used target is flagged. The "In use" integration test tags the image under a second name and asserts that all rows pointing at that target are marked, so it also covers the Docker target (where those names are collapsed into a single row).

**2. Row ordering in the new view now follows docker/cli**

docker/cli's tree view sorts by image name lexicographically and places untagged images last, while nerdctl kept its creation-time ordering. The new view is now sorted by the rendered `IMAGE` value with `<untagged>` last. This is applied to the new view only: the legacy table keeps the creation-time ordering that `--format` consumers (and the `CheckCreatedTime` subtest) rely on. Added a unit test for the comparator.

  ---

**One known divergence left, which I'd rather fix separately - let me know if you want it here**

`CONTENT SIZE` sums the descriptor sizes declared in the manifest without checking that the blobs are actually present locally, and skips attestation manifests entirely. moby walks only the present content ([walkPresentChildren](https://github.com/moby/moby/blob/v28.5.2/daemon/containerd/image_list.go#L261), [ImageManifest.PresentContentSize](https://github.com/moby/moby/blob/v28.5.2/daemon/containerd/image_manifest.go#L245)) and adds it to `TotalSize` [before](https://github.com/moby/moby/blob/v28.5.2/daemon/containerd/image_list.go#L267) the pseudo-image / attestation check, so attestation content is counted.

So nerdctl can over-report for partially pulled images - which is not exotic here, given the lazy-pulling snapshotters (stargz, nydus, SOCI, overlaybd) - and under-report when attestations are present. This math predates this PR (it is the same code behind today's `BLOB SIZE` column), but the new view now surfaces it in two columns instead of one, so it is worth calling out. I'll open a separate issue unless you prefer it folded in.

PTAL.

AkihiroSuda (approved):
Thanks

PR #5094, opened 2026-07-23
test: convert compose run linux test to Tigron framework

papidb:
Rewrite tests in compose_run_linux_test.go from testutil.NewBase to nerdtest.Setup (Tigron framework). Part of #4613.

### DIFF
--- a/cmd/nerdctl/compose/compose_run_linux_test.go
+++ b/cmd/nerdctl/compose/compose_run_linux_test.go
@@ -20,9 +20,9 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
-	"time"
 
 	"gotest.tools/v3/assert"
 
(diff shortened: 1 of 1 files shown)


cmd/nerdctl/compose/compose_run_linux_test.go
```diff
@@ -114,123 +115,135 @@ services:
...
+		// specify the name of container in order to remove
+		// TODO: when `compose rm` is implemented, replace it.
+		cmd := helpers.Command("compose", "-f", composePath, "run", "--service-ports", "--name", data.Identifier(), "web")
+		cmd.WithPseudoTTY()
+		cmd.Background()
+		nerdtest.EnsureContainerStarted(helpers, data.Identifier())
+	}
+
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("rm", "-f", "-v", data.Identifier())
```
AkihiroSuda, line 151: Doesn't this need to use `compose rm`?

papidb: Yes it is, I should ensure instead of using anyhow yes?

ChengyuZhu6 (approved):
LGTM

cmd/nerdctl/compose/compose_run_linux_test.go
AkihiroSuda: My concern was that it should use `compose rm`, not `rm`.

Haven't confirmed whether it should be "ensure" or "anyhow"

PR #5100, opened 2026-07-24
pkg/portutil/procnet: support big-endian hosts in ParseAddress

### LINKED ISSUE #4465: `pkg/portutil/procnet`lacks support for big endian  (i.e., s390x)
https://github.com/containerd/nerdctl/blob/ca6ced01e928c1d1a059e68b13c196619dce0d4d/pkg/portutil/procnet/procnet.go#L77

The support for the big endian was implemented in the Lima upstream in https://github.com/lima-vm/lima/commit/da5fb5ecafd3f6384bd8b1af66bdc62d7109ef6c

larrasket:
`/proc/net/{tcp,tcp6}` address bytes are written in the host's byte order. ParseAddress reversed each 4-byte group, which mangles the address on big-endian hosts such as s390x (e.g. 7F000001 parsed to 1.0.0.127 instead of 127.0.0.1).

Add ParseAddressWithEndian and select the byte order from golang.org/x/sys/cpu.IsBigEndian, keeping the existing behavior on little-endian hosts.

Fixes #4465

### DIFF
--- a/pkg/portutil/procnet/procnet.go
+++ b/pkg/portutil/procnet/procnet.go
@@ -17,6 +17,7 @@
 package procnet
 
 import (
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


pkg/portutil/procnet/procnetd_test.go
```diff
@@ -67,3 +67,18 @@ func TestParseTCP6Zero(t *testing.T) {
...
+	assert.Check(t, net.ParseIP("127.0.0.1").Equal(ip))
+	assert.Equal(t, uint16(80), port)
+
+	ip, _, err = ParseAddressWithEndian("FE8000000000000070A657FFFE71C75D:0050", true)
+	assert.NilError(t, err)
+	assert.Check(t, net.ParseIP("fe80::70a6:57ff:fe71:c75d").Equal(ip))
+
+	ip, _, err = ParseAddressWithEndian("0100007F:0050", false)
+	assert.NilError(t, err)
+	assert.Check(t, net.ParseIP("127.0.0.1").Equal(ip))
```
AkihiroSuda, line 86: Should test IPv6 in little endian too?

pkg/portutil/procnet/procnet.go
```diff
@@ -72,12 +74,19 @@ func removeEmpty(array []string) (results []string) {
...
 // and is licensed under the Apache License, Version 2.0
 func ParseAddress(s string) (net.IP, uint16, error) {
+	return ParseAddressWithEndian(s, cpu.IsBigEndian)
+}
+
+// ParseAddressWithEndian is like ParseAddress but takes the host endianness
+// explicitly. The kernel writes each 4-byte group of the address in the host's
+// native byte order, the bytes are reversed on little-endian hosts and kept
+// as-is on big-endian hosts (for example, s390x).
+func ParseAddressWithEndian(s string, isBE bool) (net.IP, uint16, error) {
```
AkihiroSuda: Should this take https://pkg.go.dev/encoding/binary#ByteOrder as an argument

pkg/portutil/procnet/procnetd_test.go
larrasket: I switched to `binary.ByteOrder`, `ParseAddressWithByteOrder` takes the order and `ParseAddress` passes `binary.NativeEndian`.

pkg/portutil/procnet/procnet.go
larrasket: Added little-endian IPv6 case, PTAL

AkihiroSuda (approved):
Thanks

Issue #4344, opened 2025-06-22
[CI/Windows]: TestNetworkInspect/Verify_that_only_active_containers_appear_in_the_network_inspect_output is flaky


apostasie:
### Description

```
          +------------------------------------------------------------------------------------------------------------+
          | ➡️      | ⚙️ C:\Users\runneradmin\go\bin\nerdctl.exe create --name testnetworkinspect-verify-that-only-act |
          |         | ive-containers-appear-in-the-0a54a379 --network testnetworkinspect-verify-that-only-active-conta |
          |         | iners-appear-in-the-844e8a21 registry.k8s.io/e2e-test-images/nginx:1.14-2                        |
          +------------------------------------------------------------------------------------------------------------+
          |         | 🚫 command returned a non-zero exit code                                                         |
          +------------------------------------------------------------------------------------------------------------+
          |         | ⚠️ 1                                                                                             |
          +------------------------------------------------------------------------------------------------------------+
          |         | 🟠 time="2025-06-22T21:52:24Z" level=warning msg="default network named \"nat\" does not have an |
          |         |  internal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl"  |
          |         | time="2025-06-22T21:52:24Z" level=warning msg="default network named \"nat\" does not have an in |
          |         | ternal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl"     |
          |         | time="2025-06-22T21:52:25Z" level=warning msg="networking setup error has occurred" error="plugi |
          |         | n type=\"nat\" failed (add): hcnCreateNetwork failed in Win32: The object already exists. (0x139 |
          |         | 2) {\"Success\":false,\"Error\":\"The object already exists. \",\"ErrorCode\":2147947410}"       |
          |         | time="2025-06-22T21:52:25Z" level=warning msg="default network named \"nat\" does not have an in |
          |         | ternal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl"     |
          |         | time="2025-06-22T21:52:25Z" level=fatal msg="plugin type=\"nat\" failed (add): hcnCreateNetwork  |
          |         | failed in Win32: The object already exists. (0x1392) {\"Success\":false,\"Error\":\"The object a |
          |         | lready exists. \",\"ErrorCode\":2147947410}"        
```

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

Not flake.

### What version of nerdctl are you using?

main

[...]

PR #5115, opened 2026-08-01
fix: image ls with a bare repository name should match all tags

### LINKED ISSUE #5113: `nerdctl image ls myapp` shows nothing for images without latest tag
### Description

This report is a follow-up to the discussion I opened here
- https://github.com/containerd/nerdctl/discussions/5110

The command `nerdctl image ls myapp`, where `myapp` is repository only, does not find any images of `myapp` if there is no `latest` tag present for `myapp`. This behaviour is very different to one of the Docker client, e.g. [`docker image ls java`](https://docs.docker.com/reference/cli/docker/image/ls/#list-images-by-name-and-tag). I suspect it is an unexpected behaviour or even a bug.

The current behaviour is due to the unconditional use of `TagNameOnly` which adds the default tag "latest" to a reference if it only has a repo name, from here:

1. The `referenceutil.Parse("myapp")`  yields `docker.io/library/myapp` due to https://github.com/containerd/nerdctl/blob/9145d0f1bf1afe97ee6d36b5e350138808468d84/pkg/referenceutil/referenceutil.go#L133-L134
2. The `TagNameOnly` unconditionally  appends `:latest` in https://github.com/distribution/reference/blob/0965666a6ade2e06035fe352e38344be1e68951a/normalize.go#L248-L252 
3. The filter becomes `name==docker.io/library/myapp:latest`

It's possible that this implementation of the `listOptions` function: https://github.com/containerd/nerdctl/blob/9145d0f1bf1afe97ee6d36b5e350138808468d84/cmd/nerdctl/image/image_list.go#L80-L84

should be modified to something like this:

```diff
-               filters = []string{fmt.Sprintf("name==%s", parsedReference)}
+               if parsedReference.ExplicitTag != "" || parsedReference.Digest != (digest.Digest("")) {
+                       filters = []string{fmt.Sprintf("name==%s", parsedReference)}
+               } else {
+                       filters = []string{fmt.Sprintf("name~=^%s:", regexp.QuoteMeta(parsedReference.Name()))}
+               }
```

To work around the inconvenience of the current implementation users instead of the simple invocation of
```
nerdctl image ls myapp
```

have to apply the filter trick
```
nerdctl image ls --filter 'reference=myapp:*'
```

1. Go to https://github.com/containerd/nerdctl/discussions/5110
2. Follow the sequence of examples presented there

Received results:

```
PS > nerdctl image ls myapp
```

Expected results:

```
PS > nerdctl image ls myapp
REPOSITORY                              TAG                               IMAGE ID        CREATED           PLATFORM         SIZE       BLOB SIZE
myapp                                   9.8-ltsc2022-0.1.0-dev.5011fb1    5e2b4a9d1b07    34 minutes ago    windows/amd64    11.91GB    2.113GB
```

```
PS > nerdctl version
time="2026-07-30T21:53:56+01:00" level=warning msg="unable to determine runc version" error="exec: \"runc\": executable file not found in %PATH%"
Client:
 Version:       v2.3.4
 OS/Arch:       windows/amd64
 Git commit:    0ce88b9d78b13f0caebc59c6bb01885d7df24fba
 buildctl:
  Version:      v0.31.2
  GitCommit:    e42e1bfd389af7203238cce77b1f7dad447285e9

Server:
 containerd:
  Version:      v2.3.3
  GitCommit:    aad11006b869517fcd3009450b6f82da282e1a9b
 runc:
  Version:
```

None

### Host information

Windows Server 2022 with containerd, buildkit and nerdctl

ankit090701:
## What's the bug

`nerdctl image ls myapp`, where `myapp` is a bare repository name (no tag or digest), returns nothing unless `myapp` happens to have a `:latest` tag. This is different from `docker image ls myapp`, which [lists every tag of the repository](https://docs.docker.com/reference/cli/docker/image/ls/#list-images-by-name-and-tag).

```
$ nerdctl image ls myapp
```

even though `myapp:9.8-ltsc2022-...` clearly exists.

## Root cause

`referenceutil.Parse` normalizes a bare repository name by unconditionally applying `distribution/reference`'s `TagNameOnly`, which appends an implicit `:latest` tag ([referenceutil.go#L133-L134](https://github.com/containerd/nerdctl/blob/9145d0f1bf1afe97ee6d36b5e350138808468d84/pkg/referenceutil/referenceutil.go#L133-L134)). `listOptions` in `cmd/nerdctl/image/image_list.go` then built an *exact-match* filter straight from that normalized reference:

```go
filters = []string{fmt.Sprintf("name==%s", parsedReference)}
```

which becomes `name==docker.io/library/myapp:latest` - so any image tagged with anything other than `latest` is silently excluded.

## The fix

Extracted the filter-construction logic into a small `nameFilterFor` helper:

- If the argument named an **explicit** tag or digest (`parsedReference.ExplicitTag != "" || parsedReference.Digest != ""`), keep matching it exactly with `name==...`.
- If it was a **bare repository name**, build a `name~=^<repo>:` regex filter instead (escaping the repository name with `regexp.QuoteMeta`), so it matches any tag under that repository.

This mirrors the `~=` regex-filter pattern already used elsewhere in the codebase for similar name/id matching, e.g. [`pkg/idutil/imagewalker/imagewalker.go`](https://github.com/containerd/nerdctl/blob/9145d0f1bf1afe97ee6d36b5e350138808468d84/pkg/idutil/imagewalker/imagewalker.go#L82-L83) and `pkg/idutil/containerwalker/containerwalker.go`, both of which already build `field~=regex` filters with `regexp.QuoteMeta` escaping - so this isn't introducing a new pattern, just applying an existing one to close this gap.

## Testing

- Added `TestNameFilterFor`, a focused unit test on the extracted helper (no daemon required), covering:
  - a bare repository name → now produces a `name~=^docker\.io/library/myapp:` regex filter (matches any tag)
  - an explicit tag (`myapp:v1`) → still an exact `name==...:v1` filter
  - an explicit `:latest` tag → still an exact `name==...:latest` filter
  - a digest reference → still an exact `name==...@sha256:...` filter
  - a bare name with a domain/path that needs regex-escaping (`registry.example.com/foo/my.app`)
- Verified this test fails without the fix, reproducing the exact reported bug: it produces `name==docker.io/library/myapp:latest` instead of a repo-wide match.
- `go build ./...` and `go vet ./...` pass. The rest of the package's tests are `tigron`/E2E tests that shell out to a built `nerdctl` binary against a live containerd daemon, which I don't have available in my sandboxed environment - those fail with `unable to find binary "nerdctl"` regardless of this change (verified against an unmodified checkout).

Fixes #5113

### DIFF
--- a/cmd/nerdctl/image/image_list.go
+++ b/cmd/nerdctl/image/image_list.go
@@ -18,6 +18,7 @@ package image
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/spf13/cobra"
 
@@ -81,7 +82,7 @@ func listOptions(cmd *cobra.Command, args []string) (*types.ImageListOptions, er
 		if err != nil {
 			return nil, err
 		}
-		filters = []string{fmt.Sprintf("name==%s", parsedReference)}
+		filters = nameFilterFor(parsedReference)
 	}
 	quiet, err := cmd.Flags().GetBool("quiet")
 	if err != nil {
@@ -123,6 +124,23 @@ func listOptions(cmd *cobra.Command, args []string) (*types.ImageListOptions, er
 		Stdout:           cmd.OutOrStdout(),
 	}, nil
 
+}
+
+// nameFilterFor builds the containerd image-service filter(s) matching the
+// argument to `nerdctl image ls [REPOSITORY[:TAG]]`.
+//
+// If the argument named an explicit tag or digest, it's matched exactly.
+// Otherwise the argument was a bare repository name: referenceutil.Parse
+// normalizes that to an implicit ":latest" tag (matching how most other
+// reference-consuming commands resolve a bare name), but for listing
+// purposes that would incorrectly hide every other tag of the repository -
+// unlike `docker image ls`, which matches all tags of a bare repository
+// name. Match any tag under the repository instead.
+func nameFilterFor(parsedReference *referenceutil.ImageReference) []string {
+	if parsedReference.ExplicitTag != "" || parsedReference.Digest != "" {
+		return []string{fmt.Sprintf("name==%s", parsedReference)}
+	}
+	return []string{fmt.Sprintf("name~=^%s:", regexp.QuoteMeta(parsedReference.Name()))}
 }
 
 func imagesAction(cmd *cobra.Command, args []string) error {
--- a/cmd/nerdctl/image/image_list_test.go
+++ b/cmd/nerdctl/image/image_list_test.go
@@ -38,6 +38,54 @@ import (
 	"github.com/containerd/nerdctl/v2/pkg/testutil/nerdtest"
 )
 
+// TestNameFilterFor is a regression test for
+// https://github.com/containerd/nerdctl/issues/5113: `nerdctl image ls
+// myapp`, where myapp is a bare repository name, returned nothing unless
+// myapp had a `:latest` tag, because referenceutil.Parse normalizes a bare
+// repository name to an implicit ":latest" tag and the resulting exact-match
+// filter therefore only ever matched that one tag.
+func TestNameFilterFor(t *testing.T) {
+	testCases := []struct {
+		name     string
+		arg      string
+		expected []string
+	}{
+		{
+			name:     "bare repository name matches any tag",
+			arg:      "myapp",
+			expected: []string{`name~=^docker\.io/library/myapp:`},
+		},
+		{
+			name:     "explicit tag matches exactly",
+			arg:      "myapp:v1",
+			expected: []string{"name==docker.io/library/myapp:v1"},
+		},
+		{
+			name:     "explicit latest tag matches exactly",
+			arg:      "myapp:latest",
+			expected: []string{"name==docker.io/library/myapp:latest"},
+		},
+		{
+			name:     "digest matches exactly",
+			arg:      "myapp@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+			expected: []string{"name==docker.io/library/myapp@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"},
+		},
+		{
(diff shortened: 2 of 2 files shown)


ankit090701:
The 5 red CI jobs are all pre-existing environmental flakiness, unrelated to this change - I traced each one down to the actual assertion/error, not just the test name:

- **`in-host / docker < linux`** - `TestAttachForAutoRemovedContainer`: fails on `assertion failed: strings.Contains(helpers.Capture("ps", "-a"), data.Identifier())` - a timing race where the container (started with implicit `--rm`) is auto-removed by the time `docker ps -a` runs.
- **`in-host / rootful linux`** - `TestBuildContextWithOCILayout` (`tar: Unexpected EOF in archive`) and `TestSaveMultipleImagesWithSameIDAndLoad` (`short read: expected 2811947 bytes but got 2785280: unexpected EOF`) both fail on truncated/corrupted archives in the same job - looks like runner-level disk/IO trouble, not test logic.
- **`in-host / rootful linux (arm)`** and **`in-host / rootless linux (arm)`** - fail before any test runs, in `test-integration-env.sh install`, with `E: Sub-process /usr/bin/dpkg returned an error code (1)` - pure environment provisioning failure.
- **`in-host / windows`** - `TestNetworkInspectWithContainers` fails with `hcnCreateNetwork failed in Win32: The object already exists.` - leftover Windows HNS network state from a previous run on the same runner. This matches the same flaky category as #4344.

None of these touch `cmd/nerdctl/image/image_list.go` or `nameFilterFor`: the failing tests live in the `container`, `builder`, and `network` packages, and the one failure in the same package as my change (`TestSaveMultipleImagesWithSameIDAndLoad`) calls `images` with no name argument, so it never reaches the code this PR touches. My new test, `TestNameFilterFor`, passed in every job.

I don't have permission to re-run the failed jobs on this repo (`gh run rerun` says the run "cannot be rerun"), so flagging here in case a maintainer wants to re-trigger.

AkihiroSuda (approved):
Thanks

mloskot:
@ankit090701 Thanks!

Issue #3942, opened 2025-02-27
Support `nerdctl system df` command


maxsxu:
### What is the problem you're trying to solve

It would be great helpful if we can support `nerdctl system df` command to show the containerd filesystem usage, just like `docker system df` can show docker filesystem usage.

### Describe the solution you'd like

When run `nerdctl system df`  command, it'll show filesystem usage outputs like following:
```
❯ nerdctl system df
TYPE            TOTAL     ACTIVE    SIZE      RECLAIMABLE
Images          17        1         18.25GB   17.26GB (94%)
Containers      3         3         169.2MB   0B (0%)
Local Volumes   4         3         798.66GB   22.62MB (0%)
Build Cache     44        0         0B        0B
```

yankay:
To implement this feature, I think we first need to be able to get the SharedSize of an image.  

ref: https://github.com/containerd/nerdctl/blob/main/pkg/cmd/image/list.go#L127

papidb:
Hi @AkihiroSuda @yankay can I work on adding the sharedSize of an image?

AkihiroSuda:
The image size definition has changed in Docker v29
- #5027 

We have to investigate how the size is computed for `docker system df` in Docker v29 first

PR #5125, opened 2026-08-06
fix(image): pin the image a container runs to its digest

ekalinin:
`nerdctl images` marks an image as in use by resolving the image name recorded on the container. A name is a mutable reference: `container.Image(ctx)` looks it up in the image store as it is now, not as it was when the container was created. Once a tag is moved, the `U` indicator lands on the wrong row.

```sh
 $ nerdctl tag alpine:3.13 mytag:latest
 $ nerdctl run -d --name c1 mytag:latest sleep infinity
 $ nerdctl tag nginx:alpine mytag:latest
 $ nerdctl images
 IMAGE           ID              DISK USAGE    CONTENT SIZE    EXTRA
 alpine:3.13     09538a1f51d3    5.9MB         2.7MB
 nginx:alpine    0168606be231    22.4MB        9.1MB           U
```

Both answers are wrong. The container still runs alpine, containerd still holds its snapshot and its layers, so alpine cannot be removed - yet it is shown as free, while nginx is shown as busy although nothing ever ran it. A container whose tag was removed altogether fails to resolve and drops out of the in-use set entirely.

**The fix**

Record the image target digest on the container at creation time, in a new nerdctl/image-digest label, and use it for the in-use lookup. The digest comes from the image nerdctl has already resolved, so nothing extra is fetched, and the label is read from the metadata the listing has already loaded (WithoutRefreshedMetadata), so the lookup costs no additional round trip.

**Compatibility**

Containers created before this label existed, or created outside nerdctl (ctr, kubelet, another client), have no such label. For those the previous behavior is kept as a fallback: they are still resolved by name. No state migration is needed and no existing container changes behavior for the worse.

An unparsable label value falls back the same way rather than dropping the container from the in-use set. Dropping it would be the more harmful failure: the image would look free, and anything built on this lookup would offer to reclaim space that is actually held.

**Why it matters beyond the indicator**

The same lookup backs the ACTIVE and RECLAIMABLE columns of `nerdctl system df` (follow-up, #3942). There a misattributed container is not a single letter in a column: the unique layers of an image that is actually in use get counted as reclaimable space, which is exactly the number a user acts on when deciding what to delete.

**The same digest is what `nerdctl inspect` should report**

Added after review feedback. Docker names the image a container was created from by ID, and with the containerd image store that ID is the digest of the image target, pinned on the container at creation:

```go
// moby daemon/containerd/image.go
imgV1 := dockerOciImageToDockerImagePartial(image.ID(img.Target.Digest), ociImage)
// moby daemon/create.go
imgID = img.ID()
ctr, err = daemon.newContainer(opts.params.Name, platform, ..., imgID, opts.managed)
// moby daemon/inspect.go
Image: ctr.ImageID.String(),
```

That is the value this label records. nerdctl was filling `Container.Image` with the containerd image *name*, which a retag moves just the same as the in-use lookup above, so `ContainerFromNative` now reports the pinned digest there. The reference the user asked for stays in `Config.Image`, as it does in Docker; the containers with no label keep the name.

**Tests**

  - `TestPinnedImageDigest` covers the four label states: pinned, absent, empty, unparsable.
  - `TestImages/In use survives a retag` reproduces the scenario end to end and asserts both sides: the image the container runs keeps `U`, the image the tag now points at does not.
  - `TestContainerFromNativeImage` covers both inspect fields, pinned and not.
  - `TestContainerInspectConfigImage` asserts end to end that `Image` is a digest and `Config.Image` is not.

The in-use lookup was introduced in #5093.

### DIFF
--- a/pkg/inspecttypes/dockercompat/dockercompat.go
+++ b/pkg/inspecttypes/dockercompat/dockercompat.go
@@ -340,13 +340,28 @@ var defaultCaps = map[string]struct{}{
 	"CAP_AUDIT_WRITE":      {},
 }
 
+// containerImage is the image the container was created from, the way Docker identifies it: by
+// digest, pinned when the container was created. containerd only records the image name, and a name
+// can later be retagged onto another image, so it is not an answer. The name is still the fallback
+// for the containers created before that digest was recorded, or created outside nerdctl.
+//
+// With the containerd image store, the image ID Docker reports here is the digest of the image
+// target (moby daemon/containerd/image.go, image.ID(img.Target.Digest)), which is what nerdctl
+// pins.
+func containerImage(n *native.Container) string {
+	if dgst := n.Labels[labels.ImageDigest]; dgst != "" {
+		return dgst
+	}
+	return n.Image
+}
+
 // ContainerFromNative instantiates a Docker-compatible Container from containerd-native Container.
 func ContainerFromNative(n *native.Container) (*Container, error) {
 	var hostname string
 	c := &Container{
 		ID:      n.ID,
 		Created: n.CreatedAt.Format(time.RFC3339Nano),
-		Image:   n.Image,
+		Image:   containerImage(n),
 		Name:    n.Labels[labels.Name],
 		Driver:  n.Snapshotter,
 		// XXX is this always right? what if the container OS is NOT the same as the host OS?
@@ -576,7 +591,8 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
 	c.State = cs
 	c.Config = &Config{
 		Labels: n.Labels,
-		Image:  c.Image,
+		// Docker keeps the reference the user asked for here, and the digest in Image above.
+		Image: n.Image,
 	}
 	if exposedPortsJSON := n.Labels[labels.ExposedPorts]; exposedPortsJSON != "" {
 		var exposedPorts nat.PortSet
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -87,6 +87,11 @@ const (
 	// Platform is the normalized platform string like "linux/ppc64le".
 	Platform = Prefix + "platform"
 
+	// ImageDigest is the digest of the image target the container was created from. The image name
+	// stored by containerd can be retagged to point at something else, so it is not enough to tell
+	// which image a container actually uses.
+	ImageDigest = Prefix + "image-digest"
+
 	// Mounts is the mount points for the container.
 	Mounts = Prefix + "mounts"
 
--- a/cmd/nerdctl/container/container_inspect_linux_test.go
+++ b/cmd/nerdctl/container/container_inspect_linux_test.go
@@ -278,7 +278,7 @@ func TestContainerInspectConfigImage(t *testing.T) {
 	nerdtest.Setup()
 
 	testCase := &test.Case{
-		Description: "Container inspect contains Config.Image field",
+		Description: "Container inspect names the image by digest, and by reference in Config.Image",
 		Setup: func(data test.Data, helpers test.Helpers) {
 			helpers.Ensure("run", "-d", "--name", data.Identifier(), testutil.AlpineImage, "sleep", nerdtest.Infinity)
 		},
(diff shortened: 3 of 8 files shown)


pkg/labels/labels.go
```diff
@@ -87,6 +87,11 @@ const (
 	// Platform is the normalized platform string like "linux/ppc64le".
 	Platform = Prefix + "platform"
 
+	// ImageDigest is the digest of the image target the container was created from. The image name
+	// stored by containerd can be retagged to point at something else, so it is not enough to tell
+	// which image a container actually uses.
+	ImageDigest = Prefix + "image-digest"
```
AkihiroSuda, line 93: Docker stores the image digest in:
https://github.com/containerd/nerdctl/blob/f2c15c8a7f35e764583b06cc97930e82d55a0f71/pkg/inspecttypes/dockercompat/dockercompat.go#L115

ekalinin: Good point, and it turns out the two are the same value.

I have pushed that: `ContainerFromNative` now reports the pinned digest in `Image`, and keeps the reference the user asked for in `Config.Image`, as Docker does. Containers created before the label existed, or created outside nerdctl, still fall back to the name.

Happy to split the inspect part into its own PR if you would rather keep this one to the in-use lookup.

AkihiroSuda: The label should no longer be required?

pkg/inspecttypes/dockercompat/dockercompat.go
```diff
@@ -576,7 +591,8 @@ func ContainerFromNative(n *native.Container) (*Container, error) {
 	c.State = cs
 	c.Config = &Config{
 		Labels: n.Labels,
-		Image:  c.Image,
+		// Docker keeps the reference the user asked for here, and the digest in Image above.
```
AkihiroSuda, line 594: Docker seems just storing the digest here
```console
$ docker run --name foo hello-world
$ docker container inspect foo | jq .[0].Image
"sha256:5dd0d3e6e255913fc30f90b9f2b1d359cc2cbdb48090cc4b65f1676e203243cc"
```

pkg/labels/labels.go
ekalinin: It is still required: the label is where the digest is stored, and `Container.Image` is the view built from it.

The container record containerd keeps has no field for an image digest — `WithImage` writes the name:

```go
// containerd client/container_opts.go
func WithImage(i Image) NewContainerOpts {
	return func(ctx context.Context, client *Client, c *containers.Container) error {
		c.Image = i.Name()
		return nil
	}
}
```

So without the label `ContainerFromNative` has nothing to read but that name, which is the mutable reference this PR is about.

The one way to drop the label would be to write a digested reference into `c.Image` instead (`alpine:3.13@sha256:...`). That field is passed through verbatim to the IMAGE column of `nerdctl ps` (`pkg/cmd/container/list.go`), and is keyed on by name elsewhere (`usedImages[container.Image]` in `pkg/imgutil/imgutil.go`), so it would change what every container shows, not only the ones a retag affects. A label leaves that alone.

Happy to move it somewhere better if there is one.

pkg/inspecttypes/dockercompat/dockercompat.go
ekalinin: Both fields, on the same container:

```console
$ docker run --name foo hello-world
$ docker container inspect foo | jq '{Image: .[0].Image, ConfigImage: .[0].Config.Image}'
{
  "Image": "sha256:5dd0d3e6e255913fc30f90b9f2b1d359cc2cbdb48090cc4b65f1676e203243cc",
  "ConfigImage": "hello-world"
}
```

The digest goes in `Image`, which is what the line above this one now does; `Config.Image` keeps the reference, and that is the line this comment landed on. Let me know if you meant something else.

Also worth recording: that digest is the target descriptor, not an ID derived separately.

```console
$ docker image inspect hello-world | jq -r '.[0].Id, .[0].Descriptor.digest'
sha256:5dd0d3e6e255913fc30f90b9f2b1d359cc2cbdb48090cc4b65f1676e203243cc
sha256:5dd0d3e6e255913fc30f90b9f2b1d359cc2cbdb48090cc4b65f1676e203243cc
```

(docker 29.5.2, containerd snapshotter.) So the `Image.Target().Digest` this PR records at creation is the same value Docker reports.

AkihiroSuda (approved):
Thanks

PR #5127, opened 2026-08-07
fix(image): avoid dangling-ref name collision on `rmi -f` for running images

### LINKED ISSUE #4109: Removing multiple running images is broken
### Description

When `someimage` is currently in use by a container, `nerdctl rmi -f someimage` will create a new unnamed image (name: `:`) so that we keep a ref on the layers (though this is currently buggy, see #4107).

But then, trying to remove multiple such images will fail, as image creation will bail out (`image ":": already exists`).

Does anyone have a suggestion on how to address this?
Is there a way to instruct containerd to create an image without a name? (eg: currently, passing en empty name will fail on creation)

```
nerdctl run -d debian sleep inf
nerdctl run -d alpine sleep inf
nerdctl rmi -f alpine
nerdctl rmi -f debian
```

```
FATA[0000] 1 errors:
image ":": already exists
```

main

None

### Host information

_No response_

pujitha24:
Motivation:
`nerdctl rmi -f` on an image that is still used by a running container
keeps a dangling ref alive (renaming the image before deleting its
original name) so containerd does not garbage-collect the layers.
That dangling ref was always named with the literal string ":". Since
containerd's image store requires unique names, force-removing a
second running image's image in the same run made the second `Create`
fail with "image \":\": already exists", aborting the whole command.

Approach:
Name the dangling ref after its content digest (":<digest>") instead
of the fixed ":" literal, so each kept-alive ref gets a distinct name.
The one remaining consumer that special-cased the exact ":" name
(pkg/imgutil filtering, used by `--filter reference=...` to skip
dangling images without erroring on their unparsable name) is updated
to match on the ":" prefix instead. Also tolerate AlreadyExists on the
Create call: if two different tags happen to share the same digest and
are both force-removed as running images in one invocation, the second
Create legitimately no-ops (the digest is already pinned).

Validation:
This is a name-collision defect in pkg/cmd/image/remove.go, reproducible
against containerd's real images.Store semantics without a live daemon.
Added pkg/cmd/image/remove_test.go, whose
TestDanglingImageNameIsUniquePerDigest exercises a fake images.Store
that enforces the same create-time uniqueness constraint containerd
does, and asserts the new per-digest naming avoids the collision.
I confirmed this test fails against the old behavior (temporarily
reverting the helper to return the literal ":") and passes against the
fix.

Ran:
- go build ./... (and GOOS=linux, GOOS=windows) - all pass
- go vet ./pkg/cmd/image/... ./pkg/imgutil/... - clean
- gofmt -l on the changed files - no output
- go test ./pkg/... - all pass, including the new test
- go test ./pkg/cmd/image/... -run TestDanglingImageNameIsUniquePerDigest -v - PASS

This sandbox is macOS without containerd/runc installed, so the repo's
documented integration suite (go test -p 1 ./cmd/nerdctl/... against a
real containerd store, per docs/testing/README.md) could not be run;
confirmed those tests fail uniformly here (binary not found) regardless
of this change. The added unit test is a targeted regression test for
the digest-collision defect, not a full CLI-level reproduction.

Report: https://github.com/containerd/nerdctl/issues/4109
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #4109


pujitha24:
Just checking in on this one — it's rebased and all checks are green, so let me know if anything would make it easier to review.

pkg/cmd/image/remove_test.go
```diff
@@ -0,0 +1,77 @@
+/*
```
haytok: Tests for this fix should be implemented in https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/image/image_remove_test.go instead.

cmd/nerdctl/image/image_remove_test.go
```diff
@@ -149,6 +149,37 @@ func TestRemove(t *testing.T) {
...
+			// "already exists".
+			Description: "Remove multiple images with running containers - with -f",
+			NoParallel:  true,
+			Require: require.All(
+				require.Not(nerdtest.Docker),
+			),
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("run", "--quiet", "--pull", "always", "-d", "--name", data.Identifier("a"), testutil.CommonImage, "sleep", nerdtest.Infinity)
+				helpers.Ensure("run", "--quiet", "--pull", "always", "-d", "--name", data.Identifier("b"), testutil.NginxAlpineImage)
+			},
```
haytok: Please add the followings to ensure that containers have started.

```golang
nerdtest.EnsureContainerStarted(helpers, data.Identifier("a"))
nerdtest.EnsureContainerStarted(helpers, data.Identifier("b"))
```

cmd/nerdctl/image/image_remove_test.go
```diff
@@ -149,6 +149,37 @@ func TestRemove(t *testing.T) {
...
+					ExitCode: 0,
+					Errors:   []error{},
+					Output: func(stdout string, t tig.T) {
+						helpers.Command("images").Run(&test.Expected{
+							Output: expect.DoesNotContain(repoName, nginxRepoName),
+						})
+					},
+				}
+			},
+		},
```
haytok: To verify `!errdefs.IsAlreadyExists(err)`, the following scenario should be added:

```bash
nerdcrl pull nginx
nerdctl tag nginx nginxa
nerdctl run -d --name nginx nginx
nerdctl run -d --name nginxa nginxa
nerdctl rmi -f nginx:latest nginxa:latest
```

AkihiroSuda:
Closing per https://github.com/lima-vm/lima/pull/5409#issuecomment-5328485853

You, or anybody else, may still resubmit the PR with same content after manual verification.

pujitha24:
That's fair, and I understand closing on this basis — same as the lima-vm thread, I used an AI assistant heavily here and didn't have a way to exercise this against a real containerd daemon in my environment, only the added unit test (later moved into the integration suite per haytok's comment, which I did push). I'll only resubmit this if I can actually verify it end-to-end myself first.

PR #5129, opened 2026-08-07
feat: add volume-nocopy

jiwahn:
- Part of #3867 

- This patch parses `volume-nocopy`option and skip copying the contents from the image to the volume when specified.
- Added unit and integration tests

```Dockerfile
FROM alpine
RUN mkdir -p /mnt && echo hi > /mnt/initial_file
CMD ["sh"]
```

```shell

╰─$ nerdctl run --rm -it \
  --mount type=volume,source=nocopy-vol,target=/mnt,volume-nocopy \
  nocopy-test \
  sh
/ # ls -la mount
ls: mount: No such file or directory
/ # ls -la /mnt
total 8
drwx------    2 root     root          4096 Aug  7 14:51 .
drwxr-xr-x    1 root     root          4096 Aug  7 14:51 ..
/ # exit
```

### DIFF
--- a/cmd/nerdctl/container/container_run_mount_linux_test.go
+++ b/cmd/nerdctl/container/container_run_mount_linux_test.go
@@ -234,8 +234,12 @@ CMD ["cat", "/mnt/initial_file"]
 		volName := data.Identifier("vol")
 		helpers.Ensure("volume", "create", volName)
 
+		noCopyVolName := data.Identifier("nocopy-vol")
+		helpers.Ensure("volume", "create", noCopyVolName)
+
 		data.Labels().Set("img", imgName)
 		data.Labels().Set("vol", volName)
+		data.Labels().Set("nocopy-vol", noCopyVolName)
 	}
 
 	testCase.SubTests = []*test.Case{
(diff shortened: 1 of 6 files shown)


jiwahn:
+) I thought Adding `-v...:nocopy` option makes code a little redundant, so I didn't feel like adding it here. But please let me know if it's better to be done in this PR!

AkihiroSuda:
> +) I thought Adding `-v...:nocopy` option makes code a little redundant, so I didn't feel like adding it here. But please let me know if it's better to be done in this PR!

Does it exist in Docker?
If not no need to implement it

jiwahn:
> Does it exist in Docker?

yes  https://github.com/docker/cli/blob/master/man/docker-run.1.md?plain=1#L744C1-L755C14

AkihiroSuda:
> > Does it exist in Docker?
> 
> yes https://github.com/docker/cli/blob/master/man/docker-run.1.md?plain=1#L744C1-L755C14

Then it should be implemented too

jiwahn:
Yeah I figured. I will follow up with another pr if that's ok

AkihiroSuda (approved):
Thanks

PR #5130, opened 2026-08-07
feat(system): add nerdctl system df

### LINKED ISSUE #3942: Support `nerdctl system df` command

ekalinin:
Add `nerdctl system df`, the equivalent of `docker system df`: the disk space taken by the images,
containers and local volumes of the current namespace, plus the BuildKit build cache.

```console
$ nerdctl system df
TYPE             TOTAL    ACTIVE    SIZE       RECLAIMABLE
Images           17       1         18.25GB    17.26GB (94%)
Containers       3        3         169.2MB    0B (0%)
Local Volumes    4        3         798.6GB    22.62MB (0%)
Build Cache      44       0         0B         0B
```

## The size definitions

The issue was blocked on a question: Docker v29 changed how an image size is computed (#5027), so
what should `system df` report? The formulas below were derived from moby/moby v29 with the
containerd image store (`daemon/disk_usage.go`, `daemon/containerd/service.go`) and docker/cli v29
(`cli/command/formatter/disk_usage.go`).

| Row | TOTAL | ACTIVE | SIZE | RECLAIMABLE |
|---|---|---|---|---|
| Images | unique, by target digest | with at least one container | deduplicated: every snapshot and every present blob counted once | `Σ (Size - SharedSize)` of the unreferenced |
| Containers | all, stopped included | Running / Paused / Restarting | `Σ SizeRw` (read-write layer only) | `Σ SizeRw` of the non-active |
| Local Volumes | all | with at least one container | `Σ Size` | `Σ Size` of the unreferenced |
| Build Cache | all | in use | `Σ Size` | `Σ Size` of those neither in use nor shared |

- The Images `SIZE` is **not** the sum of the image sizes: what images share is counted once, so
  the row reports the space really taken on disk.
- A single image size is its present content plus its unpacked snapshots, the value `nerdctl
  images` already shows as `DISK USAGE`. The two commands agree. Docker sizes an image by walking
  its manifests, so the index listing them is counted in the row total while no single image is
  charged for it: an unused image is never quite 100% reclaimable.
- Deduplicating by target digest is required, not cosmetic: one target can be stored under several
  names (`repo:tag`, `repo@digest`, plus the config digest under `k8s.io`), which would otherwise
  multiply every count and every size.
- An image reports the `created` of its config, not the local record time, which only says when it
  was pulled or tagged and would show an old image as brand new. For a multi-platform image it is
  the config of the manifest the host platform matcher selects, since the platforms of an index are
  not necessarily built together; the record time stays as the fallback.

## Output

`-v/--verbose` prints the four detail sections with Docker's `SHARED SIZE` / `UNIQUE SIZE` columns.

`--format` also accepts the Docker `table TEMPLATE` form, e.g. `table {{.Type}}\t{{.Size}}`: the
literal `\t` and `\n` are expanded, the columns get an aligned header, and the labels go through
docker/cli's `HeaderFunctions`, so `table {{lower .Type}}` still names the column `TYPE`. No
nerdctl command supported `table TEMPLATE` so far, so the helper lives in `pkg/formatter` for the
others to adopt. Identifiers are shortened for table output only, matching `Format.IsTable()`, so
that a custom format stays usable to look a resource up.

## Notes

- Images, containers and volumes are namespaced, the build cache is not: it comes from the BuildKit
  host of that namespace, and shows as zeros when BuildKit is down, so the shape of the output does
  not depend on which daemons happen to be up.
- Split the way `system prune` already is: `pkg/cmd/system` orchestrates, each kind of resource is
  measured by its own package.
- Unit tests cover the shared-size accounting, the reclaimable percentage (the build cache never
  gets one), the summary and verbose rendering, `table TEMPLATE`, the platform selection of the
  build time, and the parsing of `buildctl du`. The integration tests run in a private namespace
  and check the counts after a pull, a run and a stop, plus the verbose sections and the three
  output formats; the counts are asserted as a difference from a baseline taken before the setup,
  since docker has no namespaces and its daemon holds whatever the other tests left behind. Only
  the Local Volumes row needs a `_linux` file, a mount target being written differently on each
  platform, so the rest runs on Windows too.

Closes #3942

### DIFF
--- a/pkg/cmd/container/df.go
+++ b/pkg/cmd/container/df.go
@@ -1,0 +1,158 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package container
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/core/snapshots"
+	"github.com/containerd/errdefs"
+	"github.com/containerd/log"
+
+	"github.com/containerd/nerdctl/v2/pkg/api/types"
+	"github.com/containerd/nerdctl/v2/pkg/containerdutil"
+	"github.com/containerd/nerdctl/v2/pkg/containerutil"
+	"github.com/containerd/nerdctl/v2/pkg/formatter"
+	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/dockercompat"
+	"github.com/containerd/nerdctl/v2/pkg/labels"
+	"github.com/containerd/nerdctl/v2/pkg/mountutil"
+)
+
+// DiskUsage reports how much disk space the containers of the current namespace use.
+//
+// Like Docker, only the read-write layer is counted: the layers coming from the image belong to the
+// image, not to the container. Every container is counted, including the stopped ones, and the space
+// of everything that is not running can be reclaimed.
+func DiskUsage(ctx context.Context, client *containerd.Client, _ types.GlobalCommandOptions, verbose bool) (types.ContainerDiskUsage, error) {
+	du := types.ContainerDiskUsage{}
+
+	containers, err := client.Containers(ctx)
+	if err != nil {
+		return du, err
+	}
+
+	snapshottersCache := map[string]snapshots.Snapshotter{}
+	for _, c := range containers {
+		info, err := c.Info(ctx, containerd.WithoutRefreshedMetadata)
+		if err != nil {
+			// There is no guarantee that a container we just listed still exists.
+			if errdefs.IsNotFound(err) {
+				log.G(ctx).Debugf("container %q is gone - ignoring", c.ID())
+				continue
+			}
+			return du, err
+		}
+
+		snapshotter, ok := snapshottersCache[info.Snapshotter]
+		if !ok {
+			snapshotter = containerdutil.SnapshotService(client, info.Snapshotter)
+			snapshottersCache[info.Snapshotter] = snapshotter
+		}
+
+		var sizeRw int64
+		if info.SnapshotKey != "" {
+			// Only the read-write layer is wanted, so ask the snapshotter for that one snapshot
+			// rather than walking the chain: a parent missing from the image the container was
+			// created from says nothing about the size of the container, and must not fail the
+			// report over it.
+			usage, err := snapshotter.Usage(ctx, info.SnapshotKey)
+			if err != nil {
+				// The read-write layer is the container, so a NotFound here means it was removed
+				// while we were measuring it.
+				if errdefs.IsNotFound(err) && containerIsGone(ctx, client, c.ID()) {
+					log.G(ctx).Debugf("container %q is gone - ignoring", c.ID())
+					continue
+				}
+				return du, fmt.Errorf("failed to get the size of container %q: %w", c.ID(), err)
+			}
+			sizeRw = usage.Size
+		}
+
+		status := formatter.ContainerStatus(ctx, c)
+
+		du.TotalCount++
+		du.TotalSize += sizeRw
+		if isActiveStatus(status) {
+			du.ActiveCount++
+		} else {
+			du.Reclaimable += sizeRw
+		}
+
+		if verbose {
+			item := types.ContainerDiskUsageItem{
+				ID:           c.ID(),
+				Image:        info.Image,
+				LocalVolumes: localVolumes(ctx, info.Labels),
+				SizeRw:       sizeRw,
+				CreatedAt:    info.CreatedAt,
+				Status:       status,
+				Names:        containerutil.GetContainerName(info.Labels),
+			}
+			if spec, err := c.Spec(ctx); err != nil {
+				log.G(ctx).WithError(err).Debugf("failed to get the spec of container %q", c.ID())
+			} else {
+				item.Command = formatter.InspectContainerCommand(spec, true, true)
+			}
+			du.Items = append(du.Items, item)
+		}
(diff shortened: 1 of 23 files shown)


AkihiroSuda:
Please rebase

AkihiroSuda:
Needs rebasing

pkg/cmd/volume/df.go
```diff
@@ -0,0 +1,76 @@
...
+// DiskUsage reports how much disk space the local volumes of the current namespace use.
+//
+// A volume is active when at least one container mounts it, and the space of the volumes no
+// container mounts can be reclaimed. Note that, unlike `nerdctl volume prune`, this counts the named
+// volumes too: Docker reports what `docker volume prune --all` would free.
+func DiskUsage(ctx context.Context, client *containerd.Client, gOptions types.GlobalCommandOptions, verbose bool) (types.VolumeDiskUsage, error) {
+	du := types.VolumeDiskUsage{}
+
+	// The size is what we are after here, so it is always requested.
+	vols, err := Volumes(gOptions.Namespace, gOptions.DataRoot, gOptions.Address, true, nil)
```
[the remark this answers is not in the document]
ekalinin: Confirmed. `volumeStore.List` skips any volume whose `rawGet` fails and still returns a nil error (`pkg/mountutil/volumestore/volumestore.go:219-223`), so `system df` can under-report the volume total.

I would rather not do this one here. `List` is part of the `VolumeStore` interface, and its only caller is `Volumes()` in `pkg/cmd/volume/list.go`, shared by `nerdctl volume ls`, `system df` and shell completion (`cmd/nerdctl/completion/completion.go:181`). Strictness cannot be a blanket change there - completion has to stay best-effort - so it needs a new opt-in listing mode on the store, plus a decision on what `volume ls` should report when a volume is unreadable. That is a change to a shared subsystem with its own compatibility surface, and the best-effort behaviour predates this PR.

Proposal: a follow-up PR adding a strict listing mode to the volume store and switching the accounting path to it, keeping this PR scoped to `system df`. The other two comments in this review are valid and I am fixing them here.

Happy to fold it into this PR instead if maintainers prefer.

pkg/cmd/image/df.go
```diff
@@ -0,0 +1,372 @@
...
+	imageList, err := List(ctx, client, nil, nil)
+	if err != nil {
+		return du, err
+	}
+
+	var (
+		contentStore = client.ContentStore()
+		provider     = containerdutil.NewProvider(client)
+		snapshotter  = containerdutil.SnapshotService(client, gOptions.Snapshotter)
+		containers   = imagesInUse(ctx, client)
```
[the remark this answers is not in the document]
ekalinin: Fixed.

pkg/cmd/image/df.go
```diff
@@ -0,0 +1,372 @@
...
+
+	collected := make([]*imageContents, 0, len(uniqueImages))
+	index := newDiskUsageIndex()
+
+	for _, img := range uniqueImages {
+		contents, err := readImageContents(ctx, contentStore, provider, img)
+		if err != nil {
+			log.G(ctx).WithError(err).Warnf("failed to compute the disk usage of image %q", img.Name)
+			continue
+		}
```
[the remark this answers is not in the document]
ekalinin: Fixed.

pkg/cmd/image/list.go
```diff
@@ -700,28 +705,31 @@ type platformRef struct {
...
+	inUse := map[digest.Digest]int64{}
 	inUseByPlatform := map[platformRef]bool{}
 	containerList, err := client.Containers(ctx)
 	if err != nil {
-		log.G(ctx).WithError(err).Warn("failed to list containers for image in-use detection")
-		return inUse, inUseByPlatform
+		return nil, nil, fmt.Errorf("failed to list containers for image in-use detection: %w", err)
 	}
 	for _, container := range containerList {
 		if dgst, ok := containerImageDigest(ctx, container); ok {
```
ChengyuZhu6: `image.DiskUsage` aborts if `in-use` is unknown, then treats `inUse == 0` as reclaimable. Here a missing `pin + container.Image()` error just returns false. That image is reported unused and its unique size is reclaimable. Docker leaves reclaimable alone when `Containers < 0`. Fail closed, or count as in-use.

pkg/cmd/container/df.go
```diff
@@ -0,0 +1,158 @@
...
+				// parent snapshot of a container that is still there. Only the container store can
+				// tell the two apart, and dropping a live container would understate the report as
+				// silently as counting it as empty would.
+				if errdefs.IsNotFound(err) && containerIsGone(ctx, client, c.ID()) {
+					log.G(ctx).Debugf("container %q is gone - ignoring", c.ID())
+					continue
+				}
+				return du, fmt.Errorf("failed to get the size of container %q: %w", c.ID(), err)
+			}
+			sizeRw = rw.Size
```
ChengyuZhu6: `ResourceUsage` walks parents; RW `Usage` can already be in `first` when a parent `NotFounds`. You only continue if the container is gone, otherwise the whole `system df` fails. Docker keeps `SizeRw` and warns.

pkg/cmd/image/list.go
ekalinin: Fixed. `containerImageDigest` now returns the error. `NotFound` stays benign - per containerd `client/container.go:211-224` that covers a container removed mid-listing, a container created from no image (containerd wraps it in `errdefs.ErrNotFound` explicitly), and an image already gone; none of those name an image the report covers. Anything else fails the accounting instead of silently marking the image reclaimable.

pkg/cmd/container/df.go
ekalinin: Fixed. Switched to `snapshotter.Usage` on the container snapshot key: only the read-write layer is wanted here, the total was already discarded, so the parent walk could only ever fail the report over something that does not affect `SizeRw`. `containerIsGone` still covers a container removed while measuring. `ps -s` keeps `ResourceUsage`, it genuinely needs the virtual size.

ChengyuZhu6 (approved):
LGTM

AkihiroSuda (approved):
Thanks

PR #4627, opened 2025-11-29
Support `push --all-tags` to push all tags

yashkukrecha:
This PR adds Docker-compatible --all-tags support and fixes a SOCI regression when pushing multiple tags of the same image.

Key changes:
- New `--all-tags` flag for `image push`
- SOCI integration fix for multi-tag pushes

Fixes #3751

PR #5133, opened 2026-08-08
feat(push): support `--all-tags` to push all tags

### LINKED ISSUE #3751: nerdctl push does not push all tags
### Description

The push command despite making the tag optional, does not push all tagged local images like `docker push` does (https://docs.docker.com/reference/cli/docker/image/push/#all-tags). I see this is documented in the [command reference](https://github.com/containerd/nerdctl/blob/1220ce7ec2701d485a9b1beeea63dae3da134fb5/docs/command-reference.md?plain=1#L786), but is there a reason?

In addition, we need to push tags and the image needs an soci index. At the moment, I have to run 2 nerdctl push commands and the second one will overwrite the firsts soci index.

1. run nerdctl tag twice and then push without a tag
```
nerdctl tag ${{ steps.login-ecr-pull.outputs.registry }}/${{ inputs.pull_repo }}:${{ inputs.image_tag }} ${{ steps.login-ecr-push.outputs.registry }}/${{ inputs.push_repo }}:${{ inputs.image_tag }}
nerdctl tag ${{ steps.login-ecr-pull.outputs.registry }}/${{ inputs.pull_repo }}:${{ inputs.image_tag }} ${{ steps.login-ecr-push.outputs.registry }}/${{ inputs.push_repo }}:latest
nerdctl push --soci-min-layer-size ${{ inputs.soci-min-layer-size }} --snapshotter soci ${{ steps.login-ecr-push.outputs.registry }}/${{ inputs.push_repo }}
```
2. Only the `latest` tag will show up in the registry.

Only the `latest` tag is pushed, the other tag is not.

```
Client:
 Version:       v2.0.1
 OS/Arch:       linux/arm64
 Git commit:    47f31ff2c1615c1accb85c1ce4e7882ad739102f
 buildctl:
  Version:      v0.18.0
  GitCommit:    95d190ef4f18b57c717eaad703b67cb2be781ebb

Server:
 containerd:
  Version:      v2.0.0
  GitCommit:    207ad711eabd375a01713109a8a197d197ff6542
 runc:
  Version:      1.2.2
  GitCommit:    v1.2.2-0-g7cb36325
```

None

### Host information

```
Client:
 Namespace:     default
 Debug Mode:    false

Server:
 Server Version: v2.0.0
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Storage: native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
  rootless
 Kernel Version:   6.8.0-1019-aws
 Operating System: Ubuntu 22.04.5 LTS
 OSType:           linux
 Architecture:     aarch64
 CPUs:             2
 Total Memory:     3.742GiB
 Name:             ip-192-168-2-200
 ID:               cd73469d-00ea-46d9-ad10-a77473f04972

WARNING: No cpu cfs period support
WARNING: No cpu cfs quota support
WARNING: No cpu shares support
WARNING: No cpuset support
WARNING: bridge-nf-call-iptables is disabled
WARNING: bridge-nf-call-ip6tables is disabled
```

ekalinin:
Closes #3751.

`nerdctl push` accepts a bare repository name, but `referenceutil.Parse` normalizes it to `:latest`, so only that single tag is pushed. This adds the Docker-compatible `-a, --all-tags` flag, which pushes every local tag of the repository instead.

`Push()` is split into a dispatcher and `pushSingle()`. With `--all-tags` the local tags are resolved through the `name~=^<repo>:` image filter - the same idiom `nameFilterFor()` uses for `nerdctl image ls` - then sorted by name and pushed one by one. The `-tmp-reduced-platform` / `-tmp-esgz` images `push` creates for itself are skipped, so an interrupted push cannot leak one into the registry as a real tag.

A tag or a digest in the reference is rejected, as `docker push --all-tags` does. The check looks at `ExplicitTag`, not `Tag`: `Parse()` runs `TagNameOnly()`, so `Tag` is `"latest"` even for a bare repository name.

The issue also reports that pushing several tags of one image overwrites the SOCI index each time. A SOCI index is attached to the image manifest rather than to the tag, so it is now built once per distinct target digest - deduplicating by digest rather than by position keeps each image indexed when the tags differ.

Tests: four sub-tests in `TestPush` covering all tags pushed (verified against `/v2/<repo>/tags/list`), explicit tag rejected, no local tags rejected, and SOCI with `--all-tags`.

Note: #4627 is an earlier attempt at this issue, inactive since December 2025. This is an independent implementation. For whoever picks that one up: its guard is `parsedReference.Tag != ""`, which is always true after `TagNameOnly()` normalization, so `--all-tags` there always fails with `tag can't be used with --all-tags/-a`.

### DIFF
--- a/cmd/nerdctl/image/image_push_linux_test.go
+++ b/cmd/nerdctl/image/image_push_linux_test.go
@@ -17,9 +17,12 @@
 package image
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
+	"slices"
+	"strings"
 	"testing"
 
 	"gotest.tools/v3/assert"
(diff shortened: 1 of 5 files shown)


cmd/nerdctl/image/image_push_linux_test.go
```diff
@@ -278,7 +280,124 @@ func TestPush(t *testing.T) {
...
+						helpers.Anyhow("rmi", "-f", data.Labels().Get("testImageRepo")+":v2")
+					}
+				},
+				Command: func(data test.Data, helpers test.Helpers) test.TestableCommand {
+					return helpers.Command("push", "--snapshotter=soci", "--insecure-registry", "--all-tags", "--soci-span-size=2097152", "--soci-min-layer-size=20971520", data.Labels().Get("testImageRepo"))
+				},
+				Expected: func(data test.Data, helpers test.Helpers) *test.Expected {
+					return &test.Expected{
+						Output: func(stdout string, t tig.T) {
+							assertRegistryHasTags(t, registryNoAuthHTTPRandom, data.Identifier(), "v1", "v2")
```
haytok, line 368: should confirm whether SOCI index is pushed or not.

pkg/cmd/image/push.go
```diff
@@ -57,8 +60,74 @@ import (
...
+	if len(imgs) == 0 {
+		return fmt.Errorf("an image does not exist locally with the tag: %s", parsedReference.Name())
+	}
+
+	// A SOCI index is attached to the image manifest rather than to the tag, so it only needs to be
+	// built once per distinct target. Doing it per tag makes every tag overwrite the index pushed by
+	// the previous one: https://github.com/containerd/nerdctl/issues/3751
+	indexed := make(map[digest.Digest]struct{}, len(imgs))
+	for _, img := range imgs {
+		_, done := indexed[img.Target.Digest]
```
haytok: nit:

`done` doesn't say what's done, so would `alreadyIndexed` be better ?

pkg/cmd/image/push.go
```diff
@@ -57,8 +60,74 @@ import (
...
+		return strings.HasSuffix(img.Name, tmpReducedPlatformSuffix) || strings.HasSuffix(img.Name, tmpEsgzSuffix)
+	})
+	// ImageService().List does not guarantee an order, and the order decides which tag gets indexed.
+	slices.SortFunc(imgs, func(a, b images.Image) int {
+		return strings.Compare(a.Name, b.Name)
+	})
+	return imgs, nil
+}
+
+func pushSingle(ctx context.Context, client *containerd.Client, rawRef string, options types.ImagePushOptions, skipSoci bool) error {
```
haytok: nit:

`alreadyIndexed` would be better rather than `skipSoci` here.

pkg/cmd/image/push.go
```diff
@@ -57,8 +60,74 @@ import (
...
+	// built once per distinct target. Doing it per tag makes every tag overwrite the index pushed by
+	// the previous one: https://github.com/containerd/nerdctl/issues/3751
+	indexed := make(map[digest.Digest]struct{}, len(imgs))
+	for _, img := range imgs {
+		_, done := indexed[img.Target.Digest]
+		if err = pushSingle(ctx, client, img.Name, options, done); err != nil {
+			return err
+		}
+		indexed[img.Target.Digest] = struct{}{}
+	}
```
haytok, line 107: SOCI Index is deduplicated based on the image digest ID, but is it intentional that the same processing for estargz and cosign isn't performed?

pkg/cmd/image/push.go
```diff
@@ -57,8 +60,74 @@ import (
...
+	return nil
+}
+
+// localTags returns the local images tagged under the repository `name`, sorted by name.
+func localTags(ctx context.Context, client *containerd.Client, name string) ([]images.Image, error) {
+	// Same idiom as nameFilterFor() in cmd/nerdctl/image/image_list.go: a bare repository name
+	// matches every tag of that repository (pkg/ cannot import cmd/, hence the repeated filter).
```
haytok: Maybe redundant.

cmd/nerdctl/image/image_push_linux_test.go
ekalinin: Ofc. Fixed.

pkg/cmd/image/push.go
ekalinin: Agreed. Fixed.

pkg/cmd/image/push.go
ekalinin: Fixed.

pkg/cmd/image/push.go
ekalinin: eStargz is a different kind of thing and cannot be skipped the same way: the converted image is the payload pushed under each tag, so skipping it for the second tag would push the wrong content. What could be deduplicated is the conversion work (convert once per digest, push the result under each tag), but that needs `pushSingle` restructured, since it currently owns the temp image lifecycle for a single ref. I also have not verified that eStargz conversion is byte-reproducible; if it is not, two tags of one source image end up on different manifests, which is pre-existing but more visible under `--all-tags`.

Signing was not a deliberate exclusion. `signutil.Sign` gets `<pushRef>@<digest>` and passes it to `cosign sign / notation sign`, both of which key the signature by digest in the same repository, so deduplicating it by digest would be correct too. Happy to add it. It would mean renaming `indexed/alreadyIndexed` to something covering both.

pkg/cmd/image/push.go
ekalinin: Removed.

AkihiroSuda:
Please squash the commits

ktock (approved):

haytok (approved):
LGTM

ekalinin:
> Please squash the commits

Done.

AkihiroSuda (approved):
Thanks

PR #5147, opened 2026-08-19
feat(logging): add --log-file to persist nerdctl's own log

### LINKED ISSUE #4872: can nerdctl support to save log ?

containerd does't to record any useful log when nerdctl failed to create a container. we can save these log into a file.

nerdctl can save log into a file

### Additional context

_No response_

ekalinin:
Fixes #4872

## What

Adds a global `--log-file` (also `log_file` in `nerdctl.toml`, and `$NERDCTL_LOG_FILE`) that appends nerdctl's own log to a file, in addition to the standard error. Empty by default, so nothing changes unless it is set.

```console
$ nerdctl --debug --log-file /var/log/nerdctl.log run --rm alpine true
```

## Why

#4872 asks for a way to keep a record of a failing nerdctl invocation: today the diagnostics only go to the standard error and are gone once the process exits, and containerd does not record the client side of a failed container creation either.

Everything nerdctl logs funnels through the standard logger, including the error that terminates the command: `SilenceErrors: true` on the root command means cobra prints nothing itself, the error returns to `main()`, and `main()` calls `log.L.Fatal(err)`. Capturing that stream is therefore enough to capture the failure the issue is about, together with everything logged on the way there.

## Implementation notes

- The file output is attached as a **logrus hook**, not by replacing `Logger.Out` with an `io.MultiWriter`. The text formatter picks its output style by type-asserting `Logger.Out` to `*os.File`, so a MultiWriter would silently downgrade the console output from `INFO[0000] msg` to `time="..." level=info msg="..."` for everyone enabling the flag. With a hook the console is left untouched and the file gets a stable, TTY-independent format.
- No new dependency: the hook is written against the type aliases exported by `containerd/log`, so nerdctl still does not import logrus directly.
- Record format is `<RFC3339NanoFixed> <LEVEL> <message> [key="value" ...]`, fields sorted for stable output.
- Opened with `O_CREATE|O_WRONLY|O_APPEND`, mode `0600`, so concurrent invocations can share one file and a previous invocation is never truncated away. Rotation is deliberately left to `logrotate` or an equivalent - pulling in a rotation dependency for a CLI seemed disproportionate.
- The file is left open on purpose: `log.L.Fatal` calls `os.Exit`, so a deferred `Close` would not run, and the hook writes are unbuffered.
- The level keeps following the existing `--debug` / `--debug-full`, so no new `--log-level` flag is introduced.

## Not covered

- The standard error of external binaries nerdctl execs (`buildctl`, CNI plugins), which never comes back through the logger.
- containerd's own daemon-side logs, which is what containerd/containerd#13194 was about.
- Logging plugin mode (`binary://`) is dispatched in `xmain` before cobra runs, so the flag does not apply there.

## Testing

- `pkg/logging/file_hook_test.go`: record format and field ordering, append semantics, file mode, and that `Logger.Out` is left alone.
- `cmd/nerdctl/main_test.go` `TestLogFile`: a failing invocation writes the error to the file, and two invocations both end up in it.

Example output:

```console
$ nerdctl --debug --log-file /tmp/nerdctl.log version
$ cat /tmp/nerdctl.log
2026-08-19T22:57:03.776364000+03:00 WARNING unable to determine buildctl version error="exec: \"buildctl\": executable file not found in $PATH"
2026-08-19T22:57:03.776581000+03:00 FATAL cannot access containerd socket "/var/run/containerd/containerd.sock": no such file or directory
```

### DIFF
--- a/cmd/nerdctl/helpers/flagutil.go
+++ b/cmd/nerdctl/helpers/flagutil.go
@@ -85,6 +85,10 @@ func ProcessRootCmdFlags(cmd *cobra.Command) (types.GlobalCommandOptions, error)
 	if err != nil {
 		return types.GlobalCommandOptions{}, err
 	}
+	logFile, err := cmd.Flags().GetString("log-file")
+	if err != nil {
+		return types.GlobalCommandOptions{}, err
+	}
 	address, err := cmd.Flags().GetString("address")
 	if err != nil {
 		return types.GlobalCommandOptions{}, err
@@ -167,6 +171,7 @@ func ProcessRootCmdFlags(cmd *cobra.Command) (types.GlobalCommandOptions, error)
 	return types.GlobalCommandOptions{
 		Debug:            debug,
 		DebugFull:        debugFull,
+		LogFile:          logFile,
 		Address:          address,
 		Namespace:        namespace,
 		Snapshotter:      snapshotter,
--- a/cmd/nerdctl/image/image_convert_test.go
+++ b/cmd/nerdctl/image/image_convert_test.go
@@ -65,6 +65,7 @@ func addRootFlagsForConvertOptionsTest(t *testing.T, cmd *cobra.Command) {
 	flags := cmd.Flags()
 	flags.Bool("debug", false, "")
 	flags.Bool("debug-full", false, "")
+	flags.String("log-file", "", "")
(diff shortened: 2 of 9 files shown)


AkihiroSuda (approved):
Thanks

PR #5150, opened 2026-08-20
feat: add --disable-content-trust flag to improve docker CLI compatibility

### LINKED ISSUE #5053: add support for `--disable-content-trust` flag

`nerdctl run` currently lists `--disable-content-trust` as an unimplemented Docker compatibility flag.
https://github.com/containerd/nerdctl/blob/460875345229737b441a3a8618679dd0522e3d17/docs/command-reference.md?plain=1#L457-L460
Docker uses this flag to disable Docker Content Trust for a single command. However, `nerdctl` does not implement Docker Content Trust and instead provides image signing and verification through cosign.

Before implementing this flag, it would be useful to clarify its expected semantics in the context of nerdctl's existing image verification mechanisms.

TBD

### Additional context

Podman accepts `--disable-content-trust` as a Docker compatibility flag and documents it as a no-op because it does not implement Docker Content Trust (see here: https://docs.podman.io/en/latest/search.html?q=--disable-content-trust) . Since nerdctl uses a different trust model based on cosign, it would be helpful to define the desired behavior before implementation

Pratik-050:
Fixes #5053 

### Why
Docker scripts may pass --disable-content-trust when running, pulling, or pushing images. Without support for the flag, those commands fail with an unknown flag error.

nerdctl uses a different trust model:
--verify=cosign|notation for image verification
--sign=cosign|notation for image signing

Accepting the flag as a no-op improves Docker CLI compatibility without incorrectly mapping Docker Content Trust semantics onto nerdctl’s existing verification mechanisms.

### Implementation Notes
- Registers --disable-content-trust as a boolean flag on run, pull, and push.
- The flag is intentionally not used to modify image verification or signing behavior.
- Existing --verify and --sign options continue to control nerdctl’s trust operations.
- No new dependencies or trust-related behavior were introduced.

### Documentation
Updates command-reference.md to:

- Document --disable-content-trust as a Docker compatibility no-op.
- Remove it from the unimplemented Docker flag lists for run, pull, and push.
- Clarify that nerdctl uses cosign and notation instead of Docker Content Trust.

### Testing
Added TestRunDisableContentTrustFlag to verify that nerdctl run accepts the flag and parses it as true.
Verified the existing run help tests continue to pass.
Verified the image command package compiles with the new pull and push flags.

### DIFF
--- a/cmd/nerdctl/container/container_run.go
+++ b/cmd/nerdctl/container/container_run.go
@@ -304,6 +304,7 @@ func setCreateFlags(cmd *cobra.Command) {
 	cmd.RegisterFlagCompletionFunc("verify", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return []string{"none", "cosign", "notation"}, cobra.ShellCompDirectiveNoFileComp
 	})
+	cmd.Flags().Bool("disable-content-trust", false, "No-op for Docker compatibility; nerdctl uses --verify for image signing and verification")
 	cmd.Flags().String("cosign-key", "", "Path to the public key file, KMS, URI or Kubernetes Secret for --verify=cosign")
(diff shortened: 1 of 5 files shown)


AkihiroSuda:
https://github.com/containerd/nerdctl/issues/5053#issuecomment-5356316464

Issue #3577, opened 2024-10-19
Tracking CI runtime


apostasie:
With the current effort on testing, we should start looking into our run duration, platform coverage, and hard data on flakyness / disabled tests.

This issue here is to record that information over time.

This is the first sample (of #3535), ran on 10/19/2024

| Target             | Prep Time | Non-flaky                                                                      | Flaky                                                  |
|--------------------|-----------|--------------------------------------------------------------------------------|--------------------------------------------------------|
| windows            | 238 s     | 256 tests, 159 skipped in 81.173s                                              | 166 tests, 66 skipped in 263.612s                      |
| rootful-1.6        | 277 s     | 562 tests, 276 skipped in 120.279s                                             | 2 runs, 737 tests, 81 skipped, 2 failures in 1024.881s |
| rootful-1.7        | 292 s     | 562 tests, 276 skipped in 141.374s                                             | 735 tests, 76 skipped in 1134.421s                     |
| rootful-2          | 273 s     | 562 tests, 276 skipped in 124.186s                                             | 735 tests, 80 skipped in 1041.623s                     |
| rootful-2-arm      | **458 s** | **567 tests, 275 skipped in 185.060s**                                         | **735 tests, 81 skipped in 1308.355s**                                                       |
| rootless-1.6       | 283 s     | 502 tests, 305 skipped in 161.850s                                             | 727 tests, 143 skipped in 893.197s                     |
| rootless-1.7       | 272 s     | 550 tests, 274 skipped in 130.962s                                             | 727 tests, 115 skipped in 1020.648s                    |
| rootless-1.7-slirp | 288 s     | 550 tests, 274 skipped in 151.411s                                             | 727 tests, 115 skipped in 1012.942s                    |
| rootless-2         | 274 s     | 550 tests, 274 skipped in 135.481s                                             | 2 runs, 729 tests, 119 skipped, 2 failures in 937.661s |
| ipv6               | 272 s     | 341 tests, 328 skipped in 70.056s                                              | na                                                     |
| kube               | 153 s     | 33.571 s                                                                       | na                                                     |
| docker             | 25 s      | 467 tests, 302 skipped in 126.707s<hr/>ipv6: 331 tests, 325 skipped in 15.012s | CURRENTLY BUST                                                   |

Here is the list of the worth offenders (taken from the arm run):

```
github.com/containerd/nerdctl/v2/cmd/nerdctl/container TestCopyToContainer 2m38.51s
[...]

PR #5151, opened 2026-08-20
Fix foreground stdio deadlock when the internal logging process stops consuming

### LINKED ISSUE #5137: `nerdctl run` with attached stdio stops draining container stdout at ~72 KiB, deadlocking the container (nerdctl 2.3.5, containerd 2.3.3, Kata runtime-rs)
## Summary

With `nerdctl run --rm -i` (foreground, non-TTY, stdio attached), the attach-side stdout consumer stops reading after roughly 72 KiB have been delivered to the host. The container's writer then blocks forever on a full pipe and the whole chain deadlocks: the container never exits, `nerdctl` never returns, and signals sent to the `nerdctl` client do not tear the task down.

The same image, same containerd, same runtime, driven through containerd's own `ctr run` client instead of `nerdctl`, streams 64 MiB of stdout cleanly. That is what points at the nerdctl attach path rather than at the runtime or the guest.

## Environment

| Component | Version / detail |
|---|---|
| nerdctl | 2.3.5 (official release binary, checksum-verified) |
| containerd | 2.3.3 |
| Runtime | Kata Containers 4.0.0, **runtime-rs** shim (`containerd-shim-kata-v2` from the kata-static `runtime-rs/bin` tree) |
| Hypervisor actually used | QEMU (`qemu-system-x86_64`), via the shim's vendored `configuration-qemu-runtime-rs.toml` |
| Guest stdio pipe size | Kata `container_pipe_size` effective value `0` → OS default (~64 KiB) |
| Host OS | Ubuntu 24.04.4, kernel 6.17.0-1021-azure (x86_64, KVM) |
| Guest kernel | 6.18.35 |
| runc | **Not installed on this host — see "What was and was not tested"** |
| BuildKit / CNI | not installed; container had `--network none` |

## Reproduction

Everything below is generic; run as root on a host with containerd + nerdctl + a Kata runtime handler registered. `$RUNTIME` is the Kata shim handler name (ours resolved to the runtime-rs shim); `$IMAGE` is any image with a POSIX shell and `dd` already present in the content store (we used `--pull never` on an air-gapped host).

**A. Through nerdctl — stalls**

```sh
# 4 MiB of stdin, consumed to EOF first, then 8 MiB written to stdout
head -c 4194304 /dev/zero > /tmp/stdin.bin

timeout 90 nerdctl run --rm -i --pull never --runtime "$RUNTIME" "$IMAGE" \
  sh -c 'cat > /dev/null; dd if=/dev/zero bs=1M count=8 2>/dev/null' \
  < /tmp/stdin.bin > /tmp/out.bin
echo "exit=$?"
wc -c < /tmp/out.bin
```

Observed: `nerdctl` never returns; `timeout` fires (`exit=124`). Sampling `/tmp/out.bin` while the run is wedged shows delivery starting and then stopping — in our live-sampled reproduction it stopped at exactly **73,728 bytes** (72 KiB = a 64 KiB pipe plus an 8 KiB copy buffer) and never advanced again.

**B. Same image and runtime through `ctr` — completes**

```sh
timeout 90 ctr run --rm --runtime "$RUNTIME" "$IMAGE" probe-ctr \
  sh -c 'dd if=/dev/zero bs=1M count=8 2>/dev/null' > /tmp/out2.bin
echo "exit=$?"
wc -c < /tmp/out2.bin
```

Observed: `exit=0`, all 8,388,608 bytes delivered. We ran this control across a size sweep of 0.25 / 0.5 / 1 / 2 / 4 / 8 / 16 / 32 / 64 MiB, with both a single monolithic `write()` and 64 KiB chunked writes awaiting drain, and with a bidirectional shape (4 MiB in to EOF, then 30 MiB out). Every cell passed, including 64 MiB monolithic (all bytes delivered, ~8.7 s wall including VM boot) and the 30 MiB bidirectional case (~1.2 s in-guest).

## Observed vs expected

**Expected:** `nerdctl run` with attached stdio keeps draining the container's stdout for as long as the container writes, as `ctr` does; the container runs to completion and `nerdctl` exits with the container's status.

**Observed:**

1. Host-side stdout delivery stops at ~72 KiB (73,728 bytes in the sampled run) and never resumes.
2. The guest writer blocks on the now-full pipe; the container never exits and CPU goes idle.
3. Thread/wchan sampling of the wedged process chain shows the **Kata runtime-rs shim's tokio worker blocked in `anon_pipe_write`** — i.e. it is trying to write stdout *toward nerdctl* — while **every `nerdctl` thread sits idle in `futex`/`epoll`**. Nothing on the nerdctl side is consuming.
4. Signals do not get through the wedged attach: `SIGTERM` to the `nerdctl` client (via `timeout`) killed nothing; one specimen sat wedged for 19+ minutes.
5. Recovery through nerdctl is also blocked: `nerdctl rm --force <name>` hangs. Killing the `nerdctl` client leaves the shim, the hypervisor process and the virtiofsd processes alive (a leaked VM). `ctr task kill -s SIGKILL <id>` returns 0 but the task stays `RUNNING`; `ctr task rm -f` hangs. Only killing the hypervisor and virtiofsd PIDs first, then the shim, reclaims everything — after which `ctr container rm` and `nerdctl rm --force` both succeed.

Point 3 is the core of the report: the writer is blocked writing to nerdctl, and nerdctl is asleep. That is a stalled reader on the attach side rather than a runtime or guest-agent problem.

## The ~72 KiB threshold

73,728 bytes = 64 KiB + 8 KiB, which lines up with a default-sized (64 KiB) stdio pipe plus one 8 KiB copy buffer already consumed. That is consistent with the attach-side reader performing one buffer's worth of copying and then never scheduling another read, leaving the pipe permanently full. We did not bisect the threshold with a size sweep through nerdctl, so treat 73,728 as one precisely measured wedge point rather than a proven constant; the qualitative finding is that small outputs (diagnostics, short JSON) always got through and every real payload never did.

## What was and was not tested

Stated explicitly so nothing here is over-claimed:

- **Reproduced:** nerdctl 2.3.5 + containerd 2.3.3 + Kata 4.0.0 runtime-rs shim under QEMU, non-TTY, `--interactive`, with stdin fed and read to EOF before the large stdout write. Reproduced with a real workload emitting a large JSON result, and confirmed by live thread/wchan sampling of the wedged chain.
- **Ruled out (same host, same runtime, same image):** the Kata guest, the guest agent, and the runtime-rs shim's own stdio copy path — `ctr` drains the identical FIFOs at up to 64 MiB without issue, monolithic and chunked alike. Request/stdin size was never a factor (a 35 KB stdin request reproduced the stall).
- **NOT tested: runc.** `runc` is deliberately not installed on this host (the only registered runtime is the Kata shim), so we have **no** data on whether the same stall occurs with `nerdctl run --runtime runc`. We therefore cannot say whether this is Kata-specific or a general property of nerdctl's attach path. The evidence isolates the fault to nerdctl's attach-side reader *relative to `ctr` on the same runtime*, and no further.
- **NOT tested:** TTY mode (`-t`), detached mode with `nerdctl logs`, running without `--interactive`/without stdin attached, other log drivers, and other nerdctl versions. Every nerdctl reproduction we have involves `-i` with stdin attached, so we cannot say whether stdin attachment is required to trigger it.
- **NOT tested:** whether raising Kata's `container_pipe_size` (which is honored end-to-end down to `F_SETPIPE_SZ` in the guest, but is `0`/default in the shipped QEMU runtime-rs template) merely raises the threshold. It would not address a reader that stops reading.

## Impact / workaround

Any workload whose stdout exceeds ~72 KiB under `nerdctl run` with attached stdio deadlocks permanently and leaks the sandbox VM, since neither signal proxying nor `nerdctl rm --force` can reach it. Our workaround was to move all bulk output off stdio entirely (onto a bind-mounted output file), keeping stdio for small control-plane messages only, plus an out-of-band PID-kill escalation for teardown. Happy to run further probes on this host if a maintainer wants a specific variant tested (TTY, no-stdin, alternate log driver).

gsaddict91:
Fixes #5137.

`nerdctl run` in the foreground tees container stdout/stderr through a 64 KiB pipe to the forked `_NERDCTL_INTERNAL_LOGGING` process, on the same goroutine that drains the container's stdio FIFOs. If the logging process stops consuming for any reason, the tee write blocks forever — nerdctl keeps its own copies of the logger pipe read ends open, so the kernel never raises EPIPE — and the whole chain deadlocks: output freezes at pipe-capacity + one writer chunk, the container blocks on a full pipe and never exits, `nerdctl run` never returns, and `nerdctl rm -f` on the wedged container hangs too. Full analysis in https://github.com/containerd/nerdctl/issues/5137#issuecomment-5349093313.

### Changes

1. **Close the parent's copies of the logger pipe read ends after the logging process starts** (`pkg/cioutil/container_io.go`) — the equivalent of containerd's `binaryIO.CloseAfterStart`. Logger death now surfaces as EPIPE instead of an eternal block. The pipe ends are wrapped in idempotent closers so the success path and the error-path cleanup can both close them safely.

2. **Make the logger leg of the stdio tee best-effort** (`bestEffortWriter` in `pkg/cioutil/container_io.go`). While implementing (1) I found it is necessary but not sufficient: with the read ends closed, killing the logger indeed produced EPIPE — `/proc` showed the read ends gone and the stdout copy goroutine exited — but the run *still* wedged, because the EPIPE errored the `io.MultiWriter`, which aborted the `io.CopyBuffer` draining the container's stdout FIFO, and the container then blocked on the undrained FIFO chain. So the tee now warns once and stops writing to the logger on the first failed write, while continuing to stream to the attached stdout/stderr. The attach and the container survive logger death; the log file is what goes incomplete. (This is the "attach shouldn't be gated on a best-effort logging process" point from the issue — it turned out to be required, not optional.)

3. **Don't treat an errored wait-channel delivery as a container exit** (`pkg/logging/logging.go`). containerd's client sends Wait RPC failures through the exit channel as a synthetic `ExitStatus`; the logger's bare `<-exitCh` treated that as an exit, cancelled its readers, and silently stopped all logging while the container was still running — the suspected initiator of the original Kata-environment wedge. The logger now checks `status.Error()` and re-arms the wait (1s delay, aborting on SIGTERM). `getContainerWait` also closes each containerd client once its wait delivers, so re-arming doesn't accumulate open clients.

4. **Fail IO setup when a `binary-v2` logging binary exits before signalling readiness**, mirroring containerd's `n == 0` check. Plain `binary://` keeps EOF-as-ready for backward compatibility with third-party logging binaries (nerdctl's own internal logger URI is `binary://`, and with (1)+(2) a logger that dies before `ready()` now degrades gracefully instead of wedging).

### Verification

**Unit tests** (new, both fail on `main`, pass with this PR):
- `TestLoggingProcessAdapterWaitError` — feeds the adapter an errored `ExitStatus` followed by a real exit; asserts output produced after the errored delivery is still logged, the wait is re-armed, and the real exit still terminates the logger. On `main` it fails at "logger did not re-arm the container wait".
- `TestBestEffortWriter` / `TestBestEffortWriterClosedPipe` — the tee leg reports full writes and abandons the underlying writer after the first failure, including real EPIPE on a closed `os.Pipe`.

**End-to-end A/B** (privileged Linux container: containerd v2.3.3, runc 1.4.3, linux/arm64, busybox; the no-Kata reproducer from the issue — container writes 8 MiB in 8 KiB chunks with a 20 ms sleep, `kill` the `_NERDCTL_INTERNAL_LOGGING` process mid-stream):

| | unpatched `main` | this PR |
|---|---|---|
| output delivered | frozen mid-stream, never advances | **8,388,608 / 8,388,608 bytes** |
| `nerdctl run` | never returns (>45 s, killed manually) | **returns, exit 0** |
| `nerdctl rm -f` | hangs (10 s timeout) | n/a (`--rm` completed) |
| leaked containers | 1 | 0 |

Same result for SIGTERM and SIGKILL of the logger. Healthy-path controls on the patched build: untouched logger streams the full 8 MiB with exit 0 and no warning; `nerdctl logs` shows the expected output afterwards.

The runtime-side half of the original report (kata runtime-rs holding its own FIFO read end and gating exit reporting on the stdio copy tasks) is independent and tracked in kata-containers/kata-containers#13676.

### DIFF
--- a/pkg/cioutil/container_io.go
+++ b/pkg/cioutil/container_io.go
@@ -32,6 +32,7 @@ import (
 	"github.com/containerd/containerd/v2/cmd/containerd-shim-runc-v2/process"
 	"github.com/containerd/containerd/v2/defaults"
 	"github.com/containerd/containerd/v2/pkg/cio"
+	"github.com/containerd/log"
 )
 
 const binaryIOProcTermTimeout = 12 * time.Second // Give logger process 10 seconds for cleanup
@@ -50,6 +51,44 @@ var bufPool = sync.Pool{
 		buffer := make([]byte, 32<<10)
 		return &buffer
 	},
+}
+
+// closeOnce wraps f's Close so that extra calls return the first result
+// instead of "file already closed". The logger pipe ends below are closed
+// individually on the success path, but also sit in the error-path closers
+// list of NewContainerIO.
+func closeOnce(f *os.File) func() error {
+	var once sync.Once
+	var err error
+	return func() error {
+		once.Do(func() { err = f.Close() })
+		return err
+	}
+}
+
+// bestEffortWriter forwards writes to w until one fails, then silently
+// discards all further writes. The pipe feeding the logging binary is wrapped
+// in this before it joins the stdio tee of a foreground container: logging is
+// best-effort there, and a dead logging binary (EPIPE once our read ends are
+// closed, see NewContainerIO) must not error the whole tee — that would stop
+// the copier that drains the container's stdio FIFO and deadlock the
+// container. The attach keeps streaming; the log is what goes incomplete.
+// https://github.com/containerd/nerdctl/issues/5137
+type bestEffortWriter struct {
+	w    io.Writer
+	dead bool
+}
+
+func (b *bestEffortWriter) Write(p []byte) (int, error) {
+	// Only ever called from the single stdio copy goroutine of its stream, so
+	// no locking is needed.
+	if !b.dead {
+		if _, err := b.w.Write(p); err != nil {
+			b.dead = true
+			log.L.WithError(err).Warn("writing container output to the logging binary failed; further output will not be logged")
+		}
+	}
+	return len(p), nil
 }
 
 func (c *ncio) Config() cio.Config {
@@ -156,19 +195,23 @@ func NewContainerIO(namespace string, logURI string, tty bool, stdin io.Reader, 
 			if err != nil {
 				return nil, err
 			}
-			closers = append(closers, stdoutr.Close, stdoutw.Close)
+			closeStdoutR := closeOnce(stdoutr)
+			closers = append(closers, closeStdoutR, stdoutw.Close)
 
 			stderrr, stderrw, err := os.Pipe()
 			if err != nil {
 				return nil, err
 			}
-			closers = append(closers, stderrr.Close, stderrw.Close)
+			closeStderrR := closeOnce(stderrr)
+			closers = append(closers, closeStderrR, stderrw.Close)
 
 			r, w, err := os.Pipe()
 			if err != nil {
 				return nil, err
 			}
-			closers = append(closers, r.Close, w.Close)
+			closeR := closeOnce(r)
+			closeW := closeOnce(w)
+			closers = append(closers, closeR, closeW)
 
 			u, err := url.Parse(logURI)
 			if err != nil {
@@ -184,18 +227,44 @@ func NewContainerIO(namespace string, logURI string, tty bool, stdin io.Reader, 
 			closers = append(closers, func() error { return cmd.Process.Kill() })
 
 			// close our side of the pipe after start
-			if err := w.Close(); err != nil {
+			if err := closeW(); err != nil {
 				return nil, fmt.Errorf("failed to close write pipe after start: %w", err)
 			}
 
-			// wait for the logging binary to be ready
-			b := make([]byte, 1)
-			if _, err := r.Read(b); err != nil && err != io.EOF {
-				return nil, fmt.Errorf("failed to read from logging binary: %w", err)
+			// Close our copies of the stdio read ends that were handed to the
+			// logging binary; the child holds its own duplicates via ExtraFiles.
+			// This is the equivalent of containerd's binaryIO.CloseAfterStart.
+			// If this process kept the read ends open, a logging binary that
+			// stops reading (killed, crashed, ...) would never surface as EPIPE
+			// on the tee writes below: the stdio copy goroutine would block
+			// forever on the full pipe, stop draining the container's stdout
+			// FIFO, and deadlock both the container and `nerdctl run` itself
+			// (including `nerdctl rm -f` of the wedged container).
+			// https://github.com/containerd/nerdctl/issues/5137
+			if err := closeStdoutR(); err != nil {
+				return nil, fmt.Errorf("failed to close stdout pipe read end after start: %w", err)
+			}
+			if err := closeStderrR(); err != nil {
+				return nil, fmt.Errorf("failed to close stderr pipe read end after start: %w", err)
 			}
 
-			stdoutWriters = append(stdoutWriters, stdoutw)
-			stderrWriters = append(stderrWriters, stderrw)
+			// wait for the logging binary to be ready
+			// For binary-v2, readiness requires a byte to be written before close.
+			// For binary, EOF is treated as ready for backward compatibility.
+			b := make([]byte, 1)
+			n, err := r.Read(b)
+			if err != nil && err != io.EOF {
+				return nil, fmt.Errorf("failed to read from logging binary: %w", err)
+			}
+			if u.Scheme == "binary-v2" && n == 0 {
+				return nil, errors.New("logging binary did not call ready (it may have crashed or exited prematurely)")
+			}
+			if err := closeR(); err != nil {
+				return nil, fmt.Errorf("failed to close ready pipe read end: %w", err)
+			}
+
+			stdoutWriters = append(stdoutWriters, &bestEffortWriter{w: stdoutw})
+			stderrWriters = append(stderrWriters, &bestEffortWriter{w: stderrw})
 		}
 
 		streams.Stdout = io.MultiWriter(stdoutWriters...)
--- a/pkg/cioutil/container_io_test.go
+++ b/pkg/cioutil/container_io_test.go
@@ -1,0 +1,77 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


ogulcanaydogan:
Tested this on the same box/setup as my earlier repro (nerdctl 2.3.5 base, containerd 2.3.3, runc 1.5.1, rootless). Built from this branch and reran my exact reproduction (8 KiB chunks with a 20ms delay, killing `_NERDCTL_INTERNAL_LOGGING` mid-stream):

- SIGTERM on the logger: full 8 MiB delivered, `nerdctl run` exits cleanly, no leftover container.
- SIGKILL on the logger: same result, full 8 MiB, clean exit.

Also ran `go vet` and `go test -v` on both `pkg/cioutil` and `pkg/logging`, all green, including your new `TestLoggingProcessAdapterWaitError` and the best-effort-writer tests. `gofmt -l` clean too.

Nice diagnosis catching that closing the read ends alone wasn't sufficient, that's a subtle second-order effect (EPIPE killing the MultiWriter and aborting the FIFO drain) that's easy to miss. This one's solid from where I'm standing.

gsaddict91:
A note on CI for reviewers: the initial `unit / windows-2025` failure was real — my new `TestLoggingProcessAdapterWaitError` relied on `cancelreader.Cancel()` for teardown, which cannot cancel a blocked pipe read on Windows. Fixed (the test now ends its streams via EOF) and that job is green.

The three remaining red jobs appear to be pre-existing flakes unrelated to this change:

- **`in-host / nerdctl.gomodjail rootless linux` — `TestRunRmTime`** (12.6s vs 3s deadline): the identical failure (13.79s, same assertion) occurred on [run 32358304695](https://github.com/containerd/nerdctl/actions/runs/32358304695) for a PR that only adds a DCT flag and doesn't touch logging. I also verified locally that `run --rm  true` completes in ~0.1s with this branch, same as main.
- **`in-host / windows` — `TestNetworkInspectWithContainers`**: the exact same test failed on a main-branch push, [run 32338807591](https://github.com/containerd/nerdctl/actions/runs/32338807591). This PR's logger-tee code is behind a `runtime.GOOS != "windows"` guard and doesn't touch network inspect.
- **`in-host / rootful linux (old ubuntu) (ctd: v1.7.34)` — `TestIPFSCompNoBuild`**: listed as a known recurring failure in the CI tracking issue #3577 (suspected snapshotter issue).

I can't re-run the failed jobs from an outside fork — happy to rebase/re-push if a maintainer prefers a fresh run.

AkihiroSuda (approved):
Thanks

PR #5156, opened 2026-08-22
pkg/containerutil: update restart.StatusLabel to Stopped when stopping container

### LINKED ISSUE #5153: incorrect restart=always restart logic
### Description

Currently, when running a container with nerdctl run -d --restart=always busybox sh -c "top",
and then performing nerdctl stop on it, containerd still restarts the container after a few seconds.
This is inconsistent with Docker behavior and not the desired outcome.

1.nerdctl run -d --restart=always  --name=test busybox sh -c "top"
2. nerdctl stop test
3.after several seconds this container will be  restarted.

nerdtcl have same behavior  with Docker.

any

None

### Host information

_No response_

VedantMadane:
### Description

Fixes #5153.

When running a container with \--restart=always\ or \--restart=unless-stopped\, containerd's restart monitor daemon uses the \containerd.io/restart.status\ label to determine whether an exited container should be automatically restarted (when the label value is \
unning\).

Previously, while \containerutil.Start\ updated the \
estart.StatusLabel\ to \containerd.Running\ on startup, \containerutil.Stop\ and \pkg/cmd/container.killContainer\ only updated \containerd.io/restart.explicitly-stopped\ without transitioning \containerd.io/restart.status\ to \containerd.Stopped\. This caused containerd's restart monitor to still observe the status as \
unning\ and restart the container after an explicit \
erdctl stop\ or \
erdctl kill\.

This PR ensures \containerutil.Stop\ and \killContainer\ update \
estart.StatusLabel\ to \containerd.Stopped\ when a container with a restart policy is stopped or killed, aligning with Docker and containerd restart behavior.

### Checklist
- [x] Signed-off with Developer Certificate of Origin (DCO)
- [x] Passes local validation and tests

### DIFF
--- a/pkg/cmd/container/kill.go
+++ b/pkg/cmd/container/kill.go
@@ -27,6 +27,7 @@ import (
 	"github.com/moby/sys/signal"
 
 	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/core/runtime/restart"
 	"github.com/containerd/containerd/v2/pkg/cio"
 	"github.com/containerd/errdefs"
 	"github.com/containerd/go-cni"
@@ -87,6 +88,13 @@ func killContainer(ctx context.Context, container containerd.Container, signal s
 	}()
 	if err := containerutil.UpdateExplicitlyStoppedLabel(ctx, container, true); err != nil {
 		return err
+	}
+	if l, err := container.Labels(ctx); err == nil {
+		if _, ok := l[restart.PolicyLabel]; ok {
+			if err := containerutil.UpdateStatusLabel(ctx, container, containerd.Stopped); err != nil {
+				return err
+			}
+		}
 	}
 	task, err := container.Task(ctx, cio.Load)
 	if err != nil {
--- a/cmd/nerdctl/container/container_run_restart_linux_test.go
+++ b/cmd/nerdctl/container/container_run_restart_linux_test.go
@@ -383,6 +383,23 @@ func TestRunRestartStatusLabel(t *testing.T) {
 	testCase.Run(t)
 }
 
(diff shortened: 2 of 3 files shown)


AkihiroSuda:
Can we have an integration test?

VedantMadane:
Added integration tests in \cmd/nerdctl/container/container_run_restart_linux_test.go\ (\TestRunRestartAlwaysStop\ and \TestRunRestartAlwaysKill\) to verify that containers created with \--restart=always\ transition \containerd.io/restart.status\ to \stopped\ and remain in \exited\ status when explicitly stopped or killed.

ningmingxiao:
but if  containerd daemon is restarted  we should let containerd auto restart the container even the container is stoped by nerdctl when restart policy is always.

pkg/cmd/container/kill.go
```diff
@@ -88,6 +89,13 @@ func killContainer(ctx context.Context, container containerd.Container, signal s
 	if err := containerutil.UpdateExplicitlyStoppedLabel(ctx, container, true); err != nil {
 		return err
 	}
+	if l, err := container.Labels(ctx); err == nil {
```
ningmingxiao, line 92: if restart policy is always we should skip to set it because if  containerd daemon is restarted we should let containerd start the container.

pkg/cmd/container/kill.go
ningmingxiao, line 85: if failed to kill it we should restore the label in defer.

ningmingxiao:
UpdateExplicitlyStoppedLabel  and UpdateStatusLabel I think we can combine them to reduce  once container.Update  logic.

AkihiroSuda:
Merged https://github.com/containerd/containerd/pull/13993

This PR is no longer needed?

PR #5160, opened 2026-08-24
fix(image): truncate the output file of save and export

### LINKED ISSUE #5159: nerdctl save -o / export -o do not truncate an existing file
### Description

`nerdctl save -o FILE` and `nerdctl export -o FILE` open the output with `os.O_CREATE|os.O_WRONLY` and no `os.O_TRUNC`. When `FILE` already exists and is longer than what is about to be written, the tail of the previous content survives past the end of the new archive.

- https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/image/image_save.go#L98
- https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container/container_export.go#L71

`docker save` does not have this problem: it writes through an atomic writer (temp file plus rename), so the destination is replaced wholesale.

```go
// docker/cli cli/command/image/save.go
writer, err := atomicwriter.New(opts.output, 0o600)
```

1. `nerdctl save -o out.tar <a larger image>`
2. `nerdctl save -o out.tar <a smaller image>`
3. `ls -l out.tar` — it still has the size of the archive from step 1.

The same applies to `nerdctl export -o out.tar <container>`.

The open flags in isolation:

```go
func open(path string) (*os.File, error) {
	// as in image_save.go and container_export.go
	return os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
}

f, _ := open(path)
f.Write([]byte("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")) // 30 bytes, the larger image
f.Close()

f, _ = open(path)
f.Write([]byte("BBBBBBBBBB")) // 10 bytes, the smaller image
f.Close()
```

```console
wrote 10 bytes, file holds 30: "BBBBBBBBBBAAAAAAAAAAAAAAAAAAAA"
```

Received: the file keeps the size of the previous, larger archive, with the leftover bytes of that archive sitting after the end of the new one.

Note that this does not break `nerdctl load`: a tar reader stops at the end-of-archive marker and never looks at what follows, which I checked with both `archive/tar` and the system `tar`. What is wrong is the artifact itself — it is larger than the archive it is supposed to contain, and the extra bytes belong to an unrelated image, which shows up in anything that checksums the file, accounts for its size, or ships it somewhere.

Expected: `-o` replaces the file, as `docker save -o` and `docker export -o` do, and as a shell redirect does.

`pkg/healthcheck/log.go` opens a file with the same flags, but seeks to the end and appends on purpose, so it is not affected.

main, at c235f00d8a42aa810243b3ecf94feb29459aa6ed.

The flags came in with 4b8061d9 (2024-08-30) for `save` and 3e50703e (2025-08-05) for `export`.

ekalinin:
`nerdctl save -o FILE` and `nerdctl export -o FILE` open the output with
`os.O_CREATE|os.O_WRONLY` and no `os.O_TRUNC`. Writing a smaller archive over a bigger one
therefore leaves the tail of the bigger one past the end of the new archive.

```go
f, _ := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
f.Write([]byte("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")) // 30 bytes, the larger image
f.Close()

f, _ = os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
f.Write([]byte("BBBBBBBBBB")) // 10 bytes, the smaller image
f.Close()
```

```console
wrote 10 bytes, file holds 30: "BBBBBBBBBBAAAAAAAAAAAAAAAAAAAA"
```

This does not break `nerdctl load`: a tar reader stops at the end-of-archive marker and never
looks at what follows, which I checked with both `archive/tar` and the system `tar`. What is
wrong is the artifact — it is larger than the archive it is supposed to hold, and the extra
bytes belong to an unrelated image or container, which shows up in anything that checksums the
file, accounts for its size, or ships it somewhere.

`docker save` replaces the destination wholesale, writing through a temporary file and renaming
it (`atomicwriter.New` in docker/cli). Adding `O_TRUNC` gets nerdctl to the same observable
result without a new dependency: `pkg/internal/filesystem` is built for state files, with
locking and backups, not for streaming an archive that can be many gigabytes.

`pkg/healthcheck/log.go` opens a file with the same flags, but seeks to the end and appends on
purpose, so it is left alone.

**Tests**

`TestSaveReplacesExistingFile` and `TestExportReplacesExistingFile` write the bigger archive
first and the smaller one over it, then assert the file shrank. They compare against the size
measured before the overwrite rather than against a separately saved copy, so they do not depend
on two saves of the same image being byte-identical.

Both assert on the size rather than on `load` failing, since as above a reader would not notice
the tail.

Fixes #5159

### DIFF
--- a/cmd/nerdctl/container/container_export.go
+++ b/cmd/nerdctl/container/container_export.go
@@ -68,7 +68,10 @@ func exportAction(cmd *cobra.Command, args []string) error {
 
 	writer := cmd.OutOrStdout()
 	if output != "" {
-		f, err := os.OpenFile(output, os.O_CREATE|os.O_WRONLY, 0644)
+		// O_TRUNC: writing a smaller archive over a bigger one would otherwise leave the tail of
+		// the bigger one past its end. A tar reader stops at the end-of-archive marker and would
+		// not notice, but the file would carry the bytes of another container.
+		f, err := os.OpenFile(output, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
 		if err != nil {
 			return err
 		}
--- a/cmd/nerdctl/container/container_export_test.go
+++ b/cmd/nerdctl/container/container_export_test.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"testing"
 
 	"gotest.tools/v3/assert"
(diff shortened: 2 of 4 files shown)


AkihiroSuda (approved):
Thanks

Issue #4767, opened 2026-03-03
Feature request: Add necessary support for new functionality(introduce volume.type=image) from compose-spec


KristoferHansson:
### What is the problem you're trying to solve

This is a feature request for adding necessary support in nerdctl for new additions to the compose spec related to 
OCI image/artifact volume mount

Compose spec commit:
"introduce volume.type=image"
https://github.com/compose-spec/compose-spec/commit/40af7cd3e4d72368f82e8dea89e705ebc92edae1

merged docker compose pull request supporting the above compose feature addition
https://github.com/docker/compose/pull/12725

closed podman-compose issue and referenced merged pull request for supporting the above compose feature addition
https://github.com/containers/podman-compose/issues/1202

/Kristofer

KristoferHansson:
Apparently OCI Artifact volume mount support is not as wide spread as OCI image volume mount.
talked about in: https://github.com/containerd/containerd/issues/11381
CRI-O has added basic OCI Artifact volume mount support, utilizing the same type: mount.image as for OCI Image, the runtime investigates the details of the pulled OCI and handles it accordingly: https://github.com/cri-o/cri-o/issues/8953
KEP-4639 talks about OCI Objects as a term including both OCI Images and OCI Artifacts: https://github.com/kubernetes/enhancements/issues/4639 and feature now GA: https://github.com/kubernetes/kubernetes/pull/136711

papidb:
Hi, I’d like to work on this. nerdctl run --mount type=image landed in #4990, while image-subpath support is currently in #4993. My plan is to translate Compose image volumes into --mount type=image,... arguments in serviceparser, support image.subpath, and add unit and integration tests. Would you prefer that I prepare this as a stacked change on #4993, or wait until #4993 is merged?

PR #5161, opened 2026-08-25
feat: support compose image volumes

papidb:
Support Compose long-syntax image volumes by translating them to the existing `--mount=type=image` implementation.

The Compose path now resolves image sources that reference active or disabled services, ensures missing mount-source images before the child command runs with `--pull=never`, and validates image-specific options before project resources are created. Image mounts remain read-only.

`image.subpath` was intentionally left alone until the lower-level support in #4993 is available.

Tests cover external and service-referenced sources, source pulling, read-only behavior, malformed definitions, mount-field injection, unchanged legacy volume arguments, and cleanup.

Related to #4767.

### DIFF
--- a/cmd/nerdctl/compose/compose_up_linux_test.go
+++ b/cmd/nerdctl/compose/compose_up_linux_test.go
@@ -17,6 +17,7 @@
 package compose
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


cmd/nerdctl/compose/compose_up_linux_test.go
```diff
@@ -1245,6 +1246,163 @@ services:
...
+	testCase.Cleanup = func(data test.Data, helpers test.Helpers) {
+		helpers.Anyhow("compose", "-f", data.Temp().Path("compose.yaml"), "down", "--volumes", "--remove-orphans")
+		helpers.Anyhow("rm", "-f", data.Identifier("image-volume"))
+		helpers.Anyhow("rmi", "-f", testutil.NginxAlpineImage)
+	}
+
+	testCase.Run(t)
+}
+
+func TestComposeImageVolumeServiceSource(t *testing.T) {
```
AkihiroSuda, line 1311: CI failing for Docker
https://github.com/containerd/nerdctl/actions/runs/32843708128/job/97846811402?pr=5161

```
  compose_up_linux_test.go:1339: 
        
        | ➡️      | ⚙️ /usr/bin/docker compose -p testcomposeimagevolumeservicesource-image-service-source-99d59bae  |
        |         | -f /tmp/TestComposeImageVolumeServiceSource3467620079/001/b9c9a923/compose.yaml up -d            |
        |         | 🚫 command returned a non-zero exit code                                                         |
        |         | ⚠️ 1                                                                                             |
        |         | 🟠  Image source Pulling                                                                         |
        |         |  Image source Error pull access denied for source, repository does not exist or may require 'doc |
        |         | ker login'                                                                                       |
        |         | Error response from daemon: pull access denied for source, repository does not exist or may requ |
        |         | ire 'docker login'                                                                               |
        | 🌱      | (hidden: set TIGRON_DEBUG_ENV=1 to display)                                                      |
        | ⏰      | <1s (limit: 3m0s)                                                                                |
        | 📁      | /tmp/TestComposeImageVolumeServiceSource3467620079/002                                           |
        
```

papidb: Yeah, I'm on it. Thanks.

papidb:
Hi @AkihiroSuda quick question about Compose image volumes. 

The Compose spec defines source as an image reference. Docker Compose follows this, except when source matches a service that builds an image using Compose’s default <project>-<service> name, I will match that behavior.

Docker does not resolve a service’s explicit image: or a service disabled by a profile. It treats source literally and tries to pull it as an image.

Should nerdctl match Docker and the spec for these two cases, or intentionally support them as a nerdctl extension by resolving the service’s image?

AkihiroSuda:
> Should nerdctl match Docker and the spec for these two cases, or intentionally support them as a nerdctl extension by resolving the service’s image?

Let's begin with reproducing Docker's behavior.
An extension can be proposed in a follow-up issue.

AkihiroSuda (approved):
Thanks

PR #5163, opened 2026-08-28
fix: preserve equals signs in image label filter values

immanuwell:
`nerdctl images --filter label=example.payload=a=b` fails with `invalid filter`, while Docker accepts the same filter. 
Label values are plain strings, so `=` is legit here

The parser now splits into at most three parts and keeps the rest in the label value. 
Non-label filters keep their old validation

Repro before this change:

`nerdctl images --filter label=example.payload=a=b`

`FATA[0000] invalid filter "label=example.payload=a=b"`

Tested with `go test ./pkg/imgutil ./pkg/cmd/image`


AkihiroSuda (approved):
Thanks

PR #5164, opened 2026-08-31
dockercompat: key inspect Networks by real CNI network name

### LINKED ISSUE #2999: nerdctl inspect does not list networks name properly

larrasket:
`nerdctl inspect` currently shows a container's networks as `unknown-eth0` instead of the network name

It comes from `networkSettingsFromNative` in pkg/inspecttypes/dockercompat: the map key is built from the interface name with a hardcoded `unknown-` prefix, and it does not look at which network the interface belongs to. (There was already a `// TODO: set CNI name when possible` sitting there).

go-cni names the interfaces `eth0`, `eth1`, ... in the same order the networks were attached, and nerdctl keeps that ordered list in the `nerdctl/networks` annotation. I read the list from the annotation and map `eth<i>` back to `networks[i]`.

If an interface doesn't fit that pattern, host networking, something not set up by CNI, or an older container with no annotation, it keeps the old `unknown-<iface>` key, so those cases are unchanged.

Added single, and multi-network cases to `TestNetworkSettingsFromNative`; the existing cases are untouched.

Fixes #2999

### DIFF
--- a/pkg/inspecttypes/dockercompat/dockercompat.go
+++ b/pkg/inspecttypes/dockercompat/dockercompat.go
@@ -805,7 +805,7 @@ func statusFromNative(x containerd.Status, labels map[string]string) string {
 	}
 }
 
-func networkSettingsFromNative(n *native.NetNS, _ *specs.Spec) (*NetworkSettings, error) {
+func networkSettingsFromNative(n *native.NetNS, spec *specs.Spec) (*NetworkSettings, error) {
 	res := &NetworkSettings{
 		Networks: make(map[string]*NetworkEndpointSettings),
 	}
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #5165, opened 2026-08-31
ci: Print daemon logs after integration test failures

### LINKED ISSUE #4365: [CI] CI runs should output containerd / buildkitd logs

We have seen containerd crashes on the CI - and growingly, as nerdctl gets better, we are hitting containerd (or buildkitd) issues.

Debugging these (when issues are transient) is really hard without any logs.

Suggestion is to figure out a way to make daemon logs available on the CI in case of failure, either printed out directly or downloadable as an artifact.

### Additional context

_No response_

loglapa:
## Summary

- print containerd and buildkit system service logs after failed Linux host integration jobs
- collect the same diagnostics from Lima guests
- include the unprivileged user journal for rootless targets
- keep diagnostics best-effort so they do not replace the original test failure

Fixes #4365.

## Validation

- `yamllint .`
- `git diff --check`
- `actionlint` on both changed workflows (only the two existing `fromJSON(inputs.skip-flaky)` type warnings remain)

### DIFF
--- a/.github/workflows/job-test-in-host.yml
+++ b/.github/workflows/job-test-in-host.yml
@@ -333,3 +333,31 @@ jobs:
           INPUTS_IPV6: ${{ inputs.ipv6 }}
           CONTAINERD_ROOTLESS_ROOTLESSKIT_PORT_DRIVER: ${{ inputs.target == 'rootless-port-slirp4netns' && 'slirp4netns' || '' }}
           CONTAINERD_ROOTLESS_ROOTLESSKIT_IPV6: ${{ inputs.ipv6 && 'true' || '' }}
+
+      - if: ${{ failure() && contains(inputs.runner, 'ubuntu') && inputs.target != '' && env.SHOULD_RUN == 'yes' }}
+        name: "Post (linux): print daemon logs"
+        env:
+          INPUTS_TARGET: ${{ inputs.target }}
+        run: |
+          echo "::group::containerd and buildkit system service logs"
+          sudo journalctl \
+            --boot \
+            --no-pager \
+            --lines=1000 \
+            --unit=containerd.service \
+            --unit=buildkit.service \
+            --unit=test-integration-buildkit-nerdctl-test.service || true
+          echo "::endgroup::"
+
+          if [[ "${INPUTS_TARGET}" == rootless* ]]; then
+            echo "::group::rootless daemon logs"
+            journalctl \
+              --user \
+              --boot \
+              --no-pager \
+              --lines=1000 \
+              --unit=containerd.service \
+              --unit=buildkit.service \
+              --unit=test-integration-buildkit-nerdctl-test.service || true
+            echo "::endgroup::"
+          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/job-test-in-host.yml
```diff
@@ -333,3 +333,28 @@ jobs:
...
+            --no-pager \
+            --lines=1000 \
+            --unit=containerd.service \
+            --unit=buildkit.service \
+            --unit=test-integration-buildkit-nerdctl-test.service || true
+          echo "::endgroup::"
+
+          if [[ "${INPUTS_TARGET}" == rootless* ]]; then
+            echo "::group::rootless daemon logs"
+            sudo journalctl \
```
AkihiroSuda: sudo should not be needed?

loglapa: Agreed — this is the current rootless user journal, so it does not need privilege escalation. Removed `sudo` here in aaf98e3; the separate system-service journal call remains privileged.

.github/workflows/job-test-in-host.yml
```diff
@@ -333,3 +333,28 @@ jobs:
...
+            --no-pager \
+            --lines=1000 \
+            --unit=containerd.service \
+            --unit=buildkit.service \
+            --unit=test-integration-buildkit-nerdctl-test.service || true
+          echo "::endgroup::"
+
+          if [[ "${INPUTS_TARGET}" == rootless* ]]; then
+            echo "::group::rootless daemon logs"
+            journalctl \
```
AkihiroSuda, line 354: It still needs `--user`.
The rest of the flags should be same as the rootful ones.

.github/workflows/job-test-in-lima.yml
AkihiroSuda, line 1: Same comments as `job-test-in-host.yml`

loglapa: Applied the same rootless-journal change to the Lima job in 8701c82: `lima journalctl` runs as the Lima user, while the separate system-service call keeps `lima sudo journalctl`. I also amended the follow-up with the required DCO Signed-off-by trailer.

.github/workflows/job-test-in-host.yml
loglapa: Fixed in 5078408. Both rootless paths now use the user journal (`journalctl --user`, or `lima journalctl --user`) and the same `--boot`, `--no-pager`, `--lines`, and three `--unit` filters as the rootful paths; the `_UID` filters are removed.

.github/workflows/job-test-in-lima.yml
loglapa: Follow-up: 5078408 also adds `--user`, removes `_UID`, and mirrors the rootful unit filters in the Lima rootless block, matching the detailed host comment.

AkihiroSuda:
Please squash the commits

loglapa:
Squashed the follow-up into the original change as requested. The branch now has a single signed DCO commit, 5092e2b3; the resulting tree is byte-identical to the previously reviewed head.

AkihiroSuda (approved):
Thanks

Issue #3570, opened 2024-10-18
Multiple attach should be supported


apostasie:
### Description

In terminal 1:

`nerdctl run --rm -ti --detach-keys=ctrl-a,ctrl-b --name what debian`

Do _not_ detach

Then in terminal 2:

`nerdctl attach what`

Start typing in terminal 2.

Notice that some characters are displayed on terminal 1, others on terminal 2.

Do the same with docker: it works.

There is something broken either in nerdctl, or maybe in `containerd/console`.

### Describe the results you received and expected

na

### What version of nerdctl are you using?

main

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

apostasie:
Note this was actually explicitly mentioned as a limitation in the original PR introducing attach: #2108

apostasie:
@AkihiroSuda + feature

ChengyuZhu6:
@apostasie @AkihiroSuda I‘d like to take it on.

ekalinin:
I would like to work on this. It is invasive enough that I would rather agree on the design first.

@ChengyuZhu6 you volunteered for this one (and for #4374) a while back. If you have something in progress I am happy to review instead, or to split the work.

## Diagnosis

To answer the open question in the description: it is not `containerd/console`, it is the FIFOs.

`pkg/cmd/container/attach.go` attaches with `container.Task(ctx, cio.NewAttach(opt))`, which re-opens the *same* stdio FIFOs the first session already holds. A FIFO has one queue, so with two readers the kernel hands each chunk to exactly one of them: the container's output is split at random, and one terminal appears to freeze while the other prints characters it never asked for. Two writers on the stdin FIFO interleave the same way, which is the "some characters on terminal 1, others on terminal 2" above. #4374 is the same thing with a video of it.

This was called out as a limitation in #2108, the PR that introduced `attach`, and is still documented as a caveat in the help text.

## Proposal

Give each container a single owner of its stdio, and make every CLI session a client of it over a unix socket. The owner would be the process nerdctl already runs for the lifetime of every container: the internal logging process (`_NERDCTL_INTERNAL_LOGGING`, `pkg/logging`). It gains a socket, a fan-out of the container's raw output tapped *before* the log driver's line splitting, and the write end of a stdin FIFO at a stable path.

```
                    containerd-shim
                    pty master / pipes
                            |
              binary:// (spawn + fd 3,4)
                            v
   +------------------------------------------+
[...]

PR #5168, opened 2026-09-01
feat(attach): support multiple concurrent attach sessions

### LINKED ISSUE #4374: `nerdctl attach` freezes the terminal sessions randomly.
### Description

When attaching two terminal sessions using `nerdctl attach` and executing commands, the terminal randomly freezes and doesn't display the commands being entered.

https://github.com/user-attachments/assets/3e2a11ff-a710-4972-a88b-388b6c8e3d03

1. sudo nerdctl run -it --rm --name enipla alpine
2. sudo nerdctl attach enipla

This is the result when running with docker.

https://github.com/user-attachments/assets/7c5a6138-bdca-43e7-a09c-ae7f86e1a9b9

```
Client:
 Version:	v2.1.2
 OS/Arch:	linux/amd64
 Git commit:	2a2834dd1c926528aa141726a9a9132eb1665568
 buildctl:
  Version:	v0.23.0-rc1-24-g91ff73e1c
  GitCommit:	91ff73e1c26e761a64de9a05bc1e67387a7ef8d0

Server:
 containerd:
  Version:	1.7.27
  GitCommit:	05044ec0a9a75232cad458027ca83437aae3f4da
 runc:
  Version:	1.3.0
  GitCommit:	v1.3.0-0-g4ca628d1

```

Others

### Host information

```
Client:
 Namespace:	default
 Debug Mode:	false

Server:
 Server Version: 1.7.27
 Storage Driver: overlayfs
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Storage: native overlayfs
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version:   6.8.0-62-generic
 Operating System: Linux Mint 22.1
 OSType:           linux
 Architecture:     x86_64
 CPUs:             8
 Total Memory:     13.57GiB
 Name:             anton
 ID:               5616474f-8d43-470c-890a-3a1c59b1e89c
```

ekalinin:
Fixes #4374. Design discussion: https://github.com/containerd/nerdctl/issues/3570#issuecomment-5482854810 and the [follow-up on the switch](https://github.com/containerd/nerdctl/issues/3570#issuecomment-5509654759).

This was five stacked pull requests; per review feedback on the first one it is now a single PR with the history kept separate. Rebased on current `main`.

## The problem

A container's stdio FIFOs cannot be shared. A FIFO has one queue, so two `attach` sessions on the same container each receive a random subset of its output: one terminal appears to freeze while the other prints characters it never asked for. And a container started by `nerdctl run -d` cannot be attached to at all, because its output already went to the logging process.

`dockerd` does not have this problem because it owns the container's stdio and fans it out. nerdctl is daemonless, so some other process has to own it.

## The approach

The owner is the internal logging process, the one nerdctl already runs as `_NERDCTL_INTERNAL_LOGGING`. containerd spawns it from the container's log URI, so it exists for the container's whole lifetime and comes back on its own when the restart monitor recreates a task. It gains a unix socket, a fan-out of the container's raw output tapped before the log driver's line splitting, and the write end of a stable stdin FIFO that every session's input is merged into. `nerdctl attach`, `run -it` and `start -a` become clients of it.

`docs/dev/attach.md` is the long version.

## Commits

| | |
|---|---|
| `feat(attachmux)` ×4, `fix(attachmux)` ×4 | the new `pkg/attachmux`: frame protocol, broker, client, unix socket transport. Touches no existing file |
| `feat(cioutil)` | stable stdin FIFO at `<dataStore>/containers/<ns>/<id>/stdin.fifo` |
| `refactor(logging)` | tee the container's raw output before the log driver's line splitting |
| `feat(logging)` | start the broker inside the logging process; new leaf package `pkg/logging/loguri` |
| `feat(taskutil)` | declare the stable stdin FIFO on the task; `--log-driver none` goes through the internal logger |
| `fix(logging)` ×3 | listener ordering, a Windows log URI path, and keeping the `none` driver off the buffered channels |
| `feat(config)` | `disable_attach_broker` as a way back to the old stdio |
| `feat(attach)` | `nerdctl attach` dials the socket; several sessions at once |
| `feat(run)` | foreground `run -it` streams through the broker |
| `docs(attach)` | command reference, command help, `docs/dev/attach.md` |

## Invariants, each with a test

- **The container never blocks on a session.** Every session gets a byte-bounded queue and its own writer goroutine; one that falls behind is disconnected. The goroutine feeding the broker never waits on a consumer. This is the structural fix for the deadlock class in #5137 / #5151.
- **With no sessions attached the broker keeps draining**, so detaching leaves the container running normally.
- **An exit is only announced when there was one.** `<-exitCh` is not on its own evidence: containerd delivers Wait RPC failures on the same channel as a synthetic `ExitStatus`, which is what #5137 and #5151 are about. Rebased on #5151, `containerExited` is closed after its retry loop, which is reached only by the `break` on a clean status - the two returns inside the loop leave the goroutine without going past it.
- **The exit frame does not travel through the session queue.** The writer delivers it after the queue drains, so a session whose queue is exactly full still learns the container exited rather than seeing a bare EOF.
- **Nothing is recorded on disk about who owns a container's stdio.** `attach` asks containerd, through the `cio.Attach` callback that `container.Task` hands the task's recorded `Stdin`/`Stdout`/`Stderr`. It cannot be cached: the restart monitor recreates a task from the stored log URI without nerdctl running at all, so a container that had FIFO stdio can come back with URI stdio between two invocations.

## Compatibility

A container created by an older nerdctl, or on a platform where the socket is not implemented, keeps its FIFO stdio and attaches the way it always did, one session at a time. The classification and the IO construction happen in the same `container.Task` call, so the restart monitor cannot swap one for the other in between.

`disable_attach_broker` (`--disable-attach-broker`, `NERDCTL_DISABLE_ATTACH_BROKER`, `nerdctl.toml`) is the way back. It leaves the data store empty at task creation, which is what a container with a foreign log driver already looks like, so every branch takes the legacy path with no new code. Off by default. It only affects containers created while it is set: a running container's stdio is already bound.

The transport is built for `linux || freebsd`, matching where the logging process is built. Elsewhere `Probe` fails and `run -it` picks the legacy FIFO path, which is exactly today's behaviour.

## Known limitations, stated in `docs/dev/attach.md`

- **Multi-session stdin is limited to terminal containers.** For a non-terminal one the shim's `binary` scheme sets `pio.copy = false` and `binaryIO.Stdin()` returns nil, so stdin is not wired at all.
- **After a restart-policy restart, sessions are output only.** The restart monitor recreates a terminal task through `cio.TerminalLogURI`, whose config carries no `Stdin`, so the shim never opens the read end of the stable FIFO. Not a regression: such a container has no stdin today either. Closing it needs the restart monitor to preserve the full IO config, which is an upstream containerd change.

## Testing

`go test -race ./pkg/...`: green, including 53 tests in `pkg/attachmux` and the new cases in `pkg/logging`.

`go vet` and `golangci-lint` clean for linux, windows, freebsd and darwin.

New integration tests in `cmd/nerdctl/container/container_attach_linux_test.go`: `TestAttachMultipleSessions`, `TestAttachToDetachedContainer`, `TestAttachToDetachedContainerWithoutTTY`, `TestAttachAfterDetachKeepsContainerResponsive`.

The switch was exercised on a built binary: default off, env on, toml on, env beating toml, flag beating both.

### DIFF
--- a/cmd/nerdctl/container/container_attach.go
+++ b/cmd/nerdctl/container/container_attach.go
@@ -41,11 +41,14 @@ func AttachCommand() *cobra.Command {
 
 Caveats:
 
-- Currently only one attach session is allowed. When the second session tries to attach, currently no error will be returned from nerdctl.
-  However, since behind the scenes, there's only one FIFO for stdin, stdout, and stderr respectively,
-  if there are multiple sessions, all the sessions will be reading from and writing to the same 3 FIFOs, which will result in mixed input and partial output.
-- Until dual logging (issue #1946) is implemented,
-  a container that is spun up by either 'nerdctl run -d' or 'nerdctl start' (without '--attach') cannot be attached to.`
+- Several sessions can be attached to the same container at once: they all see the container's output,
+  and any of them can type into it.
+- A container whose stdio is a set of FIFOs, because it was started by a version of nerdctl that predates
+  the attach socket or on a platform without one, still allows a single session: with more than one, all
+  sessions read from and write to the same 3 FIFOs, which results in mixed input and partial output.
+- A container whose output was handed to a logging process nerdctl cannot reach, either somebody else's
+  binary through a custom '--log-driver binary://' URI or a detached container on a platform without an
+  attach socket, cannot be attached to at all.`
 
 	var cmd = &cobra.Command{
 		Use:               "attach [flags] CONTAINER",
--- a/cmd/nerdctl/container/container_attach_linux_test.go
+++ b/cmd/nerdctl/container/container_attach_linux_test.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"io"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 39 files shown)


AkihiroSuda:
> This PR adds the pkg/attachmux package and nothing else.

Thanks, but hard to review/test/merge until the actual `nerdctl attach` is implemented in the PR

ekalinin:
Done - this PR is now the whole change, rebased on current `main`. `nerdctl attach`, `nerdctl run -it` and `nerdctl start -a` all go through it, so it can be tested end to end.

The five steps are kept as separate commits so the history is still reviewable in order: `pkg/attachmux` on its own first (8 commits, no existing file touched, no containerd needed to review it), then the logging process becoming the owner of the container's stdio, then the switch, then the two commands, then the docs.

The part I would most like a second opinion on is still the one from the issue: the internal logging process as the home for the broker, and `<dataStore>/attach/<hash>.sock` as the socket location. Everything else follows from those two.

Two things worth knowing before testing:

- `disable_attach_broker` (`--disable-attach-broker`, `NERDCTL_DISABLE_ATTACH_BROKER`, `nerdctl.toml`) is the way back to the old stdio. Off by default, and it only affects containers created while it is set - a running container's stdio is already bound.
- The transport is `linux || freebsd`. Elsewhere `Probe` fails and `run -it` takes the legacy FIFO path, which is today's behaviour.

There is one overlap I could not resolve on my own: #5151 touches the same `task.Wait` result in `run`. This PR only announces an exit to sessions when `status.Error() == nil`, which is the same distinction #5151 is making, so the two should compose - but whichever lands second will need a rebase, and I am happy to be the one to do it.

ekalinin:
Rebased on current `main`, conflicts resolved - #5151 landed in the meantime, which is the overlap I mentioned above.

PR #5169, opened 2026-09-01
login: warn when the server address contains an explicit scheme

### LINKED ISSUE #3052: `nerdctl login` should warn about passing along an explicit scheme and should ignore it entirely

locker95:
Fixes #3052.

As noted by @fahedouch in the issue ("It looks like a bug. It is not the intended behavior."), `nerdctl login` silently drops an explicit scheme from the server address. Following the direction suggested by @apostasie, this change warns the user that this syntax is only supported as a convenience and that the scheme is ignored; when `http://` was requested, the warning additionally points at the global `--insecure-registry` flag. The warning is emitted only from the `login` code path — `dockerconfigresolver.Parse` is untouched, so credential lookup during pull/push stays silent.

Testing: a platform-independent unit test covers the new detection helper (verified locally on macOS); `TestLoginAgainstVariants` now also asserts the warning on stderr for the scheme variants. I could only compile-verify the Linux integration test locally (`GOOS=linux go build ./...` / `go vet` / `go test -c`), so I am relying on CI to exercise it.

This patch was developed with the assistance of an AI tool (Claude); I have reviewed and tested the change and take responsibility for it.

### DIFF
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"golang.org/x/net/context/ctxhttp"
 
@@ -41,7 +42,28 @@ Configure a credential helper to remove this warning. See
 https://docs.docker.com/engine/reference/commandline/login/#credentials-store
 `
 
+// schemeWarning returns a warning message if the provided server address contains an explicit
+// http or https scheme, and an empty string otherwise.
+// Accepting a scheme in the server address is only a convenience: the scheme itself is ignored
+// (see dockerconfigresolver.Parse), which may surprise users (https://github.com/containerd/nerdctl/issues/3052).
+func schemeWarning(serverAddress string) string {
+	sch, _, found := strings.Cut(serverAddress, "://")
+	if !found || (sch != "http" && sch != "https") {
+		return ""
+	}
+	warning := fmt.Sprintf("The %s:// scheme in the provided server address %q is accepted only as a convenience, "+
+		"and is otherwise ignored: connections to registries are made over https by default.", sch, serverAddress)
+	if sch == "http" {
+		warning += " To log in to a registry served over plain http, use the global --insecure-registry flag instead."
+	}
+	return warning
+}
+
 func Login(ctx context.Context, options types.LoginCommandOptions, stdout io.Writer) error {
+	if warning := schemeWarning(options.ServerAddress); warning != "" {
+		log.G(ctx).Warn(warning)
+	}
+
 	registryURL, err := dockerconfigresolver.Parse(options.ServerAddress)
 	if err != nil {
 		return err
--- a/cmd/nerdctl/login/login_linux_test.go
+++ b/cmd/nerdctl/login/login_linux_test.go
@@ -21,6 +21,7 @@
 package login
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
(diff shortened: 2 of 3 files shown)


AkihiroSuda:
> I could only compile-verify the Linux integration test locally (GOOS=linux go build ./... / go vet / go test -c), so I am relying on CI to exercise it.
>
> This patch was developed with the assistance of an AI tool (Claude); I have reviewed and tested the change and take responsibility for it.

Please responsibly test your PR with actual registries, by running `nerdctl login` manually.

pkg/cmd/login/login.go
```diff
@@ -41,7 +42,28 @@ Configure a credential helper to remove this warning. See
...
+// schemeWarning returns a warning message if the provided server address contains an explicit
+// http or https scheme, and an empty string otherwise.
+// Accepting a scheme in the server address is only a convenience: the scheme itself is ignored
+// (see dockerconfigresolver.Parse), which may surprise users (https://github.com/containerd/nerdctl/issues/3052).
+func schemeWarning(serverAddress string) string {
+	sch, _, found := strings.Cut(serverAddress, "://")
+	if !found || (sch != "http" && sch != "https") {
+		return ""
+	}
+	warning := fmt.Sprintf("The %s:// scheme in the provided server address %q is accepted only as a convenience, "+
```
AkihiroSuda, line 54: Does `docker login` have such a message?

locker95: no — docker login just strips the scheme in ConvertToHostname and stays quiet. I can drop the warning if you would rather match that.

locker95:
on testing: I do not have a registry handy here. the unit tests pass; the linux integration job already logs in with the http(s):// variants.

on the warning: docker login does not print one (ConvertToHostname just strips the scheme). happy to drop it if you would rather keep parity.

AkihiroSuda:
> Testing: a platform-independent unit test covers the new detection helper (verified locally on macOS); TestLoginAgainstVariants now also asserts the warning on stderr for the scheme variants. I could only compile-verify the Linux integration test locally (GOOS=linux go build ./... / go vet / go test -c), so I am relying on CI to exercise it.



> on testing: I do not have a registry handy here. the unit tests pass; the linux integration job already logs in with the http(s):// variants.

A contributor has to learn how to test their PR ahead of the submission.

PR #5173, opened 2026-09-01
Fix rootless cgroup paths on WSL 2.9

### LINKED ISSUE #5172: `user.slice` path must not be hardcoded (affects WSL2.9)
https://github.com/containerd/nerdctl/blob/833c457a45c7a0fd9e9cad9ab2d63330f1062b3d/pkg/infoutil/infoutil_linux.go#L144

https://github.com/containerd/nerdctl/blob/833c457a45c7a0fd9e9cad9ab2d63330f1062b3d/pkg/cmd/container/run_cgroup_linux.go#L238

https://github.com/containerd/nerdctl/blob/833c457a45c7a0fd9e9cad9ab2d63330f1062b3d/extras/rootless/containerd-rootless-setuptool.sh#L115

The cgroup path must not be hardcoded, as the path may differ on WSL2 since WSL2.9.

- https://github.com/microsoft/WSL/issues/41492
- https://github.com/moby/moby/issues/53531
- https://github.com/moby/moby/pull/53533

Help wanted from WSL2 users.

RRXXZZYY:
Fixes #5172.

## What

- Discover the RootlessKit child process's actual cgroup v2 path instead of assuming `/user.slice/user-<uid>.slice`.
- Read the systemd user manager's `ControlGroup` in the rootless setup check.
- Let the OCI runtime choose its rootless systemd slice default, so outer hierarchies such as WSL 2.9's `/wsl-user/distro-N/...` prefix are preserved.
- Add focused regression tests for WSL-shaped paths and rootless systemd cgroup specs.

## Validation

- `gofmt` and `git diff --check`
- `go mod tidy -diff`
- Linux-target `go vet` and test-binary compilation for the three changed Go packages with Go 1.26.3
- `golangci-lint run --new-from-rev=origin/main ./...` (0 new issues)
- `bash -n` and ShellCheck 0.11.0 for the changed setup script
- `go test ./pkg/...` on Windows passed except `pkg/netutil`, whose existing test expects a local CNI config directory that is not present on this host

I do not have a WSL 2.9 host available, so the live WSL behavior remains to be confirmed by CI or a maintainer environment. The WSL-specific hierarchy is covered with a synthetic cgroup path fixture.

## AI assistance

OpenAI Codex assisted with implementation and review. The final diff was manually inspected, checked for unrelated changes and sensitive data, and validated with the commands above.

### DIFF
--- a/extras/rootless/containerd-rootless-setuptool.sh
+++ b/extras/rootless/containerd-rootless-setuptool.sh
@@ -112,8 +112,12 @@ cmd_entrypoint_check() {
 	fi
 
 	INFO "Checking cgroup v2"
-	controllers="/sys/fs/cgroup/user.slice/user-${id}.slice/user@${id}.service/cgroup.controllers"
-	if [ ! -f "${controllers}" ]; then
+	user_cgroup="$(systemctl --user show --property=ControlGroup --value)"
+	case "${user_cgroup}" in
+	/*) controllers="/sys/fs/cgroup${user_cgroup}/cgroup.controllers" ;;
+	*) controllers="" ;;
+	esac
+	if [ -z "${controllers}" ] || [ ! -f "${controllers}" ]; then
 		WARNING "Enabling cgroup v2 is highly recommended, see https://rootlesscontaine.rs/getting-started/common/cgroup2/ "
 	else
 		for f in cpu memory pids; do
--- a/pkg/cmd/container/run_cgroup_linux.go
+++ b/pkg/cmd/container/run_cgroup_linux.go
@@ -228,14 +228,21 @@ func generateCgroupOpts(id string, options types.ContainerCreateOptions, interna
 }
 
 func generateCgroupPath(id, cgroupManager, cgroupParent string) (string, error) {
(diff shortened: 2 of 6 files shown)


RRXXZZYY:
I checked the two red jobs. The old-Ubuntu job fails in `TestPush/all_tags,_soci` and `TestIPFSCompNoBuild` because SOCI/IPFS content digests are missing. The rootless gomodjail job fails `TestBuildContextWithOCILayout` and `TestLoadStdinFromPipe` after gomodjail blocks `prctl` calls.

None of those paths are touched here; this diff is limited to rootless cgroup discovery/setup and its focused tests. The latest `main` run also has failures in the same old-Ubuntu and rootless-gomodjail matrices (with different integration tests), so this looks like matrix instability rather than a regression from this change. I haven't pushed an unrelated workaround.

AkihiroSuda:
> I do not have a WSL 2.9 host available, so the live WSL behavior remains to be confirmed by CI or a maintainer environment. 

A contributor is responsible for testing their PR ahead of the submission.
I'm closing this, but you or anybody else may resubmit the same PR after testing.

PR #5191, opened 2026-09-08
fix(image): avoid dangling-ref name collision on rmi -f for running images

### LINKED ISSUE #4109: Removing multiple running images is broken

ogulcanaydogan:
### Motivation

\`nerdctl rmi -f\` on an image that is still used by a running container renames it before deleting the original name, so containerd keeps the layers alive instead of garbage-collecting them. That dangling ref was always renamed to the literal string \`":"\`. Since containerd's image store requires unique names, force-removing a second running image's image in a separate invocation made its own rename fail with \`image ":": already exists\`, aborting the command.

### Approach

Name the dangling ref after its content digest (\`":<digest>"\`) instead of the fixed \`":"\` literal, so each kept-alive ref gets a distinct name. Also tolerate \`AlreadyExists\` on the create call: if two different tags happen to share the same digest and are both force-removed as running images in one invocation, the second create legitimately no-ops (the digest is already pinned).

The one other consumer that special-cased the exact \`":"\` name (\`pkg/imgutil\` reference filtering, used by \`--filter reference=...\` to skip dangling images without erroring on their unparsable name) is updated to match on the \`":"\` prefix instead.

### Testing

- Added \`pkg/imgutil/filtering_test.go\` coverage for both dangling-name shapes (\`":"\` and \`":<digest>"\`) in \`FilterByReference\`.
- Added an integration subtest to \`TestRemove\` in \`cmd/nerdctl/image/image_remove_test.go\` that reproduces the exact reported scenario: force-remove a running image's image, then force-remove a second, different running image's image.
- Verified end-to-end against a real containerd (rootless) daemon: reproduced the original failure on unpatched upstream main (\`image ":": already exists\`), then confirmed both force-removes succeed cleanly on this branch, with the two dangling refs ending up with distinct digest-based names.
- \`go build ./...\`, \`go vet ./pkg/cmd/image/... ./pkg/imgutil/...\`, \`gofmt -l\`, \`gci diff\`, \`gofumpt -extra -l\` all clean on the changed files.

Fixes #4109

### DIFF
--- a/cmd/nerdctl/image/image_remove_test.go
+++ b/cmd/nerdctl/image/image_remove_test.go
@@ -150,6 +150,38 @@ func TestRemove(t *testing.T) {
 			},
 		},
 		{
+			Description: "Issue #4109 - force-removing a second running image's image does not collide with the first dangling ref",
+			NoParallel:  true,
+			Require: require.All(
+				require.Not(nerdtest.Docker),
+			),
+			Setup: func(data test.Data, helpers test.Helpers) {
+				helpers.Ensure("run", "--quiet", "--pull", "always", "-d", "--name", data.Identifier()+"-1", testutil.CommonImage, "sleep", nerdtest.Infinity)
+				helpers.Ensure("run", "--quiet", "--pull", "always", "-d", "--name", data.Identifier()+"-2", testutil.BusyboxImage, "sleep", nerdtest.Infinity)
+				// Force-remove the first running image's image now: this creates a dangling ref
+				// to keep its layers alive. Before the fix, that ref was unconditionally named
+				// ":", so the second force-remove below (the command under test) would fail
+				// creating its own dangling ref with "image \":\": already exists".
+				helpers.Ensure("rmi", "-f", testutil.CommonImage)
+			},
+			Cleanup: func(data test.Data, helpers test.Helpers) {
+				helpers.Anyhow("rm", "-f", data.Identifier()+"-1")
+				helpers.Anyhow("rm", "-f", data.Identifier()+"-2")
+			},
+			Command: test.Command("rmi", "-f", testutil.BusyboxImage),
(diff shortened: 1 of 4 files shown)


cmd/nerdctl/image/image_remove_test.go
```diff
@@ -149,6 +149,38 @@ func TestRemove(t *testing.T) {
 				}
 			},
 		},
+		{
+			Description: "Issue #4109 - force-removing a second running image's image does not collide with the first dangling ref",
+			NoParallel:  true,
+			Require: require.All(
+				require.Not(nerdtest.Docker),
```
AkihiroSuda, line 156: ?

ogulcanaydogan:
Rebased and reworded the confusing description - it was garbled ("a second running image's image"), now reads "force-removing an in-use image does not collide with a dangling ref left by an earlier force-remove", which matches what the test actually does.

cmd/nerdctl/image/image_remove_test.go
AkihiroSuda: @ogulcanaydogan Please add a comment line to explain why this doesn't work on Docker, and squash the commits

PR #5192, opened 2026-09-08
gomodjail: update to v2.0.1, enforce static analysis in CI

AkihiroSuda:
[gomodjail v2](https://github.com/AkihiroSuda/gomodjail/releases/tag/v2.0.0) moves the focus from the dynamic (seccomp) mode to a static analysis gate: `gomodjail analyze` fails if a module annotated `gomodjail:confined` in go.mod can reach a denied capability (filesystem, network, process execution, raw syscalls, OS state modification, or cgo).

- Dockerfile: bump GOMODJAIL_VERSION to v2.0.1. `gomodjail pack` is unchanged in v2, so the packed `nerdctl.gomodjail` binary (dynamic mode) keeps working as before.
- Makefile: add `lint-gomodjail` / `lint-gomodjail-all` (part of `make lint`) and `fix-gomodjail` (part of `make fix`), and install gomodjail in `install-dev-tools`. The gate is only enforced for linux/amd64 and linux/arm64: these are the only platforms the packed binary is built for, and the only ones the dynamic mode supports. The verdicts are platform-dependent, hence both architectures are analyzed.
- CI: add the `gomodjail` job to the lint workflow.
- go.mod: `gomodjail fix` downgraded the annotation of the 39 modules that fail the gate to `gomodjail:unconfined`, so that the decision stays visible and reviewable. 57 modules remain confined (24 ok, 33 warnings, 0 violations). The inline annotations are also reformatted to `// gomodjail:...`, the style `gomodjail fix` writes, so that future fixes do not leave go.mod in two styles.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,18 +8,18 @@ require (
 	github.com/Microsoft/go-winio v0.6.3-0.20251027160822-ad3df93bed29
 	github.com/Microsoft/hcsshim v0.15.0-rc.4
 	github.com/compose-spec/compose-go/v2 v2.15.0 //gomodjail:unconfined
-	github.com/containerd/accelerated-container-image v1.4.4
+	github.com/containerd/accelerated-container-image v1.4.4 //gomodjail:unconfined
 	github.com/containerd/cgroups/v3 v3.1.3 //gomodjail:unconfined
 	github.com/containerd/console v1.0.5 //gomodjail:unconfined
-	github.com/containerd/containerd/api v1.12.0-rc.0
+	github.com/containerd/containerd/api v1.12.0-rc.0 //gomodjail:unconfined
 	github.com/containerd/containerd/v2 v2.4.0-beta.0 //gomodjail:unconfined
 	github.com/containerd/continuity v0.5.0 //gomodjail:unconfined
 	github.com/containerd/errdefs v1.0.0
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -7,123 +7,144 @@ require (
...
 	github.com/Microsoft/go-winio v0.6.3-0.20251027160822-ad3df93bed29
 	github.com/Microsoft/hcsshim v0.15.0-rc.4
-	github.com/compose-spec/compose-go/v2 v2.15.0 //gomodjail:unconfined
-	github.com/containerd/accelerated-container-image v1.4.4
-	github.com/containerd/cgroups/v3 v3.1.3 //gomodjail:unconfined
-	github.com/containerd/console v1.0.5 //gomodjail:unconfined
-	github.com/containerd/containerd/api v1.12.0-rc.0
-	github.com/containerd/containerd/v2 v2.4.0-beta.0 //gomodjail:unconfined
-	github.com/containerd/continuity v0.5.0 //gomodjail:unconfined
+	github.com/compose-spec/compose-go/v2 v2.15.0 // gomodjail:unconfined
```
AkihiroSuda: TODO: revert these whitespace changes

ChengyuZhu6 (approved):
LGTM

PR #5197, opened 2026-09-09
fix: match ps --filter keys exactly instead of by prefix

immanuwell:
`nerdctl ps --filter` matches the filter key with `strings.HasPrefix`, so any key that starts with a supported one gets misrouted: `--filter labels=env` runs as `label=env`, `--filter name2=foo` as `name=foo`. 
Docker rejects these, and nerdctl already rejects other unknown filters like `driver=x`

Fix: split on `=` once and match the key exactly. 
Unknown keys hit the existing `invalid filter` error. 
Valid filters (`label=k=v` included) are unchanged

Repro before this change:

```
nerdctl run -d --net none --name f1 --label env=prod alpine sleep 600
nerdctl ps --filter 'labels=env'
```

nerdctl: prints f1, exit 0
docker: `invalid filter 'labels'`, exit 1

### DIFF
--- a/cmd/nerdctl/container/container_list_linux_test.go
+++ b/cmd/nerdctl/container/container_list_linux_test.go
@@ -17,6 +17,7 @@
 package container
 
 import (
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

Issue #3452, opened 2024-09-22
volume (probably others) `filter` behavior is inconsistent with Docker


apostasie:
### Description

```
docker volume create foo1
docker volume create foo2
docker volume ls --quiet --filter name=foo1 --filter name=foo2
```

vs.

```
nerdctl volume create foo1
nerdctl volume create foo2
nerdctl volume ls --quiet --filter name=foo1 --filter name=foo2
```

Docker does "or", nerdctl does "and".

Discovered while rewriting the volume tests with the new tooling.

### Steps to reproduce the issue

na

### Describe the results you received and expected

na

### What version of nerdctl are you using?

na

### Are you using a variant of nerdctl? (e.g., Rancher Desktop)

None

PR #5206, opened 2026-09-13
fix(network): combine repeated name filters with OR

immanuwell:
Repeated `network ls` name filters used AND, so two different names returned nothing. Docker uses OR here

The fix matches any requested name while keeping label filters as AND. 
Added unit and integration coverage

Easy repro:

```bash
nerdctl network create frontend
nerdctl network create backend
nerdctl network ls -f name=^frontend$ -f name=^backend$
```

Before, no output. 
After, both names show up

Related to #3452


AkihiroSuda (approved):
Thanks

PR #5212, opened 2026-09-17
fix(compose): allow detached runs with default interactive setting

immanuwell:
`nerdctl compose run -d` bails because `-i` defaults to true and conflicts with detach. 
This sets the implicit interactive mode to false for detached runs; explicit `-i=true` still errors

Repro:
```bash
printf 'services:\n  app:\n    image: busybox:1.36\n    network_mode: none\n' > compose.yaml
nerdctl compose -f compose.yaml run -d app sleep 60
```
Before: `currently flag -i and -d cannot be specified together`. 
Now it prints a container ID

Tested with nerdctl/containerd and Docker Compose


AkihiroSuda (approved):
Thanks

PR #5223, opened 2026-09-19
fix(top): handle stopped and paused containers correctly

immanuwell:
`nerdctl top` silently exits 0 with no output for stopped and paused containers

Repro:

```sh
id=$(nerdctl run -d alpine sleep 300)
nerdctl stop "$id"
nerdctl top "$id"
echo $?
```

Before this prints nothing and exits 0. Pausing a running container has the same empty output

This returns `is not running` for stopped tasks and lets paused tasks reach the existing PID lookup. 
Running containers stay unchanged. 
Tests cover both cases with nerdctl and Docker


AkihiroSuda (approved):
Thanks

